PR #23, opened 2025-01-19
feat: API types validation

Krystian19:
tickets:
https://github.com/pathwaysdata/pathways-service/issues/43
https://github.com/pathwaysdata/pathways-service/issues/44


packages/server/src/dto/package.dto.ts
kylenesbit, line 1: We should discuss the DTO object automatically generated by the openapi generator and how to reconcile them with the definition in this package.
https://github.com/malloydata/publisher/blob/main/packages/server/src/api.ts#L50

Krystian19: I did a change so that the dto declaration references the openapi generated interface.

Krystian19: That way if we generate the api interface with a new field, we should get a compilation error if that new field is not present on the DTO class (where the actual data validation is being done).

kylenesbit: Got it!  I love that.  This is awesome.  Really happy to have your expertise on this.

kylenesbit (approved):

PR #37, opened 2025-03-25
Move publisher to bun

KantorSerhiy:
Move project from node to bun.

- Rewrite in each module(app, sdk, server) package.json to use bun.
- Update scripts in modules/main package.json
- Removed unnecessary dependencies.(chai, mocha, tsx, etc.)
- Added required dependencies (for Bun).
- Rewrite tests in server to use bun test runner instead mocha.


kylenesbit (approved):
Looks great!  Thank you so much for doing this.  I can't wait to start load testing this.

@Krystian19 implemented the tests.  Please let him take a look too.

franco-roura (approved):
This is awesome

Krystian19 (approved):
Looks awesome, thanks a lot

docker/Dockerfile.ci
```diff
@@ -1,7 +1,7 @@
-FROM node:20-alpine
+FROM oven/bun:latest
```
Krystian19, line 1: Always use a fixed version, ideally staying away from using `latest` whenever possible. The problem of not using a fixed version, is that we may find ourselves several months down the road with a broken build, just because bun introduced a breaking change out of nowhere.

KantorSerhiy: Great point, thanks!

Will fix.

PR #40, opened 2025-03-31
Add K6-compatible tests to the server

franco-roura:
This PR enhances the Malloy Publisher Server with K6 test presets and an out-of-the-box OpenTelemetry integration.
It also implements instrumentation for Malloy package load times and query execution times.

Screenshot 2025-04-11 at 3 08 21 PM

image

## Changes

- Added six different K6 tests to the server:
  - Smoke Test
  - Load Test
  - Stress Test
  - Spike Test
  - Breakpoint Test
  - Soak Test
- Updated the server's readme documenting usage of K6 presets
- Updated the main readme to say bun instead of npm


packages/sdk/README.md
```diff
@@ -1 +1,90 @@
-# Pathways SDK
\ No newline at end of file
+# Malloy Publisher SDK
+
+The Malloy Publisher SDK is a comprehensive toolkit designed to facilitate the development and testing of Malloy packages.
+
+## K6 Test Presets
```
kylenesbit: We should move the load test out of the SDK.  The SDK is a set of front end components.  The load test belongs either in the server package or in a new package.

packages/sdk/k6-tests/common.ts
```diff
@@ -0,0 +1,79 @@
...
+   patterns: open("../dist/patterns.zip", "b"),
+   bq_ga4: open("../dist/bq_ga4.zip", "b"),
+   bq_ga_sessions: open("../dist/bq_ga_sessions.zip", "b"),
+   bq_hackernews: open("../dist/bq_hackernews.zip", "b"),
+   bq_the_met: open("../dist/bq_the_met.zip", "b"),
+};
+type SampleName = keyof typeof samples;
+export const sampleNames = Object.keys(samples) as SampleName[];
+
+const SIDECAR_URL = __ENV.SIDECAR_URL || "http://localhost:4001";
```
kylenesbit: This load test should run on the OSS publisher.  It shouldn't have any dependencies on the MS2 internal components.  The sidecar is a MS internal component -- it doesn't exist in the publisher repo.

I imagine our initial load test should:
------------------
1) Start the OSS publisher with the PACKAGE_ROOT pointed at the malloy-samples submodule

2) Call get on each individual packages in malloy samples
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L93

This will cause the publisher to compile the package contents and load it into memory.  We want to know this load time.

3) List the model files in each package
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L136

Get each model file in the package:
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L181

We want to know the time it takes to get each model file.  

4) For model files of type "source"
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L475

We want to iterate through each of the sources, 
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L475

Then iterate through the views within each source,
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L488

And query the view by calling queryResult and passing the source name and the view name (as the queryName parameter)
https://github.com/malloydata/publisher/blob/main/api-doc.yaml#L230

We want to know the time it takes to query each view.

------------------

Ideally, we would be able to unload & load packages in the background -- and measure how loading packages in the background impacts query latency.  This will be pretty normal for the publisher.  Package loading is very CPU intensive and I believe it's tying up the node server's single thread and blocking other operations.

franco-roura: Thanks for the feedback! I moved it to the server, I think the new implementation fits more the intended use case
I'll udpate the PR's description as well

kylenesbit (approved):
This is awesome!

PR #61, opened 2025-04-29
Update Malloy deps to 0.0.270

quirkyllama:
* Updates malloy deps to 0.0.270.
  This updates us past the DuckDB 1.1.1 dep which caused new `bun install`s to compile DuckDB in situ.
* Adds bun script to update all malloy deps to a new version.
* Deletes `package-lock.json`. The repo uses `bun`- The package-lock is just confusing. There might should be a CI action which ensures no package-lock is ever added again?


scripts/upgrade-malloy.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
kylenesbit, line 1: Nice!

kylenesbit (approved):

PR #62, opened 2025-04-30
Enables "dev" mode for Publisher server/React development

quirkyllama:
* In dev mode, Node server proxies UI requests to the React server.
* Adds package.json targets to start Node & React servers in dev mode.
* Updates README


README.md
```diff
@@ -44,9 +44,26 @@ git submodule init
...
 
-Then build and run the package server:
+Then install the NPM deps:
 ```
 bun install
+```
+
+In development, there are 2 servers- the Node.JS API server and the React Dev server.
+In production, there's just 1 server- the Node server, and it serves the UI statically.
+The development server will hot-load changes. Changes to the Node.JS will cause the server to restart, changes to the React UI should be hot-loaded. Though note that BOTH systems sometimes fail to detect/implement changes, so don't be shy about restarting them.
```
kylenesbit, line 75: Ideally, it would hot load the malloy packages if they change as well.  I think this would require some code changes in the server, but want to put it on your radar.

kylenesbit (approved):
Nice!

README.md
quirkyllama: Yes- that's more tricky, or at least, not something I've done before...

PR #79, opened 2025-05-02
Publisher returns JSON'd MalloyResult/ModelInfo instead of QueryResult/ModelDEf

quirkyllama:


packages/sdk/src/components/Notebook/NotebookCell.tsx
```diff
@@ -129,16 +127,15 @@ export function NotebookCell({
                   }}
                >
                   <Typography
+                     component="div"
                      sx={{
                         fontSize: "12px",
                         "& .line": { textWrap: "wrap" },
                      }}
+                     dangerouslySetInnerHTML={{
```
quirkyllama, line 135: Unrelated change, but fixes a react warning about invalid DOM hierarchy (same for Typography change below as well)

packages/sdk/src/components/Notebook/NotebookCell.tsx
```diff
@@ -129,17 +127,15 @@ export function NotebookCell({
...
                         "& .line": { textWrap: "wrap" },
                      }}
-                  >
-                     <div
-                        dangerouslySetInnerHTML={{
-                           __html: highlightedEmbedCode,
-                        }}
-                     />
-                  </Typography>
+                     dangerouslySetInnerHTML={{
```
quirkyllama, line 134: Unrelated change, but fixes a react warning about invalid DOM hierarchy (same for Typography change below as well)

packages/server/src/mcp/tools/execute_query_tool.ts
```diff
@@ -191,7 +187,7 @@ export function registerExecuteQueryTool(
                      resource: {
                         type: "application/json",
                         uri: modelDefUri,
-                        text: modelDefString,
+                        text: modelInfoString,
```
quirkyllama: MCP is new- are there downstream consumers that need to be updated or is this just arbitrary JSON sent to the model context for it to "parse"?

api-doc.yaml
```diff
@@ -860,27 +855,20 @@ components:
...
       type: object
       description: A Malloy query's results, its model def, and its data styles.
       properties:
-        # Pass data styles as an opaque JSON string that is malloyVersion depdendent.
-        dataStyles:
-          type: string
-          description: Data style for rendering query results.
-        # Pass model def as an opaque JSON string that is malloyVersion depdendent.
-        modelDef:
+        modelInfo:
```
quirkyllama: ModelInfo isn't user by the renderer, but I left it in here because I figure it's useful for other purposes?

kylenesbit: ModelInfo isn't embedded in Malloy.Result?  

I would rip it out if it is not longer needed.  We can always add it later.

packages/sdk/src/components/Notebook/NotebookCell.tsx
kylenesbit: Nice!

packages/server/src/mcp/tools/execute_query_tool.ts
kylenesbit: I believe it's arbitrary text for the models.  I imagine the modelinfo is cleaner.  I think this is a fine change.

kylenesbit (approved):
Looks great!  Thank you.

api-doc.yaml
quirkyllama: No- the only thing in Malloy.Result from ModelInfo are the model annotation.
Removed- I guess there are other API endpoints which return the ModelInfo data.

PR #90, opened 2025-05-06
Adds connection panel to publisher, updates Database panel

quirkyllama:
* New "Connection" panel in package view.
* Updates embedded Database view (with DB size)
* Connection API no longer returns connection details (password/host/username)

Screenshot 2025-05-06 at 2 49 38 PM


kylenesbit (approved):
Can you update the StyledCards to be two across on large screens?  I think can just bump the lg: 6 on the package config, models, databases, & connections.

PR #93, opened 2025-05-07
Removes generated OpenAPI files

quirkyllama:
* Deletes generated OpenAPI files
* Adds then to .gitignore
* Updates build targets to automatically generate OpenAPI files
* Adds comment in README 
* Adds Java to CI dockerfile


packages/sdk/src/client/git_push.sh
quirkyllama, line 1: I assume this was a mistake? I don't know what this does or why it's in src/client

PR #98, opened 2025-05-12
Use Explorer in publisher

quirkyllama:
* Updates Publisher to return SourceInfos in ApiCompiledModel
* Updates Model.tsx to use ModelExplorer to display models. 
* 

https://github.com/user-attachments/assets/3c74c30b-942c-4f3f-b62d-c483305530c7

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -786,11 +786,6 @@ components:
           description: Array of JSON string of SourceInfo. See malloy/packages/malloy-interfaces/src/types.ts
           items:
             type: string
-        sources:
-          type: array
-          description: Array of model sources.
-          items:
-            $ref: "#/components/schemas/Source"
         queries:
           type: array
           descript: Array of named queries.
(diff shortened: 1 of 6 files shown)


api-doc.yaml
```diff
@@ -781,6 +781,11 @@ components:
         modelInfo:
           type: string
           description: JSON string of ModelInfo. See malloy/packages/malloy-interfaces/src/types.ts
+        sourceInfos:
+          type: array
+          description: Array of JSON string of SourceInfo. See malloy/packages/malloy-interfaces/src/types.ts
+          items:
+            type: string
         sources:
```
kylenesbit: Can we get rid of sources?  Or, do we still need it?

kylenesbit (approved):

api-doc.yaml
quirkyllama: Yeah- meant to remove it after got the new SourceInfo working.
Gone

PR #99, opened 2025-05-13
Add Docker instructions

quirkyllama:
Adds Docker files & README instructions for building/serving from a dockerfile


quirkyllama:
Tested this locally and it works.
I'm not sure how our publisher docker files are created for deployment, but this was straightforward enough.
They aren't the most compact- it does a hermetic build (which involves pulling all the deps etc...), but I thought that was a reasonable tradeoff.

kylenesbit (approved):
Looks great!

PR #102, opened 2025-05-14
Working version of schema explorer for BigQuery

quirkyllama:
* Server API for querying schemas & tables from a connection.
* UI for traversing schemas + tables.
* Only BigQuery is supported
* Requires specifying defaultProjectId on the sample BQ connection.

https://github.com/user-attachments/assets/c3ada7b0-1926-422e-bf19-c792b55921d0


mhwilder (approved):
This looks good to me.

api-doc.yaml
```diff
@@ -203,7 +203,7 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
   
-  /projects/{projectName}/connections/{connectionName}/tableSource:
+  /projects/{projectName}/connections/{connectionName}/{tablePath}/tableSource:
```
kylenesbit: Shouldn't this be
/projects/{projectName}/connections/{connectionName}/tableSource/{tablePath} to be consistent with the rest of the API?  Also, I would probably drop the Source and just call it "tables".  

Also, this is going to break the malloy publisher-connection type.  We will need to update it and coordinate getting it released with the malloy team.

api-doc.yaml
```diff
@@ -250,6 +244,86 @@ paths:
         500:
           $ref: "#/components/responses/InternalServerError"
 
+  /projects/{projectName}/connections/{connectionName}/{schemaName}/tables:
```
kylenesbit: Would be good to make this API consistent as well.

PR #116, opened 2025-05-23
feat: mcp/prompts

namabile:
### 1. Prompts barrel export
`packages/server/src/mcp/prompts/index.ts` now re-exports:
* `PROMPTS`, `withVersion`, arg schemas
* `promptHandlerMap`, `makePromptHandler`
* `registerPromptCapability`

### 2. Documentation
* Root `README.md` – new “MCP Prompt Capability” section with prompt list and examples.
* `packages/server/README.md` – brief note pointing to the root docs.

All unit & integration tests pass after rebase onto latest `main`.

BigQuery sample tests remain skipped; caching/logging improvements come in a follow-up.


packages/server/src/mcp/prompts/prompt_definitions.ts
```diff
@@ -0,0 +1,160 @@
+import { z } from "zod";
+
```
kylenesbit, line 2: Is it possible to move the prompts definition into a json configuration file?

I would like to be able to version the prompts with the model files.

kylenesbit (approved):
Looks great!

packages/server/src/mcp/prompts/prompt_definitions.ts
namabile: sounds good. if it's cool, i'll merge this as is for now and then circle back to add that feature.

PR #120, opened 2025-05-28
split models/notebooks in UI

quirkyllama:
Screenshot 2025-05-27 at 6 33 42 PM

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -525,6 +525,16 @@ paths:
           required: false
           schema:
             type: string
+        - name: filter
+          in: query
+          description: Filter to either "source" or "notebook" or "all"
+          required: false
+          schema:
+            type: enum
+            enum:
+              - source
+              - notebook
+              - all
       responses:
         "200":
           description:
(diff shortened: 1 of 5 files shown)


api-doc.yaml
```diff
@@ -525,6 +525,13 @@ paths:
           required: false
           schema:
             type: string
+        - name: filter
```
kylenesbit, line 528: Can we create a list notebooks operation?  To be consistent with the rest of the API?

  /projects/{projectName}/packages/{packageName}/notebooks:
    get:
      tags:
      - notebooks
      operationId: list-notebooks

api-doc.yaml
```diff
@@ -914,13 +984,17 @@ components:
...
-  
+        newSources:
+          type: array
+          description: Array of JSON string of SourceInfo made available in the notebook cell. Only *new*
+            SourceInfos are returned. The complete list of SourceInfos is available be concatenating the prior
+            notebook cells.  The SourceInfos are in the order they are made available in the notebook cell.
+          items:
+            type: string
+
     QueryResult:
```
kylenesbit, line 998: Is there a new stable type for QueryResults?

quirkyllama:
> This is awesome!
> 
> Did you run the MCP tests? I am not sure if it has any notebook dependencies.

I think there's work to do *adding* the new endpoint to MCP, but this didn't break any existing endpoint

PR #181, opened 2025-06-18
Packages publisher/app to be used as an NPM dependency by another app

quirkyllama:
* Router & render code all contained in exported `MalloyPublisherApp`
* Subsequent service PR will deploy this wrapped in Auth0
* Creates ServerProvider to transmit server/accessToken (removed from PackageProvider)
* useQueryWithApiError now pulls in the server/accessToken from the useServer hook, then creates the config and passes it back to queryFn. Lots of boilerplate removed.

### DIFF
--- a/packages/app/src/main.tsx
+++ b/packages/app/src/main.tsx
@@ -4,6 +4,8 @@ import { RouterProvider } from "react-router-dom";
 import { ServerProvider } from "@malloy-publisher/sdk";
 import { createMalloyRouter } from "./App";
 
+// Main.tsx is used to run the app locally. This is not used when the app is
+// embedded in another project.
 const server = `${window.location.protocol}//${window.location.host}/api/v0`;
 const router = createMalloyRouter("/");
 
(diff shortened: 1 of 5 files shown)


packages/sdk/src/components/Home/Home.tsx
```diff
@@ -7,22 +7,15 @@ import { useQueryWithApiError } from "../../hooks/useQueryWithApiError";
...
 }
 
-export default function Home({ server, navigate }: HomeProps) {
+export default function Home({ navigate }: HomeProps) {
    const { data, isSuccess, isError, error } = useQueryWithApiError({
-      queryKey: ["projects", server],
-      queryFn: () =>
-         projectsApi.listProjects({
-            baseURL: server,
-            withCredentials: true,
```
quirkyllama, line 20: @kylenesbit Is this an error? listProjects does not take an accessToken and has different auth semantics.
My best guess is this made sense for how app used to be distributed, but now that the whole package is behind auth, this method should use the same accessToken pattern

kylenesbit (approved):
This is awesome!

packages/sdk/src/components/Home/Home.tsx
kylenesbit: Yes. This looks like an error.  The must do the right thing wrt to listProjects -- otherwise, it wouldn't work.  Good to normalize this.

kylenesbit (approved):

PR #183, opened 2025-06-19
Fix SDK build

franco-roura:
This PR fixes a bug where the malloy SDK would always bundle the `development` version of React 19, causing apps built with the `production` version of React 19 to crash by referencing undefined methods.


packages/sdk/vite.config.ts.timestamp-1732998513502-40117fb4923d1.mjs
```diff
@@ -1,6 +1,6 @@
 // vite.config.ts
-import { defineConfig } from "file:///home/kjnesbit/publisher/node_modules/vite/dist/node/index.js";
```
quirkyllama, line 2: I don't think this file should exist at all?

quirkyllama (approved):

PR #277, opened 2025-07-18
Fix app bug where packageName was used instead of resource

franco-roura:
This PR fixes a bug where the frontend tried to read packages by their name instead of their assigned resource url, creating links in the front page that led to broken pages.

This is what happened before and after this PR when creating a malloy sample called `dirName` containing a publisher.json file such as 
```json
{
    "name": "packageName",
    "version": "1.0.0",
    "description": "This should break" 
}
```
| Before | After |
| --- | --- |
| localhost_4000_malloy-samples_packageName_ | localhost_4000_malloy-samples_ |


kylenesbit (approved):
Nice!

PR #289, opened 2025-07-29
Add GCS+S3+GH compatibility

franco-roura:
This PR lets the publisher fetch and load projects and packages from remote origins during boot. It also addresses security concerns about relative paths by disallowing them.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1105,6 +1105,9 @@ components:
         readme:
           type: string
           description: Project readme.
+        location:
+          type: string
+          description: Project location, it can be an absolute path or an URI (e.g. github, s3, gcs, etc.).
 
     Package:
       type: object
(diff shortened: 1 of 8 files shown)


api-doc.yaml
```diff
@@ -1105,6 +1105,9 @@ components:
         readme:
           type: string
           description: Project readme.
+        location:
```
kylenesbit, line 1108: Feels like this field name should have URI in it.

packages/server/src/service/project_store.ts
```diff
@@ -166,4 +183,207 @@ export class ProjectStore {
          }
       }
    }
+
+   private async loadProjectIntoDisk(projectName: string, projectPath: string) {
```
kylenesbit, line 187: Not sure if project store is the best place for the package loading code.  It's fine for now, but may want to move it into utils on the next pass.

PR #295, opened 2025-08-07
Create connections on project registration

franco-roura:
This PR makes the publisher automatically load pertinent Malloy connections in memory when a project is created through the API.

It also introduces several layers of caching at the Dockerfile level to improve DX, and a few tweaks on the SDK's vite.config to reduce build times, and fixes bugs with serialization of project DTO's.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -10,6 +10,20 @@ servers:
   - url: /api/v0/
 
 paths:
+  /status:
+    get:
+      tags:
+      - publisher
+      operationId: get-status
+      summary: Server status
+      responses:
+        '200':
+          description: Returns server status
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ServerStatus'
+
   /projects:
     get:
       tags:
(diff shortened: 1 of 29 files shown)


api-doc.yaml
```diff
@@ -1086,6 +1104,16 @@ components:
...
+      type: object
+      properties:
+        timestamp:
+          type: number
+        projects:
+          type: array
+          items:
+            $ref: '#/components/schemas/Project'
+
     About:
```
kylenesbit: Can we get rid of About?

packages/server/src/controller/schedule.controller.ts
```diff
@@ -14,8 +14,8 @@ export class ScheduleController {
       projectName: string,
```
kylenesbit, line 14: Can we rip schedules out of the Publisher?  It's a partially completed feature.  I created a separate ticket for this.
https://github.com/ms2data/service/issues/1380

api-doc.yaml
franco-roura: Sounds good, I removed it in [65e92b7](https://github.com/malloydata/publisher/pull/295/commits/65e92b7baa6bfc03c52851e58e916213625531cb)

packages/server/src/controller/schedule.controller.ts
franco-roura: Sure, may I make a separate pr for that ticket? This one covers quite a bit of surface

PR #305, opened 2025-08-18
New Publisher Config Format and Bug Fixes

sagarswamirao:


packages/server/src/service/project.ts
```diff
@@ -60,14 +74,48 @@ export class Project {
...
+               .catch(() => false))
+         ) {
             throw new ProjectNotFoundError(
                `Project path "${this.projectPath}" not found`,
             );
          }
          this.metadata.resource = payload.resource;
       }
       this.metadata.readme = payload.readme;
+      // const connections = payload.connections;
```
kylenesbit: Can you remove this rather than just commenting it out?

packages/server/src/server.ts
```diff
@@ -213,10 +210,56 @@ app.use(cors());
 app.use(bodyParser.json());
 
 app.get(`${API_PREFIX}/status`, async (_req, res) => {
-   res.status(200).json({
-      timestamp: Date.now(),
-      projects: await projectStore.listProjects(),
-   });
+   try {
+      let status: any = {
```
kylenesbit: Can we move the status logic into projectStore?

packages/server/src/service/project.ts
```diff
@@ -60,14 +74,48 @@ export class Project {
             `${API_PREFIX}/projects/`,
             "",
          );
-         if (!(await fs.promises.exists(this.projectPath))) {
```
kylenesbit, line 63: We probably should not support updating the project or resource name.

packages/server/src/service/project.ts
```diff
@@ -60,14 +74,48 @@ export class Project {
...
                `Project path "${this.projectPath}" not found`,
             );
          }
          this.metadata.resource = payload.resource;
       }
       this.metadata.readme = payload.readme;
+      // const connections = payload.connections;
+
+      // Handle connections update
+      if (payload.connections) {
```
kylenesbit, line 71: Connections should probably have their own update function.  Probably not worth changing this now, but may add a TODO?

packages/server/src/server.ts
sagarswamirao: Sure

packages/server/src/service/project.ts
sagarswamirao: Oops! Sure :)

packages/server/src/service/project.ts
sagarswamirao: True, and we might in future want to test the connections within createConnections before adding them

PR #324, opened 2025-08-28
Update docs

franco-roura:


kylenesbit (commented):
Can you add a static link from the root of repo to the packages/server/publisher.config.json ?

PR #333, opened 2025-09-01
Updated publisher.config.json package links, handled connections.json not found in NPX version, and added support for relative local file paths

sagarswamirao:
+ Updated publisher.config.json package links
+ Handled connections.json not found in NPX version
+ Added support for relative local file paths


kylenesbit (approved):
Awesome!

PR #346, opened 2025-09-09
Split app & sdk chunks for faster & smaller builds

franco-roura:
This PR includes better chunk splitting for large dependencies, it adds more efficient bundle loading and reduced duplication.

| Before | After |
| --- | --- |
| image | image |


kylenesbit (approved):
Nice!

PR #347, opened 2025-09-09
Remove navigate parameters and replace with onClick or onSelect

franco-roura:
Replaced generic navigate props with more descriptive prop names that better reflect the specific user interactions


kylenesbit (approved):
Awesome!

PR #350, opened 2025-09-11
Add connections crud

franco-roura:
This PR adds a UI for connections CRUD and fixes a bug where connection internals weren't exposed to the frontend.


packages/app/vite.config.ts
```diff
@@ -134,10 +81,38 @@ export default ({ mode }) => {
...
+            onwarn(warning, warn) {
+               if (
+                  warning.code === "MODULE_LEVEL_DIRECTIVE" ||
+                  warning.code === "SOURCEMAP_ERROR"
+               ) {
+                  return;
+               }
+               warn(warning);
+            },
+            external: [
```
quirkyllama: I don't think this is needed anymore?
App is not a re-usable component, it's standalone so there is no "host"

packages/sdk/src/components/Connections/EditConnectionDialog.tsx
```diff
@@ -0,0 +1,354 @@
...
+      MysqlConnection);
+   type: HTMLInputTypeAttribute;
+   required?: boolean;
+};
+
+const connectionFieldsByType: Record<
+   ConnectionTypeEnum,
+   Array<ConnectionField>
+> = {
+   postgres: [
```
quirkyllama: shared with createConnection?

packages/app/vite.config.ts
franco-roura: Sounds fair, addressed in [94fbd32](https://github.com/malloydata/publisher/pull/350/commits/94fbd3201783ad22799d048f569aaec5ddd57efb)

packages/sdk/src/components/Connections/EditConnectionDialog.tsx
franco-roura: Addressed in [94fbd32](https://github.com/malloydata/publisher/pull/350/commits/94fbd3201783ad22799d048f569aaec5ddd57efb) as well

PR #351, opened 2025-09-11
Added POST API variants with SQL Statements

sagarswamirao:


packages/server/src/server.ts
```diff
@@ -449,6 +496,25 @@ app.get(
    },
 );
 
+app.post(
```
quirkyllama, line 500: Isn't this the same as L459

sagarswamirao: Oops, will fix that...

PR #378, opened 2025-10-04
trino and mysql testConfig implementation

manojdeep14:
Added some checks for Trino connection and added testConfig for Trino and MySQL

### DIFF
--- a/packages/server/src/service/connection.ts
+++ b/packages/server/src/service/connection.ts
@@ -224,7 +224,7 @@ export async function createConnections(
 
             if (
                !trinoConfig.server?.startsWith("https://") &&
-               !trinoConfig.server?.startsWith("http")
+               !trinoConfig.server?.startsWith("http://")
             ) {
                throw new Error(
                   "Trino connection requires server to start with https:// or http://",
@@ -503,7 +503,7 @@ export async function testConnectionConfig(
 
          if (
             !trinoConfig.server?.startsWith("https://") &&
-            !trinoConfig.server?.startsWith("http")
+            !trinoConfig.server?.startsWith("http://")
          ) {
             throw new Error(
                "Trino connection requires: server to start with https:// or http://",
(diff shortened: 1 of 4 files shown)


packages/server/src/service/connection.ts
```diff
@@ -219,14 +219,51 @@ export async function createConnections(
...
-               schema: connection.trinoConnection.schema,
-               user: connection.trinoConnection.user,
-               password: connection.trinoConnection.password,
-            };
+
+            const trinoConfig = connection.trinoConnection;
+
+            if (
+               !trinoConfig.server?.startsWith("https://") &&
+               !trinoConfig.server?.startsWith("http://")
```
kylenesbit: Seems like this case is covered by the else below

packages/server/src/service/connection.ts
```diff
@@ -219,14 +219,51 @@ export async function createConnections(
...
+                  server: trinoConfig.server,
+                  port: trinoConfig.port,
+                  catalog: trinoConfig.catalog,
+                  schema: trinoConfig.schema,
+                  user: trinoConfig.user,
+                  password: trinoConfig.password,
+               };
+            } else {
+               throw new Error(
+                  "Trino connection requires password for https connection",
```
kylenesbit: Replace with a more generic message

packages/server/src/service/connection.ts
```diff
@@ -460,18 +497,51 @@ export async function testConnectionConfig(
             !trinoConfig.user
          ) {
             throw new Error(
-               "Trino connection requires: server, port, catalog, schema, and user",
+               "Trino connection requires server, port, catalog, schema, and user",
```
kylenesbit, line 493: Can you clean this logic up and put into a reusable method

packages/server/src/service/connection.ts
manojdeep14: Addressed in the recent commit, Kyle!!

packages/server/src/service/connection.ts
manojdeep14: Addressed in the recent commit, Kyle!!

packages/server/src/service/connection.ts
manojdeep14: Addressed in the recent commit, Kyle!!

PR #399, opened 2025-10-14
Split out ServerProvider into sdk/client

quirkyllama:
Splits out ServerProvider and ApiClient into sdk/client.
Consumers of the SDK can now depened on @malloy-publisher/sdk/client
This does not bring in the component/react/malloy deps. Makes it possible to use <ServerProvider> and useServer in top-level module code without causing the initial page-load to need to load 1.5MB of Malloy/SDK.
Also if you just want the OpenAPI code, that's in client without needing the rest of the SDK deps.


packages/sdk/TREE_SHAKING.md
sagarswamirao, line 1: Do we need this to be committed?

PR #412, opened 2025-10-15
Explorer UX: too many sources #1870

jellysaini:


jswir:
Thanks Jelly, 

Let's put all the options in a drop down - it is a bit confusing to have 5 sources, then a drop-down for additional sources. 

Let's default select the last source in the file, as this is often the primary fact table that has all the join relationships to other tables. 

let's add a search box to handle the case where there are a ton of them like my example in the issue

here's an example (but don't make it dark themed): 
image

Issue #419, opened 2025-10-17
Publisher: Presist CRUD State in DuckDB


kylenesbit:
We should store the Publisher's resource state in DuckDB so changes made via the CRUD API survive restarts. Also, need to update how Publisher initializes its config. On startup, the Publisher should not overwrite the DuckDB database unless an initialize config flag is passed.

We should build it in a way that would allow Publisher to store the resource state in an external postgres or mysql instance.

PR #503, opened 2025-11-24
Publisher: Presist CRUD State in DuckDB #419

jellysaini:


jellysaini:
@kylenesbit & @sagarswamirao 

I need to change the file storage path from the temporary directory to a persistent one. I’m considering creating a dedicated folder in the server project’s root directory and storing the files there. The project size will grow as more packages are added, but that should be acceptable.

packages/server/src/service/project_store.ts
```diff
@@ -21,74 +21,436 @@ import {
...
+         const storageConfig: StorageConfig = {
+            type: "duckdb",
+            duckdb: {
+               path: path.join(serverRootPath, "publisher.db"),
+            },
+         };
+         this.storageManager = new StorageManager(storageConfig);
+      } else {
+         // Use a mock/no-op storage manager in tests
+         this.storageManager = {
```
sagarswamirao: We have to mock it within the tests. This mocks shouldn't be here

packages/server/src/service/project_store.ts
```diff
@@ -21,74 +21,436 @@ import {
...
       followRegionRedirects: true,
    });
    private gcsClient: Storage;
 
    constructor(serverRootPath: string) {
+      this.projects = new Map();
       this.serverRootPath = serverRootPath;
       this.gcsClient = new Storage();
 
+      if (process.env.NODE_ENV !== "test") {
```
sagarswamirao: Can you remove this from here?

packages/server/src/server.ts
```diff
@@ -264,6 +266,7 @@ app.get(`${API_PREFIX}/projects/:projectName`, async (req, res) => {
 app.patch(`${API_PREFIX}/projects/:projectName`, async (req, res) => {
    try {
       const project = await projectStore.updateProject(req.body);
+      await projectStore.syncProjectToDatabase(project);
```
sagarswamirao: Can u update code to call syncProjectToDatabase() from within the Create, Update and Delete Project functions?

packages/server/src/server.ts
```diff
@@ -552,6 +555,8 @@ app.post(`${API_PREFIX}/projects/:projectName/packages`, async (req, res) => {
          req.params.projectName,
          req.body,
       );
+      const project = await projectStore.getProject(req.params.projectName);
```
sagarswamirao: Why do we have to syncProjectToDatabase again over here?
Please move this within the addPackage() impl

packages/server/src/server.ts
```diff
@@ -588,13 +593,19 @@ app.patch(
...
-         res.status(200).json(
-            await packageController.updatePackage(
-               req.params.projectName,
-               req.params.packageName,
-               req.body,
-            ),
+         const projectName = req.params.projectName;
+         const packageName = req.params.packageName;
+
+         const updated = await packageController.updatePackage(
```
sagarswamirao, line 596: rename to updatedPackage

packages/server/src/server.ts
```diff
@@ -588,13 +593,19 @@ app.patch(
...
+            projectName,
+            packageName,
+            req.body,
          );
+
+         const project = await projectStore.getProject(projectName);
```
sagarswamirao: Can you call this within the updatePackage()?

packages/server/src/server.ts
```diff
@@ -607,12 +618,15 @@ app.delete(
...
-               req.params.projectName,
-               req.params.packageName,
-            ),
-         );
+         const projectName = req.params.projectName;
+         const packageName = req.params.packageName;
+
+         await packageController.deletePackage(projectName, packageName);
+
+         const project = await projectStore.getProject(projectName);
```
sagarswamirao: Call it within the deletePackage()

package.json
```diff
@@ -18,9 +18,10 @@
     "build:server-deploy": "NODE_ENV=production && bun run build:sdk && bun run build:app:server && bun run build:server",
     "start": "cd packages/server && bun run start",
     "start:dev": "cd packages/server && bun run start:dev",
+    "start:dev:init": "cd packages/server && bun run start:dev:init",
```
sagarswamirao, line 22: If we rely on start:dev:init, then we’d also need to introduce a separate start:init command. 
Additionally, the publisher can be run either by building/pulling a Docker image or by using the npx command (from the npm registry). Have you tested --init flag works as expected for these?

packages/server/src/service/project_store.ts
```diff
@@ -233,10 +591,25 @@ export class ProjectStore {
...
+
+      if (!newProject.metadata) newProject.metadata = {};
+      newProject.metadata.location = absoluteProjectPath;
+
       this.projects.set(projectName, newProject);
-      projectConfig?.packages.forEach((_package) => {
-         newProject.setPackageStatus(_package.name, PackageStatus.SERVING);
-      });
+
+      (project?.packages || projectConfig?.packages || []).forEach(
```
sagarswamirao: Why do we need to check both project?.packages and projectConfig?.packages? Do you think the data differs between the two? If so, in what scenario would that happen?

packages/server/src/server.ts
```diff
@@ -34,6 +34,8 @@ function parseArgs() {
       } else if (arg === "--mcp_port" && args[i + 1]) {
          process.env.MCP_PORT = args[i + 1];
          i++;
+      } else if (arg === "--init") {
+         process.env.FORCE_INIT = "true";
```
sagarswamirao: Please change it to INIT or INITIALIZE_STORAGE

packages/server/src/storage/StorageManager.ts
```diff
@@ -0,0 +1,91 @@
...
+import { initializeSchema } from "./duckdb/schema";
+
+export type StorageType = "duckdb" | "postgres" | "mysql";
+
+export interface StorageConfig {
+   type: StorageType;
+   duckdb?: {
+      path?: string;
+   };
+   postgres?: {
```
sagarswamirao, line 13: Also support connectionUri.
User can enter all of these, or just use the connectionUri

packages/server/src/service/project_store.ts
```diff
@@ -21,74 +21,436 @@ import {
...
+      const dbPackages = await repository.getPackages(projectId);
+
+      for (const dbPackage of dbPackages) {
+         if (!actualPackages.includes(dbPackage.name)) {
+            await repository.deletePackage(dbPackage.id);
+         }
+      }
+   }
+
+   public async syncProjectToDatabase(project: Project): Promise<void> {
```
sagarswamirao: Can we break this up?
The function  is doing multiple things. It will be great to call specific functions that handle delete/create/update related to that particular resource (project, package, connection)

sagarswamirao (changes requested):
Please go through the comments and address them

packages/server/src/server.ts
jellysaini: This has been done.

packages/server/src/server.ts
jellysaini: This has been done.

packages/server/src/server.ts
jellysaini: This has been done.

packages/server/src/server.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/server.ts
jellysaini: This has been done.

package.json
jellysaini: I have added the start:init command and tested it. Everything is working fine on my side.

packages/server/src/service/project_store.ts
jellysaini: project?.packages comes from DuckDB, and projectConfig?.packages comes from the publisher.config.json or directory scan. We check both as a fail-safe — if DuckDB doesn’t have the latest package data, we fall back to the config so the project still loads correctly.


If you think I can only load from duckdb.

packages/server/src/storage/StorageManager.ts
jellysaini: Postgres and MySQL are just placeholders right now. I have not implemented their storage layers yet. When we actually move from DuckDB to Postgres or MySQL, we will add support for connectionUri and implement the connection + repository logic for those backends.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/server.ts
jellysaini: @sagarswamirao, updatePackage is part of the old code in packageController, and it needs to be updated in multiple places.

packages/server/src/service/project_store.ts
```diff
@@ -258,174 +258,257 @@ export class ProjectStore {
...
+         version?: string;
+         description?: string;
+         manifestPath?: string;
+         metadata?: Record<string, unknown>;
+      };
+
+      const packageData = {
+         projectId,
+         name: pkgWithExtras.name,
+         version: pkgWithExtras.version ?? "1.0.0",
```
sagarswamirao: I believe version is a mandatory field. Each package published will have a version present within the publisher.json
Use ! instead

packages/server/src/controller/package.controller.ts
```diff
@@ -42,15 +42,21 @@ export class PackageController {
...
    }
 
    public async deletePackage(projectName: string, packageName: string) {
       if (this.projectStore.publisherConfigIsFrozen) {
          throw new FrozenConfigError();
       }
       const project = await this.projectStore.getProject(projectName, false);
-      return project.deletePackage(packageName);
+      const result = await project.deletePackage(packageName);
+      await this.projectStore.syncProjectToDatabase(project);
```
sagarswamirao: Can we have a seperate function to syncPackageToDatabase(projectName, package) instead of using on generic function?

jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: /** @description Represents a Malloy package containing models, notebooks, and embedded databases */
      Package: {
        /** @description Resource path to the package */
        resource?: string;
        /** @description Package name */
        name?: string;
        /** @description Package description */
        description?: string;
        /** @description Package location, can be an absolute path or URI (e.g. github, s3, gcs, etc.) */
        location?: string;
      };
      
      
@sagarswamirao , there’s a file called src/api.ts that’s currently in .gitignore. I need to add the version field there so we can fetch the version from the publisher.json.  Right now we don’t store the publisher.json version anywhere. Please suggest me if I should remove this file from .gitignore and commit it so the version number can be properly tracked.

jellysaini: As we discussed. I have removed the version and not storing in the database.

packages/server/src/service/project_store.ts
jellysaini: This has been fixed.

packages/server/src/storage/duckdb/DuckDBRepository.ts
```diff
@@ -0,0 +1,379 @@
...
+
+      return this.getProject(id) as Promise<Project>;
+   }
+
+   async deleteProject(id: string): Promise<void> {
+      // First delete all related records
+      await this.db.run("DELETE FROM connections WHERE project_id = ?", [id]);
+
+      // Delete packages
+      await this.db.run("DELETE FROM packages WHERE project_id = ?", [id]);
```
sagarswamirao: To keep it consistent with the other functions you have defined in this file. Can u create fucntion like
```
async deleteConnectionsByProjectId(id){
  await this.db.run("DELETE FROM connections WHERE project_id = ?", [id]);
}
```
in a similar way can u create deletePackageByProjectId 
and call them into here

packages/server/src/storage/duckdb/schema.ts
```diff
@@ -0,0 +1,88 @@
...
+): Promise<void> {
+   const initialized = await db.isInitialized();
+
+   if (initialized && !force) {
+      return;
+   }
+
+   if (force) {
+      console.log(
+         "Forcing database schema initialization (dropping existing tables)...",
```
sagarswamirao: Reinitalizing database schema...

packages/server/src/storage/duckdb/DuckDBRepository.ts
jellysaini: This has been done.

packages/server/src/storage/duckdb/schema.ts
jellysaini: This has been done.

packages/server/src/storage/StorageManager.ts
```diff
@@ -0,0 +1,91 @@
...
+   private config: StorageConfig;
+
+   constructor(config: StorageConfig) {
+      this.config = config;
+   }
+
+   async initialize(forceReinit: boolean = false): Promise<void> {
+      if (forceReinit) {
+         console.log(
+            "FORCE INITIALIZATION MODE: Database will be dropped and recreated",
```
sagarswamirao: @jellysaini can you please go through the code and remove this word Force?

sagarswamirao: Initialization or re-initialzation suits better

packages/server/src/storage/StorageManager.spec.ts
```diff
@@ -0,0 +1,50 @@
+type MockData = Record<string, unknown>;
```
sagarswamirao, line 1: Why are the _mocks_ here?

packages/server/src/service/project_store.spec.ts
```diff
@@ -4,9 +4,90 @@ import * as path from "path";
...
+               }),
+               deleteConnection: async (_id: string): Promise<void> => {},
+            };
+         }
+      },
+      StorageConfig: {} as Record<string, unknown>,
+   };
+});
+
 import { ProjectStore } from "./project_store";
```
sagarswamirao: Imports always at the top please

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+                     );
+
+                     return projectInstance.listPackages();
+                  }),
+               );
+            }
+         }
+
          this.isInitialized = true;
-         logger.info(
```
sagarswamirao, line 69: Why is this removed?

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
-            `Project store successfully initialized in ${performance.now() - initialTime}ms`,
-         );
       } catch (error) {
          logger.error("Error initializing project store", { error });
-         console.error(error);
```
sagarswamirao, line 74: Why is this removed?

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+      const repository = this.storageManager.getRepository();
+
+      // Sync project metadata
+      const dbProject = await this.addProjectMetadata(project, repository);
+
+      // Sync packages
+      await this.addPackages(project, dbProject.id, repository);
+
+      // Sync connections
+      await this.addConnections(project, dbProject.id, repository);
```
sagarswamirao, line 214: add the connections first and then the packages

Swap line 231 and 234

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+      projectId: string,
+      repository: ReturnType<typeof this.storageManager.getRepository>,
+   ): Promise<void> {
+      const packages = await project.listPackages();
+      const packageNames = packages
+         .map((p) => p.name)
+         .filter((name): name is string => name !== undefined);
+
+      // Clean up orphaned packages
+      await this.cleanupPackages(projectId, packageNames);
```
sagarswamirao: Why do u want to call cleanupPackages everytime a project is added?
I remember we discussing to have call it at the startup of the code

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+      },
+      repository: ReturnType<typeof this.storageManager.getRepository>,
+   ): Promise<void> {
+      const existingPackages = await repository.getPackages(projectId);
+      const existingPackage = existingPackages.find(
+         (p) => p.name === packageName,
+      );
+
+      if (existingPackage) {
+         await repository.updatePackage(existingPackage.id, {
```
sagarswamirao: why not pass the packageData directly?

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+    * Delete a package from the database
+    */
+   public async deletePackageFromDatabase(
+      projectName: string,
+      packageName: string,
+   ): Promise<void> {
+      const repository = this.storageManager.getRepository();
+
+      // Get the project ID from database
+      const dbProjects = await repository.getProjects();
```
sagarswamirao: Use the getProjectByName()

packages/server/src/service/project_store.ts
```diff
@@ -37,59 +44,505 @@ export class ProjectStore {
...
+   private async cleanupAndCreatePublisherPath() {
+      const forceInit = process.env.INITIALIZE_STORAGE === "true";
+
+      if (forceInit) {
+         const uploadDocsPath = path.join(
+            this.serverRootPath,
+            PUBLISHER_DATA_DIR,
+         );
+         logger.info(
+            `Force init: Cleaning up upload documents path ${uploadDocsPath}`,
```
sagarswamirao: Remove the word Force

packages/server/src/service/project_store.ts
```diff
@@ -233,10 +681,20 @@ export class ProjectStore {
...
       );
+
+      if (!newProject.metadata) newProject.metadata = {};
+      newProject.metadata.location = absoluteProjectPath;
+
       this.projects.set(projectName, newProject);
-      projectConfig?.packages.forEach((_package) => {
-         newProject.setPackageStatus(_package.name, PackageStatus.SERVING);
+
+      (project?.packages || []).forEach((_package) => {
```
sagarswamirao: Remove || []

packages/server/src/storage/duckdb/DuckDBRepository.ts
```diff
@@ -0,0 +1,385 @@
...
+      return `${Date.now()}-${Math.random().toString(36).substr(2, 9)}`;
+   }
+
+   private now(): Date {
+      return new Date();
+   }
+
+   // ==================== PROJECTS ====================
+
+   async getProjects(): Promise<Project[]> {
```
sagarswamirao: Can u break it into separate repositories? One file per table and all the related CRUD + byId, byName functions go into it?

Like ProjectsRepository, PackagesRepository & ConnectionsRepository

packages/server/src/storage/StorageManager.ts
jellysaini: This has been done.

packages/server/src/storage/StorageManager.spec.ts
jellysaini: This has been done.

packages/server/src/service/project_store.spec.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/storage/duckdb/DuckDBRepository.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

packages/server/src/service/project_store.ts
jellysaini: This has been done.

PR #505, opened 2025-11-26
Dynamic Filtering on Notebooks

quirkyllama:
Implements Dynamic Filtering for Notebooks.
* Annotations in Source declarations for dimensions define the retrieval type for dimensions
* Annotations in the Notebook cell declare which dimensions are allowed to be filtered in the notebook.
* If no Notebook annotation, old UI is unchanged.
* See sdk/README.md for details on annotations and examples.

Other changes:
* Adds QueryInfo & SourceInfo to Notebook cells (for DimensionalFilter processing)
* Fixes bug in new Notebook cell fetching that used "fetch" instead of OpenAPI
* Doesn't wrap Renderer in expensive observer if no height is specified
* Uses arm64 runner to generate arm64 docker (10 min -> 3 min)
* Docker CI task can now output to a specified label (so we can publish test images not on the main version line)
* Fix to `shiki` dep- uses only 4 specified languages. Reduces # of build artifacts by 99% (only 5 output files vs 500) and build size by 75%. Dist is now 5MB instead of 21MB
* Moves most deps from `deps` to `peerDeps`. This reflects the actual config where our vite config has marked all of these things as external deps (not bundled).

https://www.loom.com/share/1d85613e10b9463dbffeb14e5113edef

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1594,7 +1594,7 @@ components:
             $ref: "#/components/schemas/NotebookCell"
         annotations:
           type: array
-          description: Array of annotations attached to the notebook
+          description: Array of file-level (##) annotations attached to the notebook
           items:
             type: string
 


packages/sdk/src/components/highlighter.ts
```diff
@@ -1,5 +1,12 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { createHighlighter } from "shiki";
+import { createHighlighterCore } from "shiki/core";
```
quirkyllama, line 2: We now use shiki/core and specific imports instead of importing *all* of shiki for language highlighting.

.github/workflows/docker-image.yml
```diff
@@ -2,13 +2,18 @@ name: Publish Docker Image to Docker Hub
...
+        required: false
+        type: string
 
 env:
   REGISTRY_IMAGE: ms2data/malloy-publisher
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.platform == 'linux/arm64' && 'ubuntu-24.04-arm' || 'ubuntu-latest' }}
```
quirkyllama, line 16: Running on an ARM platform makes generating the ARM Docker image 3x faster (3m vs 10m)

.github/workflows/docker-image.yml
```diff
@@ -2,13 +2,18 @@ name: Publish Docker Image to Docker Hub
 
 on:
   workflow_dispatch:
+    inputs:
```
quirkyllama, line 5: Add an optional input to specify the version tag. This way we can test images like "0.0.138-dimensionFilter-test" without publishing partial updates on the main version line.
This can thenbe use for the Credible Worker image tag to test new worker capabilities.

packages/sdk/src/components/filter/utils.ts
```diff
@@ -0,0 +1,354 @@
...
+ */
+export interface DimensionFilterAnnotation {
+   type: FilterType;
+}
+
+/**
+ * Get the names of all sources that are joined into a given source
+ * Looks for fields with kind: "join" in the SourceInfo
+ */
+export function getJoinedSources(sourceInfo: Malloy.SourceInfo): Set<string> {
```
quirkyllama, line 35: We need to know about joins, so we know which filter dimensions apply to which sources if the filter.dimension is on the joined source.

packages/sdk/src/hooks/index.ts
```diff
@@ -0,0 +1,31 @@
+// Filter hooks and types
+export {
```
quirkyllama, line 2: I'm exporting all this so SDK users could implement their own version of filtering outside Notebooks. Though we could restrict this to just the pieces needed to implement Retrieval

api-doc.yaml
```diff
@@ -1592,6 +1592,11 @@ components:
           description: Array of notebook cells containing raw markdown and code content
           items:
             $ref: "#/components/schemas/NotebookCell"
+        annotations:
```
kylenesbit, line 1595: These are just the file level annotations?  Not the field level annotations?  Maybe update the comment to make that clear?

quirkyllama: Fixed

Issue #415, opened 2025-10-16
Publisher: Add support for using environment variable values in Publisher config


kylenesbit:

jellysaini:
@kylenesbit,  Could you please confirm which environment variable I need to add.

PR #514, opened 2025-12-08
Publisher: Add support for using environment variable values in Publisher config #415

jellysaini:
cd packages/server
touch .env

# Added environment variable
PROJECT_NAME=malloy-samples

Here is an overview of this branch:

I created a .env file, but it is not placed at the repository root. The reason is that publisher.config.json itself is not located at the root; instead, we copy that config file and its contents as part of the setup. We intentionally do not copy the .env file due to security reasons.

malloy-samples.docker
# This is a hack to copy the publisher.config.json file in the server_root directory
# that will serve the malloy-samples directory
RUN cp packages/server/publisher.config.json /app/server_root/

In publisher.config.json, I referenced environment variables using placeholders like ${PROJECT_NAME}. Since .config.json files cannot directly consume environment variables, I introduced two helper functions: substituteEnvVars and processConfigValue. The purpose of these functions is to read values from publisher.config.json, resolve the placeholders using the .env file, and make those resolved values available to the project at runtime.

Additionally, I added fallback (failsafe) values directly in publisher.config.json. This is required because the .env file is not present on the server, and without these defaults, one of the build tests was failing. Adding these values ensures the build passes while still allowing environment-specific overrides where applicable.

### DIFF
--- a/packages/server/src/config.spec.ts
+++ b/packages/server/src/config.spec.ts
@@ -60,12 +60,8 @@ describe("Config Environment Variable Substitution", () => {
    });
 
    describe("Scenario 1: ${VAR} present in config but value not available", () => {
-      it("should document the bug: undefined variables cause string duplication", () => {
-         // This test documents the ACTUAL behavior which is a BUG
-         // The substituteEnvVars function has a parameter mismatch that causes
-         // undefined variables to be replaced with the entire original string
-
-         // Using a proper filesystem path format
+      it("should throw error when environment variable is not defined", () => {
+         // The correct behavior: throw an error when a required variable is missing
          const locationWithVar = "./path/${UNDEFINED_VAR}/end" as const;
 
          const config: PublisherConfig = {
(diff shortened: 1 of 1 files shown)


packages/server/publisher.config.json
```diff
@@ -2,7 +2,7 @@
   "frozenConfig": false,
   "projects": [
     {
-      "name": "malloy-samples",
+      "name": "${PROJECT_NAME:-malloy-samples}",
```
sagarswamirao: Can you share with me the ticket for this?

jellysaini: https://github.com/malloydata/publisher/issues/415

packages/server/src/config.ts
```diff
@@ -39,6 +39,49 @@ export type ProcessedPublisherConfig = {
    projects: ProcessedProject[];
 };
 
+function substituteEnvVars(value: string): string {
+   const envVarPattern = /\$\{([A-Z_][A-Z0-9_]*)(:-([^}]*))?\}/g;
```
sagarswamirao: Update it to remove :-DEFAULT_VALUE part

packages/server/src/config.ts
```diff
@@ -39,6 +39,49 @@ export type ProcessedPublisherConfig = {
...
+         }
+
+         throw new Error(
+            `Environment variable '\${${varName}}' is not set and no default value provided in configuration file`,
+         );
+      },
+   );
+}
+
+function processConfigValue(value: unknown): unknown {
```
sagarswamirao, line 59: Can u add tests for these below scenarios?
1) When ${VAR} is present in the config but value not available
2) When ${VAR} is present in the config and value is available
3) When ${VAR} is present in the config but not as a value, but as a key

packages/server/src/config.ts
```diff
@@ -39,6 +39,49 @@ export type ProcessedPublisherConfig = {
...
+         if (envValue !== undefined) {
+            return envValue;
+         }
+
+         if (defaultValue !== undefined) {
+            return defaultValue;
+         }
+
+         throw new Error(
+            `Environment variable '\${${varName}}' is not set and no default value provided in configuration file`,
```
sagarswamirao: Remove " and no default value provided"

packages/server/publisher.config.json
sagarswamirao: remove ${PROJECT_NAME} from the publisher.config.json

jellysaini:
@sagarswamirao, conflicts has been resolved in this one. Please review this one now.

packages/server/src/config.ts
```diff
@@ -40,6 +40,49 @@ export type ProcessedPublisherConfig = {
...
+   return value.replace(
+      envVarPattern,
+      (_match, varName, _unused, defaultValue) => {
+         const envValue = process.env[varName];
+
+         if (envValue !== undefined) {
+            return envValue;
+         }
+
+         if (defaultValue !== undefined) {
```
sagarswamirao: Why is this defualtValue piece of code still here? We don't support default values at the moment

sagarswamirao: Please clean up the code

packages/server/.env.example
sagarswamirao, line 1: Delete this file

packages/server/.env.test
sagarswamirao, line 1: Delete this file

Issue #487, opened 2025-11-13
Explorer: Show compiled SQL even if a query fails to run


kylenesbit:
When I run a query from explorer that fails (e.g., due to a timeout), I cannot see the compiled SQL.

But we can compile SQL without running a query => if i copy over to malloy file and click "show sql"

desired behavior: return the compiled sql even if the query fails due to a timeout. same with any database runtime errors => if malloy can compile, please show sql, even if the db complains about the sql.

jellysaini:
@kylenesbit , Request you to please review below PR's.

https://github.com/malloydata/malloy-explorer/pull/310

https://github.com/malloydata/publisher/pull/525

PR #525, opened 2025-12-11
 Explorer: Show compiled SQL even if a query fails to run #487

jellysaini:


packages/sdk/src/components/Model/SourcesExplorer.tsx
```diff
@@ -296,24 +296,24 @@ function SourceExplorerComponentInner({
                      }}
                   />
                </ResizableCollapsiblePanel>
-               {mutation.isError && mutation.error ? (
+               {/* {mutation.isError && mutation.error ? (
                   <Box sx={{ p: 2 }}>
                      <ApiErrorDisplay
```
sagarswamirao: Ensure to remove the code after the PR is approved

Issue #489, opened 2025-11-13
Publisher: Expose whether the Server is mutable via an API and use it to configure the ServerProvider's mutable prop


kylenesbit:
Frozen config makes API immutable.
Immutable option to all components to add/edit.
Publisher status: config is frozen?
Passed status to all components if yes.

jellysaini:
@sagarswamirao @kylenesbit , Request you to review below PR.

https://github.com/malloydata/publisher/pull/576

PR #530, opened 2025-12-15
 Publisher: Expose whether the Server is mutable via an API and use it to configure the ServerProvider's mutable prop #489

jellysaini:


packages/sdk/src/components/Project/Packages.tsx
```diff
@@ -39,10 +40,11 @@ const PackageMenu = ({
    const closeMenu = () => {
       setMenuAnchorEl(null);
    };
+   const { mutableConfig } = useIsMutable();
 
    return (
       <>
-         {mutable && (
+         {mutable && mutableConfig && (
```
sagarswamirao: Can u explain the difference between the variables mutable and mutableConfig?

api-doc.yaml
```diff
@@ -1493,6 +1520,21 @@ components:
           type: boolean
           description: Whether the server is fully initialized and ready to serve requests
 
+    FrozenStatus:
```
sagarswamirao: Not required. Add required fields to the "ServerStatus" class

api-doc.yaml
```diff
@@ -71,6 +71,33 @@ paths:
...
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/FrozenStatus"
+        "500":
+          description: Internal server error
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/Error"
```
sagarswamirao: @kylenesbit do you think we need a separate API for this?
We already have an /status API. Is it fine to add the frozenConfig field to the the ServerStatus obj?

packages/server/src/server.ts
```diff
@@ -223,6 +224,20 @@ app.get(`${API_PREFIX}/status`, async (_req, res) => {
...
+      const isFrozen = isPublisherConfigFrozen(SERVER_ROOT);
+      res.status(200).json({
+         mutable: !isFrozen,
+         frozenConfig: isFrozen,
+      });
+   } catch (error) {
+      logger.error("Error getting frozen status", { error });
+      const { json, status } = internalErrorToHttpError(error as Error);
+      res.status(status).json(json);
+   }
```
sagarswamirao: Remove this API impl

packages/server/src/server.ts
```diff
@@ -223,6 +224,20 @@ app.get(`${API_PREFIX}/status`, async (_req, res) => {
    }
 });
 
+app.get(`${API_PREFIX}/frozen-status`, async (_req, res) => {
+   try {
+      const isFrozen = isPublisherConfigFrozen(SERVER_ROOT);
+      res.status(200).json({
```
sagarswamirao: Why have frozenConfig and mutable fields. Just have one field frozenConfig

sagarswamirao (changes requested):

sagarswamirao:
We should ensue that all methods are upto date. If the namings are vague or they aren't upto date, we should ensure tosplit them out or update the methods so that they can be used, instead of having simialr logic at multiple places.

packages/server/src/service/project_store.ts
```diff
@@ -512,6 +512,7 @@ export class ProjectStore {
          timestamp: Date.now(),
          projects: [] as Array<components["schemas"]["Project"]>,
          initialized: this.isInitialized,
+         frozenConfig: isPublisherConfigFrozen(this.serverRootPath),
```
kylenesbit, line 515: Don't we need to add this field to the API spec?

jellysaini: @kylenesbit, this is not required in the API spec at the moment, and the existing parameters are not listed either. If you’d like, I can include the older parameters too.

sagarswamirao: This has to be present in the API Spec. Please add it to the API Spec Jelly

jellysaini: @sagarswamirao, I have added the value to the Components schema — I had missed it earlier. We don’t need to include this value in the API parameter because it’s already being set through the config file, and we don’t want users to provide it manually.

sagarswamirao (approved):

packages/sdk/src/components/ServerProvider.tsx
```diff
@@ -80,20 +89,54 @@ export const ServerProvider: React.FC<ServerProviderProps> = ({
...
+      let isMounted = true;
+
+      const fetchStatus = async () => {
+         try {
+            const response = await apiClients.publisher.getStatus();
+
+            if (isMounted) {
+               const data = response.data as { frozenConfig?: boolean };
+               const frozenConfig = data?.frozenConfig;
+               setMutable(!frozenConfig);
```
sagarswamirao, line 115: Perform the following check and decide the final value:
Image

PR #549, opened 2026-01-02
Added GCS in duckdb attached databases

manojdeep14:


api-doc.yaml
```diff
@@ -1854,6 +1854,23 @@ components:
...
+        keyId:
+          type: string
+          description: GCS HMAC access key ID
+        secret:
+          type: string
+          description: GCS HMAC secret key
+      required:
+        - keyId
+        - secret
+      example:
```
kylenesbit: Can you drop the examples?  We're not using throughout the API.

manojdeep14: Dropped in the recent commit, Kyle!

PR #570, opened 2026-01-08
Enhance project and service tests for better stability and cross-platform compatibility

sagarswamirao:
- Improved cleanup logic in package tests to handle EBUSY errors on Windows.
- Normalized path handling in package service for cross-platform compatibility.
- Refactored project store tests for better readability and consistency, including timeout adjustments.
- Added publisher.json manifest file creation in multiple test cases to ensure proper project setup.
- Updated temporary directory creation in project store to use a hash for shorter paths, addressing Windows MAX_PATH limitations.

### DIFF
--- a/packages/server/src/service/package.spec.ts
+++ b/packages/server/src/service/package.spec.ts
@@ -33,16 +33,32 @@ describe("service/package", () => {
    afterEach(async () => {
       sinon.restore();
       // On Windows, DuckDB connections may still have file handles open,
-      // causing EBUSY errors. Add a small delay and handle cleanup gracefully.
-      await new Promise((resolve) => setTimeout(resolve, 100));
-      try {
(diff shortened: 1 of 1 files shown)


packages/server/src/service/package.spec.ts
```diff
@@ -32,7 +32,18 @@ describe("service/package", () => {
 
    afterEach(async () => {
       sinon.restore();
-      await fs.rm(testPackageDirectory, { recursive: true });
+      // On Windows, DuckDB connections may still have file handles open,
+      // causing EBUSY errors. Add a small delay and handle cleanup gracefully.
+      await new Promise((resolve) => setTimeout(resolve, 100));
```
kylenesbit: Why do we need a delay?  That's kind of gross -- can we not retry here if we get an error on the first pass?

kylenesbit (approved):

PR #571, opened 2026-01-08
Refactor DuckDB Snowflake installation process and update CI workflow

sagarswamirao:
- Replaced the DuckDB Snowflake installation command in package.json with a Node.js script for improved reliability and cross-platform compatibility.
- Removed the old CI build steps from the GitHub Actions workflow and added a new job to build a Docker image for CI, including caching and artifact upload.
- Introduced cross-platform testing in the CI workflow to ensure compatibility across different operating systems.
- Implement project loading mutex to prevent concurrent project creation (Flakey Test: http://github.com/malloydata/publisher/actions/runs/20829225405/job/59838730020) - this also fixes the scenario where publisher at times doesn't initialize successfully at the startup

Note: Caching Bun dependencies saves only about 30–35 seconds of installation time on Windows, but it saves a lot more time on Ubuntu and macOS.

### DIFF
--- a/.github/workflows/cross-platform-tests.yml
+++ b/.github/workflows/cross-platform-tests.yml
@@ -33,13 +33,8 @@ jobs:
 
       - name: Install dependencies
         run: |
-          if [ "${{ steps.bun-cache.outputs.cache-hit }}" == "true" ]; then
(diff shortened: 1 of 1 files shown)


manojdeep14 (approved):
LGTM! Thank you!

PR #581, opened 2026-01-13
fix: Include imported sources in sourceInfos for model explorer (#569)

### LINKED ISSUE #569: Publisher: web-ui not redering both .malloy that exist and do run via VS Code Extention
### What happens?

When running Publisher via Docker image, for a project that consists of 2 .malloy files, the experience of the web-ui showing the .malloy files differs. In VS Code both .malloy files work as expected. 

But not on Publisher web-ui.

Image

Image

Image

#### Project
Project exists of:
* publisher.json 
* a source file for initial 'table definitions': `sources_psql.malloy`
* a second file `simple_calc.malloy` with simple calculations that imports source file `sources_psql.malloy`
* connecting to a postgresql db. But same issue is with a DuckDB serving local files in .data under a similar package

Expected outcome:
`simple_calc.malloy` would also be visible on Publisher web-ui `sources_psql.malloy`. Just like it shows to be working on VS Code.

### To Reproduce

- OS Windows build 26100.7462
- Docker Engine v29.1.3
- Malloy VS Code extension 0.2.1764971905
- Running Malloy Publisher via docker  
- image: ms2data/malloy-publisher:0.0.146
-sha256:608a54f281af5888143b874d15106c5cd25d57e36f8961448d62bdc85c505583

[sources_psql.zip](https://github.com/user-attachments/files/24472849/sources_psql.zip)

#### Steps
1. Developed in VS Code and validated
2. `docker run -p 4000:4000 -p 4040:4040 -v ./publisher.config.json:/publisher/publisher.config.json -v ./packages:/publisher/my-analytics ms2data/malloy-publisher`
3. Browsing to `localhost:4000` > Project  = 'default'  > Package = 'psql' > Select the .malloy file
4. Seeing content as in VS Code for `http://localhost:4000/default/psql/sources_psql.malloy` (is OK), but not redendering when browsing to the other .malloy file `http://localhost:4000/default/psql/sources_psql.malloy` (NOK)

Files

### OS:

Windows 11 Home (x64)

### Malloy Client:

Windows 11 Home (x64)

### Malloy Client Version:

0.2.1764971905

### Database Connection:

PostgreSQL

jswir:
When a Malloy model imports sources from another file, modelDefToModelInfo() only returns sources defined in the current file, not imported ones. This caused the Publisher web UI to show a blank page for models that import sources from other files.

The fix loads sourceInfos from imported models during Model.create() by:
1. Iterating through modelDef.imports
2. Loading each imported model and getting its sourceInfos
3. Merging imported sourceInfos with locally-defined ones

Fixes #569

### DIFF
--- a/packages/server/src/service/model.ts
+++ b/packages/server/src/service/model.ts
@@ -570,17 +570,24 @@ export class Model {
          const isDuckDB =
             connection instanceof DuckDBConnection ||
             connection.constructor.name === "DuckDBConnection";
-         if (isDuckDB) {
+         if (
+            isDuckDB &&
+            (connection as unknown as { workingDirectory?: string })
+               .workingDirectory !== workingDirectory
+         ) {
             // Create a new DuckDB connection with the model's directory as working directory.
-            // Using :memory: for model-specific connections to avoid conflicts.
+            // IMPORTANT: Reuse the same databasePath to share secrets and attachments via DuckDB's activeDBs cache.
+            const databasePath =
+               (connection as unknown as { databasePath?: string })
+                  .databasePath || ":memory:";
             const modelDuckDBConnection = new DuckDBConnection(
                name,
-               ":memory:",
+               databasePath,
                workingDirectory,
             );
             modelConnections.set(name, modelDuckDBConnection);
          } else {
-            // Keep non-DuckDB connections as it is
+            // Keep connection as-is (same workingDirectory or non-DuckDB)
             modelConnections.set(name, connection);
          }
       }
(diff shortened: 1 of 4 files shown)


packages/server/src/service/model.ts
```diff
@@ -154,6 +203,7 @@ export class Model {
             modelDef,
             sources,
             queries,
+            sourceInfos.length > 0 ? sourceInfos : undefined,
```
sagarswamirao, line 206: We need to use [] to be consistent. instead of undefined. I will fix it in my PR that I merge

PR #586, opened 2026-01-15
Removed logic to instantiate attached databases duckdb instance for each package

manojdeep14:
1) Removed logic to instantiate duckdb for each package. Now instantiating only in memory for each package to access the local data
2) Added a check not to have the same connection name and attached database name, as it conflicts and skips attaching
3) Implemented delete logic for project level duckdb connection

### DIFF
--- a/packages/app/package.json
+++ b/packages/app/package.json
@@ -1,7 +1,7 @@
 {
    "name": "@malloy-publisher/app",
    "description": "Malloy Publisher App",
-   "version": "0.0.157",
+   "version": "0.0.159",
    "type": "module",
    "main": "dist/index.cjs.js",
    "module": "dist/index.es.js",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,7 +1,7 @@
 {
    "name": "@malloy-publisher/sdk",
    "description": "Malloy Publisher SDK",
-   "version": "0.0.157",
+   "version": "0.0.159",
    "type": "module",
    "main": "dist/index.cjs.js",
    "module": "dist/index.es.js",
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -1,7 +1,7 @@
 {
    "name": "@malloy-publisher/server",
    "description": "Malloy Publisher Server",
-   "version": "0.0.157",
+   "version": "0.0.159",
    "main": "dist/server.js",
    "bin": {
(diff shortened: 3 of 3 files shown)


kylenesbit (approved):

packages/server/src/service/model.ts
```diff
@@ -518,36 +518,14 @@ export class Model {
...
-            );
-            modelConnections.set(name, modelDuckDBConnection);
-         } else {
-            // Keep connection as-is (same workingDirectory or non-DuckDB)
-            modelConnections.set(name, connection);
-         }
-      }
+      const duckdbConnection = connections.get("duckdb") as DuckDBConnection;
+      await duckdbConnection.runSQL(
+         `SET FILE_SEARCH_PATH='${workingDirectory}';`,
```
sagarswamirao, line 570: https://duckdb.org/docs/stable/configuration/overview#local-configuration-options

sagarswamirao: FILE_SEARCH_PATH is a comma separated string. 
Can we traverse all the paths when the package is loaded, and add use this command to set things? the order of resolution is from the left -> right of the stringified list.

packages/server/src/service/package.ts
```diff
@@ -102,13 +100,12 @@ export class Package {
          const connections = new Map<string, Connection>(projectConnections);
 
          // Add a duckdb connection for the package.
-         const duckdbConnections = await createPackageDuckDBConnections(
-            packageConnections,
+         const duckdbConnection = new DuckDBConnection(
+            "duckdb",
+            ":memory:",
             packagePath,
```
sagarswamirao, line 105: We can pass the comma separated stringified list here, and skip the doing the thing which we do at the top

Whenever a package is loaded, we need to traverse and find all the directories, and pass it in the line 106. Ensure to pass in the packagePath - as that is the roor directory

sagarswamirao (changes requested):

packages/server/src/service/model.ts
manojdeep14: Create a ticket for this and will pick this one soon: https://github.com/malloydata/publisher/issues/593

packages/server/src/service/package.ts
manojdeep14: Created a separate issue: https://github.com/malloydata/publisher/issues/593

PR #590, opened 2026-01-15
Add k6 testing framework with initial setup and client generation

sagarswamirao:
- Introduced a new k6 testing suite for the Malloy Publisher API, including various performance and load tests.
- Implemented a client factory to streamline the creation of API clients for different resources.
- Added a README file detailing installation instructions, available tests, and environment variables.
- Included TypeScript configuration and package management files for better development experience.
- Enhanced existing tests to filter available packages based on whitelisting and credentials.


packages/server/k6-tests/clients/connections-test.ts
```diff
@@ -0,0 +1,134 @@
+/**
```
kylenesbit, line 1: @quirkyllama insists we should not check these clients in.  Is it possible to make this part of the build process?

sagarswamirao: Yes we can. 
I'll remove the clients for now, and later set up and dedicated github action for this.

Issue #444, opened 2025-10-27
Publisher: Add support to create, update, delete Publisher projects, packages, connections, ... to Malloy CLI


kylenesbit:

sagarswamirao:
I have added the comments to the PR. Please resolve them.

PR #606, opened 2026-01-30
 Publisher: Add support to create, update, delete Publisher projects, packages, connections, ... to Malloy CLI #444

jellysaini:


sagarswamirao (changes requested):
1) Add Tests
2) Projects, Connections, Packages  - can you move the logic into respective files?
3) Can you change from console.log -> logger.info, logger.error, logger.warn? (There is a logger class already present within the publisher utils.
4) Gitub Action (I will work on it once the code is merged) - nothing from your side

packages/cli/src/utils/logger.ts
```diff
@@ -0,0 +1,119 @@
...
+import { AxiosError } from "axios";
+
+// Determine log level from environment variable
+// Valid levels: error, warn, info, verbose, debug, silly
+// Default: 'info' (less verbose than server for CLI)
+const VALID_LOG_LEVELS = ["error", "warn", "info", "verbose", "debug", "silly"];
+
+const getLogLevel = (): string => {
+  if (process.env.LOG_LEVEL) {
+    const logLevel = process.env.LOG_LEVEL.toLowerCase();
```
sagarswamirao, line 11: Remove this process.env variable

Can you suppress the URL.parse warning?

I still see at few places that we are using console.log
Can you change them to logger?

sagarswamirao (changes requested):

packages/cli/src/utils/logger.ts
jellysaini: image

I copied this logic from server/src/logger.ts to keep the implementation consistent. Please let me know whether I should keep it as-is or remove it.

jellysaini: @sagarswamirao, all other comments have been addressed. Request you to please review the updates.

PR #626, opened 2026-02-25
Publisher APP: Rewrite to reduce build time, bundle size, and add a structure for the components folder #490 #539 Part3: Router new design

jellysaini:


sagarswamirao:
There is already one fallback page.

PR #668, opened 2026-04-06
Refactor ConnectionController to utilize schema caching and improve e…

sagarswamirao:
…rror handling

- Introduced SCHEMA_CACHE_TTL_MS to manage schema cache expiration.
- Updated fetchTableSchema and fetchSchemaForSQLStruct methods to use fetchSchemaForTables with refreshTimestamp for better performance.
- Enhanced error handling to throw ConnectionError for specific cases.
- Increased maximum connections in createProjectConnections from 5 to 10 for improved connection management.

ENV: SCHEMA_CACHE_TTL_MS


sagarswamirao:
@kylenesbit 
FYI
Malloy-DB-Snowflake issue is fixed now. We don't need to do schema caching at Publisher. If we were to do it in the future, we can consider this PR as the reference. Note: we might need to add forceRefresh param

PR #693, opened 2026-04-22
Update DuckDB (1.4.3 -> 1.4.4) and Bun (to 1.3.13) dependencies and Dockerfile for improved compatibility

sagarswamirao:
- Bump duckdb version from 1.4.3 to 1.4.4 across bun.lock, package.json, and server package.json.
- Update Dockerfile to use bun version 1.3.13 for base and builder stages.
- Adjust GitHub workflows to reference the new bun version for consistent environment setup.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,9 @@ on:
     branches: [main]
   pull_request:
 
+env:
+  BUN_VERSION: 1.3.13
+
 # self contained tests
 jobs:
   lint:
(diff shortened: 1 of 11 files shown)


sagarswamirao:
@kylenesbit Bumping up to fix the NAPI Error
Fixed a crash during process exit when native N-API modules (seen during graceful shutdown) - https://bun.com/blog/bun-v1.3.13

image

sagarswamirao:
The Bun 1.3 / 1.3.x DuckDB fixes that came up:

Bun 1.2.2 (predates 1.3, but listed for context): "the DuckDB native module now works correctly in Bun, as previously it would crash due to a null-returning native module in napi_register_module_v1." — fixed a crash at module load where DuckDB's N-API init returned null.

Bun 1.3.13 (current): "Fixed a crash during process exit when native N-API modules (sqlite3, duckdb, kuzu, node-llama-cpp) wrap parent objects before children — finalizers now run in LIFO order matching Node.js." — fixed a process-exit segfault with DuckDB, SQLite, and a couple of other native addons. Finalizer ordering was wrong; now matches Node.

Bun 1.3 (major release, general N-API rewrite, not DuckDB-specific but affects all native addons including DuckDB): napi_async_work creation/cancellation rewrite, napi_create_buffer uses uninitialized memory, ArrayBuffer handling rewritten, 98%+ of Node's N-API tests now pass. This is the change that touched everything DuckDB depends on when running callbacks.

Bun 1.3.8: fixed napi_typeof returning wrong type for wrapped callbacks — native addons like encore.dev panicked with "expect Function, got: Object" when used inside AsyncLocalStorage.run(). Not DuckDB-specific but same class of bug.

PR #696, opened 2026-04-22
Update Project References to Environments Across Codebase

sagarswamirao:
- Marked existing project-related endpoints as deprecated in the API documentation.
- Added new environment endpoints for lifecycle management of data modeling environments.
- Updated server routes to handle both project and environment requests, allowing for backward compatibility.
- Enhanced descriptions for deprecated endpoints to guide users towards the new environment endpoints.


packages/server/src/controller/connection.controller.ts
```diff
@@ -135,12 +138,16 @@ export class ConnectionController {
...
+               ) => Promise<TableSourceDef | string | undefined>;
             }
          ).fetchTableSchema(tableKey, tablePath);
          if (!source) {
             throw new ConnectionError(`Table ${tablePath} not found`);
          }
+         // BigQueryConnection returns `error.message` as a string on failure instead of throwing.
+         if (typeof source === "string") {
+            throw new ConnectionError(source);
+         }
```
sagarswamirao, line 203: @kylenesbit, seems like this was removed during the refactor. 
This is needed for BigQuery connections. BQ returns dataset/table not found as string with 200 (doesn't throw an error)

packages/server/src/controller/connection.controller.ts
```diff
@@ -167,33 +174,44 @@ export class ConnectionController {
...
-         connectionName,
+      // Prefer the in-memory API connection (which was materialized by the
+      // environment on load and carries `attributes`). The DB row stores the
+      // raw config and FK columns, which aren't the ApiConnection shape.
+      const environment = await this.environmentStore.getEnvironment(
+         environmentName,
+         false,
       );
-      return dbConnection;
+      return environment.getApiConnection(connectionName);
```
sagarswamirao, line 243: await this.connectionService.getConnection doesn't return the Connection of type BaseConnection. In DB-Malloy-Publisher we have checks for dialect, etc. But the connection being returned here doesn't have those fields. Had to add it back.

sagarswamirao: @kylenesbit

packages/server/src/controller/connection.controller.ts
```diff
@@ -223,49 +244,55 @@ export class ConnectionController {
...
+               }) => Promise<unknown>;
+            }
+         ).fetchSelectSchema({
+            connection: connectionName,
+            selectStr: sqlStatement,
+         });
+         // BigQueryConnection returns `error.message` as a string on failure instead of throwing.
+         if (typeof schema === "string") {
+            throw new ConnectionError(schema);
+         }
```
sagarswamirao, line 339: Handling BQ Error

kylenesbit (approved):

mlennie (commented):
@sagarswamirao 
These items seemed to have popped up... Not sure if all are relevant/necessary for this PR.

  Critical                                                                                                                                                                                                           
   
  1. Existing DuckDB databases hard-crash on startup, with a data-loss foot-gun in the recovery path.                                                                                                                
  isInitialized looks for environments, finds none on a legacy DB, so CREATE TABLE IF NOT EXISTS runs without dropping old tables — environments is created empty while
  packages/connections/materializations/build_manifests keep their old project_id columns. The first WHERE environment_id = ? query crashes. The operator's natural recovery (--init) calls dropAllTables and        
  destroys all child tables along with their data.
                                                                                                                                                                                                                     
  Fix: Add a one-time migration in initializeSchema that runs before the CREATE TABLE block: detect a legacy DB by probing for the projects table, and if found, run ALTER TABLE projects RENAME TO environments plus
   ALTER TABLE <child> RENAME COLUMN project_id TO environment_id for each of packages, connections, materializations, build_manifests. Wrap in a transaction. Add a unit test that seeds a legacy schema and asserts
   a clean upgrade with row counts preserved.                                                                                                                                                                        
                  
  ---                                                                                                                                                                                                                
  2. publisher.config.json files with the old "projects" key silently parse as empty.
  config.ts only reads the environments key. Existing on-disk configs produce a zero-warning empty environment list, breaking every deployment whose config wasn't manually migrated — including every service/      
  worker in this workspace.                                                                                                                                                                                    
                                                                                                                                                                                                                     
  Fix: In the config parser, if environments is absent and projects is present, read from projects and emit logger.warn("publisher.config.json uses deprecated 'projects' key; rename to 'environments'"). Add a unit
   test covering both keys and a test asserting the deprecation warning fires.                                                                                                                                       
                  
  ---                                                                                                                                                                                                                
  High            

  3. The 1,139-line legacy /projects/... compatibility layer has zero test coverage.
  Nothing in tests/ exercises any /api/v0/projects/... path or the response remappers. The whole backward-compat claim is unverified.
                                                                                                                                                                                                                     
  Fix: Add an integration test file tests/integration/legacy_routes.integration.spec.ts that boots the server and hits at least one route per legacy URL group (status, projects CRUD, packages, connections, models,
   notebooks, databases, queries, materializations, manifest). Assert: (a) the route returns 200, (b) GET /status response contains projects not environments, (c) materialization responses contain projectId not   
  environmentId.                                                                                                                                                                                                     
                  
  ---
  4. api-doc-old.yaml is an orphan and the PR description's "marked as deprecated" claim is inaccurate.
  The 3,600-line file is read by no codegen script. The new api-doc.yaml deletes /projects paths entirely rather than marking them deprecated.                                                                       
                                                                                                                                              
  Fix: Pick one of the two paths and commit to it. Either: (a) delete api-doc-old.yaml and rewrite the PR description to say "removed /projects from the spec; kept legacy URLs in server-old.ts for runtime backward
   compatibility," or (b) restore the /projects path entries in api-doc.yaml with deprecated: true and description: "Use /environments/... instead." so generated SDKs surface the deprecation to clients. Option (b)
   is the better long-term answer.                                                                                                                                                                                   
                                                                                                                                                                                                                     
  ---             
  Medium
        
  5. MCP URI scheme rename has no compatibility layer.
  The scheme changed from malloy://project/{name} to malloy://environment/{name}. server-old.ts only handles REST. Any MCP client with stored/bookmarked URIs (agent context, saved resources) breaks silently.      
                                                                                                                                                                                                                     
  Fix: In mcp/resources/environment_resource.ts, register a second resource template malloy://project/{environmentName} alongside the new one, both pointing at the same handler. Do the same for child resource URIs
   (package, model, notebook, source, view, query). Add an MCP integration test that resolves both URI schemes against the same backing data and asserts identical responses.                                        
                                                                                                                                                                                                                     
  ---             
  Low
     
  6. server-old.ts is a misleading filename for actively-imported code.
  Reads as a stashed snapshot but is wired into server.ts:39.                                                                                                                                                        
                                                                                                                                                                                                                     
  Fix: Rename packages/server/src/server-old.ts → packages/server/src/legacy_routes.ts. Update the import in server.ts. The file's existing header comment already describes its purpose accurately.                 
                                                                                                                                                                                                                     
  ---                                                                                                                                                                                                                
  7. Branch contains DCO remediation commits and merges from main.
  Multiple "DCO Remediation Commit" entries and several merge commits clutter the history.
                                                                                          
  Fix: Use squash-merge when landing the PR, or interactive-rebase the branch into a single commit (Rename Project → Environment across REST/SDK/CLI/storage) before merging.                                        
                                                                                                                                                                                                                     
  ---                                                                                                                                                                                                                
  8. CLI projects subcommand was deleted with no alias.                                                                                                                                                              
  Users who run malloy-publisher list project or have scripts using the old subcommand get an unknown-type error instead of a deprecation message.
                                                                                                                                                  
  Fix: In packages/cli/src/index.ts, add project and projects as hidden aliases that route to the corresponding environment command and print console.warn("'project' is deprecated; use 'environment' instead.")    
  once per invocation. Add a CLI test asserting both alias paths work and emit the warning.                                                                                                                          
                                                                                                                                                                                                                     
  ---                                                                                                                                                                                                                
  9. Local publisher.config.json is --skip-worktree'd in this workspace; reviewers won't pull the rename automatically.
  Per the workspace CLAUDE.md, the file is skipped locally so DuckDB-only customizations don't show in git status. The PR changes the same file's top-level key.                                                     
                                                                                                                                                                
  Fix: Add a note to the PR description: "Reviewers / local devs with --skip-worktree set on packages/server/publisher.config.json: run git update-index --no-skip-worktree packages/server/publisher.config.json,   
  pull, update the local key from projects to environments, then re-skip with git update-index --skip-worktree ...."

sagarswamirao:
Thanks for taking a look at it @mlennie. I've made the following changes:

- Added smoke tests for the legacy /api/v0/projects/... endpoints (projects CRUD, packages, connections, models, notebooks, databases, queries, materializations, and manifest) in a separate file so it can be deleted in one motion when legacy support is dropped.
- Added mapping from projects → environments for publisher.config.json with a deprecation warning, so existing on-disk configs no longer parse silently as empty.
- Added a startup migration for legacy DuckDB databases: detects the pre-rename projects table, drops the legacy parent and child tables with a warning, and lets the new schema be created cleanly - preventing the environment_id query crash on upgrade.
- Deleted the orphan api-doc-old.yaml (no codegen referenced it) and clarified that legacy back-compat is runtime-only via server-old.ts.
- Removed the legacy /status route and the now-dead remapStatusResponse helper from server-old.ts; the single /status handler in server.ts now serves both old and new clients.
- Tagged all legacy code paths and tests with // TODO: Remove this during projects cleanup so the eventual removal is mechanical.

As CLI isn't on the NPM registry yet, no one is using it. So I think we can skip adding project alias in the cli

PR #705, opened 2026-04-24
Added deprecated options query parameter in api-doc.yaml for backward compatibility.

sagarswamirao:
Updated server.ts to support options from both body and query parameters, improving usability while maintaining legacy support.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -738,6 +738,7 @@ paths:
         - name: options
           in: query
           description: Options (deprecated; pass in body instead)
+          deprecated: true
           required: false
           schema:
             type: string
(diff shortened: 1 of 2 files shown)


api-doc.yaml
```diff
@@ -735,6 +735,12 @@ paths:
           required: true
           schema:
             $ref: "#/components/schemas/IdentifierPattern"
+        - name: options
+          in: query
+          description: Options (deprecated; pass in body instead)
```
mhwilder, line 740: Can you make this as deprecated in the spec

mhwilder (approved):
Left one comment but otherwise good.

api-doc.yaml
sagarswamirao: Good one. Sure :)

PR #724, opened 2026-05-10
Serve in-package HTML data apps from Publisher

jswir:
## Summary

Lets package authors drop an `index.html` (plus CSS/JS/assets) alongside their `.malloy` files; Publisher serves it at `/environments/<env>/packages/<pkg>/<file>`. Pages render inside the Publisher SPA chrome by default (like notebooks) and can also be embedded as iframes into any page with auto-resize via `postMessage`. With `--watch-env`, edits to your source repo live-reload the page over SSE — and `.malloy` edits now actually recompile the package. Designed for zero-build, AI-generated data apps.

## How to read this PR

Four commits, in dependency order:

1. **serve & view in-package HTML data apps** — static serving, `/pages`, `PageViewer`, runtime. Production-facing and self-contained.
2. **dev live-reload via watch-mode events** — `/events` SSE, `--watch-env`, in-place symlink mount.
3. **recompile the changed package on watch events** — the actual watch-mode-for-models fix.
4. **only signal live-reload after a successful recompile** — don't bounce open pages on a transient compile error.

Commits 1–2 are the feature; 3–4 fix watch mode (see "Watch-mode fix" below).

## What ships

**API surface (`api-doc.yaml`) — consistent with house conventions**

* `GET .../packages/{pkg}/pages` (`list-pages`) returns a **bare `Page[]`** — a named `$ref` schema (`resource` / `packageName` / `path` / `title`), matching `list-models` / `list-notebooks` / `list-databases`; full `401/404/500/501/503` set; its own `pages` tag.
* `GET .../packages/{pkg}/events` (`stream-package-events`) — the SSE live-reload stream, documented as `text/event-stream` (the spec's first SSE endpoint; consumed via `EventSource`, not the generated client).

**Server (`packages/server/src/server.ts`)**

* Static-file route `/environments/:env/packages/:pkg/*` with `index.html` directory-index fallback and `fs.realpath()` containment (path traversal + symlink-escape → 403). Owns its prefix — never falls through to the SPA on a miss.
* `publisher.json` blocked at **any depth** (basename match), not just the package root.
* Service errors mapped through `internalErrorToHttpError` (BadRequest→400, ServiceUnavailable→503) instead of collapsing to 404; a genuine missing file is still a real 404.
* `Content-Security-Policy: frame-ancestors` set **only on HTML responses**, and **configurable** via `PUBLISHER_FRAME_ANCESTORS` (default `*`).
* `/sdk/publisher.js` serves the runtime helper from disk.
* `/events` SSE: env/package names validated with the shared `assertSafePackageName` (same allowlist as every other entry point — no bespoke regex), existence-checked, then reports `event: mode` (`enabled|disabled`) and fans out `changed`. Does not auto-start the watcher.
* `--watch-env <name>` activates watch mode after env load. Multiple envs **fail fast** (watch the first, log the rest) rather than silently keeping only the last.
* Package-root redirect inserts the trailing slash before the query string; friendly 404 placeholder when the SPA bundle isn't built.

**Watch-mode fix (`watch-mode.controller.ts`)**

Watch mode previously detected edits but **never recompiled** — it ran an env-level metadata "reload" (`addEnvironment` → `Environment.update()`) that never touched compiled packages, so `.malloy` edits never showed up (matches the bug reported in #720-era testing). Now the watcher reloads the **specific package** the changed file belongs to via `Environment.getPackage(name, reload=true)`, which re-runs `Package.create` and replaces the cached entry. Scoped to watch mode by construction (only the watcher calls it). For model edits the SSE `changed` event is emitted only after a **successful** recompile, so a transient syntax error doesn't bounce open tabs into a compile error. Verified end-to-end: editing a view's `count()` → `count() + 1000` flips the query result, and a newly added view resolves.

**In-place symlink mount (`environment_store.ts`)**

With `--watch-env`, local-dir packages are mounted as **symlinks** under `publisher_data/<env>/<pkg>/` instead of copied, so the watcher sees source edits and static files are served live. Production copy semantics are unchanged when the flag is absent; remote (git/GCS/S3) sources always copy. `lstat`-aware cleanup distinguishes symlinks from real dirs so `rm` never traverses into source. Watch path uses the env's real `environmentPath` (Publisher mounts to `<root>/publisher_data/<env>/`).

**Runtime (`packages/server/src/runtime/publisher.js`)**

Hand-authored vanilla JS, no dependencies, no build step. `Publisher.query / queryFull / embed / context / setToken`; `credentials: "include"` by default. Auto-subscribes to `/events` for live reload when served from a Publisher origin. When inside an iframe, posts size updates to the parent — measuring the max bottom edge of body's children (not `body.scrollHeight`, which `min-height: 100vh` inflates), so dashboards shrink back instead of ratcheting. The `publisher:resize` protocol is defined once in `sdk/utils/pageEmbed.ts` and kept in sync here via a cross-reference comment.

**SDK (`Package.tsx`, new `PageViewer/`, `pageEmbed.ts`)**

* New "Pages" section in the package view, **above Semantic Models**, with a browser-window icon and the page `<title>` as label.
* Embedded `<PageViewer>` (default): row click routes to `/<env>/<pkg>/pages/<file>` and iframes the page inside Publisher chrome with `postMessage` auto-sizing; plus an always-available "open standalone" affordance.
* `PageViewer` renders an inline error state instead of throwing on a malformed resource URI.
* Standalone URL uses `page.resource` joined with the shared `serverBaseUrl()` helper (one definition of the `/api/v0`-stripping, used by both `Package` and `PageViewer`), so it works when the SPA origin differs from the Publisher origin.
* Pages list uses the typed generated client (`apiClients.pages.listPages`), inheriting `baseURL` / Bearer from `<ServerProvider>`; hides gracefully on older Publishers without `/pages`.

**App (`ModelPage.tsx`)** — the existing `:env/:pkg/*` catch-all dispatches a `pages/` prefix to `<PageViewer>`, mirroring the extension dispatch for `.malloy` / `.malloynb`.

**Example (`examples/html-data-app/`)** — `carriers.malloy` (real source with views) + `carriers.parquet`, a Chart.js `index.html` (dropdown filters, KPI tiles, charts, table), `embed-test.html` demonstrating `Publisher.embed()`, and a README walking the `--watch-env` dev loop.

**Tests** — `tests/integration/html_pages/` runs against the real Express app: serving, realpath containment (traversal + symlink-escape → 403), root + subdirectory manifest block, the bare-array `/pages` shape, and `/events` name-validation + SSE handshake. `pageEmbed.spec.ts` unit-tests the resize-protocol type guard and origin derivation.

## Areas that want a careful look

1. **In-place symlink mount** — only with `--watch-env`, only `local_dir` packages; `lstat`-aware cleanup. On Windows it creates a directory **junction** (no elevation needed) rather than a symlink, and **falls back to a copy** (env still serves, live-reload disabled) if even that is denied — still worth a Windows smoke test. The production copy path is untouched when the flag is absent.
2. **Recompile path** — drives the existing `getPackage(reload=true)` reload mechanism per changed package; the env-level metadata-update path is left alone.
3. **Auth deferred** — Publisher stays auth-less by design; whatever fronts it owns authentication. Embed auth flows are a follow-up.

## Test plan

* `bun test:unit` (packages/server) → 735 pass. New `html_pages` integration suite → 12 pass. `pageEmbed` unit → 8 pass. Server + SDK typecheck clean.
* Manual, single prod server (`bun run packages/server/dist/server.mjs --watch-env demo`, one terminal):
  * static serving; `publisher.json` root + subdir → 404; traversal / symlink-escape → 403; missing → 404.
  * `GET …/pages` → bare JSON array of `{ resource, packageName, path, title }`.
  * `GET …/events` → `text/event-stream`, emits `hello` + `mode`; illegal name → 400, unknown package → 404.
  * Edit a `.malloy` → log shows `Watch: recompiled package …` → a query reflects the change (recompile fix). Edit `.html`/`.css` → page live-reloads.
  * SPA: Pages section renders above Semantic Models; row → `PageViewer` (one `/pages` request); "open standalone" → raw URL.

## Known open work / accepted trade-offs

* **Compile error during a watch reload evicts the package** → it 404s until the next good save. Mitigated by success-gated live-reload (open pages don't auto-bounce into the error). Keeping the last-good compile on a *reload* failure is a deliberately-deferred follow-up (it touches the shared package-load path used by all reload callers).
* **Multi-env watch** — `--watch-env` accepts a list but watches one env at a time (fails fast + logs the rest). Multi-watcher support is a follow-up.
* **Embed auth flows** (cookie / proxy / signed-token) are not implemented.
* **`frame-ancestors`** — now HTML-only and `PUBLISHER_FRAME_ANCESTORS`-configurable; a per-environment allowlist could refine it further.
* **SDK/app `dist/`** are gitignored build artifacts — rebuild locally (`bun run build:sdk && bun run build:server`) to see the new UI in `bun run start`; verified working this branch.


jswir:
### Code review

Found 1 issue:

1. `listPackagePages` walks the package directory with `path.join` + `fs.readdir` and never resolves symlinks via `fsp.realpath`, then opens up to 4KB of each `.html`/`.htm` entry to extract a `<title>`. A package containing a symlink that points outside its root (e.g. `leak -> /etc/passwd`) will be followed (`Dirent.isFile()` returns `true` through symlinks), the target's first 4KB will be read (and may surface as the listed `title`), and the symlink's relative path will be returned in the API response. The companion function `serveFromPackage` added in this same PR explicitly defends against this with `realpath` and even calls it out in a comment: "Symlinks INSIDE the package pointing out (e.g. a malicious package shipping \`leak -> /etc/passwd\`) — realpath follows links, so the resolved target is outside realPkgRoot and is rejected." `listPackagePages` should apply the same containment check.

https://github.com/malloydata/publisher/blob/59138d6a0a9abbacd221d52bcd75233c6b27f9b5/packages/server/src/server.ts#L428-L485

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jswir:
### Code review

Found 1 issue:

1. `PageViewer` blocks the entire iframe on any `pagesQuery` error, contradicting the file's own comment that promises graceful degradation. The `/pages` query is only used to look up the header `<title>` and already falls back to `pagePath` when data is missing, but `if (pagesQuery.isError)` returns `<ApiErrorDisplay>` and prevents the iframe from rendering at all on a transient 500, network blip, or older Publishers without the `/pages` endpoint. Sibling `Package.tsx` swallows errors from the same `listPackagePages` call and degrades — `PageViewer` should do the same.

https://github.com/malloydata/publisher/blob/bdff4852b794d86ebe027aefe0070ef36e69781c/packages/sdk/src/components/PageViewer/PageViewer.tsx#L46-L84

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

jswir:
PR#813 takes this over

Issue #725, opened 2026-05-11
original overview and developer (bun) audit updates


mlennie:
Fix all issues reported in https://docs.google.com/document/d/1HaZz_ar0fE0WY9dh2Ty6Q_-UNRRizeP7vuQ-gITl1c8/edit?usp=sharing

Keep Malloy docs in sync:
https://docs.malloydata.dev/documentation/

Cross refrence with Sagar's requests here: https://credible-data.slack.com/archives/C085VT95H2T/p1778267460369349

PR #731, opened 2026-05-11
[725] Bun setup fixes

mlennie:
# Fresh-eyes fixes for first-time setup

Issue: https://github.com/malloydata/publisher/issues/725

Original audit doc: https://docs.google.com/document/d/1HaZz_ar0fE0WY9dh2Ty6Q_-UNRRizeP7vuQ-gITl1c8/edit?usp=sharing
  
  A clean-install pass on `malloydata/publisher` (clone → install → configure → build → run → verify) surfaced a set of rough edges in docs, defaults, and dev-loop ergonomics. This PR addresses 13 findings from
  that audit plus a handful of follow-ups caught in self-review. All changes stay inside the docs/defaults/build lane — no API shapes, SDK exports, CLI behavior, flag semantics, or connection config schemas were
  modified.
  
  ## Why this PR

  A first-time contributor cloning the repo today hits a few avoidable papercuts:

  - The README's `git submodule update` instruction is dead (`.gitmodules` is empty).
  - The bundled `publisher.config.json` only defines a BigQuery connection, so the bundled samples fail to query without GCP credentials.
  - `bun run typecheck` fails on a fresh clone because the SDK's `.d.ts` outputs haven't been generated yet.
  - The pre-commit hook regenerates the Python SDK on every commit, including markdown-only ones.
  - The `duckdb` connection name is silently reserved; the runtime error is the only signal.
  - No `.tool-versions` / pinned toolchain.
  - No top-level Makefile, leaving the two-terminal dev workflow undocumented and fiddly.
  - `Node`, `Java`, and the `--init` flag are required-but-undocumented in various places.

  Each of these is fixed in a focused commit, structured so reviewers can take them piecemeal.

  ## What changes
  
  ### Docs

  - **README Quick Start** — drop the dead submodule line; add a `Verify it's working` block matching CI's `/api/v0/status` smoke test.
  - **README Development** — rewrite the previous three-line section into the production-build path, two-terminal dev workflow (Express `:4000` proxying to Vite `:5173`), and which URL to visit.
  - **README Prerequisites** — Bun, Node (≥ 20), Python, Java with explicit `required for` notes.
  - **README Configuration** — single table covering every env var and matching CLI flag, with defaults and meanings.
  - **`docs/connections.md` (new)** — documents the per-package DuckDB sandboxes, the reserved `duckdb` name, uniqueness rules, and an example mixed-connections config (BigQuery + DuckDB with attached Postgres).
  - **`packages/server/README.md`** — note the `publisher.config.json` symlink to the repo root.

  ### Defaults

  - **Split `publisher.config.json`.** The default now ships three samples (`ecommerce`, `imdb`, `faa`) that run via per-package DuckDB sandboxes with **zero credentials**. The BigQuery-required
  `bigquery-hackernews` sample plus the `bigquery` top-level connection move to `publisher.config.example.bigquery.json`. Existing users who want BigQuery keep working by copying the example over the default
  (one-flag swap, no data loss).
  
  ### Build / dev-loop

  - **`.tool-versions` (new)** — pins `bun 1.3.13`, `node 24`, `python 3.12`, `java corretto-21`. Compatible with `mise` and `asdf`.
  - **`package.json`** — add `"node": ">=20"` to `engines` so the postinstall scripts' Node requirement is explicit; tighten `bun >= 1.3.13` to match `.tool-versions`; chain `generate-api-types && build:sdk` into
  the root `typecheck` script so a fresh-clone `bun run typecheck` succeeds.
  - **`.pre-commit-config.yaml`** — gate the `python-sdk-generate` hook with `files: ^api-doc\.yaml$` instead of `always_run: true`, so it only fires when the OpenAPI spec changes.
  - **`Makefile` (new)** — canonical task runner with `install`, `build`, `start`, `start-init`, `stop`, `dev` (Express + Vite in one terminal with prefixed logs), `dev-init`, `dev-server`, `dev-react`, `status`,
  `environments`, `packages`, `test`, `lint`, `format`, `prettier-check`, `typecheck`, `regen-api`, plus `help` and `doctor`. The `dev` target uses `bunx concurrently` for combined-log output.

  ### Error messages
  
  - **Reserved connection name.** `connection_config.ts` and `connection.controller.ts` both threw `"DuckDB connection name cannot be 'duckdb'; it is reserved..."`. Updated to `"Connection name 'duckdb' is
  reserved for per-package sandboxes. Choose a different name for environment-level DuckDB connections (e.g. 'shared_duckdb')."` — names the rule directly and points at a working alternative.

  ### Tests

  - Updated three reserved-name assertions to match the new message; standardized on a regex pattern so future wording tweaks don't break either site.
  - Added `config.spec.ts` smoke test that JSON-parses both committed configs and asserts the basic shape (default has no `bigquery-hackernews`; example does; both have a non-empty environment with packages).

  ## Verification

  - **Server tests:** 348 unit + 158 integration = **506 pass / 0 fail**.
  - **Fresh install → build → start cycle:** clean `bun install` + `bun run build:server-deploy` + `bun run start:init` boots, reaches `operationalState: serving`, exposes three samples on
  `/api/v0/environments/malloy-samples/packages`, and successfully compiles `flights.malloy` against the per-package DuckDB sandbox with no GCP credentials.
  - **Tested endpoints:** `/api/v0/status`, `/api/v0/environments`, package listing, model compile, static UI (`:4000/`), MCP health (`:4040/health`).

  ## Breaking changes

  **None.** The plan deliberately avoids touching API/SDK/CLI surfaces. The closest thing to a behavior change is the split-config:

  - Users running with the unmodified committed default + BigQuery creds (in practice, mostly Google-internal users) need to point `--server_root` at the new `publisher.config.example.bigquery.json` (or copy it
  over the default) to regain the `bigquery-hackernews` sample. No data loss; one-flag swap. CI paths that exercise BigQuery continue to work against the example config.
  - Adding `engines.node >= 20` formalizes an existing implicit requirement (postinstall scripts and the bin shebang both already needed Node).

  ## Out of scope

  Three items from the original audit are deliberately handled lightly or deferred:

  - **`--init`** — documented rather than made idempotent; the alternative adds code-change risk for a flag the README hadn't surfaced.
  - **DuckDB postinstall scripts** — the audit noted these could be rewritten in TypeScript + Bun to drop the Node dependency entirely. That overlaps a larger backlog item (move to prebuilt DuckDB binaries), so
  this PR ships only the cheap engine pin.
  - **Upstream docs (`docs.malloydata.dev`)** — items 3, 7, 8 may need mirrored updates on the Malloy docs site; tracked separately.

  ## Commits

  Each commit addresses one focused concern and can be reviewed in isolation. All carry DCO sign-off.

  | Commit | Topic |
  |---|---|
  | `docs(readme): fix submodule line, expand dev mode, add verify block` | README #1, #7, #8 |
  | `build: pin toolchain, fix typecheck chain, gate python-sdk hook` | #4, #9, #11 |
  | `config: split BigQuery sample out of default publisher.config.json` | #2 |
  | `docs(connections): clarify reserved-name error + add connections doc` | #3 |
  | `docs: add Prerequisites + Configuration sections, pin node engine` | #5, #6 |
  | `build: add Makefile with common dev workflows` | #10 |
  | `docs: document --init flag and publisher.config.json symlink` | #12, #13 |
  | `build(makefile): add start-init and dev-init targets` | follow-up |
  | `test: fix missed assertion for reserved-name error message` | follow-up |
  | `chore(e2e): refresh bun.lock and ignore cloned test repos` | follow-up |
  | `docs(readme): fix config-section correctness from PR review` | self-review |
  | `docs(connections): fix DuckDB schema example and use case` | self-review |
  | `fix(connections): align reserved-name error to "environment-level"` | self-review |
  | `chore: Makefile cleanups + smoke-test the example configs` | self-review |

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@
 
 ## Prerequisites
 
-| Tool | Version | Required for |
-|---|---|---|
-| [Bun](https://bun.sh/) | ≥ 1.3.13 | Primary runtime + package manager |
-| [Node.js](https://nodejs.org/) | ≥ 20 | DuckDB postinstall scripts and the `npx @malloy-publisher/server` bin shebang |
-| [Python](https://www.python.org/) | ≥ 3.12 | Only if you build the Python client (`packages/python-client`) |
-| Java | ≥ 21 (Corretto recommended) | Only if you regenerate API clients via `bun run generate-api-types` |
+| Tool                              | Version                     | Required for                                                                  |
+| --------------------------------- | --------------------------- | ----------------------------------------------------------------------------- |
+| [Bun](https://bun.sh/)            | ≥ 1.3.13                    | Primary runtime + package manager                                             |
+| [Node.js](https://nodejs.org/)    | ≥ 20                        | DuckDB postinstall scripts and the `npx @malloy-publisher/server` bin shebang |
+| [Python](https://www.python.org/) | ≥ 3.12                      | Only if you build the Python client (`packages/python-client`)                |
+| Java                              | ≥ 21 (Corretto recommended) | Only if you regenerate API clients via `bun run generate-api-types`           |
 
 The repo ships a `.tool-versions` file compatible with [mise](https://mise.jdx.dev/) and [asdf](https://asdf-vm.com/), so `mise install` (or `asdf install`) provisions all four versions at once.
 
@@ -86,13 +86,12 @@ The Publisher App is built entirely with the SDK, but the SDK is a standalone NP
 
 Publisher consists of four packages:
 
-| Package | Description |
-|---------|-------------|
-| **[packages/server](packages/server/)** | Express.js backend providing REST API (port 4000) and MCP API (port 4040). Loads Malloy packages, compiles queries, executes against databases. |
-| **[packages/sdk](packages/sdk/)** | React component library for building UIs that consume Publisher's REST API. |
-| **[packages/app](packages/app/)** | Reference implementation and production-ready data exploration tool built with the SDK. |
-| **[packages/python-client](packages/python-client/)** | Auto-generated Python SDK for the REST API. |
-
+| Package                                               | Description                                                                                                                                     |
+| ----------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------- |
+| **[packages/server](packages/server/)**               | Express.js backend providing REST API (port 4000) and MCP API (port 4040). Loads Malloy packages, compiles queries, executes against databases. |
+| **[packages/sdk](packages/sdk/)**                     | React component library for building UIs that consume Publisher's REST API.                                                                     |
+| **[packages/app](packages/app/)**                     | Reference implementation and production-ready data exploration tool built with the SDK.                                                         |
+| **[packages/python-client](packages/python-client/)** | Auto-generated Python SDK for the REST API.                                                                                                     |
 
 ## Development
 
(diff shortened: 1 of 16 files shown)


package.json
```diff
@@ -37,7 +38,7 @@
     "typecheck:app": "cd packages/app && bun run tsc --noEmit",
     "typecheck:server": "cd packages/server && bun run tsc --noEmit",
     "typecheck:cli": "cd packages/cli && bun run typecheck",
-    "typecheck": "bun run typecheck:sdk && bun run typecheck:app && bun run typecheck:server",
+    "typecheck": "bun run generate-api-types && bun run build:sdk && bun run typecheck:sdk && bun run typecheck:app && bun run typecheck:server",
```
sagarswamirao: Don't you think typecheck cmd should just be typecheck? And instead readme.md should state that one would have to do bun && bun run build and then run the typecheck?

README.md
```diff
@@ -14,6 +25,15 @@ npx @malloy-publisher/server --port 4000 --server_root .
...
 Open http://localhost:4000 to explore the sample models.
 
+### Verify it's working
+
+```bash
+curl -s http://localhost:4000/api/v0/status | jq .operationalState   # → "serving"
+curl -s http://localhost:4000/api/v0/environments | jq '.[].name'    # → list of environments
+```
+
+If `operationalState` is anything other than `serving`, the server is wedged but still responding — check the server logs.
```
sagarswamirao: if initailizing - server it loading packages and conenctions from the publisher.config.json
if draining - server is shutting down (waits until graceful_timeout env vars)

sagarswamirao (approved):

PR #736, opened 2026-05-14
  ci: integration tests for BigQuery via public-data service account

mlennie:
## Summary

  Adds two integration tests that exercise the publisher's BigQuery connection path against a real SA + real `bigquery-public-data` tables, plus the workflow plumbing to make them run in CI. Closes the gap from
  the `bun-setup-fixes` PR review where Sagar asked for the BigQuery connection AND the `bq-hackernews` package to be loaded under CI.

  ## Context

  In the `bun-setup-fixes` thread, Sagar's note was that the existing CI BQ coverage stops at the driver round-trip — it doesn't prove that a configured environment can load a BQ-backed package and compile its
  models against real BigQuery schemas. This PR addresses that.

  A new repo-level secret `BIGQUERY_PUBLIC_DATA_SA` (read-only access to `bigquery-public-data`) drives both tests. Kept on its own env-var namespace (`BIGQUERY_PUBLIC_DATA_*`) so it coexists with the existing
  `BQ_PRESTO_TRINO_KEY` SA without overwriting `GOOGLE_APPLICATION_CREDENTIALS` / `BIGQUERY_TEST_PROJECT_ID`.

  ## What's in this PR

  ### `packages/server/src/service/connection.spec.ts`

  Two new describe blocks under `createEnvironmentConnections`:

  - **`BigQuery direct connection (public-data)`** — creates a real `bigquery`-typed connection with the SA, queries `SELECT COUNT(*) FROM bigquery-public-data.samples.shakespeare`, asserts the row count is
  between 100k and 200k. Bounded both sides so we catch both auth-fail-but-empty and confusingly-large-result modes. `try`/`finally` around `connection.close()` to release the BQ HTTP/2 client.
  - **`BigQuery package end-to-end (bq-hackernews)`** — stands up a real `EnvironmentStore` against a temp `publisher.config.json` declaring **both** the BQ connection (with the SA injected via
  `serviceAccountKeyJson`) AND the `bigquery-hackernews` package (loaded from `credibledata/malloy-samples` on GitHub). Asserts:
    - `operationalState === "serving"` with no failed envs after init
    - Both the `bigquery` connection and `bigquery-hackernews` package show up in the env's listings
    - `Package.listModels()` returns at least one model that compiled cleanly (`!m.error`) — `listModels()` swallows per-model compile errors into the result, so filtering for `!m.error` is the only way to
  distinguish real success from silent auth failure. Diagnostic `console.error` dumps the per-model errors above the assertion if everything fails, so a CI red is debuggable from the log.

  The package test forces `INITIALIZE_STORAGE=true` (with `try`/`finally` env-var restore matching the existing `PGSSLMODE` pattern in the same file) to deterministically take the load-from-config init branch,
  with an inline safety-precondition comment noting why this is OK only because `tempServerRoot` is freshly empty.

  ### `.github/workflows/connection-integration-tests.yml`

  - New `Setup BigQuery public-data credentials` step decodes `${{ secrets.BIGQUERY_PUBLIC_DATA_SA }}` (base64 or raw JSON, mirrors the existing `BQ_PRESTO_TRINO_KEY` handling), validates JSON, exports
  `BIGQUERY_PUBLIC_DATA_CREDENTIALS` + `BIGQUERY_PUBLIC_DATA_PROJECT_ID` via `$GITHUB_ENV`.
  - Workflow file added to its own `pull_request:paths:` so future workflow-only edits can self-validate (mirrors the existing `push:paths:`).
  - Job-level `if:` guard skips the workflow on **fork PRs** (which can't access repo secrets) so external contributors don't see a confusing red CI for changes unrelated to credentials. Push to `main`, same-repo
  PRs, and `workflow_dispatch` all run normally.

  ## Required setup
  
  This PR requires a new repo-level secret on `malloydata/publisher`:

  - **`BIGQUERY_PUBLIC_DATA_SA`** — JSON service-account key for the "GCP SA BQ public-data for Integration Tests" account (read-only access to `bigquery-public-data`, project `vscode-demo-453414`). Already 
  uploaded.

  ## Cost

  Well under BigQuery's 1 TB/month free tier:

  - `samples.shakespeare` direct query: ~6 MB scanned per run
  - `bq-hackernews` package compile: schema introspection only, no data scan (Malloy compile resolves table types via metadata, doesn't preview rows)
  - One-time ~50 MB git clone of `credibledata/malloy-samples` per package-test run; cleaned up by the `afterEach` retry-on-EBUSY loop

  ## Verification

  - `bun run typecheck` ✓
  - `bun run lint` ✓
  - `bun run prettier:check` ✓ (touched files)
  - `bun test src/service/connection.spec.ts`: **40 pass, 0 fail** locally; both new tests log `Skipping: BIGQUERY_PUBLIC_DATA_CREDENTIALS or BIGQUERY_PUBLIC_DATA_PROJECT_ID not configured` since env vars are 
  unset locally
  - YAML parses cleanly

  ## Test plan

  - [ ] CI run on this PR exercises both new tests against the new secret
  - [ ] First-run sanity: confirm SA has BigQuery Job User role on its own project (`vscode-demo-453414`) so it can bill queries — failure mode would be a billing-permission error, fixable in IAM
  - [ ] Spot-check the workflow's "Setup BigQuery public-data credentials" step exports both env vars correctly

  ## Notes for reviewers

  - The `bq-hackernews` test asserts at least one model compiles cleanly, not strict all-models-pass, so a single broken model in the upstream sample doesn't break this CI gate.
  - If branch protection on `main` requires `Connection Integration Tests` as a passing check (not just non-failing), the job-level fork-PR `if:` guard could make fork PRs unmergeable. If that's the case, 
  switching to step-level `if:` gates on just the credentials steps (with the spec's existing `hasXCredentials()` runtime checks taking over the skip behavior) would let fork PRs go green instead of skipped.


sagarswamirao (changes requested):
Ensure Adding Connection doesn't print the creds

packages/server/src/service/environment.ts
```diff
@@ -191,7 +191,14 @@ export class Environment {
       logger.info(
          `Loaded ${malloyConfig.apiConnections.length} connections for environment ${environmentName}`,
          {
-            apiConnections: malloyConfig.apiConnections,
+            // Log only name + type. The full apiConnections array carries
+            // bigqueryConnection.serviceAccountKeyJson / snowflakeConnection.privateKey
+            // / postgresConnection.password and would leak them through winston
+            // serialization. See connection.ts:1129 for the parallel fix.
+            connections: malloyConfig.apiConnections.map((c) => ({
```
sagarswamirao, line 239: Can you make a separate PR for this?

packages/server/src/service/environment.ts
```diff
@@ -696,7 +703,15 @@ export class Environment {
          `Adding package ${packageName} to environment ${this.environmentName}`,
          {
             packagePath,
-            malloyConfig: this.malloyConfig.malloyConfig,
+            // Do NOT spread the full malloyConfig: its `connections` registry
+            // carries the live Malloy connection objects, which for BigQuery /
+            // Snowflake / Postgres hold serviceAccountKeyJson / privateKey /
+            // password in memory and may serialize them through winston.
+            // Log only the connection names + types instead.
+            connections: this.apiConnections.map((c) => ({
```
sagarswamirao: Can you make a separate PR for this?

sagarswamirao (approved):

PR #741, opened 2026-05-14
[#740]   Materialization-storage DuckLake bad-creds fast-fail (HTTP 422)

### LINKED ISSUE #740: Fast-fail on Ducklake/Postgres auth failures (incident #3)
Slack thread: https://credible-data.slack.com/archives/C085VT95H2T/p1778774806682289?thread_ts=1778630898.862899&cid=C085VT95H2T

Bad Ducklake DB credentials caused workers to hang on `ATTACH 'ducklake:postgres:...'` because the PG conn string had no `connect_timeout`. The failures surfaced as generic HTTP 500, so the upstream retry loop
  kept hammering — saturating workers until K8s killed them.

  This PR:

  - Appends `connect_timeout` (default 5s, env-overridable via `PG_CONNECT_TIMEOUT_SECONDS`) to both keyword-form and URI-form PG conn strings.
  - Classifies libpq auth/permission errors as a new `ConnectionAuthError` → HTTP **422** with `retryable: false` in the body, so retry loops have an unambiguous "don't retry" signal.
  - Fixes a pre-existing password leak in three `attachDuckLake` log lines.

  Verified end-to-end against a local Postgres 16 (wrong creds → 422 in 0.03s, password redacted) and against a TCP black-hole (`connect_timeout=3` → returns in 3.03s, confirming DuckLake forwards the key to
  libpq).

  **Scope:** PG auth path only. Bad S3/GCS creds in `attachCloudStorage` still surface as 500 and could trigger the same retry-loop risk — separate follow-up.

  **Downstream:** to actually stop the retry storm, the MS2 orchestrator needs to read `retryable: false` from the response body and back off. Separate ticket in the `service/` repo.

mlennie:
Closes #740

  ## Problem

  Bad credentials on the Ducklake catalog used for materialization storage caused workers to hang on `ATTACH 'ducklake:postgres:...'` inside `StorageManager.attachDuckLakeCatalog`. The conn string had no
  `connect_timeout`, errors surfaced as generic HTTP 500, and any retry loop treated them as transient — saturating workers until K8s killed them.

  ## Fix

  ### 1. `StorageManager.attachDuckLakeCatalog`

  - Injects `connect_timeout` (default 5s, env-overridable via `PG_CONNECT_TIMEOUT_SECONDS`) into the `catalogUrl` for PG-backed catalogs via the shared `withPgConnectTimeout` helper. Handles keyword form
  (`postgres:host=...`), URI form (`postgresql://...`), and idempotent re-runs (won't override an existing `connect_timeout`).
  - Wraps `connection.run(attachCmd)` in a try/catch that routes errors through `handlePgAttachError`. Libpq auth/permission failures get reclassified as a new `ConnectionAuthError` → **HTTP 422 with `retryable: 
  false`** in the response body, giving retry loops an unambiguous "don't retry" signal.
  - Redacts password values in the `Attaching DuckLake manifest catalog: ...` info log (pre-existing secret leak).

  ### 2. New `src/pg_helpers.ts`

  Shared utilities (`pgConnectTimeoutSeconds`, `withPgConnectTimeout`, `redactPgSecrets`, `classifyPgError`, `handlePgAttachError`, `PgAttachErrorOutcome`). Placed at the layer-neutral server root rather than 
  inside `service/` — putting them in `service/` would have made `storage/` → `service/` the first reverse-direction import in the codebase and risked a cycle once `connection.ts` ever needs a storage import.

  ### 3. New `ConnectionAuthError` class
  
  In `errors.ts`, mapped via `internalErrorToHttpError` to **HTTP 422** with `retryable: false` in the body. Non-breaking response-shape extension (clients ignore unknown fields). Not yet reflected in 
  `api-doc.yaml`.

  ### 4. Bundled security fix
  
  The pre-existing password leak in three `attachDuckLake` log lines in `service/connection.ts` (logged the full PG conn string including the password) is fixed via `redactPgSecrets`. Bundled here because we were 
  already touching the file.

  ## Scope (intentionally narrow)
  
  Only the **materialization-storage catalog ATTACH path** in `StorageManager` is instrumented. The user-facing `service/connection.ts` paths (`attachDuckLake`, `attachPostgres`) are **unchanged**. If the same 
  class of bug is worth fixing on those paths, it goes in a separate PR.

  Publisher has two parallel DuckLake/PG attach paths that don't share helpers; they failed independently in this incident.

  | Path | File | Conn-string source |
  |---|---|---|
  | User-facing connection | `service/connection.ts` | User-defined config → `buildPgConnectionString` |
  | **Materialization storage** *(this PR)* | `storage/StorageManager.ts` | `env.metadata.materializationStorage.catalogUrl` (controlplane) |

  ## Verification

  ### Automated (in-repo)

  - `pg_helpers.spec.ts` — 30 unit tests for the new shared helpers
  - `storage/StorageManager.spec.ts` — 7 wiring tests stubbing `DuckDBConnection.run` to assert the catch-block routing + `connect_timeout` injection + non-PG passthrough without a real DuckDB
  - `errors.spec.ts` — 4 mapping tests for `ConnectionAuthError → 422 + retryable: false`
  - **Full root suite: 396 unit + 158 integration tests, all green**

  ### Empirical (run locally; nothing remote touched)

  **Test A — Real Postgres 16 through the new `StorageManager` wiring.** Docker `postgres:16` on `:5440`, wrong creds, called `StorageManager.initializeDuckLakeForEnvironment()` directly. Result: rejected with
  `ConnectionAuthError` in **0.07s**. Libpq's `FATAL: password authentication failed for user "postgres"` matched `classifyPgError` verbatim; wrong password redacted to `password=***` in both the thrown error and
  the live `Attaching DuckLake manifest catalog: ...` log.
  
  **Test B — Full HTTP end-to-end through the running publisher.** Started `bun run start`, POSTed an env carrying the bad-creds `materializationStorage` payload to `/api/v0/environments`. Actual response:

  ```json
  {
    "code": 422,
    "message": "...password=***...FATAL:  password authentication failed for user \"postgres\"...connect_timeout=5...",
    "retryable": false
  }
  ```
  
  in **0.05s**. The literal wrong password did not appear anywhere in the response body. The echoed `Query: ATTACH ... connect_timeout=5 ...` confirms `withPgConnectTimeout` injected the timeout before SQL
  execution.

  **Test C — Production build.** `bun run build:server-deploy` exit 0. `dist/server.mjs` (9.8MB) and `dist/instrumentation.mjs` (5MB) produced without bundle-time errors.

  **Supporting evidence (pre-rescope ad-hoc DuckDB script):** TCP black-hole on a PG-style port confirmed DuckLake forwards `connect_timeout` through `ducklake:postgres:...` to libpq — `connect_timeout=3` produced
   3.03s wall clock. Libpq's documented 2s floor means `PG_CONNECT_TIMEOUT_SECONDS=1` yields ~2s.

  ## Out of scope (follow-ups)

  - User-facing `connection.ts` attach paths (`attachDuckLake`, `attachPostgres`) — same class of bug, intentionally deferred per review feedback.
  - Bad cloud-storage creds in `attachCloudStorage` still surface as 500; same retry-loop risk on that path.
  - Orchestrator-side change in `service/` to honor `retryable: false` and back off — separate PR in that repo.
  - Adding `retryable` to `api-doc.yaml` so typed SDK clients can read it.

  ## Notes for reviewers

  - libpq enforces a documented **2s minimum** on `connect_timeout`. Setting `PG_CONNECT_TIMEOUT_SECONDS=1` actually yields ~2s wall clock.
  - The `retryable: false` field is added to error response bodies **only for `ConnectionAuthError`**. Existing error responses are unchanged.
  - The auto-regenerated `packages/server/src/api.ts` is **not modified** by this PR (no OpenAPI schema change).

  A couple of things to tweak if you want:

  - Move Closes #740 if you've already got it linked manually via the Development sidebar — then you can drop the keyword to avoid a duplicate close.
  - If the PR title length matters for some downstream tooling, the current 60 chars should be fine — it's under the recommended 70-char ceiling and matches the commit title verbatim, so the squash-merge ends up
  consistent.

### DIFF
--- a/packages/server/src/errors.ts
+++ b/packages/server/src/errors.ts
@@ -16,6 +16,8 @@ export function internalErrorToHttpError(error: Error) {
       return httpError(400, error.message);
    } else if (error instanceof ConnectionNotFoundError) {
       return httpError(404, error.message);
+   } else if (error instanceof ConnectionAuthError) {
+      return httpError(422, error.message);
    } else if (error instanceof ModelCompilationError) {
       return httpError(424, error.message);
    } else if (error instanceof ConnectionError) {
@@ -78,6 +80,12 @@ export class ConnectionNotFoundError extends Error {
 }
 
 export class ConnectionError extends Error {
+   constructor(message: string) {
+      super(message);
+   }
+}
+
+export class ConnectionAuthError extends Error {
    constructor(message: string) {
       super(message);
    }
--- a/packages/server/src/errors.spec.ts
+++ b/packages/server/src/errors.spec.ts
@@ -1,0 +1,42 @@
+import { describe, expect, it } from "bun:test";
+import {
+   BadRequestError,
+   ConnectionAuthError,
+   ConnectionError,
+   internalErrorToHttpError,
+} from "./errors";
+
+describe("internalErrorToHttpError", () => {
+   it("maps ConnectionAuthError to 422", () => {
+      const { status, json } = internalErrorToHttpError(
+         new ConnectionAuthError("creds rejected for db_x"),
+      );
+      expect(status).toBe(422);
+      expect(json).toEqual({
+         code: 422,
+         message: "creds rejected for db_x",
+      });
+   });
+
+   it("maps BadRequestError to 400", () => {
+      const { status, json } = internalErrorToHttpError(
+         new BadRequestError("bad input"),
+      );
+      expect(status).toBe(400);
+      expect(json).toEqual({ code: 400, message: "bad input" });
+   });
+
+   it("maps ConnectionError to 502 (distinct from auth, still retryable)", () => {
+      const { status, json } = internalErrorToHttpError(
+         new ConnectionError("upstream broken"),
+      );
+      expect(status).toBe(502);
+      expect(json).toEqual({ code: 502, message: "upstream broken" });
+   });
+
+   it("falls through to 500 for unrecognized errors", () => {
+      const { status, json } = internalErrorToHttpError(new Error("boom"));
+      expect(status).toBe(500);
+      expect(json.message).toBe("boom");
+   });
+});
--- a/packages/server/src/pg_helpers.spec.ts
+++ b/packages/server/src/pg_helpers.spec.ts
@@ -1,0 +1,226 @@
+import { afterEach, describe, expect, it } from "bun:test";
+import { ConnectionAuthError } from "./errors";
+import {
+   classifyPgError,
+   handlePgAttachError,
+   pgConnectTimeoutSeconds,
+   redactPgSecrets,
+   withPgConnectTimeout,
+} from "./pg_helpers";
+
+describe("pgConnectTimeoutSeconds", () => {
+   const ORIGINAL_TIMEOUT = process.env.PG_CONNECT_TIMEOUT_SECONDS;
+
+   afterEach(() => {
+      if (ORIGINAL_TIMEOUT === undefined) {
+         delete process.env.PG_CONNECT_TIMEOUT_SECONDS;
+      } else {
+         process.env.PG_CONNECT_TIMEOUT_SECONDS = ORIGINAL_TIMEOUT;
+      }
+   });
+
+   it("defaults to 5 when env unset", () => {
+      delete process.env.PG_CONNECT_TIMEOUT_SECONDS;
+      expect(pgConnectTimeoutSeconds()).toBe(5);
+   });
+
+   it("honors PG_CONNECT_TIMEOUT_SECONDS override", () => {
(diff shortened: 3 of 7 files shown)


packages/server/src/errors.ts
```diff
@@ -16,6 +16,8 @@ export function internalErrorToHttpError(error: Error) {
       return httpError(400, error.message);
    } else if (error instanceof ConnectionNotFoundError) {
       return httpError(404, error.message);
+   } else if (error instanceof ConnectionAuthError) {
+      return httpError(422, error.message, { retryable: false });
```
sagarswamirao: API Doc doesn't have this retryable Param
Can we remove it?

sagarswamirao (approved):

Issue #749, opened 2026-05-18
fix: pin DuckLake extension via Docker bake + catalog preflight (incident #6)


mlennie:
slack thread: https://credible-data.slack.com/archives/C085VT95H2T/p1778641006406909?thread_ts=1778630898.862899&cid=C085VT95H2T

Major one: Prod had Ducklake Catalog version = 1.0.0, but the Publisher one had 0.0.3 and started to throw errors when the DMS issue was fixed ("Not implemented Error: Only DuckLake versions 0.1, 0.2, 0.3-dev1 and 0.3 are supported"). Had to downgrade the version to 0.0.3 to make the workers happy again. (we need to look into how do we upgrade to the latest one and ensure the existing Ducklake DB tables are compatible)

PR #750, opened 2026-05-18
[#749] fix: pin DuckLake extension via Docker bake + catalog preflight (incident 6)

mlennie:
## Summary
 
 Issue: https://github.com/malloydata/publisher/issues/749
  
  Fixes incident item number 6 in slack https://credible-data.slack.com/archives/C085VT95H2T/p1778641006406909?thread_ts=1778630898.862899&cid=C085VT95H2T (the May 2026 DuckLake catalog version mismatch).

  Pins the DuckLake extension at Docker build time via the `duckdb@1.4.4` npm binding so all workers boot with the same extension binary. The bake lands at `/root/.duckdb/extensions/v1.4.4/linux_arm64/`, the
  path the runtime server inspects (not the path the CLI installer in `base-deps` would use, which is for a different DuckDB engine version).

  Also adds a postgres-side preflight in `StorageManager.attachDuckLakeCatalog` that reads `ducklake_metadata.version` directly via a plain `TYPE postgres` ATTACH (not DuckLake) and throws `ConnectionAuthError`
   → 422 when the catalog format is newer than the baked extension supports. Preflight failures (missing table, query timeout, connect failure) return `undefined` and fall through to the existing ATTACH path so
   it's not load-bearing.
  
  ## Why

  In May 2026, prod's DuckLake catalog had been written at format version 1.0.0 while Publisher workers ran an older extension that only supported up to 0.3. The workaround was a manual catalog downgrade. Root
  cause: nothing pinned the DuckLake extension binary on the worker side, so workers and the catalog could drift independently.

  Verified:
  - DuckDB 1.4.4 ships the 0.3-line DuckLake (cross-referenced with the duckdb/ducklake release timeline; 1.0 requires DuckDB 1.5+).
  - Local docker build prints `DuckLake baked: [{"extension_version":"3f1b372","install_path":"/root/.duckdb/extensions/v1.4.4/linux_arm64/..."}]`.
  - `docker run --network none ... bun -e "...LOAD ducklake..."` succeeds, proving the runtime uses the baked binary (not a network re-fetch).

  ## Out of scope (separate PRs)

  - DuckLake 1.0 upgrade (bump `duckdb` to 1.5+, bump `@malloydata/db-duckdb`, add `AUTOMATIC_MIGRATION=TRUE`). Irreversible on prod data, deserves its own deliberate rollout.
  - README + public docs sync: deferred until PR #743 merges to avoid conflicts on the env-vars table.

  ## Test plan

  - [ ] CI passes (typecheck, lint, prettier, 559 server tests).
  - [ ] CI's `docker_smoke_test` build log shows the new `DuckLake baked: [...]` line with `v1.4.4` in the install path.


Dockerfile
```diff
@@ -84,6 +84,17 @@ COPY --from=builder /publisher/packages/sdk/package.json /publisher/packages/sdk
...
 
+# Bake the DuckLake extension via the `duckdb` npm binding (NOT the CLI).
+# The CLI in base-deps installs to /root/.duckdb/extensions/v<cli-version>/,
+# which is a different DuckDB version than what `duckdb@1.4.4` (the
+# runtime binding pinned in package.json) inspects at runtime
+# (/root/.duckdb/extensions/v1.4.4/). Baking via the binding here puts
+# the extension at the path the runtime server uses, so its INSTALL
+# ducklake; LOAD ducklake; at startup finds it and skips the network
+# fetch (the actual pin).
+RUN bun -e "const d=require('duckdb');const db=new d.Database(':memory:');await new Promise((r,j)=>db.exec('INSTALL ducklake; LOAD ducklake;',e=>e?j(e):r()));const rows=await new Promise((r,j)=>db.all(\"SELECT extension_version, install_path FROM duckdb_extensions() WHERE extension_name='ducklake'\",(e,x)=>e?j(e):r(x)));console.log('DuckLake baked:',JSON.stringify(rows));" || \
```
sagarswamirao, line 95: Can we get rid of the install duckdb nodebindings then?

kylenesbit (approved):

PR #756, opened 2026-05-19
fix(server): stage-and-swap package installs + extend per-package mutex to all disk reads

kylenesbit:
## Summary

Closes the race where one request mutates a package's on-disk directory (download/unzip, manifest write, delete) while another request is reading that directory to compile a model — surfacing as `Package manifest for ... does not exist`, `compiling model path not found`, or transient `ENOENT` inside Malloy's `URL_READER` import resolution.

This is the same family of races [#752](https://github.com/malloydata/publisher/pull/752) targets. The key difference: this PR also covers the **reader side** (`compileSource`, MCP `getModelText`, materialization's `compilePackageBuildPlan`, manifest `reloadAllModels`), which is the path the symptom is actually reported on. See the comparison block at the bottom of the description for details.

## Approach

Reuse the existing per-package `packageMutex` as the single serialization point and add a **stage-and-swap** for downloads so long network I/O does not block reads.

### Writes — `Environment.installPackage(name, downloader)`

- **Phase 1 (no lock):** `downloader` writes into `{env}/.staging/{pkg}-<uuid>/`. Long downloads do not block anything.
- **Phase 2 (under `withPackageLock`):**
  1. `fs.rename` canonical → `{env}/.retired/{pkg}-<uuid>/`
  2. `fs.rename` staging → canonical
  3. `Package.create` against the new canonical path
  4. Update `this.packages` and call `retireConnectionGeneration` for the old package's connections
- **Phase 3 (after lock release):** `setImmediate(() => fs.rm(retired, { recursive: true, force: true }))`. Lock-hold time stays O(rename + Package.create), not O(download + disk content).
- **Rollback:** if Phase 2 fails, restore the previous canonical from `.retired/` and rm the orphan staging.

### Reads — extended mutex coverage

All disk-touching reads now run under `withPackageLock`:

- `Environment.compileSource` (the `fs.readFile(modelPath)` + Malloy `URL_READER` import resolution)
- `Environment.getModelFileText` (new — wraps `Package.getModelFileText`, used by MCP `getModelText`)
- `Environment.reloadAllModelsForPackage` (new — wraps `Package.reloadAllModels`, used by `ManifestService` and `MaterializationService`)
- `MaterializationService.compilePackageBuildPlan` (wraps `Model.getModelRuntime`)
- `Environment.deletePackage` becomes rename-to-`.retired` under the lock + async `fs.rm`

### Re-entrancy

`async-mutex` is not reentrant. Each affected public method is split into a public method that acquires the lock and a `_xxxLocked` private sibling that assumes the lock is held. No `dirLockHeld?: boolean` flag on the public API.

### Crash recovery

`Environment.create` now runs `sweepStaleInstallDirs()` to clean up `.staging/*` and `.retired/*` orphaned by a server crash mid-install.

### Lock ordering

Preserved: `connectionMutex` (outer) → `packageMutex` (inner). No new mutex type introduced.

## Files changed

- `packages/server/src/service/environment.ts` — new `installPackage`, `withPackageLock`, `getModelFileText`, `reloadAllModelsForPackage`, `sweepStaleInstallDirs`, `_loadOrGetPackageLocked`, `_addPackageLocked`; `compileSource` / `updatePackage` / `deletePackage` wrapped.
- `packages/server/src/controller/package.controller.ts` — `addPackage`, `updatePackage`, `getPackage(reload=true)` route through `installPackage` when `body.location` is present. `downloadPackage` → `downloadInto(targetPath, …)`.
- `packages/server/src/service/manifest_service.ts` — calls `environment.reloadAllModelsForPackage`.
- `packages/server/src/service/materialization_service.ts` — calls `environment.reloadAllModelsForPackage`; `compilePackageBuildPlan` wrapped in `environment.withPackageLock`.
- `packages/server/src/mcp/tools/discovery_tools.ts` — `getModelText` routes through `environment.getModelFileText`.
- `packages/server/src/service/package_race.spec.ts` — new test file (see below).
- `packages/server/src/service/manifest_service.spec.ts` — mock update for the rerouted reload call.

## Tests

New unit-level reproductions in `packages/server/src/service/package_race.spec.ts` (no network deps, deterministic):

- **(A)** Concurrent `installPackage` and `getModelFileText` in a 30-round loop — reader must never observe a half-rewritten tree.
- **(B)** Externally-held package lock blocks a concurrent `getModelFileText` until released — verifies the reader side is actually covered by the mutex.
- **(C)** A slow staged download (deferred until the test releases it) does NOT block a concurrent `getModelFileText` — verifies Phase 1 stays outside the lock.

Verified by temporarily removing the lock from `compileSource` / `getModelFileText` and confirming Test A and Test B fail, then restoring and confirming they pass.

Lint, typecheck, and the full server unit suite are clean:

```
bunx eslint "./src/**/*.ts"
bunx tsc --noEmit -p packages/server/tsconfig.json
bun test packages/server/src
# → 392 pass, 0 fail
```

Three pre-existing integration failures (`mcp_transport.integration.spec.ts`, `mcp_resource.integration.spec.ts`) caused by `TypeError: describe.serial is not a function` reproduce on `main` and are unrelated.

## Test plan

- [ ] CI green
- [ ] Local: reproduce the original race by hammering `POST /packages` + `GET /packages/:name/models?reload=true` concurrently against a GCS-hosted package and confirm no `Package manifest for ... does not exist` errors.
- [ ] Local: kill the server mid-download and confirm `.staging/` is swept on next start.
- [ ] Verify lock ordering by running a workload that mixes connection updates with package installs (no deadlocks observed).

## Comparison with #752

[#752](https://github.com/malloydata/publisher/pull/752) shares the same architectural insight (stage download outside, swap inside). The differences worth flagging during review:

|  | #752 | This PR |
|---|---|---|
| Writer-vs-writer races (the PR's stated scope) | Fixed | Fixed |
| **Writer-vs-compile races** (the user-visible symptom) | Not fixed — `compileSource` / `URL_READER` / `getModelText` / materialization compile take no lock; `getPackage` cache-hit fast-paths past both mutexes | Fixed — all readers run under the same per-package mutex |
| Mutexes per package | 2 (new `packageDirectoryMutex` + existing `packageMutex`) | 1 (reuse existing) |
| `Environment.getPackage` signature | Adds `dirLockHeld: boolean` flag | Unchanged |
| `deletePackage` lock-hold time | Synchronous `fs.rm` inside lock — O(disk content) | rename-to-`.retired` + async `fs.rm` — O(rename) |
| Stale `.staging-*` after crash | Accumulates next to canonical path | Subdir under `.staging/`, swept on `Environment.create` |
| Reader coverage in tests | Asserts compile-side endpoints don't 5xx and don't include specific error fragments; explicitly tolerates transient 404 mid-rewrite | Asserts readers see only fully-old or fully-new state, never an intermediate |

If #752 lands first, this PR can be cleanly re-targeted on top of it: keep its `packageDirectoryMutex` and just consolidate readers onto the same lock (drop our second mutex, drop `dirLockHeld` flag in favor of `_xxxLocked` siblings, swap delete to rename + async-rm, add the startup sweep).

### DIFF
--- a/packages/server/src/controller/package.controller.ts
+++ b/packages/server/src/controller/package.controller.ts
@@ -1,6 +1,4 @@
-import * as path from "path";
 import { components } from "../api";
-import { PUBLISHER_DATA_DIR } from "../constants";
 import { BadRequestError, FrozenConfigError } from "../errors";
 import { logger } from "../logger";
 import { EnvironmentStore } from "../service/environment_store";
@@ -37,15 +35,38 @@ export class PackageController {
          environmentName,
          false,
       );
-      const _package = await environment.getPackage(packageName, reload);
-      const packageLocation = _package.getPackageMetadata().location;
-      if (reload && packageLocation) {
-         await this.downloadPackage(
-            environmentName,
-            packageName,
-            packageLocation,
-         );
+
+      if (reload) {
+         // Resolve the package's source location from the currently-cached
+         // metadata WITHOUT triggering a stale-state reload. If a `location`
+         // is set, route the reload through `installPackage` so that
+         // download-then-load happens atomically; otherwise fall back to an
+         // in-place reload of the existing on-disk content.
+         let location: string | undefined;
+         try {
+            const cached = await environment.getPackage(packageName, false);
+            location = cached.getPackageMetadata().location;
+         } catch {
+            // Not previously loaded — nothing to reinstall from.
+         }
+         if (location) {
+            const reinstalled = await environment.installPackage(
+               packageName,
+               (stagingPath) =>
+                  this.downloadInto(
+                     environmentName,
+                     packageName,
+                     location,
+                     stagingPath,
+                  ),
+            );
+            return reinstalled.getPackageMetadata();
+         }
+         const _package = await environment.getPackage(packageName, true);
+         return _package.getPackageMetadata();
       }
+
+      const _package = await environment.getPackage(packageName, false);
       return _package.getPackageMetadata();
    }
 
@@ -60,21 +81,32 @@ export class PackageController {
       if (!body.name) {
          throw new BadRequestError("Package name is required");
       }
+      const packageName = body.name;
       const environment = await this.environmentStore.getEnvironment(
          environmentName,
          false,
       );
+      let result;
       if (body.location) {
-         await this.downloadPackage(environmentName, body.name, body.location);
+         const bodyLocation = body.location;
+         result = await environment.installPackage(packageName, (stagingPath) =>
+            this.downloadInto(
+               environmentName,
+               packageName,
+               bodyLocation,
+               stagingPath,
+            ),
+         );
+      } else {
+         result = await environment.addPackage(packageName);
       }
-      const result = await environment.addPackage(body.name);
       await this.environmentStore.addPackageToDatabase(
          environmentName,
-         body.name,
+         packageName,
       );
 
       if (options?.autoLoadManifest === true) {
-         await this.tryLoadExistingManifest(environmentName, body.name);
+         await this.tryLoadExistingManifest(environmentName, packageName);
       }
 
       return result;
@@ -151,12 +183,20 @@ export class PackageController {
          false,
       );
       if (body.location) {
-         await this.downloadPackage(
-            environmentName,
-            packageName,
-            body.location,
+         // Re-install: stream the new content into a staging dir (no lock)
+         // and atomically swap it in (under the lock).
+         const bodyLocation = body.location;
+         await environment.installPackage(packageName, (stagingPath) =>
+            this.downloadInto(
+               environmentName,
+               packageName,
+               bodyLocation,
+               stagingPath,
+            ),
          );
       }
+      // Apply metadata changes (publisher.json) under the same per-package
+      // mutex via `Environment.updatePackage`.
       const result = await environment.updatePackage(packageName, body);
       await this.environmentStore.addPackageToDatabase(
          environmentName,
@@ -166,17 +206,18 @@ export class PackageController {
       return result;
    }
 
-   private async downloadPackage(
+   /**
+    * Run the right downloader for the given location into `targetPath`.
+    * This used to point at the canonical package directory, but the
+    * install pipeline now passes a sibling staging dir so the long-running
+    * download doesn't hold the per-package mutex.
+    */
+   private async downloadInto(
       environmentName: string,
       packageName: string,
       packageLocation: string,
+      targetPath: string,
    ) {
-      const absoluteTargetPath = path.join(
-         this.environmentStore.serverRootPath,
-         PUBLISHER_DATA_DIR,
-         environmentName,
-         packageName,
-      );
       const isCompressedFile = packageLocation.endsWith(".zip");
       if (
          packageLocation.startsWith("https://") ||
@@ -184,20 +225,20 @@ export class PackageController {
       ) {
          await this.environmentStore.downloadGitHubDirectory(
             packageLocation,
-            absoluteTargetPath,
+            targetPath,
          );
       } else if (packageLocation.startsWith("gs://")) {
          await this.environmentStore.downloadGcsDirectory(
             packageLocation,
             environmentName,
-            absoluteTargetPath,
+            targetPath,
             isCompressedFile,
          );
       } else if (packageLocation.startsWith("s3://")) {
          await this.environmentStore.downloadS3Directory(
             packageLocation,
             environmentName,
-            absoluteTargetPath,
+            targetPath,
             isCompressedFile,
          );
       }
@@ -207,7 +248,7 @@ export class PackageController {
          // so we need to mount them on the right place.
          await this.environmentStore.mountLocalDirectory(
             packageLocation,
-            absoluteTargetPath,
+            targetPath,
             environmentName,
             packageName,
          );
--- a/packages/server/src/mcp/tools/discovery_tools.ts
+++ b/packages/server/src/mcp/tools/discovery_tools.ts
@@ -222,8 +222,12 @@ export function registerTools(
                throw new Error(`Model not found: ${modelPath}`);
             }
 
-            // Use the new getModelFileText method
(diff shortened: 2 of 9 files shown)


mlennie:
@kylenesbit looks good 

**Nits / minor questions (none blocking)**

 1. Stale packageMutexes entries after delete. deletePackage doesn't clear the package's entry in packageMutexes. That's correct (reuse on re-install is desirable), but the comment "never replace after create"
 makes it sound permanent. If package names are highly dynamic over a long-running process, the map grows unbounded. Probably not a real concern, but worth a one-line comment that this is intentional.
 (environment.ts:449–456)
 2. TOCTOU between getPackage(false) and reloadAllModelsForPackage. In manifest_service.ts:88–92 and the materialization flow at materialization_service.ts:379–384, a concurrent deletePackage between the two
 awaits would make reloadAllModelsForPackage throw PackageNotFoundError. The caller surfaces a clean error, no corruption. Acceptable, but: if you wanted strict atomicity, both could go inside one withPackageLock
  block (with _loadOrGetPackageLocked then pkg.reloadAllModels inline). Not worth doing unless someone hits the race in practice.
 3. installPackage rollback / Package.create self-cleanup overlap. Package.create's failure path already fs.rms the package directory (per package.ts:162–173 on main). The rollback in installPackage then fs.rms
 canonicalPath again (with .catch(() => {})) before renaming retiredPath back. That's robust but worth a one-line comment in the rollback noting that Package.create may have already rm'd the path — the catch
 handles the ENOENT. (environment.ts:719–746)
 4. getModelFileText / reloadAllModelsForPackage throw PackageNotFoundError if the package isn't in the in-memory cache. Previously, callers held a pkg reference; under a concurrent delete they would have gotten
 an ENOENT on disk read. Now they get a clean PackageNotFoundError. Different error type, same outcome from the user's perspective. Worth confirming with the team that this error-shape change is fine for the
 orchestrated MS2 workers (the materialization caller does getPackage(false) first so it's protected).
 5. MCP resources still call getPackage(false) without re-checking the package post-delete. As noted in "Coverage of readers", this is safe today because Model instances are pre-compiled and resource handlers
 only do in-memory operations on them. But it's an invariant worth memorialising: if anyone ever moves disk I/O into Model.getModel() (or any resource-accessed method), the per-package mutex must cover that path
 too. Worth a comment near the fast-path in getPackage flagging this.
 6. getPackage(reload=true) controller route now triggers a full re-install when location is set. That's not a new behaviour, the previous controller also called downloadPackage after getPackage(reload=true), but
  it's now atomic + locked, which is the right thing. Also: this path bypasses publisherConfigIsFrozen (and did so before). Probably worth a separate ticket to add the frozen-config check on reload-with-location,
  but not in scope here.
 7. crypto.randomUUID() for staging/retired path suffixes. Good choice for uniqueness across concurrent installs of the same package (which serialise on the mutex in Phase 2 but may overlap in Phase 1). No
 collision risk.
 8. assertSafePackageName rejects names longer than 255 chars. The existing malloy-samples packages all fit (and PRD-named packages are short). Worth a quick check that no production package names exceed this; if
  so, callers will see HTTP 400 starting on deploy. Defensive but bounded.
  
  **Verification suggestions for the test plan**

 The PR description's test plan is solid. Two additions worth running locally before merge:

 - Test plan item: after killing the server mid-Phase-1-download, confirm both that .staging/* is swept on next start AND that the canonical tree is still intact (the failed Phase-1 download never moved it).
 - Test plan item: after killing the server mid-Phase-2 (between rename canonical→retired and rename staging→canonical), confirm the next startup's sweep cleans .retired/* and .staging/* and that the canonical
 path is gone. The caller (PUT or POST) will have errored to the client, so the next attempt is a fresh install. This is the worst-case partial-failure scenario and worth confirming behaves cleanly.

mlennie (approved):

kylenesbit:
Thanks for the careful read. Comment-only follow-ups pushed in `c255cd5` — addressed the three nits that asked for an in-code note, and triaging the rest below.

**Addressed (comments, no behaviour change):**

- **1. Stale `packageMutexes` after delete.** Expanded the docblock to make the intent explicit: `deletePackage` deliberately leaves the entry behind so a subsequent re-install serialises against any straggling readers from the deleted generation. The map grows by *distinct* names served, not install churn, so for the publisher's workload it's bounded; called out that a long-lived deployment churning unique names indefinitely would need an explicit sweep we'd add when that pattern appears.
- **3. installPackage rollback / Package.create self-cleanup overlap.** Refined the existing rollback comment to call out that ENOENT is the *expected* case the `.catch(() => {})` swallows because Package.create's failure-cleanup may have already removed the directory — the `fs.rm` is purely a best-effort wipe to leave the rename-back a clean destination.
- **5. MCP resources / fast-path invariant.** Added an INVARIANT note on the `getPackage` fast-path: any code reachable from a `Package` reference returned without taking the lock must remain in-memory only. If anyone ever pushes disk I/O into `Model.getModel()` or an MCP resource handler, that path must be bracketed by `withPackageLock`. This is the memorialisation you asked for.

**Triage on the rest:**

- **2. TOCTOU `getPackage(false)` → `reloadAllModelsForPackage`.** Agreed — `PackageNotFoundError` is the right surface and the user-visible outcome is identical. Holding it open for "wrap in one `withPackageLock`" only if it ever shows up in practice.
- **4. Error-shape change to `PackageNotFoundError`.** Worth flagging to the orchestrated MS2 workers team. The materialization caller does `getPackage(false)` first so it's protected; the only callers exposed to the new error shape are paths that skip the pre-fetch. Will raise in the team channel for confirmation.
- **6. `getPackage(reload=true)` bypassing `publisherConfigIsFrozen`.** Pre-existing behaviour, agree it deserves a separate ticket. Filing.
- **7. `crypto.randomUUID()`.** No action needed.
- **8. 255-char `assertSafePackageName` limit.** Scanned all `publisher.config*.json` files in the repo — nothing approaches the limit (longest is well under 50 chars). Comfortable that the constraint won't surprise any current deployment.

**Verification suggestions.** Both kill-mid-Phase-N tests are good additions; I'll fold them into the test plan and run them locally before merge.

PR #757, opened 2026-05-19
feat(server): RSS-based memory governor with HTTP 503 back-pressure on new package loads

kylenesbit:
## Summary

Adds an opt-in, in-process memory governor that keeps the publisher under its
Kubernetes pod memory limit by **refusing new package loads with HTTP 503 once
process RSS crosses a configured high-water mark**, instead of getting
OOM-killed mid-request.

- **Admission control only** — no soft-unload, no eviction loop. Already-loaded
  packages keep serving so dashboards stay responsive under pressure; recovery
  happens naturally as in-flight requests complete and the kernel reclaims
  pages.
- **Disabled by default.** The feature only activates when
  `PUBLISHER_MAX_MEMORY_BYTES` is set; every other knob has a documented
  default and validates at startup so a typo fails loud.
- **Hysteresis band** (low/high water marks) on a single `isBackpressured`
  boolean keeps the governor from flapping across GC cycles.
- **Optional dependency-injected** into `PackageController` so the existing
  code path is a no-op when the feature is off and unit tests can drive
  deterministic RSS sequences via an injected sampler.

### Gated endpoints

Only paths that allocate a **new** package into memory get gated; reads against
already-loaded packages stay serviceable under pressure:

- `POST /api/v0/environments/:env/packages` (`addPackage`)
- `GET …/packages/:pkg?reload=true` (`getPackage` with reload)
- `PATCH …/packages/:pkg` when the body has a `location` (forces redownload)

### Configuration

| Env var | Default | Notes |
|---|---|---|
| `PUBLISHER_MAX_MEMORY_BYTES` | unset (disabled) | RSS cap in bytes. Suggested: ~0.7 × k8s memory limit. |
| `PUBLISHER_MEMORY_HIGH_WATER_FRACTION` | `0.8` | Fraction of cap that activates back-pressure. |
| `PUBLISHER_MEMORY_LOW_WATER_FRACTION` | `0.7` | Fraction at which back-pressure clears. |
| `PUBLISHER_MEMORY_CHECK_INTERVAL_MS` | `5000` | Sampling cadence (min `100`). |
| `PUBLISHER_MEMORY_BACKPRESSURE` | `true` | When `false`, samples + emits metrics but never flips the flag (monitoring-only rollout). |

### New Prometheus metrics

| Metric | Type |
|---|---|
| `publisher_process_rss_bytes` | gauge |
| `publisher_memory_backpressure_active` | gauge (0/1) |
| `publisher_memory_backpressure_activations_total` | counter |
| `publisher_memory_{max,high_water,low_water}_bytes` | static gauges |

### Files

- `src/service/package_memory_governor.ts` (new) — the governor itself.
- `src/service/package_memory_governor.spec.ts` (new) — 10 unit tests with an
  injected `FakeRssSampler`.
- `src/config.ts` — `getMemoryGovernorConfig()` env parsing + validation.
- `src/config.spec.ts` — 7 new tests covering disabled / defaults / overrides /
  every validation error.
- `src/errors.ts` — `ServiceUnavailableError` + HTTP 503 mapping.
- `src/controller/package.controller.ts` — gates the three new-load entry points.
- `src/server.ts` — constructs / starts the governor, wires SIGTERM shutdown.
- `README.docker.md` — env-var docs and a worked k8s sizing example.

## Test plan

- [x] `bun test src/service/package_memory_governor.spec.ts` — 10/10 pass.
- [x] `bun test src/config.spec.ts` — 43/43 pass (7 new).
- [x] `bun run tsc --noEmit` — clean.
- [x] Full `bun test` — matches `main` baseline (the 10 unrelated
      `attachDuckLakeCatalog wiring` failures exist on `main` before this PR;
      confirmed by re-running with these changes stashed).
- [ ] Manual: start the server with `PUBLISHER_MAX_MEMORY_BYTES=200000000` and
      a tight `*_CHECK_INTERVAL_MS=500`, watch `/metrics` flip
      `publisher_memory_backpressure_active` once a few packages are loaded,
      confirm `POST /packages` returns 503 while loaded packages still serve.

### DIFF
--- a/packages/server/README.docker.md
+++ b/packages/server/README.docker.md
@@ -85,9 +85,11 @@ All flags exposed by `bin/malloy-publisher --help` have an equivalent env var, s
 When `PUBLISHER_MAX_MEMORY_BYTES` is unset, the governor is **disabled** and the server's behaviour is identical to prior versions. When it's set, the governor:
 
 - Periodically samples `process.memoryUsage().rss`.
-- Once RSS crosses the high-water mark, **new package loads** (`POST /api/v0/environments/:env/packages`, `GET …/packages/:pkg?reload=true`, `PATCH …/packages/:pkg` with a `location` field) return **HTTP 503**. Already-loaded packages remain fully serviceable so dashboards keep rendering under pressure.
+- Once RSS crosses the high-water mark, **any code path that would allocate a new package into memory returns HTTP 503**. The gate sits at the single choke point inside `Environment.getPackage` / `Environment.addPackage`, so it covers every controller that touches a not-yet-loaded package — including lazy loads on cache miss from `ModelController`, `ConnectionController`, `QueryController`, `DatabaseController`, etc. — not just the explicit `POST /packages` and `?reload=true` paths.
+- Already-loaded packages remain fully serviceable so dashboards keep rendering under pressure.
 - Once RSS drops back to the low-water mark, back-pressure clears automatically.
 - Recovery happens naturally as in-flight traffic completes and the kernel reclaims pages — the governor does **not** evict, unload, or interrupt loaded packages.
+- A documented `{ allowAdmission: true }` opt-out exists on `Environment.getPackage` / `addPackage` for future internal callers (e.g. warmup / health probes) that genuinely cannot tolerate 503s. No public REST endpoint sets it today.
 
 Metrics exposed on the existing `/metrics` Prometheus endpoint:
 
--- a/packages/server/README.docker.md
+++ b/packages/server/README.docker.md
@@ -74,6 +74,51 @@ All flags exposed by `bin/malloy-publisher --help` have an equivalent env var, s
 | `SHUTDOWN_DRAIN_DURATION_SECONDS` | — | `0` | On SIGTERM, how long to keep accepting requests while draining before closing server sockets. Set this to your typical request duration to avoid 502s from K8s rolling deploys. |
 | `SHUTDOWN_GRACEFUL_CLOSE_TIMEOUT_SECONDS` | — | `0` | Additional grace period after server close before `process.exit`. |
 | `GOOGLE_APPLICATION_CREDENTIALS` | — | unset | Path inside the container to a GCP service-account JSON. Required for BigQuery-backed environments. Personal user credentials don't work inside the container — use a service account. |
+| `PUBLISHER_MAX_MEMORY_BYTES` | — | unset (disabled) | Resident-set-size (RSS) cap in bytes. When set, the in-process **memory governor** polls RSS on `PUBLISHER_MEMORY_CHECK_INTERVAL_MS` and rejects new package loads with **HTTP 503** once RSS crosses the high-water mark. Designed to keep the pod under its k8s `resources.limits.memory` instead of getting OOM-killed. Set this to roughly `0.7 × resources.limits.memory` so the back-pressure band has headroom for traffic spikes and per-request DuckDB scratch. |
+| `PUBLISHER_MEMORY_HIGH_WATER_FRACTION` | — | `0.8` | Fraction of `PUBLISHER_MAX_MEMORY_BYTES` at which back-pressure activates. Must be in `(0, 1)` and strictly greater than the low-water fraction. |
+| `PUBLISHER_MEMORY_LOW_WATER_FRACTION` | — | `0.7` | Fraction at which back-pressure clears. The gap between low and high gives hysteresis so the governor doesn't flap on every GC cycle. |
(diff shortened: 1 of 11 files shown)


mlennie:
Looks good!

A couple of items worth addressing before merge. Otherwise this reads well: defaults are conservative, validation throws loud, hysteresis is simple, and the injected sampler makes the unit tests deterministic.

  **1. Coverage gap on lazy package loads**

  The description says "only paths that allocate a new package into memory get gated," but `Environment.getPackage(name, reload=false)` at `packages/server/src/service/environment.ts:402` lazy-loads via
  `Package.create(...)` whenever the in-memory map misses. Several controllers consume that path without the governor seeing it. For example, `ModelController` calls `environment.getPackage(packageName, false)`
  from at least five handlers (`packages/server/src/controller/model.controller.ts:25, 37, 51, 81, 111`), and the same pattern almost certainly exists in `connection.controller.ts`, `query.controller.ts`,
  `database.controller.ts`, and `notebook.controller.ts`. `PackageController.getPackage(reload=false)` also flows through the same lazy path; its comment ("Read-only fetches against the in-memory cache are safe
  under pressure") only holds in steady state, not after the error branch at `environment.ts:467` deletes the package on a failed load.

  In typical k8s deployments, every package is listed in `publisher.config.json` and `EnvironmentStore.addPackages` eagerly loads them at boot, so the lazy path is rarely the first allocation site. That makes this
   a soft bug rather than a hard one, but the contract advertised in the PR description is stricter than what the diff actually enforces.

  Two ways to land it:
  
  * Preferred: gate at the choke point. Add the `isBackpressured` check inside `Environment.getPackage` itself, behind a per-call `bypassGate`/`allowAdmission` boolean for the read paths that genuinely cannot
  tolerate 503s. One diff, one choke point, controllers unchanged.
  * Minimum: amend the PR description and the `README.docker.md` section so operators know the gate covers package-create and explicit-reload paths, but a first-read miss after a failed boot can still slip
  through.

  **2. Stale JSDoc on `MemoryGovernorConfig`**

  `packages/server/src/config.ts:99-115` describes an eviction loop that doesn't exist:

  ```ts
  export interface MemoryGovernorConfig {
     /** Hard ceiling for process RSS in bytes (the OOM-relevant figure). */
     maxMemoryBytes: number;
     /** Begin evicting LRU packages once RSS exceeds this fraction of `maxMemoryBytes`. */
     highWaterFraction: number;
     /** Stop evicting once RSS drops back below this fraction of `maxMemoryBytes`. */
     lowWaterFraction: number;
     /** Polling cadence for the eviction loop, in milliseconds. */
     checkIntervalMs: number;
     /** When true, return HTTP 503 on new package loads while we are above high-water with nothing idle to evict. */
     backpressureEnabled: boolean;
  }
  ```

  There's no LRU eviction, no eviction loop, and nothing "idle to evict." The commit message highlights that the design is admission control only. These comments look like leftovers from an earlier iteration and
  will mislead the next reader. Quick rewrite to match the actual semantics.

mlennie (approved):

kylenesbit:
Both items addressed in `e2fc4bc`.

**1. Coverage gap on lazy package loads** — took the preferred path. The gate is now a single choke point inside `Environment.assertCanAdmitNewPackage`, called from:

- `Environment.getPackage` at the cache-miss / reload branch — covers every controller that lazy-loads (`ModelController`, `ConnectionController`, `QueryController`, `DatabaseController`, etc.) without any of them needing to know about the governor.
- `Environment.addPackage` at the top, *after* the 404 existence check so a permanent missing-artifact failure isn't masked as a transient 503.

`EnvironmentStore` owns the single governor instance and propagates it to every `Environment` via a new `setMemoryGovernor` setter (called once from `server.ts`); environments created after the setter — including via `addEnvironment` — pick it up at construction. The redundant controller-level gates have been removed, so `PackageController` is back to its pre-PR shape.

The `{ allowAdmission: true }` opt-out lives on `Environment.getPackage` / `addPackage` for future internal callers (warmup probes, etc.) that genuinely cannot tolerate 503s. No public endpoint sets it today.

**2. Stale JSDoc on `MemoryGovernorConfig`** — rewritten to match the admission-only design. No more references to LRU, eviction loops, or "nothing idle to evict."

New test coverage: `environment_admission.spec.ts` with 8 unit tests against a stub governor — cache-miss → 503, reload → 503, addPackage post-404 → 503, missing-dir → 404 (not 503), `allowAdmission` bypass, governor flip re-admits, detach reverts to legacy admit-everything. All 61 governor + config + admission tests pass; lint, prettier, typecheck clean.

PR #744, opened 2026-05-16
feat: Add per-query givens overrides to REST API and MCP tool

adamribaudo-ff:
## Summary

Threads a `givens` parameter through the query execution and compile pipelines so API callers can override model-level [given](https://docs.malloydata.dev/documentation/experiments/givens) defaults at query time. The Malloy runtime (`@malloydata/malloy@0.0.394`) already supports `givens` in `.run()` and `.getSQL()` — Publisher was the missing link.

### Query execution endpoint
- **`api-doc.yaml`** — Add optional `givens` property (`object`, `additionalProperties: true`) to the `QueryRequest` schema
- **`packages/server/src/service/model.ts`** — Accept `givens?: Record` in `getQueryResults()` and pass it to `runnable.run({ rowLimit, givens })`
- **`packages/server/src/controller/query.controller.ts`** — Thread `givens` through `getQuery()` to `model.getQueryResults()`
- **`packages/server/src/server.ts`** — Extract `req.body.givens` from the POST request body and pass to the controller
- **`packages/server/src/mcp/tools/execute_query_tool.ts`** — Add `givens` to the `malloy_executeQuery` Zod schema and pass to both `getQueryResults()` call sites

### Compile endpoint
- **`api-doc.yaml`** — Add optional `givens` property to the `CompileRequest` schema
- **`packages/server/src/service/environment.ts`** — Accept `givens` in `compileSource()` and pass to `queryMaterializer.getSQL({ givens })`
- **`packages/server/src/controller/compile.controller.ts`** — Thread `givens` through `compile()` to `environment.compileSource()`
- **`packages/server/src/server.ts`** — Extract `req.body.givens` from the compile POST request body and pass to the controller

### Generated
- **`packages/server/src/api.ts`** — Regenerated from updated spec

## Usage

```bash
curl -X POST ".../api/v0/environments/{env}/packages/{pkg}/models/account.malloy/query" \
  -H "Content-Type: application/json" \
  -d '{
    "query": "run: account -> { group_by: name; aggregate: count }",
    "givens": {"IS_LATEST_SNAPSHOT": false}
  }'
```

Compile with givens:
```bash
curl -X POST ".../api/v0/environments/{env}/packages/{pkg}/models/account.malloy/compile" \
  -H "Content-Type: application/json" \
  -d '{
    "source": "run: account -> { group_by: name; aggregate: count }",
    "includeSql": true,
    "givens": {"IS_LATEST_SNAPSHOT": false}
  }'
```

MCP tool:
```json
{
[...]

PR #758, opened 2026-05-19
feat: per-query givens overrides (graft of #744)

Sha-Bang:
Grafted from #744 — original work by @adamribaudo-ff.

Closes #744.

### Why a graft

PR #744 was BEHIND main and "Allow edits from maintainers" wasn't enabled, so the head branch on `form-function-co/malloy-publisher` couldn't be updated from this side. Rather than block on a branch update, the 5 original commits are cherry-picked here onto a fresh branch from current main. Authors are preserved on every commit (`Adam Ribaudo <aribaudo@form-function.co>`).

Since all commits in this PR share the same author, GitHub's squash-author-preservation rule will keep @adamribaudo-ff as the Author of the resulting squash commit on main.

### Verification (carried over from #744 review)

- `bun run build:server` — green
- `bun run test:server` — unit 389 / integration 165, 0 fail
- Smoke test against a `given: target_code :: string is 'WN'` model on the `faa` sample:
  - `POST /…/compile` with `givens` → substituted literal lands in returned SQL (`WN`, `AA`)
  - `POST /…/query` with and without `givens` → correct row back; default applied when omitted
  - MCP `malloy_executeQuery` with and without `givens` → same (`WN`, `AA`, `DL`)

Composition with the legacy `#(filter)` path is intact: `model.ts` passes `givens` to `runnable.run` unconditionally and the existing filter-clause injection block is untouched. `bypassFilters` correctly scopes only to the legacy path.

### Notes
- One pre-existing behavior worth flagging (not introduced by this PR): when `includeSql: true` is requested on a model whose given has neither a default nor a supplied value, `getSQL` throws and is swallowed by the existing try/catch at `environment.ts:251`, so the caller sees `status: success` with no `sql` field. Worth a follow-up to distinguish "no runnable query" from "unsatisfied given" in the compile response.
- The 3 formatter cleanup commits from #744 are preserved, but produce the same prettier-config drift that's recurred across several recent regen rounds on `api.ts`. Not something to fix here.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -2789,6 +2789,13 @@ components:
           type: boolean
           default: false
           description: When true, skip server-side \#(filter) injection entirely.
+        givens:
+          type: object
+          description: Per-query given values that override model defaults. Keys are
+            given names declared in the model's `given:` block. Values must match
+            the declared type (string, number, boolean, date, etc.). See Malloy
+            givens documentation for accepted value shapes.
+          additionalProperties: true
 
     NotebookCell:
       type: object
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3466,6 +3466,13 @@ components:
           description: If true, returns the generated SQL alongside compilation results
             (only available when compilation succeeds and the source contains a
             runnable query).
+        givens:
+          type: object
+          description: Per-query given values that override model defaults. Keys are
+            given names declared in the model's `given:` block. Values must match
+            the declared type (string, number, boolean, date, etc.). See Malloy
+            givens documentation for accepted value shapes.
+          additionalProperties: true
       required:
         - source
 
--- a/packages/server/src/api.ts
+++ b/packages/server/src/api.ts
@@ -3,1189 +3,1192 @@
  * Do not make direct changes to the file.
  */
 
+
 export interface paths {
-   "/status": {
-      /**
-       * Get server status and health information
-       * @description Returns the current status of the Malloy Publisher server, including initialization state,
-       * available environments, and server timestamp. This endpoint is useful for health checks and
-       * monitoring server availability.
-       */
-      get: operations["get-status"];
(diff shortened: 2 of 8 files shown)


packages/server/src/api.ts
```diff
@@ -855,168 +870,661 @@ export interface components {
             /** @description PostgreSQL connection for DuckLake metadata catalog */
             postgresConnection: components["schemas"]["PostgresConnection"];
          };
+         /** @description Raw Malloy notebook with unexecuted cell contents */
```
Sha-Bang, line 873: See comment in description of PR for prettier-config drift

Sha-Bang:
PR1 from design: https://github.com/malloydata/publisher/blob/main/docs/givens-migration-plan.md

api-doc.yaml
```diff
@@ -2789,6 +2789,13 @@ components:
           type: boolean
           default: false
           description: When true, skip server-side \#(filter) injection entirely.
+        givens:
+          type: object
+          description: Per-query given values that override model defaults. Keys are
```
kylenesbit, line 2794: Can we create a givens type and reuse it in the request and notebook?  I think it's more readable that way.

Sha-Bang: Good call agreed it reads better. I'd like to defer this to PR 2 of the project plan (https://github.com/malloydata/publisher/blob/main/docs/givens-migration-plan.md) (givens introspection), which adds a Given (singular) schema for the declaration metadata anyway. Introducing both Given and Givens (plural — the runtime value map this PR adds) in the same PR keeps the naming distinction visible and lets the docs explain both together. PR 3 (notebook-cell endpoint) will then $ref the same Givens from PR 2.

This also keeps the authorship clean between @adamribaudo-ff 's authorship and myself clean.

sagarswamirao (approved):

PR #762, opened 2026-05-19
chore: stop tracking generated api.ts

Sha-Bang:
## Summary

`packages/server/src/api.ts` is already in `packages/server/.gitignore` (line 4) but was tracked in git from before that rule landed. Git keeps tracking files even after they're added to `.gitignore`, which means every regen produces an unintended diff. This commit untracks it via `git rm --cached`.

Per @sagarswamirao's review comment: *"Can you add this file to the .gitignore?"* — already gitignored, just needed untracking.

## What changes

- `git rm --cached packages/server/src/api.ts` — file stays on disk locally, just stops being tracked
- No code change, no behavior change
- Matches the existing pattern for `packages/sdk/src/client/` (also generated, also gitignored)

## Why this fixes the underlying problem

Before: `api-doc.yaml` is the source of truth, but `api.ts` is also committed. Every contributor who runs `bun run generate-api-types:server` (or whose build does it implicitly) sees an `api.ts` diff that they either commit defensively or stash. That's how PR #758 left main with a stale `api.ts` (an old `degraded` `operationalState` value lingered after #751 reverted the yaml side).

After: only `api-doc.yaml` is tracked. `api.ts` regenerates as part of the build, same as `src/client/*` in the SDK package.

## Fresh-clone contract

`bun run build:server` runs `generate-api-types` as its first step, so any contributor whose workflow involves building (always, in practice) gets `api.ts` automatically. Same implicit contract the SDK has already operated under.

If a contributor wants to run `bun run test:server` cold on a fresh clone, they need to build first. That's already true today for the SDK; making the server consistent.

## Test plan

- [x] `git rm --cached packages/server/src/api.ts` — file remains on disk locally
- [x] `bun run build:server` regenerates api.ts (verified: same content as before the untrack)
- [ ] CI passes


sagarswamirao:
Can you add this file to the .gitignore?

Sha-Bang:
Good call @sagarswamirao — the file is actually already in `packages/server/.gitignore` (line 4), it was just tracked from before that rule landed. Updated the PR: now removes `api.ts` from the index via `git rm --cached` instead of regenerating it. Same content stays on disk locally, regenerates via the existing `generate-api-types` build step. Matches how `packages/sdk/src/client/` is handled.

sagarswamirao:
Josh accidentally added it and later removed it here => https://github.com/malloydata/publisher/commit/181dd50368db8828cada016883dc1b03eed9fe3c#diff-98d1d6285dfa32fb5d44f2197b2338503d672dbbe2098b242f75d236298385c3

But on a later commit it was commited by mistake

We shouldn't be commiting these generated files

sagarswamirao (approved):

sagarswamirao:
Have approved it. Can you update the PR title and merge it?

PR #763, opened 2026-05-20
feat: add givens support to notebook-cell GET endpoint

Sha-Bang:
## Summary
- Adds an optional `givens` query parameter to the `execute-notebook-cell`
  GET endpoint (JSON-encoded record of given values), parallel to the
  existing `filter_params`.
- Threads `givens` through `model.controller → Model.executeNotebookCell
  → runnableToExecute.run({ rowLimit, givens })` so notebook cells honor
  per-query overrides.
- PR 3 of the server-side givens migration; follows #758 (per-query
  givens on POST /query and /compile) and #761 (introspection on
  CompiledModel / Source).
  
  PR 3 as outlined in: https://github.com/malloydata/publisher/blob/main/docs/givens-migration-plan.md

## Test plan
- [x] `bun run lint:server` clean
- [x] `bun run test:unit` — 487 pass / 0 fail
- [x] `bun test src/service/filter_integration.spec.ts` — 30 pass / 0 fail
- [x] New tests:
  - `applies givens to notebook cell execution` — verifies a `given:`
    override flows through cell execution
  - `applies givens on a model that also declares #(filter) annotations`
    — verifies the independence invariant (givens still work when the
    source carries both annotation types)
- [x] Manual smoke: `curl ".../cells/2?givens=%7B%22target_code%22%3A%22AA%22%7D"`
      returns AA-scoped row, SQL contains `'AA'`
- [x] Manual smoke: malformed givens (`givens=not-json`) returns HTTP 400

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -2024,6 +2024,12 @@ paths:
             enum:
               - "true"
               - "false"
+        - name: givens
+          in: query
+          description: JSON-encoded given values keyed by given name
+          required: false
+          schema:
+            type: string
       responses:
         "200":
           description: Cell execution result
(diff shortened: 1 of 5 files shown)


sagarswamirao (approved):
LGTM :)

PR #767, opened 2026-05-20
feat(server): require a package-load worker pool to serve packages

kylenesbit:
## Summary

The Kubernetes liveness probe on `/health/liveness` was timing out for multi-second package loads because Malloy compile is pure JavaScript running on V8's main event loop. PR #767's first cut moved only per-file `.malloy`/`.malloynb` compile to workers; the rest of `Package.create` — manifest read, source/query extraction, and the first-query rehydration that still recompiled on the main thread — could still blow the probe budget on a large package. **Plus**, `Package.reloadAllModels` (the post-build manifest reload path) was still compiling on the main thread.

This refactor makes the worker pool a **hard dependency** and routes both load and reload through it, leaving no main-thread compile in the production hot path:

- **`package_load_pool.ts`** (renamed from `compile_pool.ts`) drives a `PackageLoadPool` of N `worker_threads` workers. `PACKAGE_LOAD_WORKERS` must be a positive integer; defaults to 1 when unset. Setting it to **0** — the old in-process fallback — **throws at boot**. There is no fallback path. Running Malloy compile on the main event loop is what tripped the probe in the first place; we don't want a knob that re-enables that footgun. The env var was renamed from `MALLOY_COMPILE_WORKERS` to match the new "package load" framing (the worker drives the whole load, not just compile).
- **`package_load_worker.ts`** (renamed from `compile_worker.ts`) reads `publisher.json`, compiles every `.malloy`/`.malloynb` model in the package, **precomputes `modelDefToModelInfo(modelDef)`**, and ships back a structured-clonable `LoadPackageOutcome` with every `modelDef`, `modelInfo`, `sourceInfos`, `sources`, `queries`, `filterMap`, `givens`, dataStyles, and per-cell notebook payload the main thread needs to reconstitute a live `Package`. The worker owns **zero** native handles — every `lookupConnection`, schema fetch, SQL-block schema, and non-`file://` URL read proxies back to the main thread's live connection pool over RPC.
- Embedded `.parquet`/`.csv` probes stay on the main thread (`Package.readDatabases`), which already reuses the package's shared DuckDB connection (#772). Probes run in parallel with the worker compile via `Promise.all`, so wall-clock latency is unchanged for `.parquet`-heavy packages while keeping all native-DB `dlopen` on the main isolate (sidesteps Bun crash 0x20131 where `duckdb-native` can't be loaded into a second isolate).
- **`Package.loadViaWorker` rewraps pool-infrastructure failures** (worker crash, RPC timeout, pool shutting down) as `ServiceUnavailableError`, which the HTTP error mapper translates to **503 Service Unavailable** — transient, retryable — instead of an opaque 500. Real compile errors (`MalloyError` / `ModelCompilationError`) keep their existing 4xx mapping.
- **`Package.reloadAllModels` now also goes through the pool.** Previously the last main-thread compile in the prod hot path: a manifest reload on a large package could block the event loop the same way the initial load did. Reload semantics keep the historical "failed models become placeholders in the map" behaviour (unlike the initial load, which aborts on first compile error). Same 503 mapping for pool-infra failures.
- **`Model.fromSerialized`** eagerly hydrates a `ModelMaterializer` (and per-cell runnables for notebooks) from the worker's pre-compiled `modelDef` / `queryDef` via `Runtime._loadModelFromModelDef` / `ModelMaterializer._loadQueryFromQueryDef`. These are constant-time wraps, **not recompiles** — first-query latency no longer pays a recompile and the resulting `Model` is interchangeable with one from `Model.create`. Worker-emitted `modelInfo` is threaded through too, so `getStandardModel()` / `getNotebookModel()` no longer recompute `modelDefToModelInfo` on every API hit.
- `PackageLoadPool` folds in queue + per-worker active-job cap (= 1), ready/exit handling, separate spawn vs job timeouts, and terminate-worker-on-timeout (no zombie compile burning CPU after the caller rejects). The wall-clock timeout knob is also renamed: `PACKAGE_LOAD_JOB_TIMEOUT_MS`.
- `server.ts` eagerly constructs the pool at startup so a bad `PACKAGE_LOAD_WORKERS` value fails the boot rather than surfacing on the first package load.
- New `service/given.ts` centralises the `MalloyGiven` duck-typed shape and the `malloyGivenToApi` conversion so the main-thread `Model` and the worker share one source of truth instead of drift-prone copies on each side.

## Operator-visible config

| Env var | Behaviour |
| --- | --- |
| `PACKAGE_LOAD_WORKERS` | Positive int (default `1`). `0` / negative / non-numeric → publisher refuses to boot. |
| `PACKAGE_LOAD_JOB_TIMEOUT_MS` | Wall-clock cap per worker load job (default `120000`). Worker is terminated on timeout. |

The matching terraform change to rename `MALLOY_COMPILE_WORKERS` → `PACKAGE_LOAD_WORKERS` in `~/infrastructure` is staged but lives in the infra repo (separate PR).

**Sizing note:** if your operator profile expects concurrent package load + reload (e.g. a manifest build commits while a fresh package is also being loaded), bump `PACKAGE_LOAD_WORKERS` to 2 so the two jobs don't queue against each other.

## Test plan

- [x] `package_load_pool.spec.ts` — env-var validation, constructor validation, real-worker load (single-model package), in-band per-model compile error, csv ignored by worker (probing is main-thread), shutdown rejects new jobs.
- [x] `package_worker_path.spec.ts` — integration: `Package.create` on the worker path produces a live `Package`, first-query hydration runs without a recompile, per-model compile failures throw out of `Package.create`, **pool-infrastructure failures throw `ServiceUnavailableError` (HTTP 503)**.
- [x] `package.spec.ts` — pre-existing in-process integration tests still pass against the worker path (assertions are over `Package`'s externally-visible behaviour, not the path).
- [x] `manifest_service.spec.ts` — `reloadAllModels` callers still pass (manifest service stubs the package, so reload-path changes are invisible to it).
- [x] `tsc --noEmit`, `eslint`, `bun test src` — 14 failures match `origin/main` and are unrelated (Bun's `monitorEventLoopDelay` from PR #768 isn't implemented in Bun's test runtime).

## Follow-ups (out of scope)

- `materialization_service.ts:721` still calls `Model.getModelRuntime(...) → runtime.loadModel(...).getModel()` on the main thread to read `##! experimental.persistence` tags. That's a separate compile-on-main-thread spot, lower priority because it runs only on the explicit build trigger. Route through the pool in a follow-up.

### DIFF
--- a/packages/server/build.ts
+++ b/packages/server/build.ts
@@ -5,7 +5,15 @@ fs.rmSync("./dist", { recursive: true, force: true });
 fs.mkdirSync("./dist");
 
 await build({
-   entrypoints: ["./src/server.ts", "./src/instrumentation.ts"],
+   // package_load_worker.ts is bundled as a SEPARATE entrypoint so it
+   // can be loaded by `new Worker(...)` at runtime. It must NOT be
+   // inlined into server.mjs (workers can't share module state with
+   // the parent process — they get their own JS realm).
+   entrypoints: [
+      "./src/server.ts",
+      "./src/instrumentation.ts",
+      "./src/package_load/package_load_worker.ts",
+   ],
    outdir: "./dist",
    target: "node",
    format: "esm",
@@ -48,6 +56,27 @@ fs.copyFileSync(
 // Rename ESM outputs to .mjs so both Node and Bun can execute them
 fs.renameSync("./dist/server.js", "./dist/server.mjs");
 fs.renameSync("./dist/instrumentation.js", "./dist/instrumentation.mjs");
+// Bun emits package_load_worker into its source-relative subdir;
+// flatten so package_load_pool.ts's `resolveWorkerScript()` finds it
+// as a sibling of server.mjs. The path layout match is intentional —
+// keep these two in sync or the worker pool throws at boot (the
+// in-process fallback was removed; missing worker = no service).
+if (fs.existsSync("./dist/package_load/package_load_worker.js")) {
+   fs.renameSync(
+      "./dist/package_load/package_load_worker.js",
+      "./dist/package_load_worker.mjs",
+   );
+   try {
+      fs.rmdirSync("./dist/package_load");
+   } catch {
+      /* directory may be non-empty if Bun produced sourcemaps; leave it */
+   }
+} else if (fs.existsSync("./dist/package_load_worker.js")) {
+   fs.renameSync(
+      "./dist/package_load_worker.js",
+      "./dist/package_load_worker.mjs",
+   );
+}
 
 // Add shebang to server.mjs for npx/bunx compatibility
 const serverJsPath = "./dist/server.mjs";
--- a/packages/server/src/health.ts
+++ b/packages/server/src/health.ts
@@ -143,6 +143,21 @@ export async function performGracefulShutdownAfterDrain(
       /* do nothing */
    }
 
+   try {
+      // Drain in-flight compiles and terminate worker_threads before
+      // we exit so a slow compile doesn't leave orphan worker
+      // processes. Lazy-imported to avoid pulling the pool module
+      // into the health.ts dep graph for tests that don't exercise
+      // the compile path.
+      const { getPackageLoadPool } = await import(
+         "./package_load/package_load_pool"
+      );
+      await getPackageLoadPool().shutdown();
+      logger.info("Package-load worker pool shut down");
+   } catch (_error) {
+      /* do nothing */
+   }
+
    if (shutdownGracefulCloseTimeoutSeconds > 0) {
       logger.info(
          `Waiting ${shutdownGracefulCloseTimeoutSeconds} seconds after server close before exit...`,
--- a/packages/server/src/package_load/package_load_pool.spec.ts
+++ b/packages/server/src/package_load/package_load_pool.spec.ts
@@ -1,0 +1,252 @@
+/**
+ * Unit + light integration tests for the PackageLoadPool.
+ *
+ * The pool runs real `worker_threads` workers under the hood. These
+ * tests intentionally exercise that path so we catch regressions in
+ * RPC routing, queueing, lifecycle, and error propagation that
+ * wouldn't surface in a pure mock.
+ *
+ * Pool reuse strategy
+ * -------------------
+ * The "real worker" tests share a single `PackageLoadPool` across
+ * cases via Bun's `beforeAll`/`afterAll`. The worker itself owns no
+ * native handles (all duckdb work runs on the main thread); we still
+ * share to keep per-test overhead low and to match the production
+ * deployment, where the pool spawns workers up-front and reuses them.
+ */
+import {
+   afterAll,
+   afterEach,
+   beforeAll,
+   beforeEach,
+   describe,
+   expect,
+   it,
+} from "bun:test";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+   PackageLoadPool,
+   __setPackageLoadPoolForTests,
+   getPackageLoadWorkerCount,
+} from "./package_load_pool";
+
+// ──────────────────────────────────────────────────────────────────────
+// getPackageLoadWorkerCount — env var parsing
+// ──────────────────────────────────────────────────────────────────────
+
+describe("getPackageLoadWorkerCount", () => {
+   const ORIGINAL = process.env.PACKAGE_LOAD_WORKERS;
+
+   afterEach(() => {
+      if (ORIGINAL === undefined) {
+         delete process.env.PACKAGE_LOAD_WORKERS;
+      } else {
+         process.env.PACKAGE_LOAD_WORKERS = ORIGINAL;
+      }
+   });
+
+   it("defaults to 1 worker when env unset", () => {
+      delete process.env.PACKAGE_LOAD_WORKERS;
+      expect(getPackageLoadWorkerCount()).toBe(1);
+   });
+
+   it("throws when env value is non-numeric", () => {
+      process.env.PACKAGE_LOAD_WORKERS = "not-a-number";
+      expect(() => getPackageLoadWorkerCount()).toThrow(/positive integer/);
+   });
+
+   it("throws when env value is negative", () => {
+      process.env.PACKAGE_LOAD_WORKERS = "-2";
+      expect(() => getPackageLoadWorkerCount()).toThrow(/positive integer/);
+   });
+
+   it("throws when env value is 0 (no in-process fallback)", () => {
+      process.env.PACKAGE_LOAD_WORKERS = "0";
+      expect(() => getPackageLoadWorkerCount()).toThrow(/positive integer/);
+   });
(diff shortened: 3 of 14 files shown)


Sha-Bang:
Independent cross-model review (Claude + Codex). Both reviewers agree on the
headline: **pool size N does not bound concurrent compiles to N.** With your
example — 5 large requests, `MALLOY_COMPILE_WORKERS=2` — you do not get 2
active + 3 queued. You get up to 5 jobs interleaving across 2 workers, with a
likely 4/1 skew on the first worker.

### 1. No concurrency limit; broken cold-start load balancing

`compile_pool.ts:215-265` + `compile_worker.ts:626-651`

`inFlight.add(jobId)` happens AFTER `acquireWorker()` returns, so concurrent
callers all observe `inFlight.size === 0` and tiebreak to `alive[0]`. The
worker also accepts new compile messages via `void runJob(...)` while another
job is in flight, so one worker can hold multiple large CPU-bound compiles.

For your "5 large requests" scenario: w1 gets {j1, j3, j4, j5}, w2 gets {j2}.

**Fix:** real queue, cap active jobs per worker to 1 (CPU-bound work doesn't
multiplex usefully inside one event loop).

### 2. `pw.ready` has no error / exit / timeout path

`compile_pool.ts:337-338`

If the worker fails to boot (Bun can't load the bundled `.mjs`,
`resolveWorkerScript` points at a missing file, worker crashes before posting
`ready`), every subsequent compile awaits `pw.ready` forever. The job timeout
hasn't been armed yet, and `Model.create()`'s in-process fallback never fires
because `pool.compile()` never rejects.

**Fix:** attach `worker.once('error')` and `worker.once('exit')` to reject the
ready promise; add a spawn-timeout.

### 3. Job timeout rejects the caller but does not stop the worker

`compile_pool.ts:230-237`, `476-481` + `model.ts:294-317`

On timeout, `failJob` deletes from `this.jobs` but never touches
`worker.inFlight`, and the worker keeps burning CPU on the doomed compile.
Worse: `Model.create` catches the timeout and falls back to **in-process**
compile on the Bun main thread. The worker is still running the same compile,
so you now have 2× the CPU and re-introduce exactly the liveness-probe risk
this PR exists to avoid.

**Fix:** on timeout, terminate the worker and let the pool respawn it; do NOT
fall back to main while the worker job is alive. Also clean up
`worker.inFlight` so phantom entries don't permanently inflate the least-busy
ranking.

### 4. First query re-compiles on the main thread

`model.ts:336-391` + the materializerBuilder path

Worker-compiled models build a fresh `ModelMaterializer` lazily on first query
via `getModelRuntime` + `getStandardModelMaterializer`. The runtime is not
warmed with the worker's `modelDef`, so the first `loadQuery()` triggers a
full Malloy compile on the Bun main thread.

**For query traffic, this PR doesn't move compile off the event loop — it
delays it.** Worth either pushing query setup through a worker too, or scoping
the PR description so the liveness win is explicitly limited to package load
rather than query execution.

---

Happy to provide the P2 list (NaN timeout, no `resourceLimits`, no upper cap
on worker count, structured-clone completion burst, worker-death race after
`acquireWorker`, postMessage throw leak) if useful. Architecture is right;
items 1-4 are the implementation fixes that would actually deliver the goal.

kylenesbit:
Pushed [`93e1264`](https://github.com/malloydata/publisher/pull/767/commits/93e1264) extending the pool to handle inline-source compiles and routing one more main-thread Malloy bottleneck through it.

## What's new in this commit

`Package.getDatabaseInfo` (called per `.parquet`/`.csv` from `readDatabases`, via `Promise.all`) previously ran **two** synchronous Malloy compiles per data file on the main thread:

```ts
const model = runtime.loadModel("source: temp is duckdb.table('…')");
const modelDef = await model.getModel();        // (1) schema probe compile
const runner = model.loadQuery("run: temp -> { aggregate: c is count() }");
const result = await runner.run();              // (2) trivial count compile + exec
```

For a package with `N` data files, that's `2N` main-thread compiles back-to-back before `Package.create` returns — orthogonal to the `.malloy` model compiles this PR originally targeted. This commit fixes both.

## Diff at a glance

```
packages/server/src/compile/compile_pool.spec.ts |  65 ++++++++++++
packages/server/src/compile/compile_pool.ts      |  95 ++++++++++++++---
packages/server/src/compile/compile_worker.ts    |  42 ++++++--
packages/server/src/compile/protocol.ts          |  23 +++-
packages/server/src/service/package.ts           | 129 ++++++++++++++++++++---
5 files changed, 318 insertions(+), 36 deletions(-)
```

## Key changes

1. **Protocol extension** — `CompileJobRequest` gains `inlineSource?: string` and `importBaseURL?: string`. `modelPath` becomes optional. Exactly one of the two is required (worker enforces).
2. **Worker branches on `inlineSource`** — when set, calls `runtime.loadModel(source, { importBaseURL })` directly. Schema-fetch RPC, source/query extraction, and givens all work unchanged.
3. **`pool.compileInline()`** — new entry point alongside `pool.compile()`. Both share a refactored `dispatchJob()` helper for jobId allocation, JobContext registration, timeout setup, and worker `postMessage`.
4. **`Package.getDatabaseInfo` refactor**:
   - Schema probe: routed through `pool.compileInline()` when the pool is enabled; falls back to in-process on non-compile errors (worker exit, RPC timeout) and on the kill switch.
   - Row count: always direct `conn.runSQL("SELECT count(*)::BIGINT AS row_count FROM '<path>'")`. The previous Malloy compile produced exactly this SQL — bypassing Malloy here eliminates a per-file compile unconditionally. Single quotes in the path are doubled for filename-injection safety.

## API surface

Response shape (`{ name, rowCount, columns }`) is identical. `columns` continues to use Malloy type strings, not DuckDB native types — only the place where the compile runs changes.

## Validation

- `tsc --noEmit` clean
- `eslint` clean
- `bun test src/` 504 pass, 0 fail (was 502 before; 2 new tests cover `compileInline()` happy path + Malloy error propagation)
- `bun run build:server-only` produces clean `server.mjs` and `compile_worker.mjs`

## Behavioural notes

- With `MALLOY_COMPILE_WORKERS=0` (default), the schema compile runs in-process exactly as before. Only the row-count Malloy compile is always-bypassed — that's a no-op behaviourally since `count()` over a `duckdb.table` source generated `SELECT count(*) FROM '<path>'` either way.
- The kill-switch still gives you exactly today's behaviour for everything operationally observable; the worker pool's effect is now also visible on `Package.create` cold paths, not just model compiles.

kylenesbit:
Rolled back [`93e1264`](https://github.com/malloydata/publisher/pull/767/commits/93e1264). The branch is back to a single commit (`fe6ccd6`) — same scope as originally reviewed (`.malloy` model compiles routed through the worker pool, with the file-backed protocol only).

The inline-source `compileInline()` extension and the `Package.getDatabaseInfo` refactor described in the previous comment are no longer part of this PR. The earlier comment's diff and behaviour notes can be disregarded; I'll open a separate PR for that work when we're ready, so #767 stays focused on the original change.

PR #769, opened 2026-05-20
feat(server): implement schema introspection via worker pool

sagarswamirao:
This commit introduces a new schema introspection mechanism that offloads the processing to a dedicated worker pool, isolating DuckDB's native thread pool to prevent memory leaks observed in production. The `SchemaWorkerPool` manages worker threads that handle schema requests, while the `schema_worker.ts` file defines the worker's behavior for processing these requests.

Additionally, the server build process has been updated to include the new `schema_worker.ts` in the output, and a `ProcessStatsReporter` has been added to monitor memory and thread usage, providing insights into the application's performance.

Key changes:
- Added `SchemaWorkerPool` for off-main-thread schema introspection.
- Introduced `ProcessStatsReporter` for monitoring process statistics.
- Updated build configuration to include new worker scripts.

This enhancement aims to improve performance and stability during schema introspection operations.


Sha-Bang:
### Code review

Found 6 issues. The first three are in the threadpool / lifecycle code this PR exists to harden; if you only act on a few, act on these.

**1. [P1] Worker pool double-respawns on a single crash — leaks the replacement worker AND spuriously fails an in-flight request that was succeeding.**

`worker.on("error")` (L187) and `worker.on("exit")` (L211) both call `this.workers[index] = this.spawn(index)`. Node fires `error` *and then* a non-zero `exit` for the same abnormal worker death, so one crash takes both paths and cascades:

1. **Crash.** Original worker dies.
2. **`error` fires.** Rejects the in-flight request the crashed worker was handling (`Req1`), `workerCurrentId.delete(index)`, spawns replacement A into `this.workers[index]`, then calls `drain()` — which pulls `Req2` off the queue, sets `slot.busy = true`, `inFlight.set(req2.id, req2)`, **`workerCurrentId.set(index, req2.id)`**, and posts the message to A.
3. **`exit` fires** for the *original* crashed worker. Reads `workerCurrentId.get(index)` and gets **`req2.id`** (set by the drain a moment ago — slot-keyed tracking can't tell the difference between the dead worker and A), then rejects `Req2` with `SchemaWorker exited with code N` even though A is alive and currently processing it. Overwrites the slot with replacement B. Calls drain again, hands the next queued request to B.

Net per crash:
- **`Req1`** correctly rejected.
- **`Req2`** spuriously rejected with a misleading "worker exited" error even though A is alive and processing it — the caller sees a failure for work that was succeeding.
- **`A`** is orphaned with its DuckDB connection and native thread pool still live; when A eventually finishes `Req2` and posts the result, the message handler at L177 hits `if (!req)` and logs `"response for unknown request"`, silently dropping the schema result.
- **`B`** is in the slot and also alive. Total live workers = `pool_size + N_crashes` until process exit, which is the exact leak class this PR exists to fix.

Fix: respawn only in the exit handler (abnormal exit follows abnormal error, and `code !== 0` is already the gate there), OR add a per-spawn id and have exit no-op when `this.workers[index]` no longer points to the worker it fired for.

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/schema_worker_pool.ts#L185-L235

**2. [P1] `SchemaWorkerPool.stop()` and `ProcessStatsReporter.stop()` are never wired into graceful shutdown.**

`performGracefulShutdownAfterDrain` closes the HTTP servers, calls `shutdownSDK()`, and closes the logger, but the new worker pool and stats reporter started at `server.ts:178-179` are not stopped on SIGTERM. Workers get abandoned mid-query and their DuckDB native threads leak past process exit — which defeats "worker exit → native threads die with it" stated in `schema_worker_pool.ts`. Add both `stop()` calls into `performGracefulShutdownAfterDrain` alongside `shutdownSDK()`. Same omission pattern as PR #754 (winston-logger close).

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/health.ts#L113-L160

**3. [P1] `packageMalloyConfig` is declared but never assigned — the error-path DuckDB release is dead code.**

`let packageMalloyConfig: MalloyConfig | undefined;` is declared at the top of `Package.create`'s try block (L90) but no statement ever assigns it. The `if (packageMalloyConfig) { await packageMalloyConfig.shutdown("close"); … }` block in the catch (L181-189) therefore never executes, so the "release package-local DuckDB on failure" intent is silently a no-op. Either assign `packageMalloyConfig` where the package-local config is built, or remove the dead cleanup block.

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/package.ts#L86-L195

**4. [P2] Health metrics go to `logger.info` instead of OTel `ObservableGauge`s scraped at `/metrics`.**

The project already runs a `PrometheusExporter` (`instrumentation.ts:4`) and serves `/metrics` (`server.ts:330`); `PackageMemoryGovernor` already publishes `publisher_process_rss_bytes` and `publisher_memory_backpressure_active` as observable gauges (`package_memory_governor.ts:78,90`); PR #768 followed the same pattern for event-loop lag. The new reporter re-reads RSS/heap/`Threads`/`VmSize`/`VmPeak`/context-switches and emits them as one `logger.info("process stats", …)` line per tick, duplicating RSS in two channels and putting the *most diagnostic* signal for the 466-thread leak (thread count) into logs where Grafana can't plot or alert on it. Same applies to `logger.info("Memory governor status", …)` newly added inside `EnvironmentStore.getStatus()` — every UI poll of `/api/v0/status` now writes a log line for data Prometheus already scrapes. Move the new signals into `ObservableGauge` instruments (`publisher_process_threads`, `publisher_process_vm_rss_bytes`, etc.), and drop the per-`/status`-hit log.

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/process_stats_reporter.ts#L130-L170

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/environment_store.ts#L703-L718

**5. [P2] `Promise.allSettled` silently changes package-load semantics from fail-fast to partial success.**

`readDatabases` used to surface any introspection failure up to the caller; the new code `Promise.allSettled`s every database and only logs failures via `logger.warn`. Skipping a corrupt parquet so one bad file doesn't break a whole package is reasonable, but right now the only signal that a database was dropped is a warn log — `Package`'s returned `databases[]` just silently has fewer entries than `getDatabasePaths()` returned. If clients rely on knowing a database failed to load, surface it in the API response (e.g., an `error` field on the entry, or a package-level `degraded` flag) rather than only in logs.

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/package.ts#L437-L475

**6. [P2] Single quote in a database filename breaks the inline Malloy source string in the worker.**

`source: temp is duckdb.table('${normalizedPath}')` is built by string interpolation without escaping. A perfectly legal filename like `O'Brian data.parquet` produces a malformed source and the introspection fails with an opaque parse error. The pre-PR code had the same shape, but the new worker is a fresh opportunity to escape `'` → `''` (DuckDB's standard quote-doubling) before interpolation.

https://github.com/malloydata/publisher/blob/c31070ac48dc569d3bd086a558ac1a8814e93a13/packages/server/src/service/schema_worker.ts#L74-L82

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #778, opened 2026-05-22
feat(server): cap rows returned by ad-hoc connection SQL queries

kylenesbit:
## Summary

Step 1 of an incremental OOM mitigation for the publisher. Today the ad-hoc connection SQL endpoints (`/connections/.../sqlQuery` and the deprecated `queryData` GET) call `Connection.runSQL` and immediately `JSON.stringify` the entire result, with no row or byte bound.

The Malloy connectors don't give us a memory bound either:

- **Postgres**' `runSQL` ignores `RunSQLOptions.rowLimit` entirely — `runPostgresQuery(sql, _pageSize, _rowIndex, deJSON)` discards the page-size args and runs `client.query(sql)`, which buffers everything. Only `runSQLStream` honors `rowLimit` / `abortSignal`.
- **DuckDB**'s `runSQL` reads the full result then does `result.slice(0, rowLimit)` after the fact, so a billion-row scan OOMs before the slice runs.

So the only reliable way to bound row count from the publisher side without yet switching to `runSQLStream` is to rewrite the user's SQL to push a `LIMIT` to the database.

### What this PR does

`ConnectionController.getConnectionQueryData` now rewrites cappable `SELECT` / `WITH` statements as:

```sql
SELECT * FROM (<original sql>) AS __publisher_cap__ LIMIT (cap + 1)
```

The `+1` is a sentinel: if the connection returns `cap + 1` rows we know the original would have overflowed, and the controller fails the request with HTTP `413 PayloadTooLargeError` rather than serializing the response.

Cap defaults to **100,000** via `PUBLISHER_MAX_QUERY_ROWS`; setting it to `0` opts out (useful once Step 2's byte budget is the only thing enforcing the bound).

### SQL classifier

`isCappableSelect` is intentionally conservative — only single, simple `SELECT` / `WITH` statements get wrapped. Pass-through unchanged:

- DDL (`CREATE`, `INSERT`, `UPDATE`, `DELETE`, `DROP`, `ALTER`, `TRUNCATE`)
- Metadata (`SHOW`, `DESCRIBE`, `EXPLAIN`)
- Multi-statement SQL (any unquoted `;` followed by non-whitespace)
- Anything containing the DuckDB connector's `-- hack: split on this` sentinel

The classifier handles leading whitespace, `--` line comments, `/* */` block comments, single-quoted strings, double-quoted identifiers, and Postgres `$tag$` dollar quotes when scanning for unquoted semicolons.

### Known limitations (carried to later steps)

- Non-SELECT and multi-statement queries are still unbounded — Step 2 (byte budget via `runSQLStream`) covers them regardless of statement kind.
- 100k narrow rows is fine; 100k 10 KB-wide rows is ~1 GB. Step 2's `PUBLISHER_MAX_RESPONSE_BYTES` is the actual byte bound.
- Wrapper alias surfaces in DB error messages as `__publisher_cap__`, which is deliberate for greppability.

### Files changed

**New**
- `packages/server/src/sql_helpers.ts` — `isCappableSelect`, `wrapWithRowLimit`
- `packages/server/src/sql_helpers.spec.ts` — 41 unit tests
- `packages/server/src/controller/connection.controller.spec.ts` — 8 controller-level tests with a sinon-stubbed `Connection`

**Edited**
- `packages/server/src/constants.ts` — `DEFAULT_MAX_QUERY_ROWS = 100_000`
- `packages/server/src/config.ts` + spec — `getMaxQueryRows()` env parser, loud failure on malformed input
- `packages/server/src/errors.ts` + spec — `PayloadTooLargeError` → HTTP 413
- `packages/server/src/controller/connection.controller.ts` — wrap, fetch cap+1, throw 413 on overflow
- `api-doc.yaml` — shared `PayloadTooLarge` response component referenced from the four `sqlQuery` / `queryData` operations

**Generated (verified clean)**
- `packages/server/src/api.ts` — regenerated; gitignored
- `packages/python-client/malloy_publisher_sdk/` — `build-python-sdk.sh` produces no diffs (the Python generator doesn't materialize shared response components into client methods)

### Why now

The publisher has been OOM-ing in prod on large ad-hoc SQL queries. The long-term fix is real streaming / pagination through the API; this is the smallest, safest first step that closes the most common OOM path (an unintentionally large `SELECT` on a Postgres or DuckDB connection) without changing any caller contracts beyond a new 413 response.

## Test plan

- [x] `bun test src/sql_helpers.spec.ts src/errors.spec.ts src/config.spec.ts src/controller/connection.controller.spec.ts` — 104 pass
- [x] `bunx tsc -p tsconfig.json --noEmit` — clean
- [x] `bunx eslint` on all touched files — clean
- [x] `bun run generate-api-types` — clean regen with 4 expected `413` entries
- [x] `packages/python-client/scripts/build-python-sdk.sh` — clean regen, 4/4 tests pass, no SDK diffs
- [ ] CI: build, cross-platform-tests, connection-integration-tests, python-sdk
- [ ] Manual: hit `/sqlQuery` with `SELECT * FROM <wide_table>` against a Postgres connection with `PUBLISHER_MAX_QUERY_ROWS=10`, confirm 413
- [ ] Manual: hit `/sqlQuery` with `SHOW TABLES`, confirm passthrough
- [ ] Manual: hit `/sqlQuery` with `PUBLISHER_MAX_QUERY_ROWS=0`, confirm wrap disabled


mlennie:
## Review
  
  Mostly clean Step-1 plumbing (413, env knob, type guards, OpenAPI component, controller spec) but the refactor in 1d6e426 weakens the OOM bound the PR is named for. A few items below; #1 is the only blocker in
  my view.

  ### 1. (blocker) The "trust `RunSQLOptions.rowLimit`" assumption is false for Postgres / DuckDB / Snowflake

  I spot-checked the installed connector source against the refactor commit's claim that "those are upstream connector bugs we no longer work around here":

  * **Postgres** (`@malloydata/db-postgres/dist/postgres_connection.js`, `runPostgresQuery(sql, _pageSize, _rowIndex, deJSON)`): `_pageSize` is unused. The body calls `client.query(sql)` with no LIMIT and buffers
  the entire result. `rowLimit` does nothing.
  * **DuckDB** (`@malloydata/db-duckdb/dist/duckdb_common.js:78-93`): `runRawSQL` materializes the full result, then `if (result.length > rowLimit) result = result.slice(0, rowLimit)`. Memory cost is paid in full
  before the slice.
  * **Snowflake** (`@malloydata/db-snowflake/dist/snowflake_connection.js:100-111`): same shape as DuckDB. `executor.batch` materializes, then `rows.slice(0, rowLimit)`.
  * **BigQuery** (`@malloydata/db-bigquery/dist/bigquery_connection.js:174-202`): actually honors `rowLimit` via `maxResults` on the jobs API. Only connector for which the refactored approach bounds at the wire.

  Concrete consequence: `SELECT * FROM huge_table` against Postgres still buffers the whole table in the connector before `result.rows.length > maxRows` fires. The 413 only saves the subsequent `JSON.stringify`
  pass, not the materialization cost, which was the OOM path called out in the PR description. For DuckDB (shipped by default in `publisher.config.json`) a billion-row scan still OOMs before the `slice` runs.

  The deleted `sql_helpers.ts` rewriter pushed `LIMIT (cap + 1)` into the SQL itself, so the database, not the connector, enforced the bound. That actually closed the OOM. The refactor removed that protection on
  the principle that connectors should honor `rowLimit`, which is correct in the abstract but not currently true.

  Ask: revert to the rewriter (it already had 41 tests), or land an upstream Malloy fix first, or change the PR framing so the bound it actually provides matches the title/description/OpenAPI text. As written,
  "closes the most common OOM path" reads as overstated for the connectors most users hit.

  ### 2. OpenAPI text describes the deleted classifier

  `api-doc.yaml:726-730`:
  
  SELECT/WITH statements that would exceed the cap fail with HTTP 413;
  non-SELECT statements (DDL, SHOW, DESCRIBE, EXPLAIN, multi-statement)
  are passed through unchanged.

  After 1d6e426 there's no SELECT/WITH vs DDL distinction. All four endpoints reference this wording (`:726-730`, `:893`, `:1231`, `:1412`), so the misdescription propagates.

  ### 3. `PUBLISHER_MAX_QUERY_ROWS` is missing from `publisher/README.md`

  New env var, user-visible 413 response. Workspace policy is that env vars land in the README env-var table in the same PR, and user-visible behavior gets a `malloydata.github.io` notebook update under
  `src/documentation/user_guides/publishing/`. Don't see either.

  ### 4. `JSON.parse(options)` is unguarded against non-object JSON

  ```ts
  if (options) {
     runSQLOptions = JSON.parse(options) as RunSQLOptions;
  }

  The CodeQL fix in 9cdb51b validates that options is a string before it gets here, but options: "null" / '"foo"' / '42' / '[1,2,3]' all parse to non-objects. runSQLOptions.abortSignal on null throws → becomes a
  500; on an array, runSQLOptions.rowLimit = ... mutates the array and gets passed to the connector. Worth a typeof === "object" && !Array.isArray && !== null check at the same boundary as the rest of the type
  guards. Same family of risk the CodeQL alert was about.
  
  5. Caller-supplied rowLimit isn't validated

  runSQLOptions.rowLimit = Math.min(
     runSQLOptions.rowLimit ?? maxRows + 1,
     maxRows + 1,
  );
  
  Math.min(NaN, x) === NaN. rowLimit: -1 flows through unchanged. rowLimit: "10" gets coerced. Nothing exploitable, but a Number.isInteger && >= 0 guard at the boundary would be consistent with how strictly we now
   validate sqlStatement / options.

  6. No structured log on the 413 path

  throw new PayloadTooLargeError(...)
  
  Operators investigating an OOM-adjacent incident can't grep logs for "this caller hit the cap on connection X." A one-line logger.warn("Query exceeded row cap", { environmentName, connectionName, maxRows })
  before the throw would help triage and matches the existing logging style in this controller. Log only safe fields per the credential-leak audit rule (no sqlStatement, no options, no connection).

  7. (nit) Spec gap

  The cap-disabled (PUBLISHER_MAX_QUERY_ROWS=0) tests don't cover the case where the caller also supplies their own rowLimit. It should pass through verbatim. One more it.each row.

  What I checked

  Read the diff of all three commits, the post-refactor getConnectionQueryData (packages/server/src/controller/connection.controller.ts:438-510), the four route-handler callers in server.ts, all 11 controller spec
   cases, and the four connector implementations in node_modules/@malloydata/*. Did not run tests locally.
  ```

mlennie (approved):
Looks good. Left a review in a comment

Sha-Bang (approved):
api-doc.yaml claims non-SELECT statements bypass the cap, but the controller applies it uniformly. The new endpoint description says non-SELECT statements (DDL, SHOW, DESCRIBE, EXPLAIN, multi-statement) are passed through unchanged, yet getConnectionQueryData sets runSQLOptions.rowLimit = min(..., maxRows+1) and runs the result.rows.length > maxRows 413 check on every statement — there is no SELECT vs non-SELECT branching. The PR's own spec confirms it: the test forwards non-SELECT statements verbatim with rowLimit applied asserts opts.rowLimit === 11 for SHOW TABLES with cap=10. Since this OpenAPI text generates client SDKs, callers will believe SHOW/DESCRIBE/EXPLAIN bypass the cap and may be surprised by a 413.

Approving because I don't think this will actually break anything, jut a nit on documentation.

PR #783, opened 2026-05-22
feat(server): admission gate, query timeout, and per-pod concurrency cap (OOM mitigation 4-6)

kylenesbit:
## Summary

Completes the OOM-mitigation work begun in #778 (Step 1) and #779 (Steps 2-3) by adding three defense-in-depth guardrails plus operator-facing telemetry across all of them. Each guard opts out via its env var set to `0`.

| Guard | Env var | Default | HTTP on rejection |
| --- | --- | --- | --- |
| Memory admission gate | `PUBLISHER_MAX_MEMORY_BYTES` (+ fractions) | unset | 503 |
| Per-query wall-clock timeout | `PUBLISHER_QUERY_TIMEOUT_MS` | 300_000 (5 min) | 504 (new `QueryTimeoutError`) |
| Per-pod concurrency cap | `PUBLISHER_MAX_CONCURRENT_QUERIES` | 32 | 503 |

Also lands the operational guardrails from Step 6: a startup V8 heap-size warning, removal of the deprecated modern `GET …/queryData` routes (legacy `/projects/…/queryData` kept for backwards compat but now gated by the concurrency middleware), and a comprehensive telemetry pass so all guards are observable in production.

## Scope

- **Admission**: every query controller (connection SQL, model query, notebook cell, MCP `executeQuery`) calls `Environment.assertCanAdmitQuery()` before acquiring a connector.
- **Timeout**: `runWithQueryTimeout(fn, ms)` arms a timer and hands `fn` an `AbortSignal` that is forwarded to `RunSQLOptions.abortSignal` / `runnable.run({ abortSignal })`. Streaming SQL composes the timeout signal with its internal cap-abort signal via `AbortSignal.any()` so caller cancellation and overflow cancellation can both fire.
- **Concurrency**: `queryConcurrency()` Express middleware on all modern + legacy `/projects/...` query routes; MCP `executeQuery` shares the same slot pool via `tryAcquireQuerySlot()`. Slot release fires on `res.finish` and `res.close`, idempotent.
- **Heap check**: `checkHeapConfiguration()` warns at startup if V8 `heap_size_limit` is below the recommended 2 GiB.

## Telemetry (all on the existing `publisher` OTel meter)

| Metric | Type | Purpose |
| --- | --- | --- |
| `publisher_query_cap_exceeded_total{cap_type,source}` | Counter | 413 rejections, labeled by `rows`/`bytes` and `connection_sql`/`model_query`/`notebook_cell` |
[...]

PR #784, opened 2026-05-22
fix(server): close CodeQL js/path-injection findings with sanitizer barriers

kylenesbit:
## Summary

Closes the 45 pre-existing `js/path-injection` alerts CodeQL has been reporting against the server package (see [PR #783 comment](https://github.com/malloydata/publisher/pull/783#issuecomment-4521716427) for the breakdown that surfaced these as orthogonal to the OOM-mitigation work).

Applies the canonical two-layer pattern that CodeQL's `js/path-injection` query recognises as a sanitizer:

1. **Source-side allowlist** at every public entry point (`assertSafeEnvironmentName`, `assertSafePackageName`, `assertSafeConnectionName`, `assertSafeEnvironmentPath`).
2. **Sink-side containment join** (`safeJoinUnderRoot`) wherever a user- or remote-controlled segment is composed onto a server-controlled root.

### New / extended helpers (`src/path_safety.ts`)
- `assertSafeEnvironmentName` and `assertSafeConnectionName` — thin aliases on `SAFE_NAME_RE` so call sites read correctly and a future regression that diverges one rule from the others is caught.
- Exhaustive helper-level tests added to `src/path_safety.spec.ts`.

### Boundaries gated
| Location | Bucket | Treatment |
|---|---|---|
| `EnvironmentStore.addEnvironment` / `getEnvironment` | B | Source-side `assertSafeEnvironmentName` at the only two public gates that fan out into scaffolding, download, and unzip helpers. |
| `WatchModeController.startWatching` | B | `assertSafeEnvironmentName(watchName)` before `path.join` / `chokidar.watch`. |
| `deleteDuckLakeConnectionFile` | C | `assertSafeConnectionName` + `assertSafeEnvironmentPath` + `safeJoinUnderRoot` at helper entry. |
| `Environment.create`, `Environment.reloadEnvironmentMetadata` | A / C | Sink-adjacent re-asserts so CodeQL sees the sanitizer at the same data-flow node. |
| `Package.create`, `validatePackageManifestExistsOrThrowError` | A | Re-asserts at the `Package` layer where CodeQL can't see across the module boundary to the `Environment` guards. |
| `scaffoldEnvironment`, `loadEnvironmentIntoDisk`, `unzipEnvironment`, `mountLocalDirectory` | B / D | Re-asserts + `safeJoinUnderRoot` for every composed sub-path. |
| `downloadGcsDirectory`, `downloadS3Directory` | D | Destination-root assert + `safeJoinUnderRoot` for every per-object join; skip-with-warn on traversal-bearing remote object keys. |
| `downloadGitHubDirectory` | D / E | Clone-destination assert + `safeJoinUnderRoot` for the GitHub subpath and for every `readdir` entry during the post-clone shuffle. |

### Notable deliberate non-changes
- `mountLocalDirectory` source path is **not** restricted to live under `serverRootPath` — mounting host directories from anywhere on disk is a supported admin feature. The destination path *is* restricted.
- Cloud download helpers `warn-and-skip` on traversal keys rather than failing the whole sync, so a single malicious object in a bucket can't break the whole environment load.

## Behavioural impact

- Callers passing legitimate names (letters, digits, \`-\`, \`_\`, \`.\`) are unaffected.
- Callers passing traversal-bearing names now get **HTTP 400 BadRequest** with a tuning-friendly message ahead of any filesystem work, instead of either a confusing 500 or a successful escape.
- No new env vars, no telemetry changes.

## Test plan

- [x] `bunx tsc --noEmit` clean.
- [x] `bunx eslint src --fix` clean.
- [x] `src/path_safety.spec.ts` — 77 tests pass (exhaustive helper coverage incl. new aliases).
- [x] `src/service/path_injection_boundaries.spec.ts` — 20 new boundary-wiring tests pass; each gate rejects traversal-bearing inputs with `BadRequestError` before any `fs.*` access.
- [x] `src/service/environment_admission.spec.ts`, `src/service/model_limits.spec.ts`, `src/errors.spec.ts`, `src/config.spec.ts`, `src/stream_helpers.spec.ts`, `src/controller/connection.controller.spec.ts` — all green (231 pass, 0 fail).
- [ ] CI CodeQL re-run on this PR should show the path-injection alert count drop from 45 to ~0 (the readdir-loop alerts may persist as residual FPs even with `safeJoinUnderRoot` since CodeQL is conservative about `path.resolve` in some shapes; will iterate on the PR if so).
- [ ] DCO sign-off included in commit.

Pre-existing Bun test-runtime gap: `src/service/environment_store.spec.ts` fails to load due to `perf_hooks.monitorEventLoopDelay` not being implemented in the Bun shim used for unit tests. This is unrelated to this PR (reproducible on `main` with these changes stashed) and is tracked separately.

### DIFF
--- a/packages/server/src/controller/watch-mode.controller.ts
+++ b/packages/server/src/controller/watch-mode.controller.ts
@@ -1,8 +1,8 @@
 import chokidar, { FSWatcher } from "chokidar";
 import { RequestHandler } from "express";
-import path from "path";
 import { components } from "../api";
 import { logger } from "../logger";
+import { assertSafeEnvironmentName, safeJoinUnderRoot } from "../path_safety";
 import { EnvironmentStore } from "../service/environment_store";
 
 type StartWatchReq = components["schemas"]["StartWatchRequest"];
@@ -32,6 +32,15 @@ export class WatchModeController {
       res,
    ) => {
       const watchName = req.body.environmentName ?? "";
+      // Validate the request-supplied environment name BEFORE any
+      // filesystem access. `watchName` is interpolated into
+      // `path.join(serverRootPath, watchName)` below and reaches
+      // `chokidar.watch` (which spawns `fs.watch` on the resolved
+      // path); a `..`-laden value would let a caller watch any
+      // directory under the server root. The allowlist matches the
+      // one applied at `EnvironmentStore.addEnvironment` so all
+      // surfaces share a single shape rule.
+      assertSafeEnvironmentName(watchName);
       const environmentManifest =
          await EnvironmentStore.reloadEnvironmentManifest(
             this.environmentStore.serverRootPath,
@@ -53,7 +62,11 @@ export class WatchModeController {
          return;
       }
 
-      this.watchingPath = path.join(
+      // `safeJoinUnderRoot` is the resolve-and-contain sink barrier
+      // for `chokidar.watch`. `watchName` is already shape-validated
+      // above, but the secondary containment check defends against
+      // any future refactor that bypasses the source-side allowlist.
+      this.watchingPath = safeJoinUnderRoot(
          this.environmentStore.serverRootPath,
          watchName,
       );
--- a/packages/server/src/path_safety.spec.ts
+++ b/packages/server/src/path_safety.spec.ts
@@ -3,6 +3,8 @@ import * as path from "path";
 
 import { BadRequestError } from "./errors";
 import {
+   assertSafeConnectionName,
+   assertSafeEnvironmentName,
    assertSafeEnvironmentPath,
    assertSafePackageName,
    assertSafeRelativeModelPath,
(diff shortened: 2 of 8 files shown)


kylenesbit:
Superseded by #781, which landed essentially the same fix while this PR was in review (we developed them in parallel from the same CodeQL alert list).

`assertSafePackageName` is now wired through every public service entry point, `safeJoinUnderRoot` replaces template-string / bare `path.join` at every flagged sink in `environment_store.ts`, and the helpers gained `asserts x is string` for TypeScript narrowing. The current alert count on `main` after #781 is **4** (down from 45):

- `environment_store.ts:903`, `:907` — `unzipEnvironment` derives `unzippedEnvironmentPath` via `.replace(".zip","")`; CodeQL can't follow `.replace` as a sanitizer through the derived variable.
- `environment_store.ts:1385`, `:1392` — `mountLocalDirectory` mount source path (#781 deliberately left this open because operators can pass relative paths).

Closing this and opening a small focused follow-up for the 4 residuals.

PR #785, opened 2026-05-25
fix: remove modelPath filter from query annotations

adamribaudo-ff:
## Summary

Query annotations were being incorrectly filtered by `modelPath`, causing annotations on imported queries to be stripped from API responses.

When `main.malloy` imports `project.malloy`, queries defined in `project.malloy` have their `note.at.url` pointing to `project.malloy`. The `note.at.url.includes(modelPath)` filter rejects them because `modelPath` is `main.malloy`.

## Changes

Removed the `.filter((note) => note.at.url.includes(modelPath))` from the annotations mapping in:
- `Model.getQueries()` in `packages/server/src/service/model.ts`
- `extractQueries()` in `packages/server/src/package_load/package_load_worker.ts`

Annotations are now mapped directly from `blockNotes` without path filtering, falling back to `[]` when none exist.

## Why this is safe

- **Queries don't have annotation inheritance** (no `annotation.inherits` chain) like sources do
- **A query's `blockNotes` contains only the annotations directly applied to that query definition**
- **The `modelPath` filter was serving no useful purpose for queries** — it only caused annotations from imported queries to be lost

### Before

Annotations from .malloy model directly

```
GET /api/v0/environments/{orgId}/packages/operational-metrics/models/project.malloy

  index_project_taxonomy: ["#(doc) Taxonomy index for projects. `unique_projects` is the count of..."]
  index_project_fields: ["#(doc) Taxonomy index for projects. \n"]
```

Annotations from queries imported into .malloy model:

```
GET /api/v0/environments/{orgId}/packages/operational-metrics/models/main.malloy

Queries:
  index_project_taxonomy: annotations: []
  index_project_fields: annotations: []
  ...
```

### After

```
GET /api/v0/environments/{orgId}/packages/operational-metrics/models/main.malloy

  ✅ index_project_taxonomy: #(doc) Taxonomy index for projects...
  ✅ index_project_fields: #(doc) Taxonomy index for projects.

```

## Note on sources

The same `modelPath` filter pattern exists in `getSources()` for both source annotations and view annotations. The `sources[]` array in the API response likely has the same issue, but client code reads source annotations from `sourceInfos` instead, which correctly preserves them via the Malloy `modelDefToModelInfo()` path. This PR scopes the fix to queries only.

### DIFF
--- a/packages/server/src/package_load/package_load_worker.ts
+++ b/packages/server/src/package_load/package_load_worker.ts
@@ -46,26 +46,26 @@
  */
 import {
    contextOverlay,
+   isSourceDef,
+   MalloyConfig,
+   MalloyError,
+   modelDefToModelInfo,
+   Runtime,
    type Annotation,
    type BuildManifestEntry,
    type Connection,
    type FetchSchemaOptions,
    type LookupConnection,
-   MalloyConfig,
-   MalloyError,
    type ModelDef,
    type ModelMaterializer,
-   modelDefToModelInfo,
    type NamedModelObject,
    type NamedQueryDef,
    type Query,
-   Runtime,
    type SQLSourceDef,
    type SQLSourceRequest,
    type StructDef,
    type TableSourceDef,
    type TurtleDef,
-   isSourceDef,
 } from "@malloydata/malloy";
 import * as Malloy from "@malloydata/malloy-interfaces";
 import {
@@ -73,8 +73,8 @@ import {
    MalloySQLStatementType,
 } from "@malloydata/malloy-sql";
 import * as fs from "fs";
-import * as path from "path";
 import { parentPort, threadId } from "node:worker_threads";
+import * as path from "path";
 import recursive from "recursive-readdir";
 import { fileURLToPath, pathToFileURL } from "url";
 
--- a/packages/server/src/package_load/package_load_worker.ts
+++ b/packages/server/src/package_load/package_load_worker.ts
@@ -556,11 +556,10 @@ function extractQueries(modelPath: string, modelDef: ModelDef): ApiQueryWire[] {
       .map((q) => ({
          name: q.as || q.name,
          sourceName: typeof q.structRef === "string" ? q.structRef : undefined,
-         annotations: q?.annotation?.blockNotes
-            ?.filter((note: { at: { url: string } }) =>
-               note.at.url.includes(modelPath),
-            )
-            .map((note: { text: string }) => note.text),
+         annotations:
+            q?.annotation?.blockNotes?.map(
+               (note: { text: string }) => note.text,
+            ) ?? [],
       }));
 }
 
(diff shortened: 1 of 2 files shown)


mlennie:
@adamribaudo-ff great job!
## Review

Thanks for tracking this down. The diagnosis is right: filtering query annotations by `note.at.url.includes(modelPath)` strips annotations from imported queries, because an imported query's `note.at.url` points at the file it's defined in, not the model being queried. Removing the filter is the correct fix. A few things before it can land, though.

### 1. The base is stale: this is already fixed on `main`

This branch is built on a pre-0.0.403 version of Publisher. It still uses the old singular `queryObj.annotation` IR shape and a `modelPath` parameter, neither of which exists on current `main`. Upstream already removed this filter:

- #792 ("Bring publisher current with malloy 0.0.403") dropped the filter when the malloy IR moved to `queryObj.annotations` + `annotationTexts(...)`.
- #802 ("refactor: extract shared source/query introspection helper") consolidated both call sites into a single `extractQueriesFromModelDef()` in `service/source_extraction.ts`.

On current `main`, `getQueries()` and `extractQueries()` are three-line delegations with no `modelPath` and no `queryObj.annotation`, so this PR won't merge cleanly and the fix is already present. Could you rebase on the latest `main` and re-check? If the bug still reproduces there, it's a different issue and should be re-diagnosed against `extractQueriesFromModelDef` in `service/source_extraction.ts`.

### 2. `?? []` changes the empty case from `undefined` to `[]`

`Query.annotations` is optional in `api-doc.yaml`, so both conform to the contract, but the rest of the code deliberately returns `undefined` for "no annotations" (`annotationTexts` documents this as matching the optional API shape, and the sources path does the same). `[]` is client-visible (truthiness checks, JSON field presence), so I'd keep it `undefined` for consistency.

### 3. Please drop the second commit ("linting")

The second commit is purely import reordering (value imports hoisted above type imports, `path` moved, `deserializeError`/protocol imports relocated). Publisher's CI doesn't enforce import order (there's no eslint `import/order` rule and prettier doesn't sort imports), so this isn't actually required, and it widens the merge conflict for no benefit. I'd drop that commit and keep the PR to the two annotation hunks.

### 4. Add a regression test

The change is verified manually with curl only. A unit test that compiles a two-file model (one importing the other) and asserts the imported query's annotations survive would lock this in. There's no `source_extraction.spec.ts` yet, so that would be a good home for it.

### Minor / FYI

- The "queries have no inheritance" reasoning holds for the imported-query case, but note `annotationTexts` on `main` intentionally includes inherited annotations too (own notes + blockNotes + the `inherits` chain), so a blockNotes-only mapping is slightly narrower than where upstream landed.
- Good call scoping out the `getSources()` case. On `main` that path already walks the `inherits` chain with no URL filter, so it's handled there too.

adamribaudo-ff:
Resolved already in the latest publisher version

PR #786, opened 2026-05-25
feat: Theme Editor, viewer light/dark mode, and chart theming

mlennie:
## Summary
  
  Adds an end-to-end theming system for Publisher with three independent surfaces.

Walkthrough video: https://www.loom.com/share/e31e852ca1664526a1e7a355a07bddd6

Screenshot 2026-05-22 at 10 30 46 AM
Screenshot 2026-05-22 at 10 31 00 AM
Screenshot 2026-05-22 at 10 31 17 AM
Screenshot 2026-05-22 at 10 31 37 AM

  
  1. Server-side `Theme` schema in `publisher.config.json` (instance and per-environment), backed by a new SQLite `ThemeStore` and exposed via `GET / PUT / DELETE /api/v0/theme`.
  2. In-app Theme Editor at `/settings/theme` with section-by-section color pickers (background, tables, dashboard tiles, charts, maps, fonts), a Light/Dark variant toggle, and a live preview that renders the
  resolved cascade.
  3. Viewer-facing light/dark/auto toggle in the app header. Honors `defaultMode` and `allowUserToggle` from the config, persists the viewer's choice to `localStorage`, and follows `prefers-color-scheme` in
  `auto`.
  
  Cascade: SDK defaults → instance config → environment config → editor (per environment) → per-chart `# theme.*` annotation. Each layer only overrides the keys it sets.
  
  Per-mode palette keys: `background`, `tableHeader`, `tableHeaderBackground`, `tableBody`, `tile`, `tileTitle`, `mapColor`. `series` and `font` are scalars shared across modes (brand identity).
  
  ## Paired renderer change
  
  Depends on `theme` being a real prop on `MalloyRenderer`. That change lives in [malloydata/malloy](https://github.com/malloydata/malloy) on `monty/explicit-theme-prop`. Without it, chart-canvas and
  chart-internal colors won't follow the resolved theme. The SDK currently consumes the renderer via `bun link` against a local malloy checkout; switch to the npm-published version once the renderer PR merges.
  
  ## Key paths
  
  SDK
  - `packages/sdk/src/theme/*` — `defaults`, `keys`, `types`, `resolveTheme`, `buildTableCssVars`, `buildVegaThemeOverride`, `buildMalloyExplicitTheme`, `readChartAnnotations`, `ThemeContext`
  - `packages/sdk/src/components/ServerProvider.tsx` — owns mode state, `matchMedia` subscription, `localStorage` persistence
  - `packages/sdk/src/components/RenderedResult/RenderedResult.tsx` — passes the resolved theme to `MalloyRenderer` and emits scoped CSS vars
  
  Server
  - `api-doc.yaml` — `Theme` schema + `/theme` endpoints
  - `packages/server/src/service/theme_store.ts` — SQLite-backed runtime store
  - `packages/server/src/controller/theme.controller.ts` — GET / PUT / DELETE
  - `packages/server/src/config.ts` — instance + per-environment `theme` parsing with sanitizer
  - `packages/server/src/storage/duckdb/schema.ts` — `themes` table migration
  
  App
  - `packages/app/src/components/pages/ThemeEditorPage/*` — editor page, sections, previews
  - `packages/app/src/components/common/ThemeToggle.tsx` — header viewer toggle
  - `packages/app/src/theme/PublisherMuiThemeProvider.tsx`, `packages/app/src/theme/index.tsx` — MUI palette keyed off SDK mode
  
  README
  - New `### Theming` subsection under Configuration covering all seven per-mode palette keys plus the scalar `series`/`font` and `defaultMode` / `allowUserToggle` semantics. Public-docs walkthrough lives in a
  paired PR on `malloydata.github.io`.
  
  ## Test plan
  
  - [ ] `bun run typecheck && bun run lint && bun run prettier:check && bun run test` from repo root
  - [ ] `cd packages/app && bun run test:playwright` (new `theming.spec.ts` plus the existing five chromium specs stay green)
  - [ ] Editor browser sweep
    - [ ] Open `/settings/theme`, change each picker, confirm the live preview updates and edits auto-save
    - [ ] Toggle Light/Dark inside the editor, edit a value, confirm only that variant updates
    - [ ] Reset to defaults, confirm pickers return to brand defaults and the saved row is cleared
  - [ ] Render surface
    - [ ] Open a sample package (e.g. `ecommerce`), confirm chart, table, and dashboard tile colors follow the resolved theme
    - [ ] Open a `# shape_map` chart (e.g. `faa.sales_by_state`), confirm `palette.mapColor` drives the choropleth gradient
  - [ ] Viewer mode toggle
    - [ ] Click sun/moon/auto in the header, confirm the app shell and rendered charts both flip
    - [ ] Reload, confirm the choice persists via `localStorage["publisher:themeMode"]`
    - [ ] Set `theme.allowUserToggle: false` in config, restart, confirm the toggle is hidden and `defaultMode` is enforced
  - [ ] `frozenConfig: true`: confirm `PUT /api/v0/theme` returns an error and the editor surfaces as read-only
  - [ ] Per-chart annotation override: add `# theme.palette.series = [...]` to a view, confirm only that view picks up the override


api-doc.yaml
```diff
@@ -79,6 +79,66 @@ paths:
...
+  /theme:
+    get:
+      tags:
+        - publisher
+      operationId: get-theme
+      summary: Get the active instance theme
+      description: |
+        Returns the theme used by the Malloy renderer for charts/tables/dashboards.
+        Reflects edits made via the in-app Theme Editor (persisted to the server's
+        SQLite database). Empty object when no theme has been configured.
```
sagarswamirao: SQLLite DB?

mlennie: @sagarswamirao good catch. I updated to say DuckDB

sagarswamirao:
Can you add a couple of screenshots?

mlennie:
> Can you add a couple of screenshots?

@sagarswamirao thank you for the reminder. Just added screenshots to PR description.

kylenesbit (approved):

sagarswamirao:
Nice feature. A few things I ran into while going through it (9aba6e9):

**Auto-save can persist a stale theme on out-of-order PUTs.** The `saveGenRef`/`myGen` guard in `ThemeEditorPage.tsx:175-193` drops the stale onSuccess callback but doesn't cancel the in-flight request or order the writes. If two saves overlap (edit A, save-1 goes out and is slow, edit B, save-2 fires) and PUT-1 lands after PUT-2, `ThemeStore.set` full-replaces in arrival order, so the DB keeps {A} and drops B while the editor cache still shows {A+B} and reports "Saved". Narrow window since same-origin ordering usually holds and each PUT is a full snapshot that self-heals on the next save, but worth closing: either await the previous mutation before firing the next, or send a client monotonic version that `putTheme` rejects if older.

**Cascade docs vs wiring.** `resolveTheme.ts` / `ThemeContext.tsx` describe a 5-layer cascade (defaults -> instance -> env -> editor -> per-chart), but `ServerProvider.tsx:211` only feeds `[instanceTheme]` and the per-chart path appends the annotation, so it's effectively 2 layers over defaults. The merge handles N layers fine; is the intent to plumb env/editor later, or should the docstrings come down to match what's wired?

**Stale comment:** `keys.ts:3` refers to "the sanitizer" sharing this key list, but the sanitizer lives server-side in `config.ts` (there's none on the SDK side). And `readChartAnnotations.ts` says "the five PER_MODE_COLOR_KEYS" where the list is now seven.

**PUT `{}` vs DELETE** end up with different semantics (PUT suppresses the config seed, DELETE re-seeds it), and a persisted `"{}"` sanitizes back to undefined on reload. Looks intentional from the comments and isn't reachable from the editor today, but a doc line on the endpoint would save the next reader some head-scratching.

package.json
```diff
@@ -73,11 +73,12 @@
     "ssh2": "^1.17.0"
   },
   "resolutions": {
+    "pg": "8.16.3",
```
sagarswamirao, line 76: This hardcoded pin will drift on the next Malloy bump. `pg` isn't a direct dep here - it comes in transitively via `@malloydata/db-postgres`, which pins its own `pg` (8.7.3 today). This resolution overrides that to keep the server in sync, which is the right intent, but the number is a magic constant that `scripts/upgrade-malloy.sh` doesn't manage, so a future malloy upgrade will leave `pg` stranded at 8.16.3 while the connector moves on.

There's already a precedent for exactly this in the same script: after `bun install` it runs `sync-duckdb-version.js --write` to re-derive DuckDB from whatever engine Malloy resolved to ("A Malloy bump can move that engine to a new DuckDB version; syncing here keeps [it] on the same version... with no manual follow-up step"). Worth mirroring that for pg - a small `sync-pg-version.js` that reads the `pg` version `@malloydata/db-postgres` resolves to and writes it into resolutions - so it stays aligned automatically. Alternatively, if the connector's own `pg` is fine, drop the resolution entirely; if it isn't, a one-line note on why we diverge would help.

jswir:
Ran an independent review pass (bugs/security/architecture) on 9aba6e9. Re: @sagarswamirao's cascade point above — can confirm it's not just under-wired in the SDK, the environment layer is dropped server-side too:

- `getProcessedPublisherConfig()` does compute a merged per-environment theme (`config.ts:777-789`), but `addConfiguredEnvironment()` only forwards `{name, resource, connections, packages}` to `addEnvironment()` — `theme` never reaches the live `Environment` object. `Environment.update()`/`serialize()` (`environment.ts`) have zero references to `theme` anywhere in the file.
- There's also no UI path to set a per-environment theme today — `/settings/theme` isn't scoped to `:environmentName`, and `ThemeEditorPage.tsx` only calls the global `getTheme()`/`putTheme()`/`resetTheme()`.

So `environments[].theme` in `publisher.config.json` (documented in the README and `Environment.theme`'s description) has zero effect on any viewer, not just a "cascade not fully plumbed yet" gap. Given it's user-facing documented config surface, would lean toward either wiring it end-to-end or pulling the "environment config"/"editor (per environment)" claims from the README/api-doc.yaml before merge so the docs match reality.

A few other things from the pass, independent of the above:

**`GET /api/v0/status` can 500 during boot.** `EnvironmentStore`'s constructor fires `this.initialize()` without awaiting it (`environment_store.ts:173`), and the existing `/status` handler deliberately uses `skipInitializationCheck=true` so it stays available during boot. This PR adds an unconditional `await themeStore.get()` to the same handler; `ThemeStore.get()` → `StorageManager.getDuckDbConnection()` throws `"Storage not initialized"` if called before `storageManager.initialize()` resolves. A request landing in that (short) window now 500s instead of the previously-guaranteed 200.
https://github.com/malloydata/publisher/blob/9aba6e9088c6ecf5aeaba2d7ab0325be1f2e1216/packages/server/src/server.ts#L726-L741

**`sanitizeTheme()`'s `allowUserToggle` inverts malformed input.** Every other field type-checks before accepting; this one doesn't:
```ts
if ("allowUserToggle" in obj) {
   theme.allowUserToggle = Boolean(obj.allowUserToggle);
}
```
`Boolean("false")` is `true` — a config value of `"allowUserToggle": "false"` (easy to write by mistake) silently sets the opposite of what's intended, with no warning, unlike the other fields.
https://github.com/malloydata/publisher/blob/9aba6e9088c6ecf5aeaba2d7ab0325be1f2e1216/packages/server/src/config.ts#L558-L560

**`/theme` endpoints drop the existing error-response contract.** Every other mutating endpoint (`update-environment`, `delete-connection`, etc.) declares schema-backed `401/404/500/503` plus a `BadRequest`-refed `400`. The new `PUT/DELETE /theme` use bare `description:`-only `403`/`400` and omit `401/500/503` entirely, even though the same routes go through the same error-handling path that can produce them.
https://github.com/malloydata/publisher/blob/9aba6e9088c6ecf5aeaba2d7ab0325be1f2e1216/api-doc.yaml#L86-L128

**`put-theme`/`reset-theme` operationIds break the naming convention.** Every existing PUT/DELETE endpoint uses `update-<resource>`/`delete-<resource>` (zero exceptions across the spec). These generate as `putTheme()`/`resetTheme()` in the SDK client (already called that way in `ThemeEditorPage.tsx:46,125,136`) instead of `updateTheme()`/`deleteTheme()` — worth fixing now since it's shipped SDK surface.

**Palette key list is hand-duplicated in three places** (`api-doc.yaml`'s `Theme.palette` schema, `packages/sdk/src/theme/keys.ts:13-21`, `packages/server/src/config.ts:85-92`) with nothing keeping them in sync. The comment in `config.ts` says this is forced by dependency direction, but neither package actually imports the other — worth having the server import from a shared location, or at least a test that fails if the three lists diverge.

Smaller ones:

- `initializeSchema()`'s early-return guard removal (`schema.ts:7-31`) makes every boot re-run the full `CREATE TABLE`/`INDEX IF NOT EXISTS` pass, not just first-init — needed for the new `themes` table to appear on existing DBs, and the DDL itself is idempotent/safe, but this exact guard had a critical data-loss finding in #696. Worth a second look given that history, and maybe its own commit.
- `sanitizeTheme()` checks color fields are strings but not that they're valid CSS colors — a bad value silently produces broken chart/table styling instead of falling back to a default.
- `theming.spec.ts` covers toggle/mode-persistence only; the `frozenConfig` read-only editor state and `allowUserToggle:false` hiding the toggle (both called out in the PR's own test plan) have no test coverage at any level.

mlennie:
Thanks @jswir and @sagarswamirao for the thorough round-2 pass. Pushed changes addressing all of it; rundown below, with per-thread replies inline.

**Bugs and correctness**
- GET /api/v0/status no longer 500s during the storage boot window: the theme it composes is read only once storage is initialized, so /status keeps its always-200 guarantee. GET /theme now returns 503 (ServiceUnavailable, already in its documented responses) during that window rather than an empty 200, so the editor's load never mistakes "not ready" for "no theme" and auto-saves an empty theme over the real one.
- sanitizeTheme validates allowUserToggle as a real boolean instead of Boolean()-coercing it, so `"allowUserToggle": "false"` no longer flips to true; it warns and ignores malformed input the same way defaultMode does.
- Hardened the Theme Editor's auto-save vs Reset ordering: overlapping saves are serialized, a save superseded by a Reset is dropped before it hits the network, and the Reset's DELETE is chained behind any in-flight save so it is the last write to reach the server. A Reset can no longer be clobbered by a late save.

**API contract**
- Renamed the /theme operationIds put-theme/reset-theme to update-theme/delete-theme (SDK/CLI methods updateTheme/deleteTheme) to match the update-/delete- convention, done now while it is pre-publish. HTTP methods and get-theme are unchanged.
- GET/PUT/DELETE /theme now declare the standard schema-backed error responses (400/401/403/500/503 via the shared components) instead of bare description-only 400/403.

**Tests and docs**
- Added a server-side parity test that fails if the per-mode palette key list drifts across config.ts, the SDK keys.ts, and the api-doc Theme.palette schema, and corrected the comment that implied a server-to-SDK dependency (they are decoupled; api-doc.yaml is their only shared contract).
- Added Playwright coverage for allowUserToggle:false hiding the viewer toggle and for frozenConfig rendering the editor read-only (both were in the PR's test plan but untested at the UI level).
- Trimmed the per-environment theme docs to match reality: environments[].theme is computed server-side but not wired through to any viewer, so the README, api-doc (Environment.theme and the cascade wording), and the SDK docstring now describe it as reserved / not-yet-applied rather than active. Wiring it end to end is a tracked follow-up.
- Added a note in scripts/upgrade-malloy.sh that the root pg resolution is a deliberate manual hold at main's 8.16.3, since a sync-pg-version.js mirroring the DuckDB script would derive db-postgres's exact 8.7.3 and downgrade pg; the pin stays and is flagged to re-verify on the next Malloy bump.

**Deferred / no change**
- initializeSchema guard removal: verified safe. The destructive helpers stay gated exactly as before (dropAllTables on --init, dropLegacyProjectSchema on !initialized); only idempotent CREATE ... IF NOT EXISTS runs on every boot, which is what lets an existing DB pick up the themes table. The #696 data-loss path is not reintroduced.
- CSS-color validity validation: deferred to the theme-key registry follow-up. A blanket check would break the free-form keys (background, tile, tableHeaderBackground legitimately carry url()/gradients), so it needs per-key typing.
- The CLI generated client is gitignored and regenerated by CI from the spec, so it picks up the operationId rename automatically.

Verified with typecheck, lint, the full server unit suite, the theming Playwright suite, and a manual pass against a running server.

package.json
mlennie: @sagarswamirao Totally fair, and I agree the pin is a magic constant the script doesn't manage today. Quick context on why I'm leaving it as-is for this PR: the hold exists only so theming has zero net effect on pg (db-postgres@0.0.421 pins pg to an exact 8.7.3, which the family bump would otherwise pull us down to from main's 8.16.3). The catch with mirroring sync-duckdb-version.js is that a literal sync-pg-version.js would read what db-postgres resolves pg to, which is 8.7.3, so it would actually downgrade pg rather than hold it, the opposite of the intent. So the real fork here is a policy call: keep holding pg ahead of the connector (and automate that hold with a floor so it never drops below main), or let pg track db-postgres and accept 8.7.3. I didn't want to make that call inside a theming PR.

For now I've documented it (the note + echo in upgrade-malloy.sh flags the hold as deliberate and not auto-managed, to re-verify on the next Malloy bump), which keeps this PR pg-neutral so we can get theming out. Happy to do the proper version as a separate follow-up PR, either a sync-pg-version.js that pins pg to the higher of main's version and the connector's (so it stays aligned automatically without ever downgrading), or just dropping the resolution if you're comfortable riding db-postgres's 8.7.3. Let me know which you'd prefer and I'll open it.

PR #796, opened 2026-05-29
feat(sdk,app): surface materialization and manifest in the UI

mlennie:
## Summary
  
  Publisher already exposes a materialization and build-manifest REST API (create / start / stop / delete a build, read the manifest), and the SDK already generates `MaterializationsApi` and `ManifestsApi`, but
  nothing instantiated them, so neither was reachable from the web app. This PR wires them in and adds a dedicated screen so a user can run, monitor, and inspect materializations from the UI instead of calling the
  API by hand.
  
  This is client-only: no server or `api-doc.yaml` changes.
  
  ## What this adds
  
  - A "Materializations & Manifest" entry on the package page that opens a new `/:environmentName/:packageName/materializations` screen.
  - An operator view: a runs table (status, started, duration, sources built/skipped, error) with actions to start a new run, stop an active one, and delete a terminal one.
  - An author view: the build manifest (BuildID to table name) on the same screen, with a reload action.
  - A run-detail dialog with per-source build results and any error, read from the run metadata.
  - Every write action is gated on the `mutable` flag, so a read-only or frozen publisher renders no controls.
  
  ## Scope
  
  - CRUD parity is added only for resources the API already exposes. Models, notebooks, and databases stay read-only by design; no new write endpoints.
  - The runs list polls while a run is active and stops once everything is terminal; the manifest refetches when a run finishes.
  - Schedules and CLI parity are intentionally out of scope and left as follow-ups.
  
  ## Implementation notes
  
  - Wires `MaterializationsApi` and `ManifestsApi` into `ServerProvider` (the only missing piece; both clients were already generated).
  - Reuses the existing connection-management patterns: `useQueryWithApiError` / `useMutationWithApiError`, query-key invalidation, a shared Snackbar, the kebab action menu, and the page-wrapper convention.
  - Adds the SDK feature directory `packages/sdk/src/components/Materializations/`, a `MaterializationsPage` route wrapper, and a new `ContentTypeIcon` glyph.
  
  ## Testing 
  
  - New Playwright spec `package-materializations.spec.ts`: read, mutable CRUD lifecycle, mutability parity, and a keyboard-accessible detail-dialog test (6 specs, passing against a live server).
  - The materialization REST API itself is covered by the existing server integration suite.
  - typecheck, lint, prettier, and the full test suite pass locally.
  
  ## Screenshots 
  
Screenshot 2026-05-28 at 7 08 35 PM
Screenshot 2026-05-28 at 7 08 51 PM
Screenshot 2026-05-28 at 7 09 03 PM
  
  ## Follow-ups
  
  - CLI parity: a `materialization` command group reusing the same generated client.
  - Schedules: a separate, full-stack change (API + storage + scheduler, then UI).


sagarswamirao:
The add package API supports MaterializationStorage obj, and autoLoadManifest param. I think need to have UI flow for that too

mlennie:
@sagarswamirao Good call, but quick clarification first: there are two `autoLoadManifest` params. The build-time one (reload after a build) is already in this PR as the "Auto-load manifest after build" switch. The one you mean is the `create-package` param.

The catch: in the human UI that toggle does almost nothing. A brand-new package has no manifest (no-op), and for an existing package the Materializations tab already has a "Reload manifest" button that does the same thing. It only "works" if you re-add a package you already have, which is not a real user flow. Where it genuinely matters is the orchestrated path: when the control plane provisions a package onto a fresh or restarted worker, `autoLoadManifest=true` hydrates it from the shared manifest catalog with no rebuild. That is an operator/automation concern, not a user click. `materializationStorage` is the same story (the shared catalog, control-plane-set, with credentials in it), so I would keep both out of the human create dialogs.

So before adding the checkbox: do you actually want a user-facing toggle, or is `autoLoadManifest` better left as the API/control-plane param it already is? I have a working prototype if we want it, but my lean is to leave the Add Package dialog alone. The PR stays scoped to the run + manifest UI either way. 

This is what we currently have
Screenshot 2026-06-02 at 6 39 04 PM

mlennie:
@sagarswamirao maybe this makes more sense... 

For a standalone Publisher user this toggle is effectively a no-op: a brand-new package has no manifest to load, and an already-materialized one can use the existing "Reload manifest" button. 

It only does real work in an orchestrated deployment, where the control plane provisions a fresh worker that hydrates a package from a manifest already in shared storage, but that provisioning is an API call the control plane makes, not a human clicking Add Package. 

So my lean is to keep `autoLoadManifest` as an API/control-plane param and leave the human dialog alone, unless there's an add-package flow you have in mind where a person would actually want it?

kylenesbit (approved):

PR #805, opened 2026-06-02
feat(server): validate #(authorize) expressions at model-load (compile-only)

Sha-Bang:
**PR 2 of the `#(authorize)` series.** PR 0 (#802) and PR 1 (#804) are merged; this targets `main`.

3. runtime gate (the actual access enforcement) — next
4. MCP + HTTP surface alignment
5. docs
6. end-to-end + example model

### What this one does
Catches a bad authorize expression at model load instead of first query. Compiles a synthetic probe (`run: duckdb.sql("SELECT 1") -> { select: __auth_i is (expr) }`) with `getPreparedQuery` — no givens, no DB execution. Unknown givens and source-field references surface as compile errors, wrapped in `ModelCompilationError` (424) naming the source. The per-package `duckdb` sandbox is always registered, so the probe never touches the real warehouse.

Wired into **both** compile paths — `Model.create` and the package-load worker (the primary path) — via one shared `validateAuthorizeProbes`, and the worker preserves the `ModelCompilationError` classification across its serialization boundary.

**Scope note:** a type-mismatched comparison like `$ROLE = 5` is *not* a Malloy compile error, so it isn't caught here — it fails closed at the runtime gate in PR 3. Syntactic malformed annotations already fail loud in PR 1.

### Tested
- `tsc` + eslint clean; 780 unit tests pass
- Compile-time validation (`Model.create`): valid value-less given loads (probe is compiled, not run), unknown-given (source-level AND file-level `##(authorize)`) and source-field-ref fail with the source named, type-mismatch loads, and number/list givens validate (not string-only)
- Worker path via `Package.create` — both `.malloy` and `.malloynb`: valid surfaces; unknown-given rejects as a **424 ModelCompilationError** (asserts the classification survives the worker boundary)
- `filter_integration` (32) green
- Independent multi-agent code review (bugs / git-history / comment-convention lenses): no issues

### DIFF
--- a/packages/server/src/errors.ts
+++ b/packages/server/src/errors.ts
@@ -98,7 +98,10 @@ export class ConnectionAuthError extends Error {
 }
 
 export class ModelCompilationError extends Error {
-   constructor(error: MalloyError) {
+   // Accepts a MalloyError or any message-bearing object, so callers that add
+   // context around a compile failure (e.g. naming the source whose authorize
+   // annotation failed) can reuse this 424 mapping without a separate class.
+   constructor(error: { message: string }) {
       super(error.message);
    }
 }
--- a/packages/server/src/package_load/package_load_worker.ts
+++ b/packages/server/src/package_load/package_load_worker.ts
@@ -79,6 +79,7 @@ import {
    PACKAGE_MANIFEST_NAME,
 } from "../constants";
 import { HackyDataStylesAccumulator } from "../data_styles";
+import { ModelCompilationError } from "../errors";
 import { type AnnotationsDef } from "../service/annotations";
 import { validateAuthorizeProbes } from "../service/authorize";
 import { type FilterDefinition } from "../service/filter";
@@ -831,6 +832,18 @@ function serializeError(error: unknown): SerializedError {
          message: error.message,
          stack: error.stack,
          malloyProblems: error.problems as unknown[],
+         isCompilationError: true,
+      };
+   }
+   // ModelCompilationError (e.g. an invalid #(authorize) annotation caught by
+   // validateAuthorizeProbes) carries no Malloy `problems`, but it must keep its
+   // compilation-error classification across the worker boundary so the main
+   // thread re-wraps it as a 424, not a generic 500.
+   if (error instanceof ModelCompilationError) {
+      return {
+         name: error.name,
+         message: error.message,
+         stack: error.stack,
          isCompilationError: true,
       };
    }
--- a/packages/server/src/package_load/package_load_worker.ts
+++ b/packages/server/src/package_load/package_load_worker.ts
@@ -80,6 +80,7 @@ import {
 } from "../constants";
 import { HackyDataStylesAccumulator } from "../data_styles";
 import { type AnnotationsDef } from "../service/annotations";
+import { validateAuthorizeProbes } from "../service/authorize";
 import { type FilterDefinition } from "../service/filter";
 import {
    extractQueriesFromModelDef,
@@ -551,6 +552,10 @@ async function compileMalloyModel(
 
    const { sources, filterMap } = extractSources(modelDef, givens);
    const queries = extractQueries(modelDef);
+   // Validate #(authorize) at compile time (shared with Model.create). Throws
+   // on an unknown given / source-field reference; compileOneModel's catch
+   // turns it into this model's compilationError.
+   await validateAuthorizeProbes(mm, sources);
 
    return {
       modelPath,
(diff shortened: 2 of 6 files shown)


mlennie (approved):
Took a careful pass through this with the rest of the series in context. The core mechanism looks right: data flow into `validateAuthorizeProbes` is correct at all three call sites, the worker→main round-trip reconstructs a real `ModelCompilationError` (so it lands as 424, not 500), the `errors.ts` ctor widening is safe and actually makes the pool's existing `as unknown` cast honest, and the posture is fail-closed (a probe failure for any reason fails the load rather than loading ungated). Compiling the probe over a fields-only `__authorize_probe_row` source to force given-only gates is a nice trick. Good
  commit split too (feat / worker-boundary fix / tests).
  
  A couple of things before I'd merge.

  ### Probe doesn't inherit the model's `##!` experiment flags (false-reject)

  The probe compiles as a fresh internal document and doesn't pick up the base model's `##!` pragmas, so a valid model can fail to load with a 424. Two cases I hit:

  - An authorize expr using an experiment-gated feature (e.g. `sql_functions` → `sql_string(...)`) fails the probe with `Cannot use sql_function ... use sql_functions experiment`, even though the model enables
  it.
  - A model that imports a gated source, where the importing model doesn't itself declare `##! experimental.givens`, fails on `$ROLE` resolution in the probe, even though the given surfaces fine through the
  import.

  I got slightly different mechanics across two malloy versions (0.0.394 vs 0.0.403), so worth confirming against whatever we pin, but the fix is the same either way: prepend the model's active `##!` flags to
  `buildAuthorizeProbe` (at least `experimental.givens`, plus anything an expr might reference). The import + experiment-flag tests below would lock it.

  ### Test gaps

  - The in-process negative cases assert only on `err.message` substrings, never `instanceof ModelCompilationError`. Since the whole point is the 424 mapping, that path isn't actually pinned: a regression that
  threw a plain `Error` or rewrapped as 500 would stay green. The worker tests pin the type but not the message. Add `expect(err).toBeInstanceOf(ModelCompilationError)` to the in-process negatives and a message
  assertion to one worker negative.
  - No coverage for a gated source reached through `import:`, which is exactly where the false-reject above lives. Model A with `given:` + a gated source, imported and used by model B, asserting it validates
  and the given resolves (plus a negative for a given that only exists in A).
  - Lower priority: a multi-source file where a non-first source has the bad gate (assert the right source is named), and a file-level `##(authorize)` on a file with zero sources (currently parsed but never
  validated, so a typo'd file gate slips through).

  ### Optional: assert booleanness
  
  Non-boolean exprs currently pass (`$AGE + 1`, bare `$ROLE`) because `select: __auth_i is (expr)` takes any scalar. Building the probe as `where: (${expr})` over the one-row source instead would reject
  non-boolean gates at compile for basically free. (The `$ROLE = 5` type-mismatch stays uncatchable either way, that's a Malloy thing.)

  ### Design questions, not blockers, cheap to settle before the runtime PR locks the data shape

  - File + source gates are stored as one flat list and the comments say they're OR'd at runtime. Under OR, adding a source gate can only loosen access. Intended, vs AND?
  - An extending source doesn't inherit its base's gate (no inherits walk for authorize). Intended? An extension could end up unintentionally ungated.
  - Gates are given-only by design (bare field refs fail). Confirming that's the intended granularity (no row-level).
  - For the runtime-eval PR: the `(${expr})` interpolation isn't injection-proof (an author can break out into a new pipeline stage). Harmless here since it's compile-only and never reaches raw SQL, but when
  the gate actually runs, `(${expr})` has to stay in Malloy expression context (`where:` / a computed dimension), never concatenated into a SQL string, or that breakout becomes real injection against live data.

  ### Minor
  
  - Docs: a malformed `#(authorize)` now fails model load with 424, which is user-visible, and there's no README / public-doc mention of the annotation yet. Fine to fast-follow, but worth tracking so the
  failing behavior doesn't ship ahead of any description of it.
  - The "compiled, not run / never touches the warehouse" comments are slightly off: compiling does run a DuckDB `DESCRIBE` on the `:memory:` sandbox (cached after first). Not the real warehouse, so fine, just
  worth a word.
  - Stale `deserializeError` comment ("ctor expects a MalloyError-shaped input") now that the ctor takes `{ message: string }`.
  - `err.name` round-trips to `"Error"`, not `"ModelCompilationError"` (instanceof-based dispatch, so harmless; a one-line `this.name =` would help logs).
  - `authorizeMap` in `source_extraction.ts` is currently dead (everything reads `source.authorize`). From #804, not this PR, but worth dropping or commenting until enforcement uses it.

PR #806, opened 2026-06-02
feat(server): enforce #(authorize) at query time (runtime gate)

Sha-Bang:
**PR 3 of the `#(authorize)` series** — the runtime gate, i.e. the actual access enforcement. Builds on #805; base it on #805.

Remaining after this: PR 4 (MCP + HTTP `/compile` surface), PR 5 (docs), PR 6 (e2e + example).

### What this one does
Before a query touches an authorize-annotated source, evaluate the gate; deny with **403** unless at least one in-scope expression is true for the supplied givens.

- `Model.assertAuthorized(source, givens)` — empty annotations = unrestricted; otherwise runs `buildAuthorizeProbe` via `.run({ givens })` and grants on **any** true expression (OR semantics).
- **Fail closed:** any probe error (notably a referenced given with no supplied value), a missing row, or a non-true result → `AccessDeniedError`. This is how "missing given → deny" falls out for free. The 403 names only the source, never the expression.
- `isProbeTrue` is strict (`true`/`1`/`"true"` grant; null/missing/anything-else deny).
- Gate runs **first** in `getQueryResults` and `executeNotebookCell` — before filter injection / `loadQuery`, **regardless of `bypassFilters`** (authorize is an access gate, not a filter), and outside the loadQuery try so the 403 isn't rewrapped as a 400. Covers **POST /query**, the **notebook-cell GET**, and the **MCP `executeQuery`** tool (routes through `getQueryResults`). The **`/compile`** path is carried into PR 4 with the rest of the HTTP/MCP surface.

### Behavioral correction (found while testing)
The plan assumed authorize is *not* inherited through `extend` (a documented "footgun"). Reality is **safer**: Malloy carries a base's `#(authorize)` onto an extension **unless the extension declares its own**. So a base locked with `#(authorize) "false"` also locks every extension that doesn't re-expose itself with its own gate. Joins stay ungated (top-level source only). Comments corrected; both behaviors are pinned by tests.

### Tested
- `tsc` + eslint clean; 794 unit tests pass
- Runtime gate: allow when satisfied; deny when not; **deny on missing given value (fail closed)**; gate still enforced under `bypassFilters`; file-level vs source-level disjunction (each alone grants; neither denies); unrestricted source passes
- Locked-base pattern: direct query on `#(authorize) "false"` base denied; gated extension allowed; **ungated extension denied (inherits the base lock)**; join-to-locked-base allowed (documented top-level-only limitation)
- `isProbeTrue` unit matrix
- worker-path + `filter_integration` green

### DIFF
--- a/packages/server/src/errors.ts
+++ b/packages/server/src/errors.ts
@@ -6,6 +6,8 @@ export function internalErrorToHttpError(error: Error) {
       return httpError(400, error.message);
    } else if (error instanceof FrozenConfigError) {
       return httpError(403, error.message);
+   } else if (error instanceof AccessDeniedError) {
+      return httpError(403, error.message);
    } else if (error instanceof EnvironmentNotFoundError) {
       return httpError(404, error.message);
    } else if (error instanceof PackageNotFoundError) {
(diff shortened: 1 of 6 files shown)


Sha-Bang:
Moving to draft, waiting on other conflicting PR (https://github.com/malloydata/publisher/pull/807) to merge first.

Merge complete, rebased, ready for review.

kylenesbit (approved):

PR #811, opened 2026-06-03
feat(server): surface each given's default as a source literal

Sha-Bang:
Surfaces each given's declared default and shows it in the givens UI, so the input communicates the default instead of looking blank.

**API** (`server`): givens introspection now carries `Given.default` — Malloy's already-rendered source literal (`'WN'`, `2003`, `@2024-01-01`, `f'WN'`), omitted when there's no default.

**UI** (`sdk` GivensPanel): the default shows as a *hint*, not a prefill — the input stays empty so leaving it blank still means "use the model default" (the server fills it in); the user just now sees what it is. An always-visible `Default: …` helper line on every widget (string, number, date, filter, array, boolean), plus a ghost placeholder on the text inputs (a bonus MUI reveals on focus). The raw literal is rendered per type by `renderGivenDefault` (`'WN'`→`WN`, `@2024-01-01`→`2024-01-01`, `f'WN'`→`WN`, `\'`→`(empty)`).

Also unblocks the ms2data/service report-givens panel (the original #4966 review item), which consumes the same `Given.default`.

### Tested
- `server`: `givens_integration` — a defaulted given surfaces its literal (`region_filter → 'US'`, `cutoff_date → @2024-02-01`), a no-default given omits it.
- `sdk`: `renderGivenDefault` unit tests (string / escaped-quote / date / filter / number / empty-string / no-default); full sdk suite 25/25.
- `app` Playwright (`notebook-givens`): updated the description-helper assertion (the line now also carries the default) and added **"model default surfaces as a helper line"** asserting `target_code → Default: WN` (string) and `cutoff → Default: 2024-01-01` (date). 7/7 pass locally.
- **End-to-end through the live UI** against `faa-givens-demo`: `carrier` (string) shows `Default: WN`, `after` (timestamp) shows `Default: 2003-01-01 00:00:00`, inputs empty, queries still running on the model defaults (Southwest / 46,489 flights).
- `tsc` + `eslint` clean in both packages.

Coverage gap: the boolean-default caption is tsc/eslint-verified but not exercised by a browser test (no boolean given in the fixtures).

### Review
- Design decision (placeholder-hint vs prefill) recorded in a [PR comment](https://github.com/malloydata/publisher/pull/811#issuecomment-4617221414).
- `/codex review` (gate PASS) caught two display-logic edge cases — boolean defaults (a `boolean is true` given showed an unchecked box while running `true`) and explicit empty-string defaults being suppressed. Both fixed: the checkbox now carries the same `Default: …` caption, and the helper line tests `!== undefined`.

### DIFF
--- a/packages/sdk/src/components/given/utils.ts
+++ b/packages/sdk/src/components/given/utils.ts
@@ -1,0 +1,60 @@
+/**
+ * Render a given's `default` (a raw Malloy source literal) for display in the
+ * input as placeholder / helper text. Returns undefined when there is no
+ * default to show.
+ *
+ * The server surfaces `Given.default` exactly as written in the model
+ * (`'WN'`, `2003`, `@2024-01-01`, `f'WN'`). For display we unquote string
+ * literals, drop the date `@` sigil, and unwrap filter literals so the user
+ * sees `WN` / `2024-01-01` rather than the raw source form. Non-string scalars
+ * (number, boolean) and anything we don't special-case are shown verbatim.
+ *
+ * This only affects what the user *sees* — the value stays empty, so leaving
+ * the field blank still means "use the model default" (the server fills it in).
+ */
+export function renderGivenDefault(
+   type: string | undefined,
+   rawDefault: string | undefined,
+): string | undefined {
+   if (!rawDefault) return undefined;
+   const literal = rawDefault.trim();
+   if (!literal) return undefined;
+
+   const givenType = type ?? "string";
+
+   if (
+      givenType === "date" ||
+      givenType === "timestamp" ||
+      givenType === "timestamptz"
+   ) {
+      // `@2024-01-01` -> `2024-01-01`
+      return literal.replace(/^@/, "");
+   }
+
+   if (givenType.startsWith("filter<")) {
+      // `f'WN'` -> `WN`: drop the filter-literal `f` prefix, then unquote.
+      return unquoteStringLiteral(literal.replace(/^f/, ""));
+   }
+
+   if (givenType === "string") {
+      return unquoteStringLiteral(literal);
+   }
+
+   // number, boolean, array, or unknown: show the literal as authored.
+   return literal;
+}
+
+/**
+ * Strip the surrounding single quotes from a Malloy string literal and undouble
+ * any escaped quotes (`''` -> `'`). Leaves a non-quoted input untouched.
+ */
+function unquoteStringLiteral(literal: string): string {
+   if (
+      literal.length >= 2 &&
+      literal.startsWith("'") &&
+      literal.endsWith("'")
+   ) {
+      return literal.slice(1, -1).replace(/''/g, "'");
+   }
+   return literal;
+}
--- a/packages/sdk/src/components/given/utils.ts
+++ b/packages/sdk/src/components/given/utils.ts
@@ -3,11 +3,13 @@
  * input as placeholder / helper text. Returns undefined when there is no
  * default to show.
  *
- * The server surfaces `Given.default` exactly as written in the model
- * (`'WN'`, `2003`, `@2024-01-01`, `f'WN'`). For display we unquote string
- * literals, drop the date `@` sigil, and unwrap filter literals so the user
- * sees `WN` / `2024-01-01` rather than the raw source form. Non-string scalars
- * (number, boolean) and anything we don't special-case are shown verbatim.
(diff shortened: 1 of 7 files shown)


Sha-Bang:
Screenshot 2026-06-03 at 6 18 09 PM

Sha-Bang:
## UI design: two options considered

When deciding how `Given.default` should appear in the SDK `GivensPanel`, there were two ways to use it:

**Option A — show it as a hint, input stays empty (chosen)**
```
┌──────────────────────────┐
│  [ carrier            ]   │   input empty
│  …(e.g. WN, AA, DL)       │
│  Default: WN              │   ← default shown as a caption
└──────────────────────────┘
```
The field stays blank, so leaving it untouched still means "use the model default" — the server fills it in. The user just *sees* what the default is. Clean clear-to-revert semantics; no redundant override sent.

**Option B — prefill the value**
```
┌──────────────────────────┐
│  [ WN               × ]   │   value populated
└──────────────────────────┘
```
Rejected: prefilling makes every field an active override equal to the default, so `getActiveGivens()` sends `carrier=WN` even when the user never touched it, and "clear to revert to default" gets muddy (clearing a prefilled field isn't obviously different from blanking it).

**Chose A.** The server already applies the model default for any omitted given, so A keeps the existing "blank = default" contract and just makes the default *visible*.

### Refinement from end-to-end testing
First cut of A used a ghost **placeholder** for text inputs. Running it live against `faa-givens-demo` showed MUI parks the floating label in the placeholder's spot when a field is empty + unfocused, so `WN` only appeared on focus. The date picker (which used a caption) showed fine. Fixed by making the `Default: …` **caption always-visible** on every widget, keeping the ghost placeholder as an on-focus bonus.

Verified live: `carrier` (string) → `Default: WN`, `after` (timestamp) → `Default: 2003-01-01 00:00:00`, both at rest with empty inputs, queries still running on the model defaults. Boolean givens render as a checkbox (no helper slot), so their default isn't surfaced.

packages/server/src/service/given.ts
```diff
@@ -36,6 +36,12 @@ export interface MalloyGivenApi {
    name: string;
    type: string;
    annotations?: string[];
+   /**
+    * The given's default as a Malloy source literal (`'WN'`, `2003`,
+    * `@2024-01-01`, `f'WN'`), or omitted when the given has no default.
```
sagarswamirao: Could we make this comment more generic by removing the specific examples?

packages/server/src/service/given.ts
```diff
@@ -50,10 +56,12 @@ export interface MalloyGivenApi {
  *   location either; matching that floor. A future PR can add a
  *   sanitised package-relative path if a client needs it.
  *
- * - `default` / `defaultText` — Malloy's API only exposes the
- *   parsed `ConstantExpr` AST, not a rendered source string.
- *   Rendering it here would duplicate the Malloy printer. Add
- *   when Malloy surfaces a stringified accessor.
+ * - `default` is surfaced as the rendered source literal
+ *   (`given._internal.defaultText` — e.g. `'WN'`, `2003`,
+ *   `@2024-01-01`). Malloy's public surface still exposes only the
```
sagarswamirao: Could we make this comment more generic by removing the specific examples?

packages/sdk/src/components/given/utils.ts
```diff
@@ -0,0 +1,74 @@
+/**
+ * Render a given's `default` (a raw Malloy source literal) for display in the
+ * input as placeholder / helper text. Returns undefined when there is no
+ * default to show.
+ *
+ * The server surfaces `Given.default` exactly as written in the model
+ * (`'WN'`, `2003`, `@2024-01-01`, `f'WN'`). For display we unquote string
+ * literals, drop the date `@` sigil, and unwrap filter literals so the user
+ * sees `WN` / `2024-01-01` rather than the raw source form. Non-string scalars
+ * (number, boolean) and anything we don't special-case are shown verbatim.
```
sagarswamirao: Could we make this comment more generic by removing the specific examples?

Sha-Bang: Kept the concrete examples since they're traceable to the tests/samples, but broadened them to one-per-type — string `'WN'` / `"WN"`, number `2003`, boolean `true`, date `@2024-01-01`, filter `f'WN'` — so they read as illustrative of the range rather than a fixed list. 38357ce

kylenesbit (approved):

PR #813, opened 2026-06-04
  Serve in-package HTML data apps and fix watch-mode live reload

mlennie:
## Summary

Continues @jswir's #724. Package authors can drop an `index.html` (plus CSS, JS, and assets) next to their `.malloy` files; Publisher serves it at `/environments/<env>/packages/<pkg>/<file>`, lists pages at `GET .../pages`, and renders them inside the app or as an auto-resizing embedded iframe. With `--watch-env`, source edits live-reload over SSE and `.malloy` edits recompile the package.

It also fixes watch mode, which used to detect file changes but never recompile: the old reload ran an env-level metadata refresh that never touched compiled packages, and it read the staged `publisher_data` copy rather than the watched source.

## Commits

The first four are @jswir's feature and fix:

- `feat(server,sdk)`: serve & view in-package HTML data apps (static serving, `/pages`, `PageViewer`, runtime)
- `feat(server)`: dev live-reload via watch-mode events (`/events` SSE, `--watch-env`, in-place symlink mount)
- `fix(server)`: recompile the changed package on watch events (`Environment.getPackage(name, reload=true)`)
- `fix(server)`: signal live-reload only after a successful recompile

Added to land it:

- `test(server)`: an integration test for watch-mode source-edit live reload
- `docs`: README coverage for `--watch-env` / `PUBLISHER_WATCH`, `PUBLISHER_FRAME_ANCESTORS`, and in-package data apps

## How watch mode behaves

The in-place symlink mount is gated on `--watch-env` (via `PUBLISHER_WATCH`), so production keeps copy semantics and packages stay decoupled from their source. Started with `--watch-env <env>`, a local-dir package is mounted in place (a symlink, or a directory junction on Windows), the watcher recompiles only the changed package, and open pages reload over SSE after a clean recompile. Without the flag the package is copied and source edits do not propagate, by design.

## Tests

- New `packages/server/tests/integration/watch-mode/watch_mode.integration.spec.ts` boots a real server subprocess with `--watch-env` and asserts the in-place mount, that adding a new `.malloy` and editing an existing model both recompile and surface through the models API, the REST start/status/stop and validation behavior, and the copy-mode negative (no flag, no propagation). It uses a subprocess because the shared in-process REST harness can't set `PUBLISHER_WATCH` at boot or seed the config.
- Full server suite is green locally (unit 817, integration 191, 0 failures). The integration suite, including this test, runs on ubuntu, macos, and windows in CI.

## Verified manually

Built the example app and ran it under `--watch-env`: the Pages section and `PageViewer` render with live data and auto-resize, editing `carriers.malloy` hot-reloads the open tab, `embed-test.html` embeds into an external host page with cookie auth and auto-resize, and `PUBLISHER_FRAME_ANCESTORS` blocks a disallowed origin (the header is set on HTML responses only).

## Notes

- Restricting `PUBLISHER_FRAME_ANCESTORS` in production also blocks Publisher's own in-chrome page viewer unless the value includes Publisher's origin.
- Auth is unchanged: Publisher stays auth-less and whatever fronts it owns authentication; embed auth flows are a follow-up.
Screenshot 2026-06-04 at 9 38 48 AM
Screenshot 2026-06-04 at 9 38 57 AM
Screenshot 2026-06-04 at 9 39 07 AM


jswir (commented):
## Peer Review

### Summary
Solid, well-tested PR — the watch-mode recompile fix correctly closes its root cause (replacing the no-op `addEnvironment` metadata refresh with a forced `getPackage(name, reload=true)`), and the path-traversal defense (`realpath` + containment) is sound and has an escape test. Two real correctness gaps stand out: the manifest-privacy guard is case-sensitive (bypassable on macOS/Windows), and the synthetic `pages/` routing prefix collides with real model subdirectories.

### Findings

**Important:**

1. **`packages/server/src/server.ts:377-380` — Manifest-blocking guard is case-sensitive and bypassable on case-insensitive filesystems.**
   `path.basename(urlRel) === "publisher.json"` is a strict byte compare. On macOS (APFS) and Windows (NTFS) — macOS being the primary dev/deploy platform — `GET .../packages/<pkg>/Publisher.JSON` fails this equality, passes the guard, and `realpath` + `sendFile` then serve the manifest (which can hold a `secret` field). The block exists to keep the manifest private; it doesn't on the OS most people run this on. Fix: `path.basename(urlRel).toLowerCase() === "publisher.json"`.

2. **`packages/app/src/components/pages/ModelPage/ModelPage.tsx:34` — `pages/` routing prefix collides with a real model subdirectory.**
   Models are walked with `recursive-readdir` (`package_load_worker.ts:397`), so `pages/carriers.malloy` is a legitimate model path. `if (modelPath?.startsWith("pages/"))` fires before the `.malloy`/`.malloynb` checks below it, so every model inside a directory named `pages` opens in `PageViewer` (iframed as HTML) instead of the `Model` viewer. Guard with an extension check (`&& !modelPath.endsWith(".malloy") && !modelPath.endsWith(".malloynb")`) or use a collision-resistant synthetic key (e.g. `__page:/`) instead of a path-shaped prefix.

**Nit:**

1. **`packages/server/src/server.ts:1887` + `packages/server/src/service/environment_store.ts:139-143` — `PUBLISHER_WATCH=a,b` mounts *all* listed envs in-place (symlinked, live source) but watches only the first; the log says it's "ignoring the rest."** `inPlaceEnvs` is populated from the full list, so envs 2…N get live source but no auto-reload — that's neither "ignored" nor "one watched." The README/`--help` ("pass a comma-separated list") advertises a capability that only half-works. Either tighten the message to "the others are mounted in-place but will not live-reload" or drop the multi-value claim from the docs.

2. **`packages/server/src/controller/watch-mode.controller.ts:203` — `startWatching` has no `try/catch` around `ensureWatching`.** `ensureWatching` → `getEnvironment` can throw (NotFound/ServiceUnavailable), surfacing as a 500 + stack trace via Express's default handler. Sibling handlers (`/pages`, `/events`) wrap `getEnvironment` in `try/catch` + `internalErrorToHttpError` — match that.

3. **`api-doc.yaml` (list-pages endpoint, ~L2069) — missing `"400"` response.** The sibling `/events` endpoint added in this same PR documents `"400"`, and list-pages returns 400 for a malformed env/package name via `assertSafePackageName`. Add `"400": $ref: "#/components/responses/BadRequest"`.

4. **`packages/server/src/server.ts:395,501` — `await import("fs/promises")` inside the per-request handlers `serveFromPackage` and `listPackagePages`.** Every other module imports `fs/promises` statically at the top (`service/model.ts:24`, `service/connection.ts:24`). Node caches the module so cost is negligible, but it's inconsistent with house style — hoist to a top-level import.

5. **`packages/server/tests/integration/html_pages/html_pages.integration.spec.ts:250` — SSE test asserts event names but not data payloads.** It checks `event: hello` / `event: mode` but not `data: connected` / `data: disabled`, which `api-doc.yaml` specifies. A regression that sends `data: broken` would pass. Pin the `data:` values.

6. **`packages/server/src/server.ts:322` — stale comment `// Read once at module load for speed` on `PUBLISHER_RUNTIME_PATH`.** Only the path string is resolved once; `sendFile` reads the file from disk every request. Reword to `// Path resolved once at module load`.

7. **`packages/server/src/server.ts:744` — `req.on("aborted", cleanup)` is dead code on Node ≥20** (the event was removed in v19; `engines` requires `node >= 20`). `"close"` already covers both clean and abrupt disconnects. Remove it so readers don't assume two cleanup paths.

Plus ~5 similar lower-confidence items not listed: `PUBLISHER_FRAME_ANCESTORS` written to the CSP header without validation (`server.ts:423` — low risk, operator-controlled, Node throws on CRLF); `this.watcher` not nulled in `stopWatchMode`; `.json` in `ASSET_EXTS` triggers a browser reload on `publisher.json` edits; `pagePath`/`resource` URLs not percent-encoded for filenames with spaces; `PageViewer.tsx` error-handling shape doesn't mirror `Package.tsx`'s explicit try/catch.

---

Neither blocker is severe enough to hold the PR if the two Important findings are addressed (the manifest fix is a one-liner; the route guard is an extension check). The rest are follow-up quality items.

packages/server/src/server.ts
```diff
@@ -286,6 +307,279 @@
...
+      //      resolves it; legitimate accesses inside the symlinked source
+      //      stay within realPkgRoot and pass.
+      // Missing-file handling: realpath throws on ENOENT; we 404 cleanly
+      // instead of leaking via Express's default error handler.
+      const fsp = await import("fs/promises");
+      let realPkgRoot: string;
+      let realFullPath: string;
+      try {
+         realPkgRoot = await fsp.realpath(pkgRoot);
+         realFullPath = await fsp.realpath(fullPath);
```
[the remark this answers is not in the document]
Sha-Bang: Some of these comments are closer to noise, this one is probably more concerning.   This exposes raw .parquet, .csv, .duckdb, .malloy, credentials, and other source/data files directly at /environments/.../packages/..., bypassing Malloy  #(authorize) and query controls. 


There is probably some more elegant ways of verifying the path, but I would start with restricting access to an explicit public asset allowlist and(or) a public/ directory.

You will probably want to normalize the path then verify it is in the public directory.

packages/sdk/src/components/PageViewer/PageViewer.tsx
```diff
@@ -0,0 +1,164 @@
...
+         >
+            <iframe
+               ref={iframeRef}
+               src={standaloneUrl}
+               title={title}
+               // allow-same-origin so the iframe can use cookies for the
+               // Publisher API. allow-scripts so the dashboard's JS runs.
+               // allow-forms so dropdown/filter forms work. Intentionally
+               // NOT allowing top-navigation or popups.
+               sandbox="allow-scripts allow-same-origin allow-forms"
```
Sha-Bang, line 153: I think this effectively removes the sandbox from protecting from the attacks you should be concerned about... let me think on this.

Sha-Bang: allow-same-origin is probably the one that we could easiest engineer a solution to, but it gets complex (1 to N user supplied content subdomains).  We probably just need to be sure that we are only dealing with trusted authors; even then we may want to think about how to reduce blast radius across multi-tenant use cases.

Probably not blocking now.

mlennie: @Sha-Bang agreed, and not blocking. You're right that `allow-scripts allow-same-origin` effectively voids the sandbox here, so this needs a real fix rather than a flag tweak: I'm tracking it as a follow-up (isolate served pages on their own origin / per-tenant content subdomains, and/or scoped server-validated embed tokens so we can drop `allow-same-origin`), tied to the multi-tenant story. The separate raw-file data exposure from your other comment is fixed in `ffb56c4`.

Sha-Bang (approved):
Nice work

PR #814, opened 2026-06-04
feat(server): gate /compile + align authorize error surfaces

Sha-Bang:
Part of the `#(authorize)` series (#802, #804, #805, #806 merged). The runtime gate (#806) covered `/query` + notebook cells but deferred the `/compile` path and the MCP/HTTP error polish — this closes that.

### What it does
- **`/compile` gate.** `compileSource` is compile-only, but it can still act as a schema oracle (a denied caller learns a gated source's columns from compile errors) and, with `includeSql`, leak its SQL. It now gates the named source the submitted text targets **before** compiling (early surface-syntax gate, mirrors `/query`), plus a **compiled-source backstop** — run **unconditionally** — that resolves the compiled final query's `structRef` and closes named-query / multi-statement indirection (`run: ungated\nrun: gated` — the executed statement isn't the first). Reuses the package's cached `Model` via two thin public wrappers (`assertAuthorizedForText`, `assertAuthorizedForRunnable`), plus `Model.hasAuthorize()` so ungated models skip the extra final-query compile. Named-source only; inline/unnamed text gets the model-wide file-level gate (a `source: x is gated` alias is the documented "extend doesn't inherit authorize" footgun).
- **MCP.** `AccessDeniedError` already flowed through cleanly (names the source, never the expression) but landed on generic Malloy-syntax suggestions. Added an access-denied branch to `getMalloyErrorDetails` so the hint is about supplying the required givens; and a matching branch in `getModelForQuery` (`handler_utils`) so a setup-path denial is homed too, never an opaque internal error.
- **Redaction policy.** Runtime 403 stays source-only (already true); the model-load 424 keeps the full expression — author-facing at package load, needed to fix a malformed annotation. No code change; pinned by a test.

### Tested
- `errors.spec`: `AccessDeniedError`→403, `ModelCompilationError`→424.
- `error_messages.spec`: the MCP access-denied suggestion replaces the generic syntax advice.
- `authorize_integration` compile-path block: `assertAuthorizedForText` allow/deny/file-level/ungated + `assertAuthorizedForRunnable` backstop; plus a redaction pin (the 424 surfaces the expression).
- `compile_authorize.spec`: **end-to-end `environment.compileSource`** — direct-gated denied (early gate), multi-statement gated-last denied with `includeSql=false` (regression guard for the unconditional backstop), allowed with the given, ungated compilable.
- `compile_authorize_http.integration.spec`: **HTTP end-to-end** — `POST /compile` returns a real `403 {code:403, message:Access denied for source "gated".}` body through the express route, with no expression leak; 200 with the satisfying given; ungated compiles.
- `tsc` + `eslint` clean; server unit suite green.

### Known test gap (deferred — harness limitation)
No full MCP `execute_query` integration test (real `callTool` against a gated model). The MCP E2E harness binds `mcpApp` only (no REST `?reload=true` route) over a process-singleton env, so it can't seed a gated model per-test without ordering fragility in the serial suite. The `AccessDenied → isError + suggestion` path is already covered piecewise: `error_messages.spec` (the transform), existing MCP error tests (the `isError` JSON shape), the `handler_utils` branch (setup path), and the HTTP e2e above (gate → real error body). **Follow-up:** make the MCP E2E harness seedable (bind the full `app`, or a per-test `SERVER_ROOT` override, or a reload hook) — infra that unlocks clean MCP integration tests here and for every future MCP feature, and removes the singleton flake risk. Then add the `callTool` test.

### ⚠️ Deferred — restricted mode on `/compile` (tracked follow-up)
This change's authorize gate covers the **named Malloy source**, but `/compile` still compiles **unrestricted** (`loadFinalQuery`, not `loadRestrictedQuery`). A denied caller can therefore bypass the gate via raw SQL — e.g. `compile(source: "run: duckdb.sql(\"SELECT * FROM <underlying_gated_table>\") -> {…}", includeSql: true)` returns the underlying schema/SQL, because authorize never inspects raw SQL. This is the **same hole #807 closed on `/query`** via restricted mode; `/compile` was never hardened.

Decision: **defer to its own change** — a distinct hardening mechanism (like #807, with its own test surface), blast radius bounded by the trusted-tier deployment model, and pre-existing (not a regression introduced here). It is the next security item on the authorize axis.

### Other follow-ups (pre-existing, out of scope)
- `/compile` doesn't call `assertCanAdmitQuery` (memory back-pressure gate).

Out of scope (unchanged): join-to-locked-base + the extend footgun remain the documented top-level-only limitation; docs and end-to-end tests land in later changes in the series.


Sha-Bang:
## Manual testing

Ran the `/compile` gate live against a local server with a gated fixture package (`#(authorize) "$ROLE = 'analyst'"` on `gated`, plus an ungated `open_src`).

```
POST .../models/model.malloy/compile

1. run: gated -> { aggregate: c }           givens {}                 → 403 {"code":403,"message":"Access denied for source \"gated\"."}
2. run: gated -> { aggregate: c }           givens {ROLE:"analyst"}   → 200 {"status":"success"}
3. run: open_src ... \n run: gated ...      givens {}, includeSql:false → 403  (gated is the LAST run; early gate only saw open_src — backstop caught it)
4. run: open_src -> { aggregate: c }        givens {}                 → 200
```

Confirmed:
- Deny is a clean **403** whose message names **only the source** — no `ROLE` / `analyst` / expression leak (runtime redaction).
- The compiled-source backstop fires **without `includeSql`** (case 3), closing the multi-statement path the early surface-syntax gate misses.
- Ungated sources compile freely.

The **424** model-load case (a malformed `#(authorize)` keeps the full expression for the author) is covered by the unit test (`authorize_integration` asserts the `ModelCompilationError` message contains the raw expression). Demonstrating it through the live package-reload path fights the install/manifest machinery, so it's verified by the test rather than by hand.

sagarswamirao (approved):

PR #816, opened 2026-06-04
chore(deps): upgrade @malloydata/* to v0.0.405

kylenesbit:
## Summary

Upgrades all `@malloydata/*` dependencies from `^0.0.403` to **v0.0.405** (via `scripts/upgrade-malloy.sh 0.0.405`), rebased on current `main`. `malloy-explorer` stays pinned since 0.0.405 isn't published for it.

## Breaking changes handled (malloy 0.0.405)

| Change | Migration |
|---|---|
| `ModelDef.annotations` removed — model (`##`) annotations now live in `ModelDef.modelAnnotations` (a `modelID → {ownNotes, inheritsFrom}` registry), folded across the import/extend lineage by malloy's `getModelAnnotations` (which is **not** on the public barrel) | Added a centralized `modelAnnotations(modelDef)` helper in `service/annotations.ts` that replicates the fold (cycle-safe post-order DFS; target's own notes at the top of the `inherits` chain) so existing `.notes`/`.texts()` readers work unchanged. Applied at the three former `modelDef.annotations` sites. |
| `FetchSchemaOptions.modelAnnotations` removed | Dropped from the worker serialize/build helpers and the worker RPC protocol wire types. |

## DuckDB runtime pin (macOS CI fix)

`@malloydata/db-duckdb@0.0.405` bumps the duckdb native runtime (`@duckdb/node-api`) from `1.4.4-r.1` → `1.5.3-r.2`. In 1.5.3 the core extensions (httpfs, ducklake, postgres, aws, …) are no longer bundled and are fetched on demand from `extensions.duckdb.org`. The GitHub **macOS** runner can't reach that host ("Could not establish connection"), so every test that loads a DuckDB extension failed (and the native layer aborts, exit 133) — while ubuntu/windows, which can download, passed. duckdb 1.4.4 ships those extensions in the bindings, which is why `main` is green on macOS.

Fix: pin `@duckdb/node-api` and `@duckdb/node-bindings` to `1.4.4-r.1` via `resolutions` (the version `main` already runs). `db-duckdb@0.0.405` is ABI-compatible with it — the full server `test:unit` suite is 840 pass / 0 fail. Remove the pin once 1.5.x extensions are bundled again or the runner can fetch them.

## Verification

- ✅ All CI checks green, including Cross Platform Tests on **macOS / ubuntu / windows**, Typecheck, Lint, Docker smoke, Playwright, DCO
- ✅ Locally (CI Bun 1.3.13): server `test:unit` 840 pass / 0 fail

Pre-existing `main` failures unrelated to this bump (verified identical before it): the `authorize`/`default` API-schema type errors in integration specs + `model.ts`, and the `GivenInput` `Given.default` error — all from un-regenerated API schema types.


Sha-Bang (approved):

packages/server/src/service/annotations.ts
```diff
@@ -25,6 +25,44 @@ export function isReservedRoute(route: string): boolean {
...
+   };
+   visit(modelDef.modelID);
+
+   // Fold most-ancestral → local so the local model lands at the top of the
+   // resulting `inherits` chain.
+   let folded: AnnotationsDef | undefined;
+   for (const id of order) {
+      folded = { ...registry[id].ownNotes, inherits: folded };
+   }
+   return folded ?? {};
```
sagarswamirao, line 77: **Inherited `##` notes are dropped here vs. malloy's own fold (authz fail-open).**

This fold keeps a link for *every* id in `order` and spreads `ownNotes`, so the top link is always the local model and `.notes` returns the local model's own `##` notes (possibly empty). malloy's `getModelAnnotations` (`@malloydata/malloy/dist/model/annotation_utils.js`) instead **skips** any id whose `ownNotes` has no `notes`/`blockNotes`, so when the local model has no own `##` but an imported model does, malloy's `.notes` is the nearest ancestor's notes.

Reproduced:

```
CASE A: local model has no own ## notes, an imported model does
  this fold .notes: []            <- inherited ##(authorize) dropped
  malloy    .notes: ["## base"]
```

Both `.notes` readers feed file-level `##(authorize)` enforcement (`model.ts` `fileLevelAuthorize`, `source_extraction.ts` `fileLevelAuthorize`). Since `inheritsFrom` includes imported models, a `##(authorize)` declared in an imported file is meant to flow into the importing file's gate; under this fold it is silently dropped whenever the importing file declares no `##` of its own -- a fail-open direction for an authorization gate.

(`.texts()` consumers such as the notebook `allAnnotations` are unaffected -- both folds flatten the same lineage.)

Suggest matching malloy: skip empty-`ownNotes` links and copy only `notes`/`blockNotes` (don't spread `ownNotes`, which can carry a stale `inherits`):

```ts
let folded: AnnotationsDef | undefined;
for (const id of order) {
   const own = registry[id].ownNotes;
   if (!own.notes?.length && !own.blockNotes?.length) continue;
   folded = { notes: own.notes, blockNotes: own.blockNotes, inherits: folded };
}
return folded ?? {};
```

Worth a regression spec: a model that imports a file carrying `##(authorize)` and declares no own `##`, asserting the gate still enforces.

sagarswamirao:
**Coverage gap: file-level `##(authorize)` inherited from an `import` is untested -- and this PR's fold changes exactly that path.**

The `##(authorize)` enforcement was built across #804 (parse/surface), #805 (compile-time validate), #806 (query-time runtime gate), and #807 (restricted-mode hardening). #804 introduced the first file-level read of `modelDef.annotations.notes` (in `source_extraction.ts`), and #806 added the enforcing read in `model.ts` plus the bulk of these integration specs.

Across all of them the specs only cover `##(authorize)` declared **in the same file** and source-level inheritance through `extend` (the `does NOT inherit a base source's authorize through extend` case above). None use `import:` -- `grep "import:"` finds zero matches in either authorize spec. So a `##(authorize)` declared in an **imported** model, flowing into the importing file's file-level gate, was never tested.

That gap was latent and harmless on malloy <= 0.0.404, where `ModelDef.annotations` was already the correctly-folded bundle. It bites now because this PR's hand-rolled `modelAnnotations` fold diverges from malloy's `getModelAnnotations` on exactly that path (see the comment on `annotations.ts`): when the importing file declares no `##` of its own, `modelAnnotations(modelDef).notes` returns `[]` and the inherited gate is silently dropped (fail-open).

Suggested regression spec: model A `import`s model B; B carries `##(authorize) "$ROLE = 'admin'"`; A declares no own `##` and a source; assert A's source enforces the gate (denies without the role, allows with it). With this PR as-is the test fails; with the `annotations.ts` fold fixed to match malloy's `getModelAnnotations`, it passes.

packages/server/src/service/annotations.ts
kylenesbit: Great catch — fixed in 1d7d4fd. You're right that this is a fail-open direction for the authorize gate. I matched malloy's `getModelAnnotations` exactly: skip any model whose `ownNotes` has no `notes`/`blockNotes`, and copy only `notes`/`blockNotes` rather than spreading `ownNotes` (so a stale `inherits` can't leak in). Now `.notes` falls back to the nearest ancestor that actually has notes, so an imported `##(authorize)` flows into an importing file that declares no `##` of its own.

Added `annotations.spec.ts` coverage too, including the exact regression you described (imported `##(authorize)`, no local `##` → gate still sees it), plus a cycle-safety case. The old fold returned `[]` for that case; it now returns the inherited note.

kylenesbit:
Added in 7d7099a — `authorize_integration.spec.ts` now covers the import path you described. Model A `import`s model B; B carries `##(authorize) "$ROLE = 'admin'"`; A declares no `##` of its own and a source. Three assertions across all layers:

- **Introspection**: `getAuthorize("inherited_gate")` surfaces the inherited gate (previously `[]`).
- **Runtime gate, allow**: query runs with `{ ROLE: 'admin' }`.
- **Runtime gate, deny**: query throws `AccessDeniedError` with `{ ROLE: 'intern' }` — not fail-open.

All three fail against the previous fold and pass with it matching malloy's `getModelAnnotations` (the fix in 1d7d4fd). Also added `annotations.spec.ts` unit coverage for the fold itself (inherited-fallback + cycle-safety). Thanks for the thorough catch.

PR #820, opened 2026-06-05
fix(duckdb): bake all extensions offline (CA certs + verification), drop external ADBC driver

sagarswamirao:
## Summary

Removes the external ADBC Snowflake driver install from the image, fixes the CA-cert gap that broke the extension bake, and adds offline verification of the baked extensions. Stacked on #819.

## Changes

- **Remove the external ADBC driver install** (`install-adbc-driver.sh` + `ldconfig`) from the Dockerfile. The Snowflake community extension on 1.5.x bundles its own driver (verified offline in built `linux/amd64` and `linux/arm64` images: with no external driver present, `snowflake_query` loads and reaches credential validation). This also drops a network dependency from the build.
- **Install `ca-certificates` in the builder stage.** The `bun:slim` base ships without them, so the extension bake (run by `packages/server`'s build in the builder) failed every download with an SSL CA-cert error and baked only 1 of 7 extensions. base-deps already had the certs; the builder did not.
- **Verify baked extensions in the bake script.** After `INSTALL`/`LOAD`, query `duckdb_extensions()` to confirm each reports `loaded`, and print an `N/7 loaded` summary (non-fatal -- the bake is an optimization, not a build gate).
- **Verify baked extensions in the Docker smoke test.** With the network cut off (`--network none`), assert the CLI version matches the baked version and that every extension (httpfs, aws, azure, postgres, ducklake, bigquery, snowflake) loads from the on-disk cache. This catches a broken bake, a version/dir mismatch, or a Snowflake bundled-driver regression at build time rather than at a user's first attach.

## Verification

- Built `base-deps` on `linux/amd64` and `linux/arm64`; Snowflake loads offline via the bundled driver with no external `.so` present.
- The smoke-test offline check correctly fails when an extension is not baked (proven against a stage that omits it).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -270,6 +270,47 @@ jobs:
               || { echo "✗ Missing/empty OCI label: org.opencontainers.image.$key"; exit 1; }
           done
           echo "✓ All five OCI image labels present"
+      - name: Verify baked DuckDB extensions
+        run: |
+          # The image bakes the DuckDB extensions the server loads at runtime
+          # (packages/server build runs bake-duckdb-extensions; the builder's
+          # ~/.duckdb/extensions is copied into the final image). This asserts
+          # they are actually present and LOAD *offline* -- with the network cut
+          # off, so a load can only succeed from the baked on-disk cache, never a
+          # fallback download. Catches a broken bake, a version mismatch between
+          # the baked dir and the CLI, or the snowflake bundled-driver regression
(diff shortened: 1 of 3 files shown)


sagarswamirao:
**Baking DuckDB extensions at build time:**
Since DuckDB extension versions are baked in at build time, they remain fixed after deployment and stay consistent across environments.

**NPM Build:**
image


**Docker Build:**
image

kylenesbit (approved):

PR #823, opened 2026-06-06
feat(server): explore visibility + query boundary via publisher.json explores

jswir:
## What

An opt-in `explores` field in `publisher.json` designating which `.malloy` files
form a package's public discovery surface, plus an optional `queryableSources`
boundary that turns that surface into an enforced one. Everything is gated on
declaring `explores`: **with no `explores`, discovery and querying behave exactly
as they do today** — all models listed, full source set, nothing blocked. Opt-in
and backward compatible by default.

This lets customers build layered models (base/bridge sources in their own files)
without those internals showing up in chat/discovery, and optionally stops them
being queried directly.

> **Naming:** the field is `explores` — the established term for a directly-
> queryable starting point, vs. the building-block sources it composes.

### Discovery surface — gated on `explores`

Two granularities, both active **only when `explores` is declared**:

- **File level — `explores`** decides which files are listed. Declared →
  `listModels()` returns only those models; other `.malloy` files still compile
  and stay joinable but are hidden from listings. Absent/empty → every file
  listed (today's behavior). Notebooks are always public.
- **Within a file — `export {}`** decides which sources within a listed file are
  surfaced — accessors list only `modelDef.exports`, matching Malloy's export-
  aware `modelInfo`/`sourceInfos`.

The complete source/query set is retained internally, so `#(filter)`/`#(authorize)`
and join/extend resolution are unaffected. Discovery hiding is not access control
on its own.

### Query boundary — `queryableSources` (`declared` | `all`, default `declared`)

Once `explores` is declared, the surface can also bound what's directly queryable:
- `declared` (default): queryable == discoverable. Hidden files/sources return
  `NotQueryableError` (404), indistinguishable from not existing.
- `all`: discovery curated, everything stays queryable by name (softer migration).

With **no `explores`, the boundary is inert** in both modes. Enforcement mirrors
the authorize gate: an early pre-compile gate that positively denies what it can
resolve (schema-oracle defense), and a compiled backstop that settles everything
else against the run target Malloy actually executes. Notebooks are exempt.

## Rollout — opt-in, no flag

The earlier revision shipped `export {}` curation unconditionally and added a
temporary `PUBLISHER_LEGACY_DISCOVERY` env flag to protect existing packages.
That's been **dropped** in favor of gating all curation on `explores`: a flag only
protects non-adopters while every deployment remembers to set it, which doesn't
hold for packages we don't own or control. Gating on `explores` makes "no
`explores` = today's behavior" true structurally — existing packages upgrade with
zero config, and migration is per-package and gradual.

## Invalid `explores` policy (asymmetric by design)

- **Strict at publish** — create and update reject an unresolved entry with an
  actionable `400`; create rolls back, update restores prior metadata before
  rejecting, so a bad write never persists or mutates the served surface.
- **Fail-safe at load** — warns and serves; an unresolved entry lists *nothing*
  rather than falling back to exposing everything. Re-emitted on reload.
- `exploresWarnings` is surfaced read-only on the package API so consumers can
  see why a package is empty; clean packages omit it.

## Tested

- Server unit suite: **887 pass**; explores/query-boundary targeted specs: **35
  pass**; typecheck + lint clean.
- Integration (real worker pool + HTTP/MCP): **225 pass** (1 unrelated failure —
  a fixture pulling from a private GCS bucket that needs local reauth).
  `queryableSources` verified serializing end-to-end through the live REST API;
  authorize compile flow 403→200 intact.
- `explore_visibility.spec`: legacy behavior without `explores`, file + `export {}`
  curation with `explores`, direct `getModel` on a hidden file, `export {}`
  without `explores` (no-op), soft migration via `queryableSources: "all"`.
- `query_boundary.spec`: `declared` vs `all`, inert without `explores`,
  multi-statement settled by the compiled run target, ad-hoc derivation laundering
  denied, and a regression test pinning that a hidden source can't be reached via a
  view whose name collides with an exported top-level query (the source-prefixed
  request gates on the *source*, not the query name).
- `package.controller.spec`: publish `400` on invalid `explores`. `errors.spec`:
  `NotQueryableError` → 404. Contract test pins Malloy's `ModelDef.exports`.
- Also fixes a pre-existing CI flake: `health.spec` drained the worker-pool
  singleton and poisoned later specs; now reset in `afterAll`.

## Docs

- README "Controlling the discovery surface": `explores` + `export {}`, how they
  compose, discovery-vs-access-control, `queryableSources`, publish-vs-load policy.
- `api-doc.yaml`: `explores`/`queryableSources` documented; `exploresWarnings`
  `readOnly`; create/update document the `400`.

## Downstream

- Consumers use `explores` / `exploresWarnings` / `queryableSources`.
- Retiring Credible's `.malloy_hidden` / `#(agent-hidden)` lives in separate
  Credible repos; this PR is the upstream source of truth.

## Note

Supersedes #825 — the `queryableSources` query boundary is folded in here so the
discovery surface and its enforcement land together behind the single `explores`
opt-in.


kylenesbit:
Took a full pass — the design is solid and the careful bits are genuinely careful: the fail-safe-at-load vs strict-at-publish asymmetry, the one-sided path normalization, and (importantly) the discovery/enforcement decoupling. I confirmed `getAuthorize`/access checks read the complete `this.sources` set, so gates aren't weakened by curation. 👍

A few non-blocking notes, then a bigger naming + semantics point.

**Non-blocking**
1. The **unconditional `export {}` curation** is the one behavior change that hits packages that never opted in — worth confirming no in-repo example/package relies on `export {}` still listing everything.
2. `getModel()` now curates, but the raw-notebook serialization still reads `this.sources`/`this.queries` directly. Probably fine (notebooks are always public), but a one-liner on why notebook-embedded sources don't curate while model ones do would help.
3. Rollback-by-delete on a bad publish runs before `addPackageToDatabase`, so it's removing a not-yet-persisted package — works, just worth a comment to that effect.

**Naming + semantics (the main thing)**
I don't love `entryPoints` — it's a bundler/program-`main()` metaphor, and it names the wrong axis (files), when the concept we actually care about is *the sources a consumer queries directly* vs. the internal sources they compose.

Two asks:

1. **Rename the concept to `explores`.** It's the established term (Looker) for exactly this: a source exposed as a directly-queryable starting point, as opposed to building-block views. Reads naturally in `publisher.json` and in chat/discovery. (Yes, Malloy historically used `explore` as a keyword — now `source` — but at the manifest layer the collision is mild and we're reclaiming the familiar concept.)

2. **Make it a true queryability boundary, not just a discovery filter.** Today hiding is discovery-only — an unlisted source stays fully queryable by name, with `#(authorize)` offered as the mitigation. I think the cleaner model, and what "explore" implies, is: non-explore sources are building blocks — importable/joinable/extendable, but **not valid top-level query targets**. That closes the "queryable by anyone who knows the name" gap directly rather than asking authors to bolt on a second gate.

If we go that way, a few things to scope:
- It becomes access control, so it needs **query-time enforcement** (reject a run/ad-hoc/named query whose target source isn't an explore), parallel to the authorize gate — not just list filtering.
- Define how it **composes with `#(authorize)`**: explore-membership decides *what* is directly queryable; authorize decides *who*; both must pass.
- **Compat sharp edge:** turning restriction into an *access* change means the unconditional `export {}` curation would silently make existing users' internal sources non-queryable. I'd keep the queryability boundary opt-in (tied to an explicit `explores` declaration) and/or keep `export {}` discovery-only.
- **Granularity:** an explore is a *source*, but `explores` currently lists *files*. Worth deciding whether `explores` should enumerate sources (truest to the concept, and it would subsume the `export {}` axis) or stay file-level (smaller change, but the name then implies sources it doesn't directly name).

None of this takes away from the implementation quality — it's mostly about landing on the right concept/word before this becomes the public manifest API and harder to change.

kylenesbit (approved):
Comment is mostly food for thought.  I am fine with this PR as is.

Sha-Bang:
Closing in favor of a fresh PR from the same branch. GitHub stopped syncing this PR's head to the branch ref: it stayed pinned at `fea90809` while the branch advanced to `57589657`, so CI never ran on the current code. Three real pushes plus a close/reopen didn't recover the head pointer (the branch ref itself is correct — `git ls-remote` and the commit object both confirm `57589657`). Reopening as a new PR so the head resolves and checks run; same branch, same commit. Discussion and approval here remain the record.

Sha-Bang:
Continued in #833 (same branch/commit `57589657`; this PR's head pointer was stuck).

PR #824, opened 2026-06-08
fix(sdk): givens UI correctness (mid-run, empty/boolean overrides, date dedup) + e2e authorize coverage

Sha-Bang:
Started as end-to-end validation for the `#(authorize)` series, then picked up a cluster of givens UI fixes that review surfaced along the way.

## Authorize coverage
- Playwright spec writes a gated model + notebook into the `faa` package and proves the gate at the notebook-cell HTTP surface (403 naming the source, redacted; 200 with the satisfying given) and through the UI (result hidden until `role` is supplied, then renders).
- Integration block proves a given can be both a runtime query parameter and the subject of an `#(authorize)` gate on the same source: gate fires before filtering; once it passes, the parameterized filter applies.

## Givens UI fixes
- **Apply changes made mid-execution.** The re-execution effect recorded the latest inputs before the `isExecuting` guard, then skipped the run — so a given/filter edited while cells were running was silently dropped. Now the change isn't consumed until it can be acted on; the effect re-runs when execution finishes with the latest inputs. (Surfaced by Codex review.)
- **Empty-string and boolean overrides.** Givens used absence/`null` to mean "use the model default," and the inputs collapsed an empty string and an untoggled boolean into that same state — so a deliberate `""` (matters for gates like `$region != ''`) and an explicit boolean couldn't be expressed. `GivenInput` now models three states: unset (omitted, server default), explicit override (any concrete value incl. `""`/`false`, sent verbatim), and revert (the × drops the override). The boolean reflects the model default when unset and gains a revert affordance.
- **Date dedup key.** The change-detection effect serialized dates as full ISO while the wire sends `YYYY-MM-DD`, so two times on the same day triggered a redundant re-run. The dedup key now uses the same granularity as the payload.
- **Remove dead `clearGiven`** from `useGivensForm` (exported, never wired).

### Tested
- `tsc` + `eslint` clean (server, sdk, app)
- `authorize_integration` 40/40 (incl. the compose block)
- Playwright `notebook-authorize` 4/4 and `notebook-givens` 9/9 against a local stack on malloy 0.0.405 (new tests cover the mid-execution re-run, empty-string override, and boolean default/toggle/revert; the mid-execution test is confirmed to fail without its fix)
- Drove the authorize flow in a real browser: no result with `role` unset, `WN` / Southwest Airlines once `role=analyst` is set

### DIFF
--- a/packages/sdk/src/components/Notebook/Notebook.tsx
+++ b/packages/sdk/src/components/Notebook/Notebook.tsx
@@ -443,7 +443,11 @@ export default function Notebook({
    const prevInputsRef = useRef<string>("");
 
    useEffect(() => {
-      // Serialize activeFilters + givenValues to detect actual value changes
+      // Serialize activeFilters + givenValues to detect actual value changes.
+      // Encode dates to the same YYYY-MM-DD granularity buildGivens sends on the
+      // wire — otherwise JSON.stringify renders Date as a full ISO timestamp, so
+      // two times on the same day would look like a change and trigger a
+      // redundant re-run even though the sent value is identical.
       const serialized = JSON.stringify({
          filters: activeFilters.map((f) => ({
             dim: f.dimensionName,
@@ -451,9 +455,12 @@ export default function Notebook({
             val: f.value,
             val2: f.value2,
          })),
-         givens: Array.from(givenValues.entries()).sort(([a], [b]) =>
-            a.localeCompare(b),
-         ),
+         givens: Array.from(givenValues.entries())
+            .sort(([a], [b]) => a.localeCompare(b))
+            .map(([name, v]) => [
+               name,
+               v instanceof Date ? v.toISOString().slice(0, 10) : v,
+            ]),
       });
 
       // Skip if no actual change
--- a/packages/sdk/src/components/Notebook/Notebook.tsx
+++ b/packages/sdk/src/components/Notebook/Notebook.tsx
@@ -471,12 +471,18 @@ export default function Notebook({
          return;
       }
 
-      prevInputsRef.current = serialized;
-
-      // Re-execute with current filters and givens (or empty if cleared)
-      if (!isExecuting) {
-         executeCells(activeFilters, givenValues);
+      // Don't consume the change until we can act on it. If a run is already in
+      // flight, leave prevInputsRef stale and bail: this effect re-runs when
+      // isExecuting flips back to false (it's a dependency), and will then pick
+      // up the latest inputs. Updating prevInputsRef here instead would mark the
+      // change "seen" and drop it — a given/filter edited mid-run would never
+      // re-execute.
+      if (isExecuting) {
+         return;
       }
+
+      prevInputsRef.current = serialized;
+      executeCells(activeFilters, givenValues);
    }, [activeFilters, givenValues, isExecuting, executeCells]);
 
    // Handle filter change using composite key
--- a/packages/app/tests/playwright/notebook-authorize.spec.ts
+++ b/packages/app/tests/playwright/notebook-authorize.spec.ts
@@ -107,14 +107,27 @@ test.describe("notebook-authorize", () => {
    }
 
    test("UI: result is gated until the given is supplied", async ({ page }) => {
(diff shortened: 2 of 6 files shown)


packages/sdk/src/components/Notebook/Notebook.tsx
```diff
@@ -471,12 +471,18 @@ export default function Notebook({
...
-      // Re-execute with current filters and givens (or empty if cleared)
-      if (!isExecuting) {
-         executeCells(activeFilters, givenValues);
+      // Don't consume the change until we can act on it. If a run is already in
+      // flight, leave prevInputsRef stale and bail: this effect re-runs when
+      // isExecuting flips back to false (it's a dependency), and will then pick
+      // up the latest inputs. Updating prevInputsRef here instead would mark the
+      // change "seen" and drop it — a given/filter edited mid-run would never
+      // re-execute.
+      if (isExecuting) {
```
sagarswamirao, line 487: Nice fix. One thing I wanted to confirm: if inputs change twice mid-run (A then B), only the final state B gets serialized and re-executed once `isExecuting` flips false -- A is coalesced away. I'm reading that as intentional (you want the latest, which matches the "executes with the latest inputs" line in the description), just wanted to double-check that's the desired behavior and not something to queue up.

sagarswamirao:
Not related to this PR's changes, but while reading through the givens flow a few other potential rough edges stood out -- leaving them here for visibility, might be worth a look at some point:

1. **Can't override a non-empty default back to empty string.** In `GivenInput`, the string/number inputs map `""` -> `null`, and `null` means "revert to model default" in `useGivensForm`. So for a given like `given: region :: string is 'us-west'`, clearing the field reverts to `'us-west'` rather than `''` -- there's no way to express a deliberate empty override when a default exists. Matters for gates like `#(authorize) "$region != ''"`. (Symmetric to the empty-default *display* handling that's already accounted for.)

2. **Date dedup key vs wire encoding mismatch.** The re-exec change-detection effect serializes givens via `JSON.stringify`, which renders `Date` as a full ISO timestamp, while `buildGivens` truncates dates to `YYYY-MM-DD` on the wire. Two times on the same calendar day produce different dedup keys but identical sent values, so a date-given can trigger a redundant re-run of all cells.

3. **Boolean givens have no unset state.** `checked = value === true`, so a `boolean is true` given can't distinguish "untouched, use default" from an explicit `false`, and can't be reverted to default once toggled. (Already acknowledged in a code comment -- noting it as part of the same "null conflates two meanings" theme.)

4. **`clearGiven` is dead code.** `useGivensForm` exports `clearGiven`, but the panel only wires up `updateGiven`/`clearAll` (per-given clear goes through the `onChange(null)` adornment instead). Minor cleanup.

None of these are regressions from this PR; just adjacent givens behavior worth tracking.

packages/sdk/src/components/Notebook/Notebook.tsx
Sha-Bang: Yes, intentional — last-write-wins. When `isExecuting` flips false the effect re-runs with the latest committed inputs, so only the final parameter set executes. Queuing each intermediate edit would re-run all cells once per mid-run change with no user benefit (they only care about the final values), and it is already strictly better than the prior behavior, which dropped the mid-run change entirely.

sagarswamirao (approved):

Sha-Bang:
@sagarswamirao thanks — all four are handled in this PR.

- **#2 (date dedup key)** — fixed; the change-detection key now encodes dates as `YYYY-MM-DD`, matching what `buildGivens` sends, so two times on the same day no longer trigger a redundant re-run.
- **#4 (dead `clearGiven`)** — removed from `useGivensForm`.
- **#1 (empty-string override) and #3 (boolean unset)** — both fixed too. `GivenInput` now models three distinct states: *unset* (omitted from the request → server default), *explicit override* (any concrete value including `""` and `false`, sent verbatim), and *revert* (the × drops the override). The boolean reflects the model default when unset and gains a revert affordance. So a deliberate `""` (for gates like `$region != ''`) and an explicit boolean are both expressible now.

All covered by new `notebook-givens` tests (empty-string override cycle + boolean default/toggle/revert).

PR #825, opened 2026-06-10
feat(server): enforce explores as a query boundary (queryableSources)

jswir:
> **Stacked on #823** (the `entryPoints` → `explores` rename). The first commits here are #823's; **review only the last commit** (`feat(server): enforce explores as a query boundary`). Will rebase once #823 merges, which collapses the diff to just the boundary work.

## What

Follows up on the review discussion in #823: promotes the `explores`/`export {}` **discovery surface** into an optional **query boundary**, via a new package-level `queryableSources` field in `publisher.json`.

- **`"declared"` (default)** — *queryable == discoverable*. When `explores` is declared, only `explores` files — and within them only the `export {}` closure — are valid top-level query targets. Every other source still **compiles, imports, joins, and extends**; it just isn't a valid `run:` target. Denials are a generic **404** (`NotQueryableError`), deliberately indistinguishable from a non-existent target — no enumeration/existence oracle.
- **`"all"`** — decouples the axes: `explores`/`export {}` gate discovery only; every compiled source stays directly queryable (the behavior in #823).
- **No `explores` declared ⇒ inert in both modes** — there's no curated surface to enforce, so existing packages are unaffected even under the default. That's what makes the safe default non-breaking.

This is the **what** axis (identity-free), orthogonal to and composed with `#(authorize)` (the **who** axis): the boundary runs first (404), then authorize (403) — both must pass.

## How enforcement works (two steps, mirroring the authorize gate)

A deliberate design constraint: Publisher must **never alter or restrict Malloy compilation**. Reading the *output* of compilation is fine — and that's exactly what the authorize gate already does.

1. **Early pre-compilation gate** — positively denies explicit names (`sourceName`/`queryName`) and surface-resolved ad-hoc targets outside the curated surface, *before* `loadQuery`, so a denied target's compile errors can't be used as a schema oracle. Anything it can't pin is **deferred**, not denied.
2. **Compiled backstop** — settles deferred admissions against the run target Malloy *actually executes*, read off the prepared query's `structRef` (the same `resolveAuthorizeSourceFromRunnable` the authorize gate uses). Survives multi-statement forms (the LAST `run:` wins — a curated decoy can't smuggle a hidden target) and named-query indirection. **Unresolvable targets fail closed.**

Notable semantics:
- **Ad-hoc derivation over a curated source is itself queryable**: `source: x is my_explore extend { … }` then `run: x` is admitted, via the same `source: NAME is BASE` alias walk filter inheritance already uses (extracted into a shared `buildAliasMap`). Laundering a *hidden* source through an alias is denied — the chain never reaches the curated surface.
- **An exported named query stays runnable by name even when the source it reads is hidden** — exporting it is the author's deliberate exposure. Explicit-name admissions skip the backstop so re-deriving the underlying source can't re-deny them. (Malloy's `export {}` accepts query names; pinned by test.)
- **The file-level gate is the robust line**: the model file is named by the request URL, so there is nothing to resolve and nothing to evade. Building blocks in their own non-`explores` files are hard-blocked.

## Where it's enforced

Single chokepoint per surface, placed exactly like the authorize gate (pre-compile + post-compile, outside the `loadQuery` try, independent of `bypassFilters`):

| Surface | Gate |
| --- | --- |
| `POST /…/query` | early + compiled backstop in `Model.getQueryResults` |
| MCP `malloy_executeQuery` | routes through `getQueryResults` — same gates |
| `POST /…/compile` | early text gate + materializer backstop (`assertQueryBoundaryForRunnable`) |
| Notebooks | exempt — always public, never gated |

The `Package` pushes the policy onto each `Model` (`applyQueryBoundaryToModels`) at load, reload, and metadata PATCH — no Model→Package back-reference.

## Plumbing

- `queryableSources` threaded through the worker protocol, pool outcome, worker manifest read (**invalid values fall back to `"declared"`**, fail-safe), DTO, `api-doc.yaml`, and the metadata-PATCH preserve/write path (preserved across a name/description-only PATCH, same as `explores`).
- New `NotQueryableError` → **404** with a redacted, generic message.

## Tests

`query_boundary.spec.ts` (worker-pool, end-to-end), 9 cases:
- **declared admits**: exported source (incl. join-through a hidden module), ad-hoc on a curated source, all-curated multi-statement, derivation over a curated source, exported-named-query-over-hidden-source.
- **declared denies**: non-`explores` file (named + file-level), non-exported source (named + ad-hoc), curated-decoy multi-statement (backstop catches the last statement), hidden-first multi-statement (early gate, pre-compile), hidden source laundered via ad-hoc derivation.
- **`"all"`**: listing still curated; hidden file/source stay queryable.
- **no `explores`**: boundary inert (backward compatible).

`tsc` + `eslint` clean; **879 server unit tests pass**.

## Out of scope (tracked separately)

- Identity-bound / reserved "system givens" for `#(authorize)` (the *who*-axis hardening; givens are caller-asserted today — see `docs/authorize.md`).
- Malloy-native enforcement of non-exported-source non-queryability (the long-term home for the within-file boundary).


Sha-Bang:
Folding this into #823. With discovery curation now gated on declaring `explores` (and the `PUBLISHER_LEGACY_DISCOVERY` flag dropped), the `queryableSources` boundary shares the same single opt-in and the same worker-serialization plumbing — so it's cleaner to land the discovery surface and its enforcement together than as a stacked pair. The query-boundary code, `NotQueryableError` → 404, and `query_boundary.spec` are all in #823 as of 18c48290. Closing; no work lost.

PR #828, opened 2026-06-11
feat(server): validate renderer tags at package load

mlennie:
Supersedes #827.

Renderer tags (e.g. `# big_value`, `# line_chart`) were only validated by the renderer at render time, so a misconfigured tag stayed invisible until a view was queried and could render as `[object Object]`. One case: a child-only `# big_value { sparkline=... }` placed on a view with no activating big_value, which the renderer declines to match, letting the result fall through to a `String()` coercion.

This validates renderer tags during the package-load compile pass. For each top-level query and each source view, the query is prepared (compile only, no execution), its stable result schema is taken, and the renderer's headless `validateRenderTags` runs against it. Any error-severity finding throws a `ModelCompilationError`, surfaced as the model's compilationError (HTTP 424), so a bad tag fails the load with a clear message instead of producing garbage at render time. Only errors are fatal; warnings stay on the query-time `renderLogs` surface so a benign render lint never blocks a load.

pr_before_after

## Why the main thread (vs #827)

#827 ran this inside the package-load worker, but importing `@malloydata/render-validator` there pulls the ~3MB solid-js renderer (which mutates DOM globals at import) into every worker isolate at spawn. That destabilizes the deliberately pure-CPU worker on Linux and turned the cross-platform unit job red on ubuntu (macOS and Windows passed). This version runs validation on the main thread, where the renderer is already used at query time (`query.controller` / `execute_query_tool`):

- `Model.validateRenderTags()` dynamically imports the validator, so the heavy module never loads for packages with no renderable views.
- It runs from `Package.loadViaWorker` after each model hydrates (the create path aborts the whole load on a bad tag) and from `Package.reloadAllModels` on its success branch, where a render-tag error is recorded as that model's compilationError rather than thrown, preserving reload's per-model-placeholder semantics.
- The worker stays pure-CPU and free of the renderer import.

The companion renderer fix (malloydata/malloy) makes the renderer surface this error instead of `[object Object]`, so the two together fail at publish time and fail gracefully at render time.

## Testing

`package_worker_path.spec.ts` covers the good/bad render-tag, notebook, and reload paths. Local gate green: typecheck, lint, prettier, unit (865 pass), integration (229 pass).

Includes Sagar Swami Rao Kulkarni's original package-load validation commits, which this builds on.


packages/server/src/service/model.ts
```diff
@@ -700,6 +700,83 @@ export class Model {
...
+               label: query.name,
+               queryString: `run: ${query.name}`,
+            });
+         }
+      }
+      for (const source of this.sources ?? []) {
+         for (const view of source.views ?? []) {
+            targets.push({
+               label: `${source.name} -> ${view.name}`,
+               queryString: `run: ${source.name} -> ${view.name}`,
```
sagarswamirao: Interpolating the raw identifiers here breaks for any name that needs backtick-quoting in Malloy: a hyphen, space, reserved word, or leading digit. For a source named `gated-source` (which this repo already exercises in `authorize_integration.spec.ts`), this builds `run: gated-source -> card`, which fails to lex; `loadQuery(...)` throws and is swallowed by the `catch { continue }` below, so the view is silently skipped and never render-validated -- the exact case this feature exists to catch.

Suggest backtick-quoting each interpolated identifier (a bare-valid name stays valid backticked, so it's strictly safer):

```ts
queryString: `run: \`${source.name}\` -> \`${view.name}\``,
```

and likewise `run: \`${query.name}\`` for the named-query target above. The sibling `extractSourceName` (model.ts:365-378) already handles backtick-quoted names for this reason.

packages/server/src/service/model.ts
```diff
@@ -700,6 +700,83 @@ export class Model {
...
+      const targets: { label: string; queryString: string }[] = [];
+      for (const query of this.queries ?? []) {
+         if (query.name) {
+            targets.push({
+               label: query.name,
+               queryString: `run: ${query.name}`,
+            });
+         }
+      }
+      for (const source of this.sources ?? []) {
```
sagarswamirao, line 765: This prepares (compile + `getPreparedResult`) every view and every named query in the package, even ones with no render tags at all. The annotation data needed to skip them is already in hand: each view carries its own `annotations` (`ApiSource.views[].annotations`, from `source_extraction.ts:194`), as does each query.

Suggest gating on the target's own annotations before pushing/preparing:

```ts
for (const source of this.sources ?? []) {
   for (const view of source.views ?? []) {
      if (!view.annotations?.length) continue;
      targets.push({ ... });
   }
}
```

and `if (!query.name || !query.annotations?.length) continue;` for the query loop.

I traced the renderer's tag resolution (`@malloydata/render`): tags are read per-field via `tagFromAnnotations(field.annotations, '')` (`data_tree/fields/base.ts:111`) plus a file-level `modelTag` (`render-field-metadata.ts:55`) -- there is no source-to-view tag inheritance. So the gate belongs on the view's (and query's) own annotations, not the source's: a source-level tag is not a renderable target, and source annotations don't flow down to a view's render decision. This also shrinks the blast radius of the unquoted-identifier issue, since untagged views are no longer prepared at all.

mlennie:
Pushed `b1b326d4`, which addresses both review comments plus one extra hardening I caught on a self-review.

Backtick-quoting (comment 1): agreed, the unquoted identifiers silently disable the feature for any name that needs Malloy quoting, and we already ship `gated-source` in `authorize_integration.spec.ts`. Each interpolated identifier now goes through a `quoteMalloyIdentifier` helper. I also made it escape-safe while I was in there: it escapes backslashes then backticks, mirroring malloy's internal `identifierCode`/`escapeIdentifier`, so a name containing a literal backtick can't break out of the quotes either. Added two regression tests (a hyphenated source name and an escaped-backtick source name), each of which would be silently skipped without the quoting, and verified both fail without the fix.

Annotation gate (comment 2): also applied. Each loop now skips a view or query that has no annotations of its own before compiling. Agreed it belongs on the target's own annotations: tags resolve per-field from `field.annotations` plus the file-level `modelTag` with no source-to-view inheritance, so a source-level tag isn't a renderable target. The one case it doesn't reach is a model-level `##` tag (theme/config, not the child-only chart tags this guards), which I noted in a code comment. Nice side effect that it also shrinks the quoting issue's blast radius, since untagged views are never prepared now.

Full root gate is green: typecheck, lint, prettier, unit 867 and integration 229, 0 fail. Thanks for the careful review.

sagarswamirao (approved):

PR #831, opened 2026-06-14
feat(server): support publisher proxy connections

jswir:
## Problem

The Malloy CLI and both VS Code extensions can run queries against a remote Publisher dataplane (e.g. Credible) via a `publisher`-type connection. The standalone Publisher server (`malloy-publisher --watch-env <env>`) could not — it only knew the 9 native warehouse types, so a `publisher` connection hit `Unsupported connection type`.

This blocks the local data-app authoring loop: run a local Publisher to serve a package's `public/` app with watch/live-reload, while queries proxy to the user's real Credible connection — no need to replicate warehouse credentials locally.

## Change

No new connection implementation needed — `@malloydata/db-publisher`'s `PublisherConnection` already POSTs SQL to `connectionUri` with `Authorization: Bearer <token>` (the same class the CLI/extensions use). This PR makes the server depend on it, register it, and let it through the config assembler.

- **Register the type** — add `@malloydata/db-publisher` dependency + side-effect import in `connection.ts` so `registerConnectionType('publisher', …)` runs at startup.
- **Config assembler** (`connection_config.ts`) — validate (`publisher` branch with an actionable missing-`connectionUri` error) and assemble to `{ is: "publisher", connectionUri, accessToken }`.
- **SSRF gate (default-deny / fail-closed)** (`connection_config.ts`) — a `publisher` connection makes the *server* issue outbound HTTP to a tenant-controlled `connectionUri`, which is an SSRF surface in a hosted multi-tenant deployment. `validateConnectionShape` now refuses the type unless the operator sets `PUBLISHER_ALLOW_PROXY_CONNECTIONS=true`. Because every connection passes through `validateConnectionShape` at config-load, this one choke point shuts off config-load, query, *and* introspection at once. The signal is an operator/deployment env var rather than a `publisher.config.json` key, so the same untrusted config that supplies `connectionUri` cannot also enable the type. (Resolves the open question previously flagged in this PR.)
- **OpenAPI** (`api-doc.yaml`) — add `publisher` to the type enum and a `PublisherConnection` schema (`connectionUri` required, `accessToken` optional). `api.ts` is generated at build (`generate-api-types`), so it is not committed.
- **Schema/table introspection** (`db_utils.ts`) — proxy `getSchemasForConnection` / `listTablesForSchema` to the remote dataplane's identical `/schemas` and `/tables` endpoints, so the Connection Explorer works instead of throwing `Unsupported connection type: publisher`. The proxy applies a request timeout, redacts any URL-embedded credentials from error messages, and matches the table-name filter across schema- and catalog-qualified `resource` shapes.
- **SDK** (`Connections/common.ts`, `Add`/`EditConnectionDialog.tsx`) — `publisher` is a valid `ConnectionTypeEnum` but is configured in `publisher.config.json`, not created through the UI, so it is excluded from the connection-form dropdown (`uiCreatableConnectionTypes`). The exhaustive `Record<ConnectionTypeEnum, …>` maps gain `publisher` entries so the SDK type-checks.
- **Docs** — document the `publisher` connection block, the `PUBLISHER_ALLOW_PROXY_CONNECTIONS` gate + SSRF rationale, and the token-staleness limitation in `docs/connections.md`.

Usage (in `publisher.config.json`, with the server started as `PUBLISHER_ALLOW_PROXY_CONNECTIONS=true malloy-publisher --watch-env <env>`):
```json
{
  "name": "analytics",
  "type": "publisher",
  "publisherConnection": {
    "connectionUri": "https://org.data.example.com/api/v0/environments/<env>/connections/<name>",
    "accessToken": "<jwt>"
  }
}
```

`readOnly` is intentionally omitted — the proxied dataplane enforces auth and read-only; `PublisherConnection` does not reject writes client-side.

## Testing

- **Unit — config** (`connection_config.spec.ts`): valid assemble shape, optional `accessToken`, missing-`connectionUri` error, and reserved-name / no-name validations still fire for a publisher connection.
- **Unit — SSRF gate** (`connection_config.spec.ts`): default-deny when the flag is unset, deny for any non-`"true"` value (fail-closed), the error names `PUBLISHER_ALLOW_PROXY_CONNECTIONS`, the gate fires *before* the `connectionUri` shape check, and the enabled path (`"true"`) assembles. The existing assembly tests opt in via `beforeEach` and restore the prior value in `afterEach`.
- **Unit — introspection proxy** (`db_utils.spec.ts`): the request URL and bearer header sent to the dataplane, the bare-table-name filter across schema-/catalog-qualified resources, the non-ok error surface, credential redaction from error messages, and network-failure handling — exercised with a stubbed `fetch`. The remote dataplane remains the oracle for introspection *correctness* and is not simulated.
- **Manual end-to-end against real Credible staging** (the real oracle): a local `--watch-env` server (with `PUBLISHER_ALLOW_PROXY_CONNECTIONS=true`) and a `bq_demo` publisher connection executed raw SQL, ran a Malloy model query (`run: users -> { aggregate: count() }` → 100k rows), and browsed schemas/tables through the Connection Explorer — all proxied to BigQuery via Credible and returning real rows. Also verified the gate denies at config-load when the flag is unset (the whole environment is skipped with the actionable error), and that an in-package `public/index.html` data app serves and live-reloads via the watch-mode symlink.

## Out of scope / known limitations

- Reading `malloy-config.json` directly (the CLI/extension shape) — deferred; authors add the block to `publisher.config.json` for now. Note: the server reads connections from its persisted store on normal boot, so `--init` is required to pick up new config-file connections.
- Token refresh/expiry — the JWT is user-scoped and short-lived; the server uses it as configured and does not refresh, so a long `--watch` session can outlive the token (the CLI/extension own refresh today).
- Creating publisher connections through the SDK UI — config-file only for now (see SDK note above).
- Editing an existing `publisher` connection through the SDK Edit dialog — the type is excluded from the dropdown (correct for *Add*); a `publisher` row opened in *Edit* has no matching menu item. Config-file-only today, so low-impact, but a follow-up should hide/disable Edit for `publisher` rows.
- Fail-closed granularity — a disallowed `publisher` connection fails the *entire* environment load (consistent with every other connection-validation error), rather than being skipped while the rest of the environment loads.


kylenesbit (approved):
Approving. Clean, minimal change that reuses the existing `PublisherConnection` class — the right call. Good hardening (request timeout, URL-credential redaction, schema-/catalog-qualified table filtering) and solid unit coverage of the proxy's own logic, with the remote dataplane left as the oracle for introspection correctness plus manual e2e against real Credible staging. Verified the two integration points are clean: `getStaticConnectionAttributes` returns undefined via its default, and `lookupConnection` falls through to the registry-resolved `base.lookupConnection` for `publisher`.

One organizational ask (not code, already flagged as an open question in the PR): a `publisher` connection makes the server issue outbound HTTP to a tenant-controlled `connectionUri`, which is an SSRF surface in a hosted multi-tenant deployment. Fine for local `--watch-env` authoring and opt-in/config-file-only today, so deferring the gate is reasonable — but please ensure the `validateConnectionShape` gate lands (ideally default-deny / fail-closed) before this type is accepted in any hosted Credible deployment.

jswir:
Added the SSRF gate that was flagged as the one pre-ship ask (commit 4a3a38c2).

**What landed:**
- `validateConnectionShape` now refuses a `publisher` connection unless `PUBLISHER_ALLOW_PROXY_CONNECTIONS=true` — **default-deny / fail-closed**, as recommended. Since every connection passes through `validateConnectionShape` at config-load, this single choke point shuts off config-load, query, *and* introspection at once.
- The gate is an **operator/deployment env var**, not a `publisher.config.json` key, so the same untrusted config that supplies `connectionUri` cannot also enable the type. A hosted multi-tenant deployment (e.g. Credible) gets the type denied by default; local `--watch-env` authoring opts in explicitly.
- The denial error is actionable and names the exact var: `Publisher proxy connection '<name>' is disabled in this deployment. … Fix: set the environment variable PUBLISHER_ALLOW_PROXY_CONNECTIONS=true to enable them.`
- Tests cover default-deny when unset, non-`"true"` values (fail-closed), the gate firing *before* the `connectionUri` shape check, and the enabled path. Docs (`docs/connections.md`) note the flag and the SSRF rationale.
- Verified live: with the flag unset, a `bq_demo` publisher connection is refused at config-load and the environment is skipped; with `PUBLISHER_ALLOW_PROXY_CONNECTIONS=true` it loads and proxies queries to Credible.

The PR description's "Open questions" section has been updated accordingly. Two minor follow-ups are now tracked under "Out of scope": hiding `publisher` rows from the SDK *Edit* dialog, and the fail-closed granularity (one bad `publisher` connection fails the whole environment, consistent with all other connection-validation errors).

packages/server/src/service/connection_config.ts
```diff
@@ -336,6 +336,36 @@ function validateConnectionShape(connection: ApiConnection): void {
...
+         if (process.env.PUBLISHER_ALLOW_PROXY_CONNECTIONS !== "true") {
+            throw new Error(
+               `Publisher proxy connection '${connection.name}' is disabled in this deployment. ` +
+                  `'publisher' connections make the server issue outbound requests to a configured connectionUri, ` +
+                  `which is only appropriate for local --watch-env authoring. ` +
+                  `Fix: set the environment variable PUBLISHER_ALLOW_PROXY_CONNECTIONS=true to enable them.`,
+            );
+         }
+         const publisher = connection.publisherConnection;
+         if (!publisher?.connectionUri) {
```
jedibrillo, line 360: publisher should check for malformed URLs here, at validation time, rather than runtime (which can potentially leak raw user info in thrown errors)

jswir: Fixed in 652aac9 — but to be precise about what this does and doesn't do:

**It's input validation, not an SSRF control.** `validateConnectionShape` now parses `connectionUri` with `new URL()` at config-load and requires an `http:`/`https:` scheme. That gives a fast, actionable error on a malformed URI at config-load instead of an opaque `fetch` failure deep in the request path, and it keeps the raw `connectionUri` out of the thrown message (only the connection name and parsed scheme are interpolated) — closing the credential-echo path you flagged, since `redactUrlCredentials` returns the URI unchanged when it can't parse it.

**It does not mitigate SSRF.** It only checks the scheme, not the host, so internal targets still pass validation — e.g. `http://169.254.169.254/latest/meta-data/`, `http://localhost:8080/...`, `http://10.0.0.5/...`. The SSRF control for the planned implementation is the off-by-default, fail-closed `PUBLISHER_ALLOW_PROXY_CONNECTIONS` gate (and the fact that hosted Credible never enables it); this URL check isn't load-bearing there.

**Validation contract — accepted:** parses as an absolute URL *and* scheme is `http`/`https`.

Valid:
- `https://org.data.example.com/api/v0/environments/prod/connections/bq_demo`
- `http://localhost:4000/api/v0/environments/dev/connections/bq_demo` (http allowed for local dev)
- `https://user:token@org.data.example.com/connections/bq_demo` (userinfo parses fine)

Invalid:
- `connections/bq_demo` / `org.data.example.com/connections/x` — no scheme, `new URL` throws → "not a valid URL"
- `not a url` / `https://bad host/x` — unparseable → "not a valid URL"
- `file:///etc/passwd`, `ftp://host/x`, `gopher://host/` — parses but non-http(s) scheme → "must use http or https (got 'file:')"

kylenesbit (approved):

PR #833, opened 2026-06-17
feat(server): explore visibility + query boundary via publisher.json explores

Sha-Bang:
> **Supersedes #823.** GitHub stopped syncing that PR's head to the branch — it stayed pinned at the old `fea90809` commit while the branch advanced to `57589657`, so no CI would run. Three real pushes and a close/reopen didn't recover it. This is a fresh PR from the same branch (`jswir/entry-point-visibility`) so the head resolves correctly and checks run. Same commit, same content; #823's discussion and Kyle's approval live there.

## What

An opt-in `explores` field in `publisher.json` designating which `.malloy` files
form a package's public discovery surface, plus an optional `queryableSources`
boundary that turns that surface into an enforced one. Everything is gated on
declaring `explores`: **with no `explores`, discovery and querying behave exactly
as they do today** — all models listed, full source set, nothing blocked. Opt-in
and backward compatible by default.

This lets customers build layered models (base/bridge sources in their own files)
without those internals showing up in chat/discovery, and optionally stops them
being queried directly.

### Discovery surface — gated on `explores`

Two granularities, both active **only when `explores` is declared**:

- **File level — `explores`**: `listModels()` returns only the listed files;
  other `.malloy` files still compile and stay joinable but are hidden. Absent/
  empty → every file listed (today's behavior). Notebooks always listed.
- **Within a file — `export {}`**: accessors list only `modelDef.exports`,
  matching Malloy's export-aware `modelInfo`/`sourceInfos`.

The complete source/query set is retained internally, so `#(filter)`/`#(authorize)`
and join/extend resolution are unaffected. Discovery hiding is not access control
on its own.

### Query boundary — `queryableSources` (`declared` | `all`, default `declared`)

Once `explores` is declared, the surface can also bound what's directly queryable:
- `declared` (default): queryable == discoverable. Hidden files/sources → 404
  (`NotQueryableError`), indistinguishable from non-existent.
- `all`: discovery curated, everything stays queryable by name (softer migration).

With **no `explores`, the boundary is inert** in both modes. Enforcement mirrors
the authorize gate (early pre-compile gate + compiled backstop). The early gate
only takes the exported-query fast-path for a *pure* named query (no source
prefix); a source-prefixed request is gated on the named source's curation, so a
hidden source can't be reached via a view whose name collides with an exported
top-level query.

## Rollout — opt-in, no flag

The earlier revision shipped `export {}` curation unconditionally plus a temporary
`PUBLISHER_LEGACY_DISCOVERY` env flag. That's been **dropped** in favor of gating
all curation on `explores`: a flag only protects non-adopters while every
deployment remembers to set it, which doesn't hold for packages we don't own or
control. Gating on `explores` makes "no `explores` = today's behavior" true
structurally.

## Non-destructive validation rollback

Strict-at-publish `explores` validation rejects a bad manifest with a `400`
without destroying content:

- **No-location create** registers a *pre-existing* user directory, so a bad
  manifest unloads it from memory (`unloadPackage`) instead of `deletePackage`
  (which would delete the user's files).
- **Location install/update** validates the *effective* explores INSIDE
  `installPackage`'s swap window, so a rejected update rolls back to the previous
  tree instead of swapping the bad one in and 400-ing after. Reload passes no
  validator and stays fail-safe.

## Tested

- Server unit suite: **899 pass / 0 fail**; typecheck + lint clean.
- `explore_visibility.spec` / `query_boundary.spec`: legacy behavior without
  `explores`; file + `export {}` curation; `declared` vs `all`; multi-statement
  settled by the compiled run target; ad-hoc-derivation laundering denied; the
  collision regression (hidden source not reachable via a view name colliding
  with an exported query); backtick-quoted exported source still queryable.
- `package_rollback.spec` (real `Environment` + temp dirs): `unloadPackage` keeps
  the on-disk dir; a rejected location reinstall restores the previous tree; a
  valid reinstall still swaps.
- `package.controller.spec`: no-location invalid → `unloadPackage` not
  `deletePackage`; location paths pass a validator; update validates effective
  explores. `errors.spec`: `NotQueryableError` → 404. Contract test pins Malloy's
  `ModelDef.exports`.

## Docs

- README "Controlling the discovery surface": `explores` + `export {}`,
  `queryableSources`, discovery-vs-access-control, publish-vs-load policy.
- `api-doc.yaml`: `explores`/`queryableSources` documented; `exploresWarnings`
  `readOnly`.

## Downstream

- Consumers use `explores` / `exploresWarnings` / `queryableSources`.
- Retiring Credible's `.malloy_hidden` / `#(agent-hidden)` lives in separate
  Credible repos; this PR is the upstream source of truth.

### DIFF
--- a/README.md
+++ b/README.md
@@ -328,6 +328,8 @@ The two compose: `explores` decides which files are listed, and `export { … }`
   For gradual migration, use `explores` with `queryableSources: "all"` to curate listings while keeping every source queryable by name; switch to `"declared"` when ready to enforce the boundary.
 
 > **`explores`/`export {}` are a discovery filter; `queryableSources` decides if they also gate queries; `#(authorize)` is the identity gate.** With `queryableSources: "all"`, hiding a source only removes it from listings — it stays queryable by name. To restrict *who* can query (as opposed to *what* is queryable), gate the source with `#(authorize)` (see [docs/authorize.md](docs/authorize.md)); those gates are enforced against the complete source set and are never weakened by listing or boundary curation.
+>
+> The `queryableSources` boundary applies to the *query* surface (`getQueryResults`, the MCP query tool, and `/compile`). It does **not** cover raw retrieval by exact path — a hidden model's file text and its compiled metadata are still fetchable by path — by design; use `#(authorize)` when the contents themselves must be protected, not just removed from discovery.
 
 Validation is asymmetric by design: **publishing** a package with an `explores` entry that doesn't resolve to a real model is rejected with a `400`, while at **startup/reload** the package still serves but hides the unresolved entry (it never falls back to listing everything) and surfaces the reason in the package's `exploresWarnings` field.
 
--- a/README.md
+++ b/README.md
@@ -299,6 +299,38 @@ A package can ship a `public/` directory of web files (an `index.html` plus CSS,
 
 For local development, start the server with `--watch-env <env>` (or `PUBLISHER_WATCH=<env>`). Publisher then mounts that environment's local-dir packages in place and watches them: editing a `.malloy` recompiles the package, editing an asset refreshes the page, and open pages live-reload over a server-sent-events stream at `GET .../packages/<pkg>/events`. See `examples/html-data-app/` for a worked example, and [docs/html-data-apps.md](docs/html-data-apps.md) for the full authoring reference (the `Publisher.query` / `Publisher.embed` API, the page and event contracts, and the security model).
 
+## Controlling the discovery surface
+
+Declaring `explores` in `publisher.json` is the **single opt-in** for curated discovery. When absent or empty, every model is listed with its full source set — today's backward-compatible behavior.
+
+A package's manifest can scope which models and sources appear in listings (the surface that drives discovery and chat), at two granularities that **both apply only after `explores` is declared**:
+
+- **File level — `explores`.** An optional `string[]` of `.malloy` file paths (relative to the package root) that form the package's public surface. When present, only those models are returned by `listModels()`; every other `.malloy` file still compiles for import/join resolution and stays queryable, but is hidden from listings. When absent or empty, every model is listed. Notebooks are always listed regardless of this field (they can't be imported, so they have nothing to hide behind).
+
+  ```json
+  {
+    "name": "sales",
+    "description": "Sales models",
+    "explores": ["index.malloy"]
+  }
+  ```
+
+- **Within a file — `export { … }`.** Once `explores` is declared, the discovery accessors list only the model's re-export closure (`modelDef.exports`), matching what Malloy's `modelInfo`/`sourceInfos` expose. A model with no `export { … }` exports all of its locally-declared top-level sources; declaring `export { customers }` lists only `customers` and keeps imported/internal helpers out.
+
+The two compose: `explores` decides which files are listed, and `export { … }` decides which sources within a listed file are shown.
+
+- **Query boundary — `queryableSources`.** Controls whether that discovery surface is *also* a query boundary. `"declared"` (the default) makes **queryable == discoverable**: when `explores` is declared, only `explores` files — and within them only the `export {}` closure — are valid top-level query targets; every other source still compiles, imports, joins, and extends, but a direct query against it is denied with a `404` (indistinguishable from a non-existent target). `"all"` decouples the axes — `explores`/`export {}` gate discovery only and every compiled source stays directly queryable. When `explores` is absent there is no curated surface, so both modes are equivalent (everything queryable).
+
+  ```json
+  { "name": "sales", "explores": ["index.malloy"], "queryableSources": "all" }
+  ```
+
+  For gradual migration, use `explores` with `queryableSources: "all"` to curate listings while keeping every source queryable by name; switch to `"declared"` when ready to enforce the boundary.
+
+> **`explores`/`export {}` are a discovery filter; `queryableSources` decides if they also gate queries; `#(authorize)` is the identity gate.** With `queryableSources: "all"`, hiding a source only removes it from listings — it stays queryable by name. To restrict *who* can query (as opposed to *what* is queryable), gate the source with `#(authorize)` (see [docs/authorize.md](docs/authorize.md)); those gates are enforced against the complete source set and are never weakened by listing or boundary curation.
+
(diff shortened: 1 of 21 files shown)


jswir:
## Peer review

Well-designed, heavily-tested change. The query boundary itself is sound — verified at the PR head it's an allowlist (deny-by-default), fails closed on invalid config/unresolved targets, uses a two-gate (pre-compile + compiled backstop) structure mirroring the authorize gate, and its tests run against the real Malloy compiler. One issue worth fixing before merge, rest are nits.

### Important

**1. API-body `explores` aren't normalized → misleading 400s + publish-vs-reload divergence**
`packages/server/src/service/package.ts:399-422` (`getInvalidExplores`), `controller/package.controller.ts:234`

The worker normalizes on-disk `explores` (`package_load_worker.ts:396` → `normalizeModelPath`, strips `./`, forward-slashes), and the comment at `package_load_worker.ts:421-424` says `getInvalidExplores()`/`listModels()` compare against those normalized keys. But `getInvalidExplores` compares the raw `body.explores` against normalized `this.models.keys()` with no normalization. So:
- `explores: ["./index.malloy"]` (or backslash paths) is rejected at publish with a misleading `file not found in the package` 400, even though `index.malloy` exists.
- `writePackageManifest` writes the raw value; on reload the worker normalizes it and it resolves — same manifest behaves differently at publish vs. after restart.

Fails *closed* (unmatched entry hides its model, nothing leaks), so it's correctness/UX, not security. Fix: normalize `body.explores` via `normalizeModelPath` before validation/persist so all input channels share one representation.

### Nits

**1. `/compile` against a notebook 404s once `explores` is declared** — `environment.ts:376` calls `assertQueryBoundaryEarly` for any `gateModel`; a `.malloynb` is never in `explores` so `isQueryEntryPoint` is false → `NotQueryableError` (`model.ts:815`). `compile.controller.ts` passes the path through with no `.malloy` restriction, so it's reachable via the API (latent for the SDK). Contradicts the documented "notebooks are always public" invariant — exempt notebooks in the boundary methods or at the call site. Confirm `/compile` is even meant to accept notebook paths.

**2. Raw model text / compiled-model-by-path aren't part of the boundary** — `getModelFileText` (`environment.ts:1106-1119`) and `getModel` (`model.controller.ts:44-62`) still return a hidden file's contents by path under `queryableSources: "declared"`. Consistent with the design (`#(authorize)` is the access gate, README:330) and these endpoints are pre-existing — but the README leads with "queryable == discoverable," so one sentence noting raw-text/by-path retrieval is *not* part of the boundary would close the gap.

**3. "explores is declared" derivation duplicated ~4×** — `package.ts:122-124` (and `applyDiscoveryPolicyToModels`, `listModels`, warnings). A single `exploresDeclared()`/`getExploreSet()` helper removes the hand-sync risk.

**4. `updatePackage` preserves one concept two ways** — `environment.ts:1191-1223` sets in-memory metadata with the *effective* values but passes raw `body.*` to `writePackageManifest`; stays consistent only because the writer skips `undefined`. Pass the same resolved locals to both.

**5. Stray rename artifact** — `package_load_pool.ts:251` comment reads "Locate the worker explore." (was "worker entrypoint"); revert that word.

### Scope note (non-blocking)
Bundles three separable changes — discovery curation, the query boundary, and the non-destructive publish-rollback refactor. The rollback fix could stand alone, though it's a defensible prerequisite for validating `explores` at publish.

Sha-Bang:
Thanks James — all addressed in 84519e40 (plus the `/compile` follow-up in 6c5ba2ea).

**Important #1 — `explores` normalization.** Fixed. Lifted `normalizeModelPath` out of the worker into `constants.ts` so it's the single normalizer for every input channel. `updatePackage` now normalizes `body.explores` before *both* validation and `writePackageManifest` (passing the same resolved value to the in-memory metadata and the persist call, which also closes Nit 4). So `["./index.malloy"]` / backslash paths validate and persist identically to on-disk explores — no misleading publish-time 400, no publish-vs-reload divergence. Regression test added (`./`-prefixed body explores → accepted, persisted as `model.malloy`).

**Nit 1 — `/compile` 404s a notebook once `explores` is declared.** Exempted notebooks at the top of `assertQueryBoundaryEarly`/`assertQueryBoundaryCompiled` (always public, regardless of the call site), with a test.

I also chased down the open question I'd raised — *should `/compile` accept notebook paths at all?* **Answer: no.** `/compile` appends the submitted source to the *target model's* content for namespace context (`api-doc.yaml` ~1745-1798), so a `.malloynb` (markdown + cells) isn't a valid target — it only produces a confusing parse error. It was reachable, but nothing actually calls it that way: the SDK/app compile against `.malloy` model paths and run notebooks via the separate `executeNotebookCell` endpoint; MCP doesn't use `/compile`; tests pass only `.malloy` paths. So in 6c5ba2ea I added an explicit guard in `compileSource` that rejects a notebook path up front with a clear `400` ("Cannot compile against a notebook… /compile takes a .malloy model path for namespace context"), with a test. Consider this one closed.

**Nit 2 — raw text / by-path retrieval.** Added a README note that `queryableSources` gates the query surface (`getQueryResults`, the MCP tool, `/compile`), not raw retrieval by exact path — by design; use `#(authorize)` when contents must be protected, not just hidden from discovery.

**Nit 3 — duplicated `explores` derivation.** Extracted `exploresDeclared()` / `exploreSet()` helpers; all four sites use them now.

**Nit 5 — stray rename.** Reverted "worker explore" → "worker entrypoint".

**Scope note.** Fair — it does bundle discovery curation, the query boundary, and the rollback refactor. Leaving them together since the branch was collapsed to a single commit to work around a GitHub PR-head desync on the original #823 (this PR continues it); splitting now would be churn for no review benefit. Happy to revisit if you'd prefer otherwise.

kylenesbit (approved):
Approving. Thorough, well-scoped change with strong test coverage on the security-relevant paths (schema-oracle defense, multi-statement last-`run:`-wins, name-collision and ad-hoc-alias laundering, backtick identifiers). The opt-in-on-`explores` gating makes "no explores = today's behavior" structurally true, and the strict-at-publish / fail-safe-at-load asymmetry with non-destructive rollback is well-reasoned. CI green on typecheck/lint/tests.

One thing for whoever wires up the downstream Credible migration to internalize (documented in the PR, not a blocker): `explores` hiding curates listings and (under `declared`) queries, but a hidden model's file text and compiled metadata remain fetchable by exact path. If `#(agent-hidden)` implied "an agent cannot see this content at all," use `#(authorize)` for true content protection — `explores` alone is a discovery filter, not a retrieval gate.

PR #837, opened 2026-06-19
feat(mcp): add agent retrieval tools, a separate agent MCP server, and dual-channel skills

mlennie:
## Summary

Adds two read-only agent retrieval tools to Publisher, served on a separate, isolated MCP server, plus a set of agent skills delivered both as files and as MCP prompts.

- **`malloy_getContext`**: lexical (lunr/BM25) retrieval over a package's model entities (sources, views, named queries, and dimension/measure fields). Two-phase: discover with a plain-English `query`, then optionally drill down with `sourceName`. The entities it returns map directly onto `malloy_executeQuery`. The per-package index is cached on the `Package` instance and rebuilds on reload.
- **`malloy_searchDocs`**: lexical search over a bundled index of the Malloy documentation (no runtime network dependency).
- **Agent MCP server** (`packages/server/src/mcp/agent_server.ts`): a third in-process listener (`AGENT_MCP_PORT`, default `4041`) using the official MCP SDK with a stateless StreamableHTTP transport. It hosts only these two tools, kept separate from the core MCP server.
- **Dual-channel skills**: the 24 skills under `skills/` are also registered as MCP prompts on the agent server, so hosts that ingest MCP but do not load skill files get the same guidance. A contributor guide lives under `docs/agent-skills/`.

## Why a separate server

The new tools and their tests stay isolated from the existing MCP tool surface, so this does not change the core server's behavior. Both servers share the one store and the same shutdown handling.

## Configuration and deployment

- New env var `AGENT_MCP_PORT` (default `4041`), bound to `PUBLISHER_HOST`.
- `Dockerfile` exposes `4041`; the example Compose file and the docker run snippets map it.
- The agent listener is registered for graceful shutdown alongside the others.

## Tests

- Unit specs for the tool utilities (`docText`, `sanitize`, `searchDocsIndex`), the skills-bundle parser (frontmatter, CRLF, colons), and bundle integrity.
- An in-memory MCP protocol spec drives the agent server over the real MCP client and transport (tool isolation, the skill prompts, a live `searchDocs` call, and the `getContext` error path).
- A `get_context` retrieval eval harness establishes a lexical baseline.

## Notes

- Both tools are read-only; no state mutation.
- The agent server is unauthenticated, mirroring the existing MCP server; deployments front it with their own gateway.
- No `api-doc.yaml` change (MCP tools do not touch the OpenAPI surface).


kylenesbit (approved):
## 🚀 This is excellent work

Really impressive PR, Monty. A few things that stand out:

- **The architecture is exactly right.** Standing the agent tools up on their own isolated MCP server (`agent_server.ts`, port 4041) keeps the core MCP surface completely untouched — new tools and their tests can evolve without any risk to the existing tool surface. Clean separation, shared store + shutdown handling, no spillover.
- **`malloy_getContext` is thoughtfully built.** Two-phase discover→drill-down maps directly onto `malloy_executeQuery`, the lunr/BM25 index is cached per-`Package` in a `WeakMap` (so a reload swaps the instance and the stale index is GC'd automatically — no manual invalidation), `name` is boosted over `doc`, and user input is sanitized of lunr operators. And critically, it goes through the **curated** `listModels`/`getSourceInfos`/`getQueries` accessors, so it respects the #833 discovery surface — hidden sources never leak through retrieval. That interaction was easy to miss and you nailed it.
- **Dual-channel skills are a great call.** Shipping the 24 skills both as files and as MCP prompts means skill-aware hosts (Claude) and MCP-only hosts (Codex/Cursor/ChatGPT) all get the same guidance. The bundled docs index keeps `malloy_searchDocs` free of any runtime network dependency. Nice.
- Strong test coverage — tool utilities, the skills-bundle parser, an in-memory MCP protocol spec over the real client/transport, and a retrieval eval harness for a lexical baseline.

Approving on the strength of the design and implementation. 🎉

## ⚠️ One blocking item to land first — CI failures

The 3 red watch-mode E2E tests (ubuntu + windows; macOS happens to pass on timing) are a real regression from this PR, not flake. Root cause:

- The new agent server binds a **fixed default port `4041`**, but the integration harness only allocates free ports for `PUBLISHER_PORT`/`MCP_PORT` — never `AGENT_MCP_PORT`. So every spawned server child binds `4041`, and when a prior SIGKILL'd server hasn't released it yet, the next boot hits `EADDRINUSE`.
- `startAgentMcpServer` attaches **no `'error'` handler** to its listener, so that bind failure throws and crashes the whole process → the test's fetch gets "Unable to connect."

Two small fixes:
1. **Harness:** in `watch_mode.integration.spec.ts` (and any other server-spawning harness), allocate and pass `AGENT_MCP_PORT: String(await getFreePort())`, distinct from the other two ports — same pattern already used for `PUBLISHER_PORT`/`MCP_PORT`.
2. **Robustness (also matters in prod):** add an `'error'` handler in `startAgentMcpServer` so a bind failure on the *auxiliary* agent listener logs and degrades gracefully instead of taking down the main API + core MCP server at boot.

#1 turns CI green; #2 is worth doing regardless. Happy to push these if it's faster — just say the word.

## 📋 Follow-up: Quick Start

Separately, we'll want a great quick-start so users can go from zero to "agent tools answering questions against my package" in a few steps. Looping in Ofer to help shape that — tracking as a fast-follow, not a blocker for this PR.

mlennie:
Thanks @kylenesbit , and great catch on the root cause. Both fixes are pushed:

1. Harness: watch_mode.integration.spec.ts now allocates a distinct free port for AGENT_MCP_PORT alongside PUBLISHER_PORT/MCP_PORT. It's the only harness that spawns a full server child; the others import the app in-process.
2. Robustness: startAgentMcpServer now attaches an 'error' handler so a bind failure on the auxiliary agent listener logs and degrades gracefully instead of taking down the main API + core MCP server.

Verified locally: the watch-mode spec passes 11/11, and with 4041 deliberately occupied the server stays up (main API + core MCP serving) and just logs the degradation. I rebased the fix onto your merge-main commit so your work is preserved and the history stays linear.

Quick Start sounds great as a fast-follow. Happy to pair with Ofer on it once this lands.

PR #839, opened 2026-06-21
Materialization quality pass: simplify, tests, telemetry, serve-time binding

kylenesbit:
## Summary

Publisher-side quality pass over the Malloy Persistence v0 materialization subsystem. Behavior-preserving except for the serve-time manifest-binding alignment (called out below). Builds on the auto-run work merged in #836.

- **Simplify** (`materialization_service.ts`): centralized `computeBuildId`, extracted a shared `commitManifest` finalizer used by both the auto and round-2 paths, removed a duplicate recompile, and fixed a misleading path-specific log message.
- **Serve-time manifest binding alignment** (`environment.ts`, `model.ts`, `package.ts`, `api-doc.yaml`): the worker now threads the bound build manifest into the serve runtime and reports read-only binding state — `manifestBindingStatus` (`unbound`/`bound`/`live_fallback`), `manifestEntryCount`, and `boundManifestUri` — so the control plane can confirm a worker actually bound a distributed manifest instead of inferring it from logs. *(Contract change: additive, read-only response fields on the serving API.)*
- **Telemetry** (`materialization_metrics.ts`): manifest bind success/failure/timeout counters, per-source build-duration, and drop/dropTables counters; structured transition logging that distinguishes timeout from failure.
- **Tests**: characterization for build-instruction derivation + the staging→swap→rename SQL sequence; unit coverage for skip-if-unchanged / `forceRefresh` and in-flight cooperative cancel; extended manifest-binding and lifecycle integration specs.
- **Cleanup**: inlined the always-true `ENABLE_LIST_MODEL_COMPILATION` flag.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `bun test src/service/package.spec.ts src/service/materialization_service.spec.ts` — 45 pass / 0 fail
- [ ] `bun run test:integration` (manifest-binding + lifecycle suites; needs Docker/Postgres)
- [x] Exercised end-to-end via the service repo's Postgres materialization e2e against a locally built image (publish → materialize → physical table created with correct data → on-demand rebuild → drop on teardown)

## Notes

- Serve-time query **routing** is not yet provable end-to-end: the worker fetches and binds the manifest (`manifestBindingStatus=bound`, `manifestEntryCount=1`), but a query on a persisted source is still served live rather than read from the materialized table, and multi-worker manifest redistribution is racy. Tracked as a follow-up; the e2e routing assertion remains skipped on the service side with this documented.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -2451,6 +2451,37 @@ components:
             URI (gs:// or s3://) of the control-plane-computed manifest for this package.
             On (re)load the worker reads it and binds persist references
             (buildId -> physicalTableName). Null = serve live.
+        manifestBindingStatus:
+          type: string
+          readOnly: true
+          enum:
+            - unbound
+            - bound
+            - live_fallback
+          description: >-
+            Server-computed, read-only: whether the control-plane build manifest
+            is currently bound to this package's served models. `unbound` = no
+            manifest configured, so the package serves live. `bound` = a manifest
+            was fetched and applied, so persist sources route to their
+            materialized physical tables. `live_fallback` = a `manifestLocation`
+            is configured but the fetch/bind failed or timed out, so the package
+            is serving live despite intending to be materialized-routed. Lets the
+            control plane confirm a worker actually bound a distributed manifest
+            rather than inferring it from logs.
+        manifestEntryCount:
+          type: integer
+          readOnly: true
+          description: >-
+            Server-computed, read-only: number of buildId -> physical-table
+            entries currently bound (0 when unbound or on live fallback).
+        boundManifestUri:
+          type: ["string", "null"]
+          readOnly: true
+          description: >-
+            Server-computed, read-only: the manifest URI actually bound to the
+            served models. Usually equals `manifestLocation`, but can differ
+            after an in-memory auto-load following a materialization build (no
+            URI), in which case it is null. Null whenever the package is unbound.
 
     Model:
       type: object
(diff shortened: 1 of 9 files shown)


kylenesbit:
Pushed fixes for the two checks owned by this PR:

- **DCO** — re-applied all commits with `Signed-off-by` (was failing: no sign-off).
- **Lint and check formatting** — ran Prettier on the three flagged files (`materialization_metrics.ts`, `model.ts`, `materialization_service.spec.ts`).

Both now pass, along with Typecheck, `bun-direct test`, and `npx test`.

### The remaining red check (`Cross Platform Tests`) is pre-existing and unrelated to this PR

The only failures are two assertions in `tests/integration/concurrent_package/concurrent_package.integration.spec.ts` (lines 140 and 230):

```
expect(meta.name).toBe("gcs_faa")  ->  Received: "faa"
```

This is not caused by this branch:

- The package `name` returned by the API comes from `Package.loadViaWorker` (`name: outcome.packageMetadata.name`), which reads the extracted package's on-disk `publisher.json`. That code path — and `installPackage` — are **byte-identical to `main`** and untouched by any commit here (this PR only touches the materialization service, manifest binding, telemetry, and their tests).
- The `name` value is therefore determined entirely by the external test fixture `gs://publisher_test_packages/gcs_faa.zip`, whose `publisher.json` now resolves to `name: "faa"`.
- "concurrent POST … all succeed" isn't even a race for this assertion — it does N POSTs then a single GET, so `name` is deterministic; it fails on every run regardless of timing.
- An unrelated PR (#837) from 2026-06-19 passed this exact test, while runs from 2026-06-21 fail it — consistent with the fixture's `publisher.json` having changed between those dates. A re-run on `main` should reproduce the failure.

Happy to open a separate issue/PR to make the test robust to the fixture's internal name (e.g. assert against the registered package key rather than the zip's `publisher.json` name), but that's out of scope for this materialization pass.

PR #846, opened 2026-06-25
refactor(model): update render tag validation to log warnings instead of throwing errors

sagarswamirao:
- Adjusted the `validateRenderTags` method to log warnings for misconfigured renderer tags instead of throwing a `ModelCompilationError`, allowing package loads to proceed without failure.
- Updated related tests to verify that warnings are logged correctly for invalid renderer configurations in various scenarios, including packages and sources with bad render tags.
- Cleaned up imports in `model.ts` for better organization and clarity.

This change enhances the user experience by providing more informative logging while maintaining the integrity of the package loading process.

### DIFF
--- a/packages/server/src/service/model.ts
+++ b/packages/server/src/service/model.ts
@@ -25,11 +25,6 @@ import * as fs from "fs/promises";
 import { createRequire } from "module";
 import * as path from "path";
 import { components } from "../api";
-import { deserializeError } from "../package_load/package_load_pool";
-import type {
-   SerializedModel,
-   SerializedNotebookCell,
-} from "../package_load/protocol";
 import {
    getDefaultQueryRowLimit,
    getMaxQueryRows,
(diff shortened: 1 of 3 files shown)


mlennie (approved):
LGTM, this is the right fix and the tests are solid. Nice touch asserting the specific target name shows up in the warning, that keeps the quoted-identifier coverage from the old reject tests. Core behavior is exactly what we want: a bad render tag logs a warning and the package still loads.

One optional tightening, non-blocking, fine to merge as-is to unblock: validateRenderTags can still throw for an unexpected reason, the await import("@malloydata/render-validator") and the validateRenderTags(result) call sit outside a try. The reload path's defensive catch handles that, but the create path (loadViaWorker) has no catch, so an unexpected validation failure there would still propagate to Package.create's outer catch and clean up / fail the whole package, the same failure class we're removing here. Wrapping the create-path call (or the body of validateRenderTags) in the same defensive try/catch would give the full "render-tag validation can never fail a load" guarantee. Happy to push that as a follow-up so it doesn't hold this up.

Minor: we lean on structured logging elsewhere (logger.warn("msg", { target, issues })) rather than string concat, but not important.

Follow-up for a separate PR, not this one: we don't surface the finding anywhere the UI can read, so a genuinely-misconfigured package still shows the generic "couldn't be loaded" message instead of the render-tag reason (the thing James flagged). Recording the findings on the model and adding them to the model response would let the UI name the bad tag.

PR #856, opened 2026-06-29
Sagark/validate render tags on load

sagarswamirao:


sagarswamirao:
Implemented in other PR - Monty pulled in these changes into his PR

PR #858, opened 2026-06-30
feat: reach databases through an SSH bastion (connection proxy)

Sha-Bang:
## What

Adds an optional, connection-type-agnostic **SSH bastion proxy** so a database reachable only through a jump host can be queried. The publisher opens an `ssh2` connection to the bastion plus a local `127.0.0.1` `net.Server` that forwards each socket via `forwardOut()`, and points the `pg` driver at the local endpoint — no driver / `pg` changes.

## Shape

- `Connection.proxy` → `ConnectionProxy { type: 'ssh', ssh: SshProxyConfig }` (discriminated union so future proxy types slot in).
- `openProxy(proxy, target)` in `src/service/proxy.ts` — reusable transport; the per-type builder points the driver at the returned endpoint. The tunnel opens lazily in `lookupConnection`, is cached per connection, and is torn down in `releaseConnections` (forwarded sockets force-destroyed so teardown can't hang).

## Robustness

- Forwarded traffic uses `pipe()` so SSH/TCP **backpressure** is honored — neither side buffers unbounded on a slow peer.
- The bastion connection sends **SSH keepalives** (15s interval) so an idle tunnel isn't silently reaped by Cloud NAT / stateful-firewall / `ClientAlive` idle timeouts, which would otherwise surface only as a confusing error on the next query.

## Security

- **No env-flag gate.** The SSH proxy is a normal connection capability, authorized by whoever configures the connection — deliberately not behind an env flag, and kept separate from the `publisher` HTTP multi-hop type's `PUBLISHER_ALLOW_PROXY_CONNECTIONS` (a distinct publisher-to-publisher decision). The trust boundary is host-key pinning below.
- **Host key pinned** and verified on every connect (`ssh2` `hostVerifier`); fail-closed unless `PUBLISHER_SSH_ALLOW_UNKNOWN_HOSTKEY=true`. Accepts a `known_hosts` line or a bare base64 blob — plain and hashed (`|1|…`) lines both work (only the key blob is compared, never the hostname).
- Load-time validation in `validateConnectionShape` (postgres-only today; explicit host/port required; the `connectionString` form is rejected outright so the tunnel can't silently target a different database than configured).

## Tests / docs

Hermetic `proxy.spec.ts` (in-process ssh2 + TCP echo): byte round-trip, host-key mismatch, fail-closed, env opt-out, `known_hosts`-line parsing, unsupported type. Gate tests in `connection_config.spec.ts` — including rejecting a proxied postgres connection that carries a `connectionString`, and one missing explicit host/port. Operator guide in `docs/connections.md`. Cross-platform CI green.

## Verification (local, end-to-end)

Beyond the hermetic specs, the full `lookupConnection` proxy branch was exercised against real infrastructure locally:

- Stood up **Postgres reachable only through a dockerized SSH bastion** — the DB port is not published to the host, so traffic can only reach it via the tunnel.
- Drove the actual feature path — `createEnvironmentConnections` → `lookupConnection` → `openProxy` (ssh2 local forward) → `buildProxiedPostgresConnection` → a pooled `pg` connection at the local endpoint — with the bastion's real pinned host key.
- Ran a live Malloy query through the tunnel and got real rows back (`SELECT count(*) … FROM widgets` → `{n:3, last:"gamma"}`).

This confirms SSH public-key auth, host-key pinning (verified on connect), the `forwardOut` channel to an in-VPC-style DB, and a full query round-trip — with no changes needed to the feature code.

## Follow-ups

Move `ssh2` from the root to `packages/server` (works today via hoisting); commit the local docker-based integration harness as an opt-in (docker-gated) test so the `lookupConnection` proxy branch has durable end-to-end coverage; reconnect-on-drop (re-dial the tunnel after a post-handshake SSH close, beyond the keepalive that already prevents idle reaping); `ssl.servername` passthrough for fully-verified TLS through the tunnel (tracked in malloydata/malloy#2960).

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3078,7 +3078,7 @@ components:
 
     ConnectionProxy:
       type: object
-      description: "Optional proxy (e.g. SSH tunnel) that the publisher opens before connecting to the database. Proxied connections require the operator to set PUBLISHER_ALLOW_SSH_PROXY=true; default-deny."
+      description: "Optional proxy (e.g. SSH tunnel) that the publisher opens before connecting to the database. Proxied connections require the operator to set PUBLISHER_ALLOW_PROXY_CONNECTIONS=true; default-deny."
       required:
         - type
       properties:
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3053,6 +3053,8 @@ components:
             ]
         attributes:
           $ref: "#/components/schemas/ConnectionAttributes"
+        proxy:
+          $ref: "#/components/schemas/ConnectionProxy"
         postgresConnection:
           $ref: "#/components/schemas/PostgresConnection"
         bigqueryConnection:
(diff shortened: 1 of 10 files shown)


sagarswamirao (approved):

housejester:
Took a pass through the proxy path. The happy path and hermetic tests look solid, and the commit history shows most of the sharp edges (teardown hangs, backpressure, connectionString precedence, retry-on-failed-build) were already caught. A few **minor / non-blocking** notes — none of these should hold the merge:

- **Theoretical tunnel leak on a post-`openProxy` throw.** In the proxy branch of `lookupConnection` (`connection.ts` ~L1102), if `openProxy` resolves (endpoint registered in `proxyEndpoints`) but a later step in the IIFE throws, the rejected-promise `catch` evicts the *connection* cache entry and a retry calls `openProxy` again — overwriting `proxyEndpoints.set(name)` and orphaning the first tunnel. Today the only post-`openProxy` step is `buildProxiedPostgresConnection`, which only throws on `!pg` (already impossible after validation), so this is defense-in-depth, not a live bug. If we want to close it, register endpoints keyed on the endpoint object (or close-on-throw inside the IIFE). Noting, not requesting.

- **Doc caveat is slightly stricter than the parser.** `extractHostKeyBase64` (`proxy.ts:52`) picks the `AAAA…` blob token, which is present even in a hashed (`|1|…`) `known_hosts` line — the hash only obscures the *hostname*, not the key, and this code never matches hostnames. So the "hashed entries not supported" note in the docs/schema is more conservative than the code actually requires. Harmless; just flagging the mismatch.

- **Host-key comparison non-constant-time — verified fine.** The base64 compare in the `hostVerifier` (`proxy.ts:120`) is non-constant-time, but the host key is public, so there's no timing concern here. Noting pre-emptively in case a scanner flags it.

PR #862, opened 2026-07-02
feat(proxy): per-connection sslmode + optional multi-key host-key pinning

Sha-Bang:
Follow-up to #858 (SSH-bastion connection proxy). Makes proxied Postgres connections robust for real-world bastions and databases — specifically load-balanced/HA bastions and force-SSL RDS — by adding a per-connection TLS mode, relaxing host-key pinning from required-single-key to optional-multi-key, and validating all proxy config up front.

## Why

Three things #858's design got wrong for real targets, found testing against a customer's AWS setup:

1. **Required single-key host-key pinning breaks load-balanced bastions.** Their bastion is an ELB fronting ≥2 backends, each with a *different* SSH host key (confirmed: 8 scans → 2 distinct keys). A single required pin fails ~50% of connects. Mainstream self-service tools (Metabase, Looker, Redash, Hex, Retool, Superset) don't pin bastion host keys at all — encrypted SSH transport is the baseline, pinning is opt-in.
2. **No per-connection TLS mode.** Proxied connections inherited the deployment `PGSSLMODE`. Through the tunnel the driver connects to `127.0.0.1`, so any hostname-verifying mode fails the cert ALTNAME check even with the CA trusted — a force-SSL RDS target couldn't connect.
3. **Static config errors surfaced at connect time**, after dialing the tenant's bastion — which fails *open* for a degenerate hostKey and re-dials the bastion indefinitely for a bad sslmode.

## What

**Per-connection `postgresConnection.sslmode`** `{disable, no-verify, verify-ca}`, applied only on the proxied path (self-built connectionString to the local endpoint; the non-proxied path keeps using env `PGSSLMODE`).
- **`no-verify` (default when a proxy is set)** — encrypt without verifying, so a force-SSL target isn't rejected for plaintext.
- **`verify-ca`** — chain-verify against `NODE_EXTRA_CA_CERTS` (the baked RDS roots from #860) via `uselibpqcompat`, skipping the hostname the tunnel can't satisfy.
- **`disable`** — emits an explicit `?sslmode=disable` so pg can't silently fall back to the deployment `PGSSLMODE` and verify.

**`hostKey` is now optional and multi-key.** Absent → connect unpinned (still encrypted; a warning is logged). Present → fail-closed, **match-any** over one-or-more `known_hosts` lines, so an LB/HA bastion can pin every backend's key. Removes the `PUBLISHER_SSH_ALLOW_UNKNOWN_HOSTKEY` env flag (its meaning inverts under the new default).

**Fail-fast validation + fail-closed boundary** (from a 5-reviewer pass — the general fix for #3). `validateConnectionShape` now rejects, at config load *before any bastion is dialed*:
- a hostKey that is set but parses to zero keys (blanks/`#` comments only) — was a silent fail-*open* downgrade to an unpinned tunnel;
- an unsupported `sslmode`, or `verify-ca` when the CA bundle is unset/missing — was thrown only inside the lazily-dialed builder, so a permanent typo re-dialed the bastion on every query;
- `sslmode` set on a non-proxied connection — was silently ignored;
- a proxied database name with URI-reserved chars (pg decodes the path with `decodeURI`) — was a silent wrong-database.

And `hostVerifier` gates "pinned" on hostKey being *configured*, not on it parsing to ≥1 key — so a set-but-degenerate pin fails closed at the security boundary regardless of how config reached it (defense-in-depth), and the mismatch error carries the presented-key fingerprint (needed to know which LB backend's key to add).

Docs (`docs/connections.md`) and the `SshProxyConfig.hostKey` / `PostgresConnection.sslmode` schema descriptions updated.

## Robustness matrix

| Config | Behavior |
|---|---|
| Direct postgres, no sslmode | Unchanged — deployment PGSSLMODE |
| Direct postgres + sslmode set | Rejected at load (no silent-ignore) |
| Proxy, no hostKey | Connect unpinned (encrypted), warn |
| Proxy + single / multi-key hostKey | Fail-closed match-any (LB/HA bastion) |
| Proxy + hostKey comment/blank only | Rejected at load; fails closed at boundary |
| Proxy + no-verify (default) | Encrypt, skip verify (force-SSL RDS) |
| Proxy + verify-ca (+CA present) | Chain-verify, hostname skipped |
| Proxy + verify-ca, CA missing | Rejected at load (no bastion re-dial churn) |
| Proxy + disable / invalid sslmode | Explicit plaintext / rejected at load |
| Proxy + reserved-char db name | Rejected at load with guidance |

## Security note

Unpinned + `no-verify` means the publisher→bastion→DB path is unauthenticated end-to-end — a network MITM could capture DB credentials. This matches the default posture of the tools above; the compensating control is the customer allowlisting our published egress IPs on the bastion's inbound SSH. Stronger-than-default customers should prefer **`verify-ca`** (authenticates the actual database endpoint) and/or pin host keys. `verify-full` through a tunnel awaits per-connection `servername` passthrough (malloydata/malloy#2960).

## Tests

- `buildProxiedSslQuery`: default→no-verify, explicit `disable`→`?sslmode=disable` (regression: never empty), `verify-ca` query shape, throws without a CA bundle, throws on unknown mode, and a `pg-connection-string` version guard (verify-ca must still parse to an ssl config if pg is ever downgraded).
- `connection_config` proxy validation: multi-line hostKey + no-verify accepted; comment-only hostKey, unsupported sslmode, verify-ca-without-CA, sslmode-without-proxy, and reserved-char db name all rejected at load.
- `proxy.spec`: unpinned-connect (default), multi-line match-any accept, all-mismatch reject, set-but-degenerate hostKey fails closed; existing single-key / known_hosts-line / hashed-line / mismatch cases retained.

`tsc` clean; proxy 9/9, connection_config 37/37, buildProxiedSslQuery 7/7.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3189,10 +3189,11 @@ components:
             hostname can't be checked; `verify-ca` validates the server cert
             chain against the trusted CA bundle (e.g. the baked Amazon RDS roots)
             without the hostname, `no-verify` encrypts without verifying, and
-            `disable` uses no TLS. Defaults to `no-verify` when a proxy is set (so
-            a force-SSL target isn't rejected for plaintext). Only valid on a
-            proxied connection — a direct connection uses the deployment PGSSLMODE
-            and rejects this field.
+            `disable` uses no TLS. The server defaults it to `no-verify` when a
+            proxy is set (so a force-SSL target isn't rejected for plaintext) — a
+            server-applied default, not a schema default. Only valid on a proxied
+            connection — a direct connection uses the deployment PGSSLMODE and
+            rejects this field.
 
     MysqlConnection:
       type: object
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3130,12 +3130,14 @@ components:
         hostKey:
           type: string
           description: |
-            Pinned bastion host public key, as an OpenSSH known_hosts line or its
-            base64 blob (e.g. from `ssh-keyscan <bastion>`), verified on every
-            connect. Plain and hashed (`|1|…`) known_hosts lines both work — only
-            the key blob is compared, never the hostname. A missing or mismatched
-            host key fails the connection closed, unless
-            PUBLISHER_SSH_ALLOW_UNKNOWN_HOSTKEY=true is set.
+            Optional pinned bastion host public key(s), as one or more OpenSSH
+            known_hosts lines (or bare base64 blobs), verified on every connect.
+            List multiple lines to pin a load-balanced/HA bastion that presents a
+            different key per backend — any listed key is accepted; a mismatch
+            fails the connection closed. Plain and hashed (`|1|…`) lines both work
+            — only the key blob is compared, never the hostname. When omitted, the
+            tunnel connects without host-key verification (the self-service
+            default); the SSH transport is still encrypted.
 
     ConnectionAttributes:
       type: object
@@ -3178,6 +3180,18 @@ components:
           description:
             Complete PostgreSQL connection string (alternative to individual
             parameters)
+        sslmode:
+          type: string
+          enum: [disable, no-verify, verify-ca]
+          description:
+            TLS mode for a connection reached through a `proxy` (SSH bastion).
+            Because the driver connects to a local tunnel endpoint, the cert
+            hostname can't be checked; `verify-ca` validates the server cert
+            chain against the trusted CA bundle (e.g. the baked Amazon RDS roots)
+            without the hostname, `no-verify` encrypts without verifying, and
+            `disable` uses no TLS. Defaults to `no-verify` when a proxy is set (so
+            a force-SSL target isn't rejected for plaintext); ignored on the
+            non-proxied path, which uses the server's PGSSLMODE.
 
     MysqlConnection:
       type: object
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3190,8 +3190,9 @@ components:
             chain against the trusted CA bundle (e.g. the baked Amazon RDS roots)
             without the hostname, `no-verify` encrypts without verifying, and
             `disable` uses no TLS. Defaults to `no-verify` when a proxy is set (so
-            a force-SSL target isn't rejected for plaintext); ignored on the
-            non-proxied path, which uses the server's PGSSLMODE.
+            a force-SSL target isn't rejected for plaintext). Only valid on a
+            proxied connection — a direct connection uses the deployment PGSSLMODE
+            and rejects this field.
 
     MysqlConnection:
       type: object
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -97,8 +97,8 @@ VPC; it is not an IP-restriction mechanism (restrict the database directly for t
 > tenant-configured bastion, so it is authorized by whoever configures the connection. It
 > is **not** behind an env-flag gate, and is deliberately kept separate from the
 > `publisher` type's `PUBLISHER_ALLOW_PROXY_CONNECTIONS` (that flag is about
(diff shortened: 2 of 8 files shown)


api-doc.yaml
```diff
@@ -3178,6 +3180,19 @@ components:
           description:
             Complete PostgreSQL connection string (alternative to individual
             parameters)
+        sslmode:
```
sagarswamirao, line 3183: The schema and the code/description disagree on `sslmode` here, and since api-doc.yaml is the contract we feed to openapi-generator (the Pydantic/Java models the downstream services consume), that drift can break the generated clients even though the server tests are green. Flagging it early so it doesn't bite us downstream.

Two small things the code relies on that aren't in the schema yet:

1. The code defaults it (`sslmode ?? "no-verify"`) and the description says "Defaults to `no-verify`" -- but there's no `default:` in the schema, so that default only lives in the TS builder. A generated Pydantic/Java client would materialize `Optional[SslMode] = None`, not `= no-verify`, so two clients end up disagreeing on what "unset" means. If we want the default to be part of the API, adding `default: no-verify` here would cover it.

2. It's an optional enum with no null-union, so a strict generated deserializer would reject `sslmode: null` (and `sslmode: ""`, which isn't an enum member) at parse time -- before `validateConnectionShape` ever runs. If any producer can serialize null/empty, making it nullable (`type: [string, "null"]` for 3.1) would handle that.

Context on why it slips past CI: our server uses openapi-typescript (structural types, no runtime enum validation), so it passes here, but the SDK/CLI use openapi-generator-cli where it'd surface. The rule of thumb that's bitten us before: if the code null-checks/defaults a field, the schema should say `nullable:`; if the API is meant to expose the default, the schema should carry `default:`.

Could you add whichever one matches the intent (likely `default: no-verify`, plus the null-union if empty is ever sent)? Happy to help if useful. Thanks!

Sha-Bang: Thanks Sagar — dug into how both suggestions land in the generated clients and settled on leaving the schema shape as-is (with a description clarification, pushed in e31bb7f3):

**`default: no-verify`** — `sslmode` is proxy-only and the default is *conditional* (no-verify only when a proxy is set), so a static schema default would apply unconditionally. typescript-axios (the SDK + CLI) ignores `default:` (JSDoc only, never serialized), so it's inert for TS users. openapi-python-client, though, materializes a schema default as the field's value → it's no longer `UNSET`, so `to_dict()` always serializes it. Every `PostgresConnection` built via the python SDK would then send `sslmode: "no-verify"`, including on **direct** (non-proxied) connections — which `validateConnectionShape` rejects. So it'd break the common direct-postgres case to make the rare proxied case marginally more self-describing. The default is intentionally server-applied (`sslmode ?? "no-verify"`); I clarified the description to say so explicitly rather than imply a schema default.

**nullable** — no producer emits `null`: both generated SDKs omit unset optionals (UNSET / undefined), and the server tolerates null anyway (`if (sslmode != null)` + `?? "no-verify"`). The only exposure is a hand-authored `sslmode: null` echoed back on read, which is symmetric with every other optional field in this spec — none of which are nullable — so a one-field null-union would be inconsistent and generates awkward `Union[None, Unset, Enum]` types for SDK users.

Two things I did fold in:
- Description now states `no-verify` is a server-applied default, not a schema default.
- Empty-string edge: `sslmode: ""` was slipping the truthiness check and only failing at tunnel-build; it's gated on `!= null` now so it's rejected at config load, with a regression test.

Appreciate the early flag — the generated-client-contract angle is exactly right; it just pointed the other way once the python-vs-TS generator split and the proxy-only rejection are in the picture.

sagarswamirao (approved):

PR #863, opened 2026-07-03
fix(runtime sdk): encode model path as a single URL segment (%2F)

jswir:
## Problem
The served `/sdk/publisher.js` builds the query URL as:
```js
"/models/" + modelPath.split("/").map(encodeURIComponent).join("/") + "/query"
```
This keeps **literal slashes**, so a model in a subfolder (`model/x.malloy`) becomes a **multi-segment** URL path. Servers that match only a single-segment `{path}` variable (e.g. a Spring `AntPathMatcher` route `…/models/{path}/query`) don't match the POST and fall through to a GET-only handler → **HTTP 405 "Method 'POST' is not supported."**

This hits **in-package HTML data apps** (which load this vanilla SDK). It does *not* hit the typed axios SDK, which `encodeURIComponent`s the **whole** path (slashes → `%2F`, one segment) — so app/notebook/MCP queries against the same subfolder model work fine.

## Fix
Encode the whole model path as one segment, matching the typed SDK:
```js
"/models/" + encodeURIComponent(modelPath) + "/query"
```
- Safe on the open-source publisher: the server route (`packages/server/src/server.ts`, `…/models/*?/query`) matches the `*?` wildcard and decodes the `%2F` param back to `model/x.malloy`.
- Also works against single-segment servers that accept encoded slashes.
- One-liner in `packages/server/src/runtime/publisher.js` (`rawQuery`, which backs both `query` and `queryFull`). `dist/` is build-generated.

## Test
Query a package whose model is in a subfolder from a data app (or `POST …/models/model%2Fx.malloy/query`): 200 instead of 405. Flat models are unaffected (`encodeURIComponent("x.malloy")` == `"x.malloy"`).

### DIFF
--- a/packages/server/src/runtime/publisher.js
+++ b/packages/server/src/runtime/publisher.js
@@ -79,7 +79,12 @@
          "/packages/" +
          encodeURIComponent(target.pkg) +
          "/models/" +
-         modelPath.split("/").map(encodeURIComponent).join("/") +
+         // Encode the whole model path as ONE path segment (slashes -> %2F). A model
+         // in a subfolder ("model/x.malloy") must not become two URL segments: some
+         // servers only match a single-segment {path} and would 405 a multi-segment
+         // path. This matches the typed SDK (encodeURIComponent) and the server's
+         // encoded-slash handling.
+         encodeURIComponent(modelPath) +
          "/query";
       var body = { compactJson: compactJson };
       if (malloyQuery) body.query = malloyQuery;


Sha-Bang (approved):
LGTM — clean, well-scoped fix.

### Verified

1. **Typed SDK already does this.** The generated OpenAPI client (`packages/sdk/src/client/api.ts:5429`) builds the URL with:
   ```js
   .replace(`{${"path"}}`, encodeURIComponent(String(path)))
   ```
   So `model/x.malloy` → `model%2Fx.malloy` (one segment). This PR aligns the vanilla runtime SDK to match.

2. **Server route is compatible.** The Express route (`server.ts:1501`) uses a `*?` wildcard:
   ```
   …/models/*?/query
   ```
   Express stores the wildcard match in `params['0']` and auto-decodes `%2F` back to `/` — so the server handles both encodings transparently.

3. **No-op for flat models.** `encodeURIComponent("x.malloy") === "x.malloy"`, so non-subfolder models are completely unaffected.

4. **Fixes the real bug.** Servers with single-segment `{path}` patterns (e.g. Spring `AntPathMatcher`) were 405-ing because `model/x.malloy` created two path segments instead of one.

One minor nit (non-blocking): the 4-line comment in the diff is thorough, but given how self-explanatory `encodeURIComponent(modelPath)` is (and that it now mirrors the typed SDK), a single-line comment would suffice. Not worth a round-trip though.

Sha-Bang (approved):

PR #873, opened 2026-07-08
docs(ai-agents): document the :4041 agent MCP server

mlennie:
## What

Rewrites `docs/ai-agents.md` so it documents the agent MCP server on `:4041`, which shipped in #837 but was never mentioned in this guide. The page previously covered only the core MCP server on `:4040` and the Python bridge for Claude Desktop, so a reader following the canonical AI-agents doc had no way to discover the agent server.

## Changes

- Adds a section for the agent MCP server on `:4041` (`AGENT_MCP_PORT`): the two read-only retrieval tools `malloy_getContext` and `malloy_searchDocs`, and the bundled agent skills served as MCP prompts (noting that MCP prompts are on-demand). Cross-links `docs/agent-skills`.
- Documents connecting over HTTP (for example Cursor and VS Code), and keeps the Python-bridge path for Claude Desktop, noting that current Claude Desktop connects through the bridge or its Connectors UI rather than an HTTP entry in `mcpServers`.
- Corrects the core discovery tool name `malloy_projectList` to `malloy_environmentList` (renamed in the project-to-environment migration), and the quick-start startup line to the real output `0.0.0.0:4040`.
- Adds a prerequisites note (Node.js for `npx`, Python 3 for the bridge) and a note that the first run downloads the sample packages over the network.
- Strengthens the security guidance: both servers are unauthenticated and `malloy_executeQuery` runs against the databases your models connect to, so bind to loopback for local use and front it with an authenticating gateway before exposing it.
- Drops the horizontal-rule separators and tightens the prose.

## Notes

Docs-only change, no code touched. Every technical claim was verified against source in `packages/server/src` (`server.ts`, `mcp/agent_server.ts`, `mcp/tools/*`) and confirmed live against a running agent server on `:4041` (`tools/list` returns the two tools, `prompts/list` returns the 24 skills). All cross-links resolve and both config examples are valid JSON.

### DIFF
--- a/docs/ai-agents.md
+++ b/docs/ai-agents.md
@@ -1,148 +1,117 @@
-# AI Agents with the Model Context Protocol (MCP) Server
+# AI Agents with the Model Context Protocol (MCP)
 
-## 1. Overview
+## Overview
 
-The Malloy Publisher includes a server implementing the **Model Context Protocol (MCP)**, a standardized interface designed to connect Large Language Models (LLMs) and AI agents directly to governed, semantic data models.
+Publisher speaks the [Model Context Protocol (MCP)](https://modelcontextprotocol.io), so an AI agent can work with your Malloy models over a standard interface. Because a Malloy model already carries the business logic and the relationships between entities, an agent grounds its answers in your definitions instead of guessing at table and column names.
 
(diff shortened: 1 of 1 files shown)


jswir (approved):
lgtm

PR #875, opened 2026-07-08
fix(sdk): repair dead docs links and in-package link routing

mlennie:
## What

Fixes several developer-experience link/navigation bugs in the Publisher web UI.

### Dead "Learn more" docs links

The Home screen feature cards, the empty-state button, and the Add-Package dialog linked to README anchors (`#ad-hoc-data-analysis`, `#notebook-based-dashboards`, `#mcp-based-ai-data-agents`, `#server-configuration`, `#architecture-overview`) that no longer exist after the README was restructured to link out to docs.malloydata.dev. GitHub silently loads the README top, so every card was a dead link on the app's first screen. They now point at the live docs pages (Explorer, MCP for AI Agents, Publish Your Models) through a shared `DOC_LINKS` module that the sidebar also uses, so a future docs move needs one edit.

### In-package navigation dropped the package segment

Two paths built URLs relative to the current page and dropped the package segment, landing on `/{env}/{file}` which resolves the filename as a package and 404s with a misleading "Package manifest ... does not exist":

1. Links inside a rendered notebook or README (for example an imdb package README's table of contents, `[Steven Spielberg](spielberg.malloynb)`) were plain browser anchors resolved against the current page URL. They now route through the SPA, resolved against the source file's own directory so nested notebooks keep their subdirectory, while external, hash, and absolute links stay as normal anchors with a safe-scheme guard for hrefs in untrusted package READMEs. `Notebook`/`Package` stay router-agnostic: the host threads an optional `onNavigate` handler (the app supplies react-router), and without one, links fall back to plain absolute anchors, so embedding the SDK still needs no `<Router>` context.

2. The package page's file-list rows (notebooks, models, pages) navigated with a bare relative path and relied on react-router relative resolution, which drops the package segment depending on trailing-slash and route state. They now use absolute paths, matching the sibling "Back" and "Materializations" links that were already absolute.

## Testing

New Playwright regression tests: clicking a README table-of-contents link and clicking a package file-list row both keep the package segment and render, instead of 404ing. The README case uses `README.malloynb`, which shares no name overlap with the package, so a dropped segment is caught (the pre-existing `imdb.malloynb` test could not catch it).

`bun run typecheck`, `bun run lint`, and `bun run test` (226 pass) are green, and every new docs URL was verified to return HTTP 200.


jswir (approved):
lgtm!

PR #878, opened 2026-07-09
fix(server): redact connection secrets from request and response logs

mlennie:
## Problem

`loggerMiddleware` logs every non-health request at `info` with the full request body (`payload: req.body`) and, when response logging is enabled, the full response body. On the connection and environment endpoints (for example `PATCH /api/v0/environments/:name` and the connection create/update routes) that body carries provider credentials, so secrets were written to the logs in plaintext:

- BigQuery `serviceAccountKeyJson` (the full service-account private key)
- Snowflake `privateKey`, `privateKeyPass`, `password`
- Postgres / MySQL / Trino `password`, Postgres `connectionString`
- S3 `secretAccessKey` / `sessionToken`, GCS `secret`
- Azure `clientSecret` / `sasUrl`, Databricks `token` / `oauthClientSecret`, Trino `peakaKey`, MotherDuck `accessToken`

Winston re-serializes the payload (re-escaping quotes, inserting whitespace), so downstream secret masking in CI or a log sink cannot be relied on to catch these. The source-level fix is the real protection.

## Fix

Add a key-based redactor in `logger.ts` and apply it to both the logged `payload` and `response`. It deep-copies the body and replaces any property whose key names a credential with `[REDACTED]`, recursing through nested objects and arrays (connections arrive nested under `connections[]`). Primitives and `Date` values pass through unchanged, the input is never mutated, and circular references are handled.

Non-secret fields (host, project id, connection name, and so on) are preserved, so the request logs remain useful for debugging.

## Tests

`logger.spec.ts` covers: every credential field masked at the top level, secrets masked inside the environment-PATCH payload shape, non-secret fields preserved, arrays, primitive pass-through, `Date` preservation, no input mutation, and circular references.

## Notes

Internal logging behavior only. No new env var, CLI flag, config key, or endpoint, so no README or docs change is triggered.


packages/server/src/logger.ts
```diff
@@ -102,14 +156,14 @@ export const loggerMiddleware: RequestHandler = (req, res, next) => {
       const logMetadata: Record<string, unknown> = {
          statusCode: res.statusCode,
          duration: formatDuration(durationMs),
-         payload: req.body,
+         payload: redactSensitive(req.body),
          params: req.params,
          query: req.query,
```
sagarswamirao, line 161: `params` and `query` are logged as-is while `payload`/`response` are redacted. That is fine today -- no connection route carries a credential in the path or query string (they all take the config in the body). The gap is that a future route that does put a token in the query (a signed URL, an `?access_token=` callback) would leak it here with no obvious signal, since this line looks intentional. A one-line comment noting the assumption -- "query/params are not redacted; keep credentials out of the URL" -- would make that contract explicit for the next person adding a route. Non-blocking.

packages/server/src/logger.ts
```diff
@@ -84,6 +84,60 @@ export function formatDuration(durationMs: number): string {
    return `${durationMs.toFixed(2)}ms`;
 }
 
+// Connection endpoints carry provider credentials in their request/response
+// bodies (BigQuery serviceAccountKeyJson, Snowflake privateKey/password, Postgres
+// password/connectionString, S3 secretAccessKey/sessionToken, Azure clientSecret/
+// sasUrl, Databricks token, ...). The middleware below logs the full body, so
+// those values are masked by key name before they reach a log transport.
+const SENSITIVE_KEYS: ReadonlySet<string> = new Set([
```
sagarswamirao, line 92: This set is now a second source of truth that has to stay in sync with the connection DTO by hand. Matching is exact-key, so the day a provider adds a differently-named secret (`apiSecret`, `refreshToken`, `clientKey`, ...) it will log in plaintext with nothing failing to signal it -- the leak is silent and only this set gates it. Two low-cost ways to prevent drift: (1) a comment here pointing maintainers to update this set when adding a credential field to `connection.dto.ts`, or (2) a unit test that reflects over the DTO's credential fields and asserts each is present in `SENSITIVE_KEYS`, so adding an un-redacted secret field turns CI red. Non-blocking, but (2) is what actually holds the invariant over time.

sagarswamirao (approved):
Correct, well-scoped secret-redaction fix. I cross-checked `SENSITIVE_KEYS` against every credential-bearing field in `api-doc.yaml` -- all real secrets are covered, and the fields left unredacted (`keyId`, `accessKeyId`, `hostKey`) are public identifiers, not secrets. `redactSensitive` handles nesting, arrays, primitives, Date, non-mutation, and circular refs; tests are thorough. Left two non-blocking notes inline (unredacted query/params assumption, and keeping the key set in sync with the DTO).

mlennie:
Thanks Sagar, both fair.

- query/params: agreed, no connection route carries a credential in the path or query today (configs go in the body). I'll add the one-line "query/params are not redacted, keep credentials out of the URL" comment so the assumption is explicit for the next route.
- SENSITIVE_KEYS drift: good catch, exact-key matching means a future differently-named secret would log silently. I'll add the maintainer note and a drift guard that fails CI if a connection credential field isn't covered.

Both as a small follow-up so I don't churn this approved PR. Merging as-is per your sequence.

PR #879, opened 2026-07-09
fix(server): redact secrets from axios error logs

mlennie:
Follow-up to the request/response log redaction PR. Same class of leak, a different code path.

## Problem

`logAxiosError` logs the details of every failed outbound call, and those details can carry credentials:

- `error.response.data` can echo a connection config (with `serviceAccountKeyJson`, `password`, and so on) when an upstream worker returns it in an error body.
- `error.request` (the client-side-error branch) is the raw request object, which holds the outgoing `Authorization` header.
- the setup-error branch logged the whole `error`, whose `config.headers` also holds `Authorization`.

## Fix

Extract `buildAxiosErrorLog(error)` as the single place that shapes the log metadata, with secrets removed:

- server-side error: `response.data` and `response.headers` run through `redactSensitive`.
- client-side error: log only `method`, `url`, and `code`, never the raw request object.
- setup error: log only `error.message`.

`redactSensitive` now matches keys case-insensitively and covers credential HTTP headers (`authorization`, `proxy-authorization`, `cookie`, `set-cookie`, `x-api-key`), so an `Authorization` bearer token cannot slip through on header casing.

## Tests

`logger.spec.ts` adds: case-insensitive matching, HTTP-header redaction, and `buildAxiosErrorLog` for all three branches (response data and headers redacted, the raw request never logged, the setup error reduced to its message).

## Stacked PR

This branch is built on the request/response redaction PR because both share the `redactSensitive` helper. Review and merge that one first; this diff is only the axios-error change and will retarget to `main` once the base lands.

## Verification

Full publisher gate green (typecheck, lint, prettier, tests). Redaction confirmed end to end by running the real `logAxiosError` through winston with secret-bearing errors and checking the emitted log: no bearer token, no service-account key, non-secret fields preserved.

### DIFF
--- a/packages/server/src/logger.ts
+++ b/packages/server/src/logger.ts
@@ -87,9 +87,12 @@ export function formatDuration(durationMs: number): string {
 // Connection endpoints carry provider credentials in their request/response
 // bodies (BigQuery serviceAccountKeyJson, Snowflake privateKey/password, Postgres
 // password/connectionString, S3 secretAccessKey/sessionToken, Azure clientSecret/
-// sasUrl, Databricks token, ...). The middleware below logs the full body, so
-// those values are masked by key name before they reach a log transport.
-const SENSITIVE_KEYS: ReadonlySet<string> = new Set([
+// sasUrl, Databricks token, ...), and axios error logs can carry credential HTTP
+// headers (Authorization, Cookie). Those values are masked by key name before
+// they reach a log transport. Matching is case-insensitive so header casing
+// (Authorization vs authorization) does not slip through.
+const SENSITIVE_KEY_NAMES = [
+   // connection config credentials
    "password",
    "connectionString",
    "serviceAccountKeyJson",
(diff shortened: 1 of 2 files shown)


packages/server/src/logger.ts
```diff
@@ -184,23 +197,53 @@ export const loggerMiddleware: RequestHandler = (req, res, next) => {
...
-      // Something happened in setting up the request that triggered an Error
-      logger.error("Axios unknown error", { error });
+      // that falls out of the range of 2xx.
+      return {
+         message: "Axios server-side error",
+         meta: {
+            url: error.response.config.url,
+            status: error.response.status,
+            headers: redactSensitive(error.response.headers),
+            data: redactSensitive(error.response.data),
```
sagarswamirao, line 221: `response.data` is only walked when it is an object. If an upstream echoes a secret inside a JSON *string* body (an unparsed response, e.g. `responseType` left as text or a non-JSON content type), it passes through `redactSensitive` as a primitive and lands in the log verbatim. The object case is the common one for a JSON API, so this is a narrow residual rather than a blocker -- worth being aware of. If you want to close it, a string-body branch that skips or truncates `data` when it is not an object would cover it.

packages/server/src/logger.ts
```diff
@@ -184,23 +197,53 @@ export const loggerMiddleware: RequestHandler = (req, res, next) => {
...
-   } else {
-      // Something happened in setting up the request that triggered an Error
-      logger.error("Axios unknown error", { error });
+      // that falls out of the range of 2xx.
+      return {
+         message: "Axios server-side error",
+         meta: {
+            url: error.response.config.url,
+            status: error.response.status,
+            headers: redactSensitive(error.response.headers),
```
sagarswamirao, line 220: At runtime `error.response.headers` is an `AxiosHeaders` instance, not the plain object the tests construct. `redactSensitive` uses `Object.entries`, which does pick up AxiosHeaders' own enumerable properties, so redaction still works here -- no change needed. Flagging only so it is a conscious choice: the `buildAxiosErrorLog` server-side test builds `headers` as a plain literal, so the real `AxiosHeaders` shape is never exercised. An optional test that passes a `new AxiosHeaders({ authorization: ... })` would lock the behavior in against an axios internals change.

sagarswamirao (approved):
Correct follow-up. `buildAxiosErrorLog` is a clean, testable extraction; all three branches are handled right -- server-side redacts `response.data`/`response.headers`, client-side drops the raw `error.request` (which carried the outgoing Authorization header) for safe descriptors only, and the setup branch is reduced to `error.message`. Case-insensitive matching plus the credential-header additions close the header-casing gap. Two non-blocking notes left inline (string-body `response.data`, and exercising a real `AxiosHeaders` instance in the test). Merge #878 first, then this.

mlennie:
Thanks Sagar, both noted.

- string-body response.data: right, a non-object body (responseType text or a non-JSON content type) carrying a secret would pass through redactSensitive as a primitive and land verbatim. I'll add a branch that skips or truncates data when it isn't an object.
- AxiosHeaders: good flag, the test builds headers as a plain literal so the real AxiosHeaders shape isn't exercised. I'll add a test constructing new AxiosHeaders({ authorization: ... }) to lock the redaction against an axios internals change.

Both in a quick follow-up PR.

PR #881, opened 2026-07-11
feat(server): per-load compile vs schema-fetch timing breakdown

housejester:
## What & why

`malloy_package_load_duration` reports only the *total* time to load a package. When a load is slow, an operator can't tell whether the time went to **Malloy compilation** or to **connection schema fetches** — two costs with very different remedies (worker CPU/sizing vs. warehouse round-trips). This adds a per-load phase breakdown so that's answerable from metrics and logs.

Additive instrumentation only — no HTTP/OpenAPI shape changes.

## Changes

- **New histograms** (labelled `status` only — deliberately *not* by package name, to keep cardinality bounded):
  - `malloy_package_load_compile_duration`
  - `malloy_package_load_schema_fetch_duration`
  - `malloy_package_load_schema_fetches`
- The package-load **worker** measures these per load and reports them on its result; `service/package.ts` records them and adds the split to the existing "load completed" log line.
- **Bucket boundaries.** All load-duration histograms now use explicit boundaries out to 5 min. OTel's defaults top out at 10 s, so any slower load landed in `+Inf` and the tail was unresolvable — this also widens the existing `malloy_package_load_duration`.
- **Test harness:** added `collectHistogram` (it had counter/gauge collectors only).

## How compile time is measured

Compile time = compile-region wall-clock − time awaiting proxied schema fetches. `process.cpuUsage()` is process-wide across worker threads, so it can't give a per-load CPU number; this wall-minus-wait "self time" is the honest proxy — treat it as a **ceiling** on compile CPU (non-fetch stalls like GC also land in it). Worker setup (manifest read + file listing) is bracketed out so it can't inflate the figure; the remainder (`total − compile − schema-fetch`) stays derivable. The accounting is `jobId`-scoped so a straggler fetch from a prior (reused-worker) load can't corrupt the next load's numbers.

Example completion log:

```
workerDurationMs: 5.05, compileDurationMs: 2.68, schemaFetchDurationMs: 2.21, schemaFetchCount: 1
```

## Testing

- New unit tests for the wait accountant (overlap/union, reset isolation, straggler rejection, clamp) and the metrics module (records + tail-past-10s bucket boundaries).
- Existing package-load pool + service specs pass; `tsc` / eslint clean.

### DIFF
--- a/packages/server/src/package_load/package_load_pool.ts
+++ b/packages/server/src/package_load/package_load_pool.ts
@@ -247,6 +247,8 @@ export interface LoadPackageOutcome {
       }
    >;
    loadDurationMs: number;
+   /** Per-load phase timing breakdown (see {@link LoadPackageResult.timings}). */
+   timings: LoadPackageResult["timings"];
 }
 
 // ──────────────────────────────────────────────────────────────────────
@@ -844,6 +846,7 @@ function adaptResult(result: LoadPackageResult): LoadPackageOutcome {
          sourceInfos: m.sourceInfos as Malloy.SourceInfo[] | undefined,
       })),
       loadDurationMs: result.loadDurationMs,
+      timings: result.timings,
    };
 }
 
--- a/packages/server/src/package_load/package_load_worker.ts
+++ b/packages/server/src/package_load/package_load_worker.ts
@@ -99,6 +99,7 @@ import {
    type MalloyGivenApi,
 } from "../service/given";
 import { ignoreDotfiles } from "../utils";
+import { RpcWaitAccountant } from "./rpc_wait_accountant";
 import type {
    ConnectionMetadata,
    ConnectionMetadataRequest,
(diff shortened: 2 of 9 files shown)


kylenesbit (approved):
Approve. Well-scoped additive instrumentation with strong documentation and test coverage. Verified the two riskiest bits hold up: jobId gating is consistent (schemaWait.begin uses job.requestId, matching ProxyConnection's jobId), and the module-level singleton is safe under the pool's one-load-per-worker invariant.

Minor, non-blocking follow-ups:
- The "error" status path is never exercised — recordPackageLoadPhases is only called with "success", and in-band model compilation errors also record as "success". Worth clarifying what status="success" means or wiring up the error path.
- compileDurationMs subtracts whole-load schema-fetch wait rather than compile-region wait; correct today only because fetches are lazy. Consider hardening.

housejester:
Thanks for the review! Both follow-ups addressed in `d1a5501b`.

**Status labeling (1):** good catch — and it was actually recording a *wrong* label, not just leaving one unused. The phase record ran before the model-build loop, so an in-band model compilation error was tagged `success` on the phase metrics while `malloy_package_load_duration` recorded `compilation_error` for the same load. Now a shared classifier feeds both, and the phases are recorded after the model build settles with the load's terminal status (`success` / `compilation_error` / `error`). Pool-infrastructure failures throw before the worker returns timings, so they're intentionally not recorded (documented in the record helper).

**Compile-region wait (2):** moved `schemaWait.begin()` to the compile-region boundary so `compileDurationMs = compileRegion − schemaFetchWait` stays exact regardless of whether a fetch ever happens during setup (none do today, as you noted) — removes the latent coupling.

Added a test asserting the phase histograms are separable by status.

PR #887, opened 2026-07-14
feat: close the agent authoring loop (compile + reload tools, doc corrections)

mlennie:
## What and why

Follow-up to #884 and #885, from feedback on a demo of the clone-and-go flow. The theme is closing the authoring loop for an agent: after #885 an agent could discover a deployment and answer questions, but changing a model still meant firing throwaway queries to see whether it parsed, and a saved change could not be run at all without restarting the server.

## Changes

**Two MCP tools**

- `malloy_compile`: compile-check Malloy source against a model and get structured diagnostics back (severity, message, code, line and character) without running a query, so an agent validates a change while authoring instead of firing a throwaway query. Wraps the existing in-process compile path (`CompileController.compile`, the same one behind REST `/compile`). `includeSql` only extracts the generated SQL; the query is never run.

- `malloy_reloadPackage`: recompile a package from its on-disk content so a source or view added after boot becomes queryable by name, without restarting the server. This is the other half of the loop: validate, save, reload, run. It wraps `PackageController.getPackage(reload=true)`, the same path behind REST `GET /environments/:env/packages/:pkg?reload=true`, so it adds no capability the HTTP surface did not already expose.

**Orientation**

- Serve the MCP `instructions` field, the on-the-wire analog of AGENTS.md, so an agent that reached the server over npx without cloning still knows to start with `getContext` and can fetch the skills as prompts. The text is static and pre-auth safe (it interpolates no live environment, package, or connection data), with a test pinning it.

**Docs**

- Document the edit-and-run loop and its safety gate. Compile first is not just a speed-up: a reload whose models do not compile removes the package's on-disk copy under `publisher_data/` and drops the cached package. Watch-mode symlink mounts are exempt. `--init` restores packages as configured but wipes all of `publisher_data/` first, so it does not bring an edit back.
- Refresh the README agent-tools section, which still described the pre-#884 two-tool surface, and update `docs/ai-agents.md`, which still listed three tools and described the endpoint as read-only. `malloy_reloadPackage` mutates server state, so that sentence needed correcting.
- Widen AGENTS.md "What you can do" and the getting-started skill to name building models and data apps as first-class. A live agent asked what it could do answered with querying alone and never mentioned data apps, though the `html-data-apps` skills ship here.
- Document what `--watch-env` actually needs, so people stop hitting the silent no-op: a local-directory env, and the in-place symlink mount established on a fresh server root or an `--init` run. Later boots keep watching without `--init`.
- Note that MCP clients discover tools at connect time, so start the server before launching the agent, or reconnect (`/mcp` in Claude Code) if it was already running.
- Add the two tools and the loop to the `malloy-modeling` skill, which several surfaces route agents to as the modeling authority but which mentioned neither tool.
- After building a model, the `malloy-model` skill suggests a browser link to the package at `localhost:4000`, guarded by a check that the running server serves the package.

## Verification

Full gate green: typecheck, lint, prettier (CI's root scan), and 195 tests passing including the integration suite. New unit tests cover the compile tool (clean/error/warning diagnostics, includeSql, the error path, a missing location, givens forwarding) and the reload tool (payload shaping, warnings and exploresWarnings, reload=true forwarding, a compile failure on reload, the error path). The protocol spec asserts both tools over a real MCP client.

Verified live against a running server over MCP, not only in tests:

- The full loop: edit a model on disk, watch the new view fail to resolve, `malloy_compile` it clean, `malloy_reloadPackage`, then run the new view by name and get real data back, with the on-disk edit intact.
- The destructive path, by doing it: a reload with a deliberately broken model removes the package directory, and a second reload cannot recover it. That is why the compile-first gate is documented as a safety rule on every surface rather than a suggestion.
- Boot behavior, from the server log: each configured package compiles at boot before the server reports `serving`, which is why a source added afterwards needs a reload.
- The `instructions` field arrives over the wire and carries no interpolated data.

Security: both tools are thin wrappers forwarding verbatim to the controller methods the unauthenticated REST routes already call, so neither adds capability to the MCP surface. They log only scalar identifiers, never config objects or row data, and the reload payload is narrower than the REST response (it hand-picks name, description, and warnings, and does not echo `location`). Path traversal stays contained by the existing `assertSafePackageName` and `assertSafeRelativeModelPath` checks. Worth flagging for a maintainer: `malloy_reloadPackage` makes the MCP surface state-mutating for the first time, though the same effect is already reachable through the equivalent REST endpoint, so the existing "stateless and unauthenticated, run it behind your own gateway" guidance still applies, and `docs/ai-agents.md` now says so explicitly.

## Follow-ups (not in this PR)

- A failed reload deletes the package directory: `Package.create`'s cleanup treats a reload like a failed fresh install, and only skips the removal for a watch-mode symlink. The fix belongs in the reload primitive rather than in a tool that wraps it; documented here as a compile-first gate in the meantime.
- Both tools funnel caught errors through the Malloy error helper, so a back-pressure 503 reads as a syntax problem. A shared classifier would fix both.
- The reload payload cannot distinguish an in-place recompile from a re-fetch; a `mode` discriminator would make that observable.


jswir:
**Worth fixing before merge: error-signal conflation in both new tools.**

`compile_tool.ts:141-166` and `reload_package_tool.ts:108-133` funnel every caught error through `getMalloyErrorDetails()`, whose regexes only match Malloy-syntax shapes. A `ServiceUnavailableError` from `assertCanAdmitNewPackage` (`environment.ts:736-753`, reachable on every reload under memory pressure) or a plain `PackageNotFoundError` falls through to the generic "check your syntax/typos" suggestions — indistinguishable from a real compile failure, with no retryable signal. The discrimination this needs already exists next door: `handler_utils.ts:53-88` has the `instanceof` cascade (`EnvironmentNotFoundError` / `PackageNotFoundError` / `ServiceUnavailableError` with back-pressure-specific guidance) that `execute_query` uses. Extracting that into a shared classifier and calling it from both new catch blocks is small enough to land here rather than as the deferred follow-up — as shipped, the two most common operational failures (typo'd name, server under load) both return misleading remediation advice.

Nits, none blocking:

- `server.protocol.spec.ts:40-41` only asserts the new tool names appear in `listTools()`, while the same file does a real `callTool` for `malloy_searchDocs` (L57) and `malloy_getContext` (L72) — add `callTool` coverage for both new tools. Related: the error tests in both new specs stop at `expect(parsed.error).toBeDefined()` (`compile_tool.spec.ts:129`, `reload_package_tool.spec.ts:106,123`); pinning the suggestion text would have caught the conflation above.
- `compile_tool.ts:100` takes no `tryAcquireQuerySlot`/timeout, unlike `execute_query_tool.ts:137-146` (whose comment explains why MCP shares the concurrency pool), even though `compileSource` can trigger live schema fetches while holding the package lock and the tool description invites tight-loop use. Pre-existing parity with REST `/compile`, so fine to defer — flagging because the fix pattern is already in this PR.
- The failed-reload-deletes-the-package fact is hand-restated in six places (`AGENTS.md:90`, `README.md:112`, `docs/ai-agents.md:27`, `skills/malloy-modeling/SKILL.md:116`, `reload_package_tool.ts:26`, `server.ts:37`) and already needed synchronized edits twice within this PR's own commit history. At minimum `RELOAD_DESCRIPTION` and `MCP_INSTRUCTIONS` are TypeScript in the same package and can share one exported fragment.

jswir:
**Design direction, not blocking merge — how should the authoring loop work long term?**

These tools close the loop today, but stepping back: `malloy_reloadPackage` reads like a workaround for three gaps in watch mode, and the cleaner end-state may be to fix those gaps instead of growing an agent-facing mutating surface.

The gaps as they stand:

1. **Watch mode keeps no readable compile state.** The watcher's `reloadPackage` swallows failures — logs and returns `false` (`watch-mode.controller.ts:139-145`) so it doesn't bounce browser tabs — and nothing stores "last compile outcome" anywhere a client can read. This is missing for humans too: a person in watch mode whose save fails to compile gets silence unless they tail the server log.
2. **Watch mode is single-env** — one chokidar instance, first `--watch-env` wins, the rest are mounted live but never auto-reloaded (`server.ts:1835-1846`).
3. **In-place symlink mounts only happen at boot**, so `POST /watch-mode/start` on a running server watches the `publisher_data/` copies, not the source repo.

Proposed end-state:

- **Editing session ⇒ watch mode.** The watcher stays the single recompile engine; agents don't get a second trigger into the same `Package.create` path.
- **Watcher records per-package `lastCompile` state** (`ok | failed`, diagnostics, `compiledAt` or a generation counter). Fixes the silent-swallow problem for everyone, and models compile status as what it is — state on the package resource, rather than a GET with a side effect (`?reload=true`).
- **MCP surface stays read-only:** a `malloy_getPackageStatus` (or compile status folded into `malloy_getContext`'s package payload) replaces `malloy_reloadPackage`. The destructive-delete trigger disappears from the agent surface entirely — no compile-first safety rule to document in six places.
- **`malloy_compile` stays** as the non-mutating, synchronous pre-flight.

Cost worth naming: determinism. Today an agent calls reload and knows the recompile finished before its next query; with watcher-owned recompile it has to confirm the watcher settled, which is why the status payload needs `compiledAt`/generation so the agent checks "newer than my write" instead of polling blind.

The residual case reload genuinely covers — files changed by an external process (git-sync, volume update) on a server without watch mode — is an operator concern the REST `?reload=true` endpoint already serves; it doesn't need to be an MCP tool aimed at agents.

Concretely: the follow-up already planned for `Package.create`'s destructive cleanup could be the moment to flip to watcher-owned recompile + readable compile status, and deprecate `malloy_reloadPackage` before anything depends on it. If there's an agent-editing scenario where watch mode fundamentally can't work (beyond the three gaps above), that would change the calculus — genuinely interested in the counter-argument.

mlennie:
**On the error conflation:** you and my reviewer independently landed on the same fix, extract the `instanceof` cascade out of `handler_utils` and call it from both catches. It's already built as `classifyToolError`, along with the rest of this, in a branch stacked on #887 that I'm about to push. Missing package now reports not-found, `ServiceUnavailableError` reports retryable, neither mentions the Malloy file. If you'd rather it land in #887 itself than in the stack, say so and I'll move it.

Your nits, taken:

- **The weak error assertions were the actual bug.** `expect(error).toBeDefined()` is the shape that let the conflation through unnoticed, so the tests were green while both of the most common failures gave misleading advice. Now pinning the remediation text, and both new assertions fail if the classifier is reverted.
- **`callTool` coverage** for both new tools added to the protocol spec.
- **The six restatements**: fair, and worse than you could see. They needed resynchronizing twice in this stack's history and disagreed both times. `RELOAD_DESCRIPTION` and `MCP_INSTRUCTIONS` now share one exported fragment.
- **compile's missing back-pressure gate**: agreed, deferring.

**On the design direction**, one thing changes the premise: the destructive cleanup is already fixed in that same stacked branch. `Package.create`'s cleanup is opt-in now (only `installPackage` asks), a failed reload keeps the last-good model serving, and the docs already say a failed reload is safe. So "the destructive-delete trigger disappears, no compile-first safety rule in six places" is no longer a reason to drop the tool. That's gone either way.

Strong agree on gap 1 regardless. A human in watch mode whose save doesn't compile gets silence and keeps staring at a stale chart. `lastCompile` fixes that whether or not an agent reads it, and `compiledAt`/generation is a better settle answer than I expected: "newer than my write" beats polling blind.

The counter-argument you asked for: **the default deployment has no watch mode.** Clone-and-go, `npx`, `bun run start`, none of it passes `--watch-env`, and `malloy-samples` is github-sourced so it isn't watch-eligible for the mount anyway. That's the exact scenario #885 and #887 exist for, and it's where a user asks an agent to build a model mid-session. Under watcher-owned recompile that agent either can't make its saved edit live at all, or something has to `POST /watch-mode/start` first, which is itself a mutating call. The mutation doesn't disappear, it moves: to another endpoint, or to the human having guessed right at launch. Gap 2 has the same shape, one env gets a watcher and the rest silently don't.

"Editing session ⇒ watch mode" presumes we know it's an editing session at launch. We don't; the user decides mid-session by asking.

Also worth separating: read-only MCP isn't a read-only deployment. `GET ...?reload=true` is already unauthenticated and is exactly what the tool wraps, so removing the tool narrows what agents can reach, not what the server exposes.

Where I'd land: `lastCompile` now, since it fixes the human case today and gives agents visibility into watch envs they didn't start. Keep the explicit reload while the default deployment has no watcher. If always-on watching is on the table, that genuinely changes the calculus and I'd rather have that conversation than defend a tool. Either way it's the same engine, both call `environment.getPackage(pkg, true)` (`watch-mode.controller.ts:134`), so this is about the trigger, not a second recompile path.

mlennie:
Pushed as #890, stacked on this branch so its diff is only its own commits.

Mapping it to your review:

- **The classifier** is in `56f07ca5`: `classifyToolError` in `handler_utils`, the `instanceof` cascade extracted and called from both new catch blocks, exactly as you described. Same commit carries the `Package.create` destructive-cleanup fix, so the compile-first safety rule is already gone from all six surfaces.
- **Your nits** are in `1d939e6c`: remediation text pinned (missing package reports not-found, `ServiceUnavailableError` reports retryable, neither mentions the Malloy file, and both fail if the classifier is reverted), `callTool` coverage for both new tools in the protocol spec, and `RELOAD_DESCRIPTION` / `MCP_INSTRUCTIONS` sharing one exported fragment instead of restating it.

Still happy to move the classifier into this PR instead, if you'd rather it land before #887 merges rather than right after.

jswir (approved):

PR #888, opened 2026-07-14
feat(materialization): standalone scheduler, environment-scoped listing, and schedule management UI

girishjeswani:
## Summary

Adds standalone materialization scheduling to the publisher, plus the API and UI to manage it:

- **Standalone scheduler** — an opt-in (`PUBLISHER_LOCAL_MATERIALIZATION_SCHEDULER`, default off) periodic sweep that fires a loaded package's `materialization.schedule` cron so an open-source publisher can rebuild on a cadence. Two independent orchestration guards (disabled by default; skips control-plane-driven packages with a `manifestLocation`), in-memory arming that skips fires missed during downtime (no double-fire on restart), a per-tick stampede cap, and conflict-coalescing when a build is already active.
- **5-field UTC cron evaluator** — a thin wrapper over `cron-parser` pinned to exactly 5 fields and UTC, mirrored on the client for the UI.
- **Run trigger** — materialization runs record `trigger` (`ON_DEMAND` vs `SCHEDULER`) so a scheduled rebuild is distinguishable from a manual one, preserved through the build commit.
- **Environment-scoped listing** — `GET /environments/{env}/materializations` returns every materialization across all packages (newest first, each carrying `packageName`), bounded by a default limit.
- **PATCH-editable persistence policy** — a package PATCH can now edit `scope` and `materialization.schedule`, enforcing the same scope/schedule/freshness rules a publish enforces (plus a valid 5-field cron); changing the schedule re-arms the running scheduler on its next tick.
- **SDK UI** — a schedule card + editor on the package materializations view (cronstrue describes the cadence and next UTC run), an environment-scoped materializations view with a Trigger column, and shared `TriggerChip`/`SectionLabel` components.

## Testing

- `typecheck` (server + SDK) and `eslint` clean on changed files.
- Materialization integration suite: **28 tests / 7 files pass** (includes new `environment_list`, `schedule_edit`, `scheduler` specs; real DuckDB builds).
- `cron_evaluator` + `materialization_scheduler` unit specs pass (17 tests).
- Python SDK: `build-python-sdk.sh` runs clean (spec validates, client in sync with `api-doc.yaml`, tests pass) — no drift.

## Notes

- `api-doc.yaml` changed; the checked-in Python client was verified in sync (regeneration produces no diff).

---

## Review response (round 2)

Addressing @kylenesbit's two reviews.

**Config-parity (the local ↔ hosted control-plane contract):**
1. **Restart recovery** — the scheduler now re-anchors on first arm from the newest `trigger=SCHEDULER` run, firing one catch-up and jumping forward (matches "recovery fires once and jumps forward") instead of skipping a missed occurrence.
2. **Cron grammar + publish gate** — `CronEvaluator` (and the client preview) reject non-UNIX extensions (`L`/`W`/`#`/`?`); cron validity moved into `persistencePolicyWarnings()` so publish/PATCH/load/scheduler all enforce it identically (publish no longer accepts a cron PATCH would reject).
3. **Scope flip** — scope is now an explicit `package|version` control on the schedule card; clearing a schedule no longer strands `scope: version`.
4. **Write-once ingest** — schedule card + dialog note that, in a hosted deployment, a schedule change takes effect on the next publish.
5. **Tables-only** — the scheduler fires the table build only (the OSS publisher has no index pipeline); captured in code comments now, with the operator-facing write-up deferred to the docs follow-up below.

**Orchestrated-deployment safety:**
1. PATCH treats `scope:null`/`materialization:null` as absent (preserve, not wipe), so the control plane's rebind PATCH never trips the policy gate — with an integration test on a package carrying a pre-existing load-tolerated warning.
2. Both scheduler guards documented as independent + load-bearing; env-var doc warns it must never be set on an orchestrated worker (serve-live gap noted).
3. `api-doc` notes `trigger` stays in `metadata`; promoting it is a breaking change.
4. `validateCreateBody` comment + unit test: a client cannot forge a `SCHEDULER` run.
5. Default `LIMIT` confirmed env-scoped only; per-package list stays unbounded (comment + repo test).

## Follow-ups (tracked separately, not this PR)

- **CLI** — `schedule view/set/clear` (shared validation), `materializations list` (per-package + environment-scoped, Trigger column), trigger a run, inspect a run's manifest/status.
- **Agent skills** — cost/performance optimization (persist candidates, schedule/scope changes) from the materialization history + APIs; depends on the CLI.
- **Full docs pass** — Malloy docs site (getting-started, config/env-var reference, publish-gate rules, fire semantics) + repo README operator depth (including a standalone-vs-hosted divergence write-up); depends on how the gap-closers land.
- **Orchestrated worker `trigger`** — a control-plane-scheduled worker run still records `Manual`; carrying the trigger through the orchestrated `buildInstructions` path is a control-plane-side change.


kylenesbit:
This is an awesome feature — it completes the local half of the story we care most about: **develop and test your materializations locally on the open-source publisher, then publish to Credible for production serving, with the same config behaving the same way.** The core of that holds up really well in this PR: the manifest surface is identical (`scope` + `materialization.schedule`), the PATCH path enforces the same rules a publish enforces via the same `persistencePolicyWarnings()`, fires are `forceRefresh` (matching "a cron re-fires regardless"), a failed fire waits a full occurrence, and an in-flight build coalesces the occurrence. Nice work threading `trigger` through the build commit too.

Reviewing against `persistence.md` (§3.1, §9.2, §9.4), there are a handful of places where the standalone behavior diverges from the Credible/control-plane behavior. **Please close these as best you can — it's okay if some can't be fully closed**, but each one should end either fixed or explicitly documented as a divergence. My take on which side each fix belongs on:

**1. Missed occurrences are skipped on restart (fix in the publisher).**
The settled cron contract (§9.4) is "recovery from an outage **fires once and jumps forward**" — the control plane persists `next_fire_at`, so a due stream fires on the first sweep after downtime. The standalone scheduler arms in memory and re-arms from `nextAfter(now)` on restart, so a missed occurrence is skipped entirely. On a dev machine or a nightly-restarted container, a daily cron can systematically under-fire relative to production. The publisher already has the DuckDB repository — persisting `nextFireAt` per package (or deriving "did I miss one?" from the newest `trigger=SCHEDULER` materialization row) would match the contract with modest code.

**2. Cron grammar parity + publish-gate validation (fix in the publisher).**
`cron-parser` accepts extensions (`L`, `#`) that the control plane's `cron-utils` UNIX parser rejects — and on the Credible side an invalid cron is warn-only at ingest, so `"0 6 L * *"` validates locally, fires locally, and after publish **silently never arms in production**. Two composable fixes: (a) restrict the publisher's `CronEvaluator` to the plain UNIX grammar (reject `L`/`#`), and (b) move the cron-validity check from the PATCH path into `persistencePolicyWarnings()` itself, so publish, PATCH, load-warn, and the scheduler all enforce the same rule (right now a publish accepts a garbage cron that PATCH would reject).

**3. Asymmetric scope flip in the UI (fix in the publisher/SDK).**
Setting a schedule auto-sets `scope: version` (necessary, and the dialog says so — good), but clearing the schedule leaves `scope: version` behind. In Credible that's a real cost posture change (no cross-version artifact reuse), with no way back in this UI. Suggest clear-schedule offer to restore `scope: package`, or make the scope an explicit choice rather than a side effect.

**4. Schedule edits vs Credible's write-once ingest (docs/UI copy, no code change).**
Standalone, "Edit schedule" is immediate and durable. In Credible the schedule is ingested **write-once per version** (§9.2), so the same edit takes effect on the **next publish**. The design composes correctly (the PATCH writes `publisher.json`, so the edit rides the next publish) — just add a line to the `ScheduleCard`/dialog: "takes effect in Credible on your next publish."

**5. Scheduled fires are tables-only standalone (inherent — document it).**
In Credible a scheduled refresh is a full two-phase run: tables, then indexed dimensions. The OSS publisher has no index pipeline, so `#(index)` behavior (index-from-table sequencing, index refresh cadence) can't be exercised locally. Nothing to fix, but the docs should say so plainly so nobody is surprised after publish.

**6. Orchestrated worker rows always read "Manual" (optional, Credible-side later).**
`trigger` is deliberately not client-settable, so a CP-scheduled fire produces a worker-side record labeled Manual. Fine for now (orchestrated deployments don't live in the publisher UI); if worker rows are ever surfaced, we'd let the orchestrated `buildInstructions` path carry the trigger — that's a Credible-side change, not this PR.

Two asks beyond the gap-closing:

**CLI support.** What's the plan for the CLI here? For local devex this matters a lot — at minimum: view/set/clear the schedule (with the same validation), list materializations (per-package and the new environment-scoped view, including the Trigger column), trigger a run, and inspect a run's manifest/status. A modeler iterating locally shouldn't need the UI open to manage their rebuild cadence. If it's follow-up work, let's track it — I'd consider it part of this feature being "done."

**Agent skills for cost/performance optimization.** Once the CLI is there, please create agent skills that leverage the CLI + publisher APIs to help folks optimize their models for cost and performance: e.g., look at what's persisted vs not, run timings and sources built/reused from the materialization history, and recommend `#@ persist` candidates, schedule/scope changes, or removals. This is the local, recommendations-only version of where the platform is headed (usage-driven optimization), and the publisher now has all the raw signals a skill needs.

**Docs.** Once the gaps above are resolved and local devex is great (CLI included), let's do a proper documentation pass — both the Malloy docs site (getting-started guide, config/env-var reference, the publish-gate rules stated as rules, fire semantics) and this repo's README/docs (operator depth, the environment-scoped view, and an honest list of whatever local-vs-Credible divergences remain). We're tracking the full docs checklist on our side; the content depends on how items 1–2 land, so docs come last.

None of this blocks the direction — the config-parity core, which is the hard part, is right. Thanks Girish, this is great work.

kylenesbit:
One more ask before this merges: **please validate that nothing here changes behavior for an orchestrated (control-plane-driven) deployment.** I did a pass from the Credible side and I believe the PR is safe as written, but a few of the safeguards it relies on are subtle enough that I'd like them locked in deliberately rather than by accident. Top risks, in order:

**1. The PATCH policy gate must not fire on the orchestrator's manifest-rebind PATCH.** This is the highest-risk change in the PR. After every orchestrated build, the control plane re-PATCHes the package with a body containing only `name`/`location`/`manifestLocation` — on the finalize path a rejection **fails the run**. Safety currently hangs on the `editingPolicy` check (`body.scope !== undefined || body.materialization !== undefined`) staying false for that body, which also depends on the orchestrator's client omitting unset fields rather than sending `"scope": null`. Two asks: (a) confirm a body that carries an explicit `scope: null` / `materialization: null` behaves sanely (it currently counts as "editing policy" — is a null-wipe intended or should null be treated as absent?), and (b) add an integration test asserting a `manifestLocation`-only PATCH on a package with a **pre-existing load-tolerated policy warning** succeeds and preserves `scope`/`materialization` unchanged. That test is what protects the orchestrated rebind path from a future refactor of this block.

**2. Keep both scheduler guards, and note the serve-live gap.** The scheduler is protected by two independent guards: the opt-in env var (default off) and the `manifestLocation` skip. Note that a control-plane-loaded package *serving live* has `manifestLocation` **null**, so for those packages only the env-var guard applies. That's fine today, but please (a) keep both guards (don't ever collapse them into just the manifestLocation check), and (b) document in the env-var description that the flag must never be set on an orchestrated worker.

**3. No new top-level response fields — keep it that way.** `trigger` lives inside `metadata` (a free-form object), which is why an orchestrated client with strict model deserialization is unaffected. If a follow-up ever promotes `trigger` (or anything else) to a top-level `Materialization` field, that becomes an API-doc/schema change orchestrated clients need to pick up first — worth a comment in the code or api-doc so it isn't done casually.

**4. Create-body whitelist stays intact.** `trigger` being service-level-only (not client-settable) is correct — please make sure `validateCreateBody` continues to reject/ignore it so an API caller can't forge a `SCHEDULER` run, and that the orchestrated create body (`buildInstructions`/`referenceManifest`/`strictUpstreams`) is untouched by any of the new validation.

**5. Per-package list semantics unchanged.** The always-applied `LIMIT` (default 500) applies only to the new environment-scoped query, and the per-package route only gained safer param parsing — confirming that stays true, since existing callers of the per-package list may rely on unbounded history.

If you can add the test in (1) and the doc lines in (2)–(3), I'm comfortable this is non-breaking for orchestrated deployments.

girishjeswani:
Thanks for both passes -- really helpful framing. Pushed a round addressing everything:

Config-parity: (1) restart recovery now re-anchors from the newest SCHEDULER run and fires one catch-up then jumps forward; (2) CronEvaluator + client preview reject L/W/#/?, and cron validity moved into persistencePolicyWarnings() so publish/PATCH/load/scheduler all enforce it (publish no longer accepts a cron PATCH would reject); (3) scope is now an explicit package|version control, so clearing a schedule no longer strands scope: version; (4) card + dialog note the next-publish timing; (5) tables-only + all divergences documented in docs/materialization-scheduling.md.

Orchestrated safety: (1) PATCH treats scope:null/materialization:null as absent (preserve, not wipe) -- new integration test drives a rebind PATCH (manifestLocation-only and explicit-null) on a package with a pre-existing load-tolerated warning and asserts scope/materialization survive; (2) both guards documented as independent + load-bearing, env-var doc warns against orchestrated workers, serve-live gap noted; (3) api-doc notes trigger stays in metadata; (4) comment + unit test that a client can't forge a SCHEDULER run; (5) default LIMIT is env-scoped only, per-package stays unbounded (comment + test).

CLI, agent skills, and the full docs-site pass are tracked as follow-ups (listed in the PR description); #6 (orchestrated worker trigger) is a control-plane-side change.

api-doc.yaml
```diff
@@ -2105,6 +2105,44 @@ paths:
         "503":
           $ref: "#/components/responses/ServiceUnavailable"
 
+  /environments/{environmentName}/materializations:
```
kylenesbit: Can we move this endpoint under packages?

kylenesbit:
Did another pass over the current head (`f816c67`), checking each item from my two earlier reviews against the code. Almost everything landed cleanly — nice work. One item went missing, and I have one more ask on testing before this merges.

## Status of the earlier asks

**Config-parity:**

1. **Restart recovery — done.** First arm anchors to the newest recorded `SCHEDULER` fire (`getLatestScheduledFireAt`), fires exactly one catch-up when an occurrence was missed, then jumps forward; a never-fired schedule arms from `now`; a lookup failure degrades to `now`. The residual divergence (a schedule set while the scheduler was disabled isn't caught up on first enable) is documented in the `arm()` docblock. This matches the contract.
2. **Cron grammar + publish gate — done.** `isPlainUnixGrammar` rejects `L`/`W`/`#`/`?` while keeping month/day names, and cron validity is now Rule 4 inside `persistencePolicyWarnings()`, so publish, PATCH, load-warn, and the scheduler enforce the identical rule. The client-side preview mirrors it.
3. **Scope flip — done.** The explicit `package|version` toggle on the schedule card, locked while a schedule is set and unlocked on clear, is exactly what I was after.
4. **Next-publish copy — done** on both the card and the dialog.
5. **Tables-only divergence — missing.** Your reply says this is "documented in docs/materialization-scheduling.md", but that file isn't in the diff — the only non-package files changed are `api-doc.yaml` and `bun.lock`, and I couldn't find a tables-only / index-pipeline note in the scheduler or config comments either. I suspect it was lost in the force-push. Please restore it (a short docs file or a comment block on the scheduler's `fire()` is fine): a scheduled fire standalone is tables-only; in a hosted deployment it's a two-phase run (tables, then indexed dimensions), so index behavior can't be exercised locally.

**Orchestrated safety: all five landed.** Verified the null-as-absent PATCH handling in `environment.ts` plus the `orchestrated_rebind` integration test on a fixture carrying a load-tolerated warning (both the `manifestLocation`-only and explicit-null bodies); the two-guards docblock and the "never set this on an orchestrated worker" env-var warning; the `metadata.trigger` breaking-change note in `api-doc.yaml`; the controller test that a client-supplied `trigger` can't forge a `SCHEDULER` run; and the repo tests locking the default LIMIT to the env-scoped query with the per-package list unbounded.

Also, my inline question about moving `/environments/{env}/materializations` under the packages hierarchy is still open — let's resolve that thread one way or the other before merge.

## One more ask: testing depth

The scheduler and its state machine are the most complex new logic in the PR, and the coverage is good but has a few structural holes. What exists today: solid unit specs (12 scheduler cases, cron evaluator, service, controller, repo bounds) and an integration suite that drives a hand-constructed scheduler's `tick()` against the real store with a real DuckDB build. What I'd like added:

1. **A true e2e test through the real wiring, like the lifecycle/freshness suites.** The current scheduler integration test constructs `MaterializationScheduler` directly and calls `tick()` by hand — deterministic and worth keeping, but it bypasses `getMaterializationSchedulerConfig` and the `server.ts` startup wiring entirely, so the env-var path (`PUBLISHER_LOCAL_MATERIALIZATION_SCHEDULER` + interval config + start/unref) that operators actually use has zero test coverage. One test that boots the server with the flag set and a short tick interval and observes a scheduled build fire through the real timer would close that.
2. **Restart recovery is only tested against a stub.** The unit spec stubs `getLatestScheduledFireAt`, and the repo spec doesn't cover it — so the `json_extract_string(metadata, '$.trigger')` query and the actual restart sequence are untested end to end. Please add: (a) repo-level tests for `getLatestScheduledFireAt` (returns the newest `SCHEDULER` row, ignores `ON_DEMAND` rows and rows with no trigger, null when none), and (b) an integration test that does a real `SCHEDULER` fire, constructs a *second* scheduler instance (simulated restart), and asserts exactly one catch-up when an occurrence was missed — and no double-fire when one wasn't.
3. **Garbage collection.** `deleteMaterialization` + `dropTables` has good unit coverage and a happy-path e2e, but please cover: the drop-failure path (record still deleted, failure metered), staging-table cleanup when a build fails or is cancelled mid-run (are `_staging` tables orphaned?), and what happens to physical tables on environment/package delete (`deleteByEnvironmentId` removes records — do tables leak?). If tables intentionally leak on env delete, a test asserting that documents the decision.
4. **State-machine transitions at integration level.** The unit specs cover the per-tick guards well; the transitions worth an integration test each: a cron edit between ticks re-arms from `now` (no stale fire from the old cadence), a package unload/reload prunes and re-anchors without a spurious fire, a `FAILED` fire retries at the next cron occurrence (not every tick), and `maxFiresPerTick` carryover — capped packages fire on the following tick.

None of this changes the design — the config-parity core is right and the orchestrated-safety work is exactly what I asked for. Restore item 5, close out the endpoint-location thread, and land the testing above (or split 3–4 into an immediate fast-follow if you'd rather keep this PR moving), and I'm ready to approve.

girishjeswani:
Pushed 648a1995 covering the round-3 items, and merged latest main (fd5fab94) to clear the branch conflict, mergeable now.

Endpoint location (your inline thread): nested it -- GET /environments/{env}/packages/materializations, the collection-level sibling of the per-package list. It registers ahead of /packages/:packageName so the literal segment wins the match, which reserves materializations as a package name (noted in the route comment + api-doc description). TS + Python client regen is clean, and the env-scoped integration test drives the new path. Resolving that thread.

Item 5 (tables-only) -- restored as a comment on the scheduler's fire() rather than reintroducing the doc: a standalone fire is single-phase (tables only), while a hosted deployment runs the two-phase job (tables, then indexed dimensions), so index behavior isn't exercisable locally.

Testing depth 1–2 -- landed:

Real-wiring e2e: extracted startMaterializationSchedulerFromEnv() so the env-var → getMaterializationSchedulerConfig → start()/unref() → timer path is exercised; the test boots the scheduler from the flag + a 1s interval and observes a build fire through the real timer.
getLatestScheduledFireAt repo tests against a real in-memory DuckDB: newest SCHEDULER row, ignores ON_DEMAND/trigger-less rows, package-scoped, null when none.
Restart-recovery e2e: a real SCHEDULER fire, then a second scheduler instance (simulated restart) re-anchors from the stored row, exactly one catch-up when an occurrence was missed, and no double-fire when it wasn't.
Testing depth 3–4 (GC + integration state-machine transitions), taking you up on the split to keep this moving; tracked in #895. Note one thing that surfaced while scoping it: deleteByEnvironmentId/deleteByPackage remove the records but don't DROP the physical tables, so materialized tables currently leak on env/package delete -- captured in #895 as a decision (document vs. clean up) rather than silently.

Phase 2 follow-ups: CLI (schedule view/set/clear, materializations list, trigger/inspect), the cost/perf agent skills that build on it, and the full docs pass (Malloy site + repo README, incl. the standalone-vs-hosted divergences), tracked in #896. Docs come last, since they depend on how the CLI and remaining gap-closers land.

Local: server unit 1159/0, materialization integration 34/0, typecheck (server + sdk + app) clean.

kylenesbit (approved):
Approving — every item from my two earlier reviews is now closed against head 648a199 (+ the main merge), and CI is green.

**Round-3 asks, all landed:**
- **Tables-only divergence (item 5)** — restored as a docblock on the scheduler's `fire()`: a standalone fire is single-phase (tables only); a hosted deployment runs the two-phase (tables → indexed dimensions) job, so index behavior isn't exercisable locally.
- **Endpoint location** — nested to `GET /environments/{env}/packages/materializations`, registered ahead of `/packages/:packageName` (literal segment wins; `materializations` reserved as a package name). api-doc + Python client in sync, inline thread resolved.
- **Testing depth 1–2** — real-wiring e2e boots through the extracted `startMaterializationSchedulerFromEnv()` (env var → config → start/unref → real timer) and observes a build fire; repo-level `getLatestScheduledFireAt` tests against real DuckDB (newest SCHEDULER row, ignores ON_DEMAND/trigger-less, package-scoped, null when none); restart-recovery e2e with a second scheduler instance catching up exactly one missed occurrence and no double-fire otherwise.

**Confirmed still intact** from the earlier rounds: restart-recovery anchor, plain-UNIX cron grammar as Rule 4 in `persistencePolicyWarnings()`, the two independent scheduler guards + serve-live-gap docblock, and the `metadata.trigger` breaking-change note.

Splitting GC + state-machine-transition coverage to #895 and CLI/docs to #896 is the right call. One thing worth keeping visible from the #895 scoping: `deleteByEnvironmentId`/`deleteByPackage` remove records but don't DROP the physical tables, so materialized tables leak on env/package delete — good that it's tracked as an explicit decision rather than silent.

Nice work threading all of this through. LGTM.

PR #890, opened 2026-07-14
fix(server): a failed reload must not destroy a serving package

mlennie:
> Follow-up to #887, which is now merged. Rebased onto `main`, so this diff is only its own commits.

## What and why

`Package.create`'s failure cleanup was written for a fresh install, where a half-built tree is Publisher's to remove. A reload runs that same code against a directory that already exists and is already serving. So saving a model that does not compile and reloading deleted the package directory and evicted the last-good compiled model, taking the package offline until it was re-provisioned.

That is reachable from `GET /environments/:env/packages/:pkg?reload=true` and from the `malloy_reloadPackage` tool that #887 added, by anyone who saves a broken model and reloads. That is the ordinary authoring mistake, not an exotic one.

I verified it by doing it rather than by reading the code: a reload of a deliberately broken model removed `publisher_data/malloy-samples/ecommerce`, and a second reload could not recover it.

## The fix

Directory cleanup is now opt-in, and only `installPackage` asks for it, because it staged the tree itself and rolls the previous one back on failure. The other two callers load directories that already existed and no longer delete them: the reload path, where the directory is the live source and the package was serving a moment ago, and `addPackage`, which registers a pre-existing user directory (the same defect class, one caller over).

A failed reload also keeps serving the last-good compiled model instead of evicting it, and surfaces the compile error to the caller. That applies to both reload paths:

- **In-place** (`_loadOrGetPackageLocked`), which is the path every package from `publisher.config.json` takes.
- **Reinstall** (`installPackage`'s rollback), for a package whose stored metadata carries an install location. This one previously dropped the package from `packageStatuses` while `packages` kept it, which stranded a ghost: it answered `getPackage` but vanished from `listPackages` and from `malloy_getContext`, permanently, because the cached fast path never restored the status.

Keeping the previous model served is only honest when the rollback actually put its tree back, so the status is gated on that. If the rename-back threw, or there was no old tree to retire in the first place, the cached package no longer matches disk and is evicted instead, with its connections retired through the same drain every other eviction here uses. Dropping it from `packages` without retiring would orphan its `MalloyConfig` and native DuckDB handle for the life of the process, since `closeAllConnections` only releases what is still in the map.

## Reload mode

With the in-place path safe, re-fetching is the only path left that can still overwrite on-disk edits: a package whose stored metadata carries an install location is re-downloaded over whatever is there. The payload said `"reloaded"` for both, so a caller could not tell whether its saved edit had been recompiled or overwritten.

`PackageController.reloadPackage` now returns the metadata plus a `mode` of `"in-place"` or `"reinstalled"`, and `malloy_reloadPackage` surfaces it. `getPackage` delegates to it and still returns bare metadata, so the REST contract is unchanged.

The re-fetch branch also had no test at all, which is a poor place to have none, since it is the one branch that overwrites data. Both routings are now covered, plus a check that the payload never echoes the package `location` (that allowlist is what keeps a deployment detail out of an agent-facing response; REST spreads the whole object).

## Diagnosability

Two things that were observed misleading a real agent:

- Both MCP tools funneled every caught error through the Malloy error helper, so an unknown package told the agent to check its Malloy syntax and a back-pressure rejection sent it hunting for a typo. `classifyToolError` homes each error class instead.
- A name the model does not define reads as a typo, and the suggestions said so. But the same error is what an author gets after saving a new view, because the served model is the one compiled at boot. It now points at `malloy_reloadPackage` rather than at a typo that is not there.

## Docs

#887 documents the hazard this removes, so its docs move with the behavior: a failed reload is now safe, and `malloy_compile` is framed as faster feedback rather than a data-safety gate.

Two further doc fixes came out of live testing rather than review:

- An agent that starts the server mid-session cannot get the `malloy_*` tools: its tool list was fixed when it connected, and only the user can reconnect. The guidance said "then reconnect", which an agent cannot act on, so in a real session it diagnosed the cause correctly and then quietly ran everything on curl against REST, and the user never learned the session had no tools. It is now told what it cannot do, what to ask for, and not to route around it.
- "Reconnect with `/mcp`" pointed users at the wrong button. `/mcp` opens a server manager; you select the server, and the panel offers `Authenticate` first while reporting `Auth: not authenticated` for an endpoint that has no auth. `Reconnect` is the option that works, and it does re-attach tools mid-session. Also, first boot was documented as 30 to 60 seconds and really takes about 13.

## Verification

Gate green: typecheck, lint, prettier, unit 1110 passing, integration 195 passing.

Verified live against a running server over MCP, not only in tests:

- The safety fix, with a clean before and after: the identical scenario that deleted the package now leaves the directory intact, the edit on disk, the package listed over REST and discoverable via `getContext`, and the last-good model still serving.
- Recovery: fixing the model and reloading returns the package to health, reporting `mode: "in-place"`.
- Five consecutive fail-then-fix reload cycles, each failing and recovering cleanly with the server still serving.
- Error classification: an unknown package returns `Resource not found: malloy-samples/nope` instead of Malloy syntax advice.
- A full agent session on a fresh clone: it stops and asks the user to reconnect instead of silently falling back to curl, and after reconnecting makes only tool calls.

Every regression test here was checked against its own bug by reverting the fix and confirming the test fails. The first version of the reload test asserted only `toBeDefined()` on the status, which a package stranded at `LOADING` would have satisfied while silently disappearing from `listPackages`, so it now asserts the real invariant: the same Package instance is still served, the status is SERVING, the package is still listed, and the user's model is still on disk.

## Follow-ups (not in this PR)

- `Package.create` leaks a per-package `MalloyConfig`, and with it a native DuckDB handle, on every failed load: it is built inside the try and the catch rethrows without releasing it. Pre-existing, but this PR changes its exposure class, since a failed reload is now survivable and retryable. The obvious fix is unsafe: that config is shared by both legs of a `Promise.all` that rejects early, so shutting it down in the catch can pull connections out from under a still-running worker compile. It needs the legs settled first, or the release deferred through the same drain used for evictions. Worth its own PR; `installPackage`'s catch never releases a validator-rejected package either.
- `classifyToolError`'s not-found and back-pressure branches are pinned here, but the connection error classes (and `AccessDeniedError`) still have no branch, so an expired-credentials reload gets Malloy syntax advice.
- Its `ServiceUnavailableError` advice assumes back-pressure, but the worker pool throws the same class for crashes and RPC timeouts, where "raise the limit" is wrong.
- The `executeQuery` reload hint matches upstream compiler prose and has no spec, so a rewording upstream would silently disable it.
- The in-place and reinstall paths treat a vanished tree differently (one keeps serving from the in-memory model, the other evicts). Degraded rather than corrupt, but it deserves a comment explaining why they differ, or a narrower catch.


jswir:
## Pre-submission review

The core fix is correct and well-tested. Traced every failure path in `_loadOrGetPackageLocked` and the reinstall/rollback branch: a failed reload always leaves the package either serving the last-good compiled model or cleanly evicted, with `packages`/`packageStatuses` in agreement and connections drained. The ghost-package bug is genuinely closed. `package_reload_safety.spec.ts` asserts the real invariant (same `Package` instance still served, `SERVING`, still listed, model still on disk) against a real `Environment`/`Package.create` rather than a mock. No blockers.

The remaining items are all in the secondary error-hint cleanup, not the safety fix — they affect whether an agent gets pointed at the real problem, not correctness or data safety.

### Worth addressing

**1. `execute_query` doesn't route through `classifyToolError` — the busiest tool still misclassifies back-pressure.** [`execute_query_tool.ts:282`](https://github.com/malloydata/publisher/blob/94cc2b3f8bc251d1cf9f0dce278789da4dad3084/packages/server/src/mcp/tools/execute_query_tool.ts#L276-L299)
`tryAcquireQuerySlot("mcp:executeQuery")` at `:158` is inside the try, so at the concurrency cap it throws `ServiceUnavailableError` into the `:282` catch, which funnels it through `getMalloyErrorDetails`. That doesn't match, so an at-capacity 503 tells the agent to "verify your Malloy syntax / check for typos" instead of "retry." The PR built `classifyToolError` for exactly this and wired it into compile + reload but not the highest-traffic tool, whose catch this PR already edits. Route this catch through `classifyToolError`, then append the undefined-name reload hint. (Not-found/memory errors are fine — they're homed earlier in `getModelForQuery`; this is specifically the concurrency-slot path.)

**2. `classifyToolError`'s catch-all sends internal errors to Malloy advice, and the docstring says the opposite.** [`handler_utils.ts:54` + docstring `:22-26`](https://github.com/malloydata/publisher/blob/94cc2b3f8bc251d1cf9f0dce278789da4dad3084/packages/server/src/mcp/handler_utils.ts#L22-L54)
The `else` routes every unclassified error (filesystem `EACCES`, `TypeError`, worker crash) to `getMalloyErrorDetails` → "check your Malloy syntax." The pre-existing `getModelForQuery` correctly routes its unexpected `else` to `getInternalError` ([`:130`](https://github.com/malloydata/publisher/blob/94cc2b3f8bc251d1cf9f0dce278789da4dad3084/packages/server/src/mcp/handler_utils.ts#L130)) — mirror that. And the docstring's "Only errors that really are about the Malloy fall through to the Malloy advice" is false given the catch-all; fix the comment either way.

**3. `compile_tool` error test uses the weak assertion pattern flagged on #887.** [`compile_tool.spec.ts:129`](https://github.com/malloydata/publisher/blob/94cc2b3f8bc251d1cf9f0dce278789da4dad3084/packages/server/src/mcp/tools/compile_tool.spec.ts#L120-L130)
It throws a generic `Error` and asserts only `expect(parsed.error).toBeDefined()`, so compile's not-found/back-pressure branches through `classifyToolError` are never exercised — it'd pass even if compile handed back wrong advice. The reload spec pins the text ("Resource not found"/"Retry", not "Malloy file"); do the same here (throw a typed `PackageNotFoundError`/`ServiceUnavailableError`, assert the classified message).

### Nits / follow-ups

- **Duplicated, divergent back-pressure text.** `classifyToolError` (`handler_utils.ts:48-51`) and `getModelForQuery` (`:123-125`) give different remediation wording for the same `ServiceUnavailableError`. Extract one shared constant — same drift `RELOAD_FAILURE_IS_SAFE` just fixed.
- **`ServiceUnavailableError` advice is wrong for worker crashes / RPC timeouts** (same class, "raise the limit" doesn't apply). Already noted as a follow-up in the description — flagging for tracking.
- **Comment altitude:** `reload_package_tool.ts:28` — "...resynchronized twice, and disagreed both times" is change-history; move to the commit message, keep the single-source-of-truth rationale.
- **`package.ts:300`** — `fs.lstat` is computed on every failed load but only used in the `cleanupDirectoryOnFailure` branch; move it inside that branch.

None of these block. #1 is the one I'd fold in now (small, high-traffic path); 2–3 are quick and worth doing; the rest are fine as follow-ups.

jswir (approved):

mlennie:
@jswir Thanks, this was a good catch on the classifier. All three "worth addressing" items are in, plus three of the four nits. One of them turned out to be a trap, which is the interesting part.

## 1. executeQuery routes through classifyToolError

Done. You were right that this was the tell: we built the thing for exactly this case and then wired it into compile and reload but not the tool whose catch the PR was already editing.

## 2. The catch-all, and why this is not a literal mirror of getModelForQuery

The catch-all is fixed and the docstring no longer lies, but mirroring `getModelForQuery` exactly would have regressed executeQuery, so I did something slightly different and want to flag it explicitly.

`getModelForQuery`'s list is right *for `getModelForQuery`*: it only wraps model setup, where a raw engine error never escapes, because `model.ts:576` wraps it in `ModelCompilationError`. executeQuery's catch is a different population. I probed it rather than reasoned about it:

```
bad syntax      -> MalloyError: unexpected '@'
undefined name  -> MalloyError: Reference to undefined object 'nonexistent_source'
failed reload   -> ModelCompilationError: Error(s) compiling model:
```

A bad query throws a **raw `MalloyError`**, not `ModelCompilationError`. So applying #1 and #2 together with a literal mirror would have sent every syntax error to `getInternalError` ("an unexpected internal error occurred during executeQuery") on the highest-traffic tool, which is worse than the bug we're fixing. `MalloyError` and `BadRequestError` are therefore homed alongside the compile errors and authorize denials, and everything else goes internal.

I also tried the version that looked more principled: gate the fallback on `getMalloyErrorDetails`' own `refined` flag, so "did it actually recognize this?" decides. That is wrong too, and it is worth knowing why. Its patterns do not match what the shipped engine emits: `syntaxErrorMatch` wants `/no viable alternative at input/` where the real error says `unexpected '@'`, and `invalidRequestMatch` is `/Invalid query request\\. Query OR queryName must be defined/` (double-escaped `\.`, and the real message reads "(Query AND !sourceName) OR (queryName AND sourceName)"). Neither refines on a real error, so a refined-gated fallback would route genuine compile errors to internal. Routing by class is the only safe option, and the comment now says so.

Both branches are mutation-tested: dropping the `MalloyError` branch fails the executeQuery guard, and restoring the catch-all fails the internal test.

## The reroute had no test, which my own review caught late

Worth flagging plainly: my first pass at this added tests for the *classifier* and none for the *wiring*. Reverting the executeQuery reroute (your #1, the actual reported bug) left the whole suite green, because `handler_utils.spec.ts` calls `classifyToolError` directly and only passes `"executeQuery"` as an operation string. Every other MCP tool has a unit spec; the one this change is named after did not.

There is now an `execute_query_tool.spec.ts` that pins the wiring. One subtlety in case it looks over-built: the obvious version of that test has no teeth. `getModelForQuery` calls `environment.assertCanAdmitQuery()` outside the tool's try and already routes that 503 to the Retry advice, so a test tripping the memory-admission path passes with the reroute reverted. The throw has to originate inside the try, so the mock's model loads cleanly and its `getQueryResults` throws. Reverting the reroute now fails 3 tests.

## Two classes I had mis-homed

Enumerating what can actually reach that catch turned up two errors my class list got wrong, both of which the old code got wrong differently:

- **`QueryTimeoutError` / `PayloadTooLargeError`** were falling to the internal branch, which says "Try the request again later." That is the one thing a timed-out or oversized query must not be told: retrying it unchanged fails identically. Both classes already carry their own "refine the query" remediation in the message, so the suggestions now say the part the message leaves out and the internal branch had backwards: that this is not transient. That distinction is the whole reason they are 504/413 rather than the retryable 503, which `QueryTimeoutError`'s own docstring spells out.
- **`NotQueryableError`** was also falling through, so a query-boundary denial reported as "an unexpected internal error occurred ... contact support", i.e. a working security boundary presenting as a server fault. It is a deliberate 404, so it now routes to `getNotFoundError` alongside the other not-found classes. That also drops the `No queryable source "x"` echo, which fits what the class is for.

## 3. compile_tool.spec

Fixed, and pinned the way the reload spec does: typed `PackageNotFoundError` / `ServiceUnavailableError` / `TypeError`, asserting the classified text rather than `toBeDefined()`. There is also a new `handler_utils.spec.ts` covering the classifier's branches directly, since it now has more of them.

## Nits

- **Shared back-pressure constant**: done. It names both the memory governor and the concurrency slots, since `ServiceUnavailableError` covers both and the two copies disagreed on that.
- **Comment altitude**: done, history moved to the commit message.
- **`fs.lstat`**: done, moved inside the cleanup branch.
- **`ServiceUnavailableError` advice for worker crashes / RPC timeouts**: still a follow-up, agreed.
- **Connection classes**: left as a follow-up deliberately. Re-routing them to the Malloy helper does not help, because their messages mostly do not match its connection pattern (`/Cannot connect to database/i`), so an expired-credential reload still gets generic Malloy advice. They need their own remediation text, which felt like its own change.

Gate green: typecheck, lint, prettier, unit 1124 passing (up from 1110), integration 195 passing.

Verified live rather than only in tests, since the reported symptom is a runtime one. I rebuilt the pre-fix code and saturated a server with `PUBLISHER_MAX_CONCURRENT_QUERIES=1`:

- **Before**: an at-capacity 503 returned "Verify the structure and syntax within the Malloy file: malloy-samples/ecommerce/ecommerce.malloy", "Ensure all referenced sources, queries, or views are correctly defined", "Check the database connection configuration".
- **After**: "Retry once the publisher is below its configured memory or concurrency limit.", "If this persists, raise the limit or scale up the pod."

Also confirmed live that a syntax error keeps its Malloy advice, the undefined-name reload hint still fires, and an unknown package still reports "Resource not found".

The two re-homed classes were verified against a real server too, not just in unit tests:

- **`NotQueryableError`**: the boundary only engages when a package declares `explores` (`model.ts`: `if (mode === "all" || !exploresDeclared) return "cleared"`), and no sample does, so I added one to a `publisher_data` copy and reloaded it with `malloy_reloadPackage`. Querying the gated model returned `An unexpected internal error occurred during executeQuery.: No queryable model "brand_synergy.malloy"` + "Try the request again later" before the fix, and `Resource not found: malloy-samples/ecommerce/brand_synergy.malloy` after.
- **`PayloadTooLargeError`**: with `PUBLISHER_MAX_QUERY_ROWS=1`, a real overflow now returns the error's own message (which names the cap) plus "This is not transient. The same query will fail the same way, so change the query rather than retrying it."

One thing that fix only narrows rather than closes, worth its own change: a hidden model and a missing one are still distinguishable by the shape of the message (`Resource not found: env/pkg/model` from this catch, versus `Resource not found: model 'x' in package 'y' for environment 'z'` from `getModelForQuery`). Aligning those identifier formats reshapes every error from that catch, so I left it alone here. Before this PR the gated case announced itself outright, so this is strictly better either way.

One unrelated thing I fixed while in here: this branch and #892 both rewrite the same `AGENTS.md` first-boot line, so whichever landed second would have conflicted. This branch now carries #892's wording, which is the measured one, so both sides make the identical change.

PR #891, opened 2026-07-15
feat(sdk): forward Malloy given: values through Publisher.query → data API

Sha-Bang:
## What

Lets in-package HTML data apps pass Malloy `given:` runtime parameters at query time:

```js
Publisher.query("model.malloy", "run: orders -> by_region", { givens: { target_region: "EU" } })
```

The browser SDK previously forwarded only `#(filter)` params (`opts.filterParams`). The **server-side givens path already existed** (`QueryRequest.givens` in the spec, the `/query` route reads `req.body.givens` and threads it to `runnable.getPreparedResult({ givens })`), so this wires the browser SDK and hardens the error handling around givens.

## Why publisher-only (no service change)

- The Credible **router already forwards and hardens** givens (trusted-given injection, reserved-name stripping, fail-closed) and its `QueryRequest` models `givens` from the shared spec — nothing drops it.
- Embedded-in-Credible data apps are a plain same-origin iframe (no `window.Publisher` shim, no transport swap), and the SDK URL is origin-relative, so the identical request body works both standalone on localhost and embedded.

## Design: Malloy is the single given validator

There is **no publisher-side given validation**. Malloy validates supplied givens when it prepares the query and throws a `runtime-given-*` error for an unknown name (with a "did you mean?" hint), a wrong-typed value, or a finalized-override. The publisher just **maps any `runtime-given-*` code to a clean 400** — in `getQueryResults`, `executeNotebookCell`, and the `/compile` `getSQL` path.

Supporting error-handling changes (translation, not validation):
- **`runtime/publisher.js`** — forward `opts.givens` onto the request body (absent → byte-identical request).
- **`service/model.ts`** — `getPreparedResult` runs *inside* the run try/catch, so a bad given at prepare time is a **400, not a 500**.
- **`service/environment.ts`** — the `/compile` `getSQL` path surfaces a bad given (400) instead of silently omitting `sql`.

## Error-status model

| Case | Status |
|---|---|
| Unknown given name, or a bad value, on an **ungated** source | 400 |
| Unknown given name, or a bad value, on an **`#(authorize)`** source | 403 |
| `#(authorize)` denies the caller | 403 |

The gated **403** is intentional: the authorize probe runs before prepare and binds the supplied givens, failing closed on a bad given. This also keeps Malloy's did-you-mean name hint from leaking declared given names to an unauthorized caller. `opts.givens` is safe *parameterization*, not an authorization boundary.

## Tests

- `tests/integration/query_givens` — 7 ungated acceptance cases (retarget, unknown name → 400, value mismatch → 400, compose with filterParams, default, `null` = SQL-NULL override, numeric-string for a `number` given).
- `tests/integration/query_givens/query_givens_authorize` — gated status matrix (unknown name → 403, authorized → 200, deny → 403, bad value → 403).
- `tests/integration/sdk_givens` — evaluates the served `publisher.js` in a `vm` sandbox and asserts givens reach the wire.
- `model.spec.ts` — `runtime-given-finalized` → 400 unit tests on the query and notebook paths.

## Docs

`docs/html-data-apps.md`, `docs/givens.md`, `examples/html-data-app/AGENTS.md`, and the `html-data-app-runtime` skill now document forwarding, the validation/error model, and the gated 403; the filterParams/givens conflation is corrected. `skills_bundle.json` regenerated.

## Verification

- `tsc --noEmit` 0 errors · eslint clean
- Unit: 1126 pass / 0 fail
- Integration: query_givens 7/7, gated authorize 4/4, sdk 1/1, authorize + html_pages no regression


sagarswamirao (approved):
Traced the givens forwarding and the `runtime-given-*` error mapping across the `/query`, notebook-cell, and `/compile` paths. The load-bearing invariant holds: on an `#(authorize)`-gated source the early authorize gate runs before prepare, so a bad given fails closed as a 403 and Malloy's "did you mean?" given-name hint can't leak to an unauthorized caller. The `null`/empty-map wire handling and the prepare-inside-the-try timing both trace clean. Nice test coverage -- the gated-source 403/200 matrix and the vm-evaluated served `publisher.js` are exactly the right things to pin.

Two small hardening notes for a follow-up (non-blocking):

1. **Bad-given handling diverges between `/compile` and `/query` if Malloy ever surfaces a bad given as a bare error without a `runtime-given-*` code.** On `/query` a `MalloyError` is re-thrown and mapped to a 4xx; on the `/compile` getSQL path (`environment.ts`) a `MalloyError` falls through to the outer catch and returns `{ problems }` as a 200. Under Malloy's current documented behavior both paths tag bad givens with `runtime-given-*` and hit the 400 branch, so this doesn't bite today -- just flagging the asymmetry in case the error shape ever shifts.
2. **The `.code.startsWith("runtime-given-")` duck-type is a stringly-typed coupling to Malloy.** If those codes are renamed upstream, bad givens silently revert to 500 (`/query`) or a silently-omitted `sql` (`/compile`), and the integration tests assert on status only so a rename wouldn't be caught. Worth a comment pinning the Malloy-version assumption.

Also worth a quick refresh of the PR description before merge -- it describes an earlier `validateGivens`/`getDeclaredGivens()`/pre-authorize name-check design that isn't in the final diff (Malloy is now the single validator).

PR #775, opened 2026-05-21
feat: complete givens migration (docs + deprecation surface)

Sha-Bang:
Final PR of the [givens-migration-plan](https://github.com/malloydata/publisher/blob/main/docs/givens-migration-plan.md). Combines docs + deprecation messaging (plan PR 5) and end-to-end validation (plan PR 6) into one PR.

## What this ships

- New `docs/givens.md` canonical reference with `#(filter)` → `given:` migration recipe table.
- `docs/filters.md` reduced to a redirect; `docs/ai-agents.md` + `RELEASE_NOTES.md` updated.
- `api-doc.yaml`: `deprecated: true` on `Filter`, `filterParams`, `bypassFilters`, `filter_params`, `bypass_filters`.
- Runtime: `Model` constructor logs once per instance when `#(filter)` annotations are present. `POST /…/query` and `GET /…/notebooks/.../cells/N` emit RFC 8594 `Deprecation` + `Link` response headers when the request actively opts into the deprecated surface (`filterParams` set, or `bypassFilters: true`).
- Customer sample lives in sibling PR [credibledata/malloy-samples#17](https://github.com/credibledata/malloy-samples/pull/17).

## Test plan

- [x] `bun run build` / `bun run lint` clean
- [x] `bun test src/service/filter_integration.spec.ts` — 30/30 pass
- [x] `bun test src/service/model.spec.ts` — 13/13 pass
- [x] Manual smoke against running server — REST query, REST compile, MCP `malloy_executeQuery`, notebook-cell GET all accept `givens` and apply overrides; Deprecation header fires on opt-in and not on no-op opt-out.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #732, opened 2026-05-12
[726] Npx quickstart fixes

mlennie:
## Summary

Issue: https://github.com/malloydata/publisher/issues/726
  
  Addresses the four publisher-side findings from [`publisher-audit-npx.md`](https://docs.google.com/document/d/1Q9Zpq_Mm9-8bI7eCwZ1J3VVWrJ4qdzWqGTXFaqkikXY/edit?usp=sharing) (2026-05-07) — the audit that walked the `npx @malloy-publisher/server` Quick Start path end-to-end and
  documented every rough edge. Branched off `bun-setup-fixes` (the previous audit PR); will rebase onto `main` after that lands.

  Six commits, each scoped to one finding plus a documentation correction and a small post-review polish pass. Independently reviewable / revertable.

  ## Audit findings addressed
  
  | # | Finding | Commit | Status |
  |---|---|---|---|
  | #3 | Empty `attachedDatabases: []` rejected by validation, even though downstream code (`attachOnce`, `testConnectionConfig`) and the per-package sandbox already handle the empty case | `a5a8c9e` | ✅ fixed
   |
  | #5 | `operationalState: "serving"` reported even when env init failed, with no signal for K8s probes / CI / dashboards | `1828ced` | ✅ fixed — new `degraded` enum value + `failedEnvironments` array on
  `/api/v0/status`; readiness probe returns 503 when degraded |
  | #8 | `npx --server_root .` re-clones malloy-samples internally; no way to point at a config elsewhere; first-run requires the user to know they need a config file | `b080a24` | ✅ fixed — new `--config
  <path>` flag; bundled DuckDB-only `default-publisher.config.json` ships inside the package so zero-arg `npx @malloy-publisher/server` works |
  | #9 | ~50 `npm warn deprecated` lines on install (cryptiles, hawk, har-validator, request@2.x, har-validator, q@1.5.1, etc.) | `4ea9918` | ✅ partially fixed — removed unused `trino` direct dep, killing the
  cryptiles/hawk/har-validator/request/translate@0.7.x chain. ~25 upstream-owned warnings remain (documented in `packages/server/README.md`) |

  Audit items #1, #2 belong in `credibledata/malloy-samples` — separate PR, separate repo.

  Audit items #4 (`Node ≥20` engine pin), #6 (MCP :4040 mention), #7 (verify block) already landed in the parent PR (`bun-setup-fixes`).

  ## Heads up — known limitation in the npx install path

[...]

PR #892, opened 2026-07-15
docs: fix the quick start, which warned about a crash that cannot happen

mlennie:
Fixes the first thing a reader sees after the install command. This is the page the blog post's CTA lands on, and `npx` only became real for users last night with 0.0.227, so the lines under it are worth getting right. Six line changes, no new sections.

## Two defects, one paragraph apart

The quick start said to expect a 30-60s wait before `serving`, and immediately below it warned that `npx` exits with `Cannot find module ...duckdb.node`. Together they are worse than separately: a reader waits past the number we gave them, still sees `initializing`, and concludes the warning underneath was right.

## The warning is not stale, it is dead

It describes a module missing from the `duckdb` package. But #818 ("unify on `@duckdb/node-api` and pin the build-time CLI to the engine version", 2026-06-05) dropped `duckdb` entirely. The published `@malloy-publisher/server@0.0.227` that `npx` installs declares `@duckdb/node-api` and no `duckdb`, and the legacy package is absent from the installed tree. So `node_modules/duckdb/lib/binding/duckdb.node` is not missing, it was never there. The failure it warns about cannot occur.

The same commit deleted `install-duckdb-bindings`, which is the script the warning's own workaround tells readers to rely on ("the workspace's `install-duckdb-bindings` script handles the binding install during `bun run build`"). That script exists nowhere in the repo; the only match today is the README line itself. So the warning offered a phantom fix for an unreachable failure.

I added that warning in #732 on 2026-05-12, three weeks before #818 removed the thing it described. It was true when written and quietly rotted. I had planned to move it to Troubleshooting rather than delete it, on the reasoning that failing to reproduce a bug is not proof it never happens. That reasoning is right for absence of evidence, but this is evidence of absence, so it is deleted.

## The 30-60s figure blames the wrong phase, and pins a number to something that will not hold still

Measured over five `npx @malloy-publisher/server@0.0.227` boots on Node 24.15.0:

| Run | npm download | sample clone | total to `serving` |
|---|---|---|---|
| cold npm cache, fresh data | 55s | 15s | **70s** |
| cold npm cache, fresh data | 55s | 33s | **88s** |
| warm npm cache, fresh data | 2s | 16s | **18s** |
| warm npm cache, fresh data | 2s | 12s | **14s** |
| warm, samples on disk | 2s | 1s | **3s** |

Two things fall out. First, the sentence attributes the whole wait to the sample clone and never mentions the npm download, which is 55s of a 70s first run and is the part that actually dominates. Second, the clone it does describe ranged from 12s to 33s across four runs on one machine on one network, so a fixed range was never going to be right. I also saw ~108s end to end the night before, well outside what the README promised.

So the quick start now names both phases, says which one dominates, and sends readers to the existing "Verify it's working" poll instead of timing themselves against a number that is wrong on their connection. I deliberately did not swap in my own numbers: my first draft said the clone takes "10 to 20 seconds", based on three measurements, and the fourth came in at 33s. The replacement number was as fragile as the one it replaced.

## AGENTS.md carried the same claim

`AGENTS.md` also said "about 30 to 60 seconds", and that is the file the agent reads. In a live session an agent quoted our own figure back at the user, so fixing only the README would have left the wrong number in the one surface that actively repeats it. It now separates the two paths, because from a clone the wait really is seconds; it is `npx` that adds the download.

## Three adjacent claims that were wrong the same way

- **Prerequisites** justified the Node floor with "DuckDB postinstall scripts". No DuckDB package has an install hook any more: `@duckdb/node-api` ships prebuilt platform binaries, and it, `@duckdb/node-bindings`, and the platform packages all declare none. The floor itself stands, because the published bin's shebang is `#!/usr/bin/env node`, so only the stale reason is dropped.
- **The `initializing` bullet** is where a reader who polled and saw `initializing` is already standing, so the one genuinely useful hint goes there rather than into a new section. Note what it says: the clone emits a single `Cloning GitHub repository` line and then runs silently for the whole 12-33s, so **a quiet log is the healthy state**. My first draft said the opposite ("if it sits there with no log progress, check your network"), which would have had readers diagnose a working clone as a failure.
- **`publisher.config.json` has shipped four samples** since #775 added `faa-givens-demo`; the README still said three. Note this is not a contradiction with the quick start's "three": `npx` without a local config resolves to a bundled DuckDB-only default that carries exactly `ecommerce`, `imdb`, and `faa`, while a clone loads the four-sample `publisher.config.json`. Happy to split this one out if you would rather keep the PR to the quick start.

## Verification

Full root gate on a fresh worktree off `main`: typecheck, lint, prettier, unit 1098 passing, integration 195 passing. (Root `README.md` and `AGENTS.md` are matched by no test, lint, or prettier rule, so the gate speaks to the tree, not to this text.)

The claims are measured rather than reasoned. Timings come from the boots in the table against the published artifact; the dependency claims were checked against the tree `npx` actually installs rather than this repo's `package.json`; the sample count came from `publisher_data/` on disk and the `/environments` response; and the log-silence claim came from a cold boot log, where the only output during the clone turned out to be my own polling requests.


mlennie:
Closing this in favour of #893, which merged first and rewrote the README wholesale. The npx/DuckDB warning this PR removed is already gone from main, the boot-time wording got rewritten as part of that pass, and the samples note here no longer applies now that the default env is `examples` rather than `malloy-samples`. Nothing left for this to add, so closing rather than rebasing onto a README it no longer matches. Thanks all.

PR #893, opened 2026-07-15
docs: update docs and examples pass

kylenesbit:
## Summary

An end-to-end pass over Publisher's docs, examples, and visual assets so the repo lands as exemplary open-source documentation — for prospective users, analysts, and AI agents alike.

**Examples (the first-open experience)**
- New deterministic data generator (`scripts/generate-example-data.mjs`): storefront grows to 1,000 customers / 200 products / ~25k order lines over three years with growth + seasonality; governed-analytics to ~4,900 orders over two years; html-data-app gets a new 800-account SaaS subscriptions dataset.
- **storefront**: enriched model (margin, seasonality, YoY views, chart tags on views, `business_overview` dashboard), guided-tour notebook, polished HTML data app.
- **governed-analytics**: visual `overview` dashboard so the givens Parameters panel demo feels alive; README aligned to the default `examples` env with the full governance curl matrix.
- **html-data-app**: retargeted from the toy `carriers` dataset to a SaaS subscriptions story (model, Chart.js dashboard, README, in-package AGENTS.md).
- **data-app** (React SDK): migrated legacy `projects` paths to `environments`, fixed stale hardcoded KPIs, renamed the leftover `MalloySamples*` component/widgets/keys to `Storefront*`, real README.

**Consistency & correctness**
- Synced `default-publisher.config.json`, both `publisher.config.example.*.json`, and the Makefile to the shipped `examples` environment (npx, Docker, and clone paths now agree).
- Aligned `--init`/`INITIALIZE_STORAGE`, shutdown-drain, memory-governor, and `--host` docs with actual server behavior; removed a phantom env var (`PG_CONNECT_TIMEOUT_SECONDS`); documented real but missing knobs.
- Compile route now accepts nested model paths (wildcard, matching the query route and the OpenAPI spec).
- Purged all carriers/malloy-samples-era leftovers from docs, skills, and the MCP tool docstring.
- Cross-linked every feature doc with its runnable example and every example README with its docs; house-style banners added where missing.

**Visual assets (Playwright)**
- Re-captured all screenshots against the enriched examples (README hero, storefront dashboard, givens panel, SaaS app, API explorer).
- New screen-recording GIFs embedded where motion sells the feature: givens Parameters panel re-running live (`givens.md`) and HTML-app filtering (`html-data-apps.md`).
- Capture scripts checked in as `docs:screenshots` / `docs:recordings`.

## Test plan

- [x] Server boots against local examples; all three packages compile and render (notebook charts, HTML apps, Parameters panel)
- [x] Governance curl matrix: no-identity → 403, admin → 3 tenants, tenant → 1 tenant, hidden source → 404, REGION given filters
- [x] `config.spec.ts` (76 pass) and compile HTTP integration spec pass; data-app typechecks
- [x] Link/image check: 272 local links across 100 markdown files, 0 broken


mlennie:
@kylenesbit This is a big step up, and the parts I care most about are exactly the ones that were hurting: dropping the fabricated `install-duckdb-bindings` workaround, and replacing the boot-time number with "give the server a moment to report `serving`" rather than a figure that is wrong on someone else's connection. Two things worth a look before it lands, one of which I have measured, plus a heads-up on overlap.

## The npx/DuckDB warning describes a crash that cannot happen (README, "Heads up")

> on some Node 24 setups `npx` skips DuckDB's native binding and the server exits with `Cannot find module ...duckdb.node`. Until that upstream fix lands, run from a clone instead

That failure is unreachable on anything we ship today. #818 ("unify on `@duckdb/node-api`", 2026-06-05) dropped the legacy `duckdb` npm package. The published `@malloy-publisher/server@0.0.227` that `npx` installs declares `@duckdb/node-api` and no `duckdb`, and the legacy package is absent from the installed tree, so `node_modules/duckdb/lib/binding/duckdb.node` is not missing, it was never there. I ran `npx @malloy-publisher/server@0.0.227` five times on Node 24.15.0 and it booted to `serving` every time, loading all three DuckDB packages. (The same commit also deleted the `install-duckdb-bindings` script the old version pointed at, which you have already dropped.)

The reason I would not carry it forward: this version tells readers to **run from a clone instead**, which steers them off `npx`, and `npx` is what the blog post's CTA sends them to. We cut 0.0.227 last night specifically so that CTA would be honest. So the warning now costs us the first-run path it was meant to protect, for a failure mode that was architecturally removed six weeks ago.

I added that warning myself in #732, three weeks before #818 removed the thing it described, so this is my rot, not yours. #892 is the PR that deletes it. Since your rewrite replaces the whole README, the cleanest thing is probably for you to drop those three lines here and I close #892. Happy either way.

## The new default env full-clones this repo on first boot

`default-publisher.config.json` now points at `https://github.com/malloydata/publisher/tree/main/examples/*`. The package loader clones with no depth limit (`environment_store.ts:1723`: `simpleGit().clone(repoUrl, absoluteDirPath, {}, ...)`), so a fresh `npx` run clones **all of `malloydata/publisher`, full history**, to extract three example directories.

For scale, measured today across four cold boots of the current default: the `malloy-samples` clone alone takes 12s to 33s (it varies 3x on one machine), inside a 70s to 108s first run. This repo's pack is ~88 MiB against malloy-samples' ~72 MiB, so first-run gets somewhat slower now, and unlike malloy-samples it will keep growing with our own commit history.

Two options if you think it is worth it: ship `examples/` inside the npm tarball via `files` (no clone at all, which would make first run just the npm download), or pass `--depth: 1` in the clone. The second is a one-line change in the loader and helps every remote package, not just ours. Neither is a blocker, but the first-run number is the thing this PR is trying to make good, so it seemed worth flagging rather than discovering later.

## Heads-up on overlap

@jswir you asked in Slack today (on #892) for the samples to be fixed: `faa2` instead of `faa`, drop `faa-givens-demo`, and add "a package with a data app". This PR arguably answers all three by replacing the default env with local examples including `html-data-app`, so it is worth the two of you reconciling before anyone does samples work. Worth knowing either way: `faa-givens-demo` is our only demo of givens plus row-level ACL, and its newest commit (Jun 30) added the `#(secure)` example that `docs/givens.md` points at three times, so if the default env stops carrying it, that doc link needs somewhere to land.

mlennie:
@kylenesbit You asked me to take it for a spin, so rather than hand you a list I did the pass and pushed it to `docs/exemplary-oss-pass`: 9 commits, fast-forward, nothing of yours rewritten, all DCO-signed. Green at the end: 1126 unit, 207 integration, 53 Playwright, zero fail, and I drove the UI and the data-app by hand. The docs themselves read really well. Everything below is the stuff that only shows up when you actually boot it.

Two things were load-bearing.

**The examples don't load from a clone.** The config points each package at `github.com/malloydata/publisher/tree/main/examples/<pkg>`, but `storefront` and `governed-analytics` don't exist on `main` yet, because this PR is what adds them. So a fresh clone boots with nothing to serve. It's a chicken-and-egg that resolves itself the moment this merges, which is almost certainly why it slipped through. I split the two configs: the clone-run `packages/server/publisher.config.json` now points at local `../../examples/<pkg>`, which works before and after the merge and has the nice side effect that you're serving your working copy rather than whatever's committed on main. The published npx default (`src/default-publisher.config.json`) stays remote on purpose, since npx has no repo to read.

**The data-app's charts render nothing.** This one predates your PR, it's been broken. `@malloydata/render` at 0.0.415 and earlier ships an exports map with no `import` condition, so `import("@malloydata/render")` resolves to the UMD bundle, `MalloyRenderer` comes back undefined, and every chart dies on `t is not a constructor`. The SDK loads render lazily and swallows it with a `console.error`, so it degrades silently and a passing build tells you nothing, which is why nobody caught it. The fix was to move the example onto the current published SDK (0.0.227) and declare the five `@malloydata` peers it asks for, which are exactly the versions that work. Dashboard now renders its line chart, bar charts, tables and scalars. Related: the example was pinned at SDK ^0.0.137, whose resource URIs are `publisher://projects/...`, while its own constants already use `publisher://environments/...`, so it was mismatched with itself. `examples/data-app` isn't a workspace member, so neither the root typecheck nor CI's `--frozen-lockfile` covers it.

The rest: your env swap from `malloy-samples` to `examples` left `packages/app/tests/playwright/helpers/fixtures.ts` still pointing at `DEFAULT_ENV = "malloy-samples"` with `imdb`/`faa`, so the whole Playwright suite was targeting an environment that no longer exists. Retargeted it and the executeQuery integration spec onto storefront. I also corrected a few claims my own config change falsified in your new docs: `development.md` said packages are fetched at runtime and that you get main's version rather than your working copy, `AGENTS.md` said first boot clones them, and `configuration.md` still advertised three MCP tools when the server registers five.

Closing out the two things from my earlier comment.

**The npx/DuckDB warning is gone**, I deleted it in this push rather than ask you to. Zero hits repo-wide now. That was my rot from #732 and it described a crash that's been architecturally impossible since #818. Which leaves the overlap: #892 deletes the same lines, so whichever of us lands second hits a trivial conflict. Say the word and I'll close #892 in favour of this.

**The first-boot clone got better and worse in different places.** From a clone there's now no clone at all, since the config reads `examples/` off disk. But the npx path still full-clones all of `malloydata/publisher`, full history, to extract three directories (`environment_store.ts:1723`, still no depth limit), and npx is exactly where the blog CTA sends people. So the point is narrower than I first wrote it but sharper: npx is now the only path that pays it. Not a blocker for this PR and I haven't touched it here. I'll file `--depth: 1` separately, it's a one-line change in the loader and helps every remote package, not just ours.

@jswir the samples question from Slack is still open and worth reconciling before anyone does samples work: this PR arguably answers all three of your asks (`faa2` over `faa`, dropping `faa-givens-demo`, adding a data-app package) by replacing the default env with local examples including `html-data-app`. One thing to know either way: `faa-givens-demo` is our only demo of givens plus row-level ACL, and its newest commit (Jun 30) added the `#(secure)` example that `docs/givens.md` points at three times, so if the default env stops carrying it, that doc link needs somewhere to land.

Last thing, flagging rather than deciding: my merge of `main` into your branch carries no DCO sign-off. The bot skips merge commits so it should be fine, but it's your branch, squash it however you like.

kylenesbit:
Thanks Monty can you approve and I will merge

mlennie (approved):

mlennie:
@kylenesbit just approved!

PR #894, opened 2026-07-15
skills: neutral, namespaced, verbatim-shareable Malloy skill set

Sha-Bang:
This branch brings the bundled Malloy skills up to date and reshapes the shared set so it stays in sync with the upstream source of truth (ms2data/agent-skills) with no manual branding step — the two copies are byte-identical and copy verbatim in either direction.

**What it does**
- The shared "engine" skills are one neutral body of text: no Credible-platform-specific content, bare tool names (`get_context` / `execute_query` / `search_malloy_docs`) with a per-skill portability note mapping them to this server's `malloy_*` tools.
- **Every skill is namespaced** with a `malloy-` or `credible-` prefix (the root `malloy` index keeps its name as the entry-point router): `gotchas-*` → `malloy-gotchas-*`, `html-data-app*` → `malloy-html-data-app*`, and `analysis-pitfalls` / `analysis-report` / `lookml-review` / `notebook-chat` / `phrase-detection` / `getting-started` → their `malloy-*` names.
- Publisher's own host/router skills (`malloy-modeling`, `malloy-publish`, `malloy-document`, `malloy-getting-started`, the `malloy` index) keep the `malloy_*` tool names.
- Adds `malloy-analysis` as the neutral analysis workflow.
- `skills/README.md` documents the shared-set contract; `AGENTS.md` links it.
- `.gitignore` blocks stray `credible-*` skills/rules from this open-source tree.

Companion change in Credible's skill repo: ms2data/agent-skills#64.

### Tested
- Verbatim gate: `diff -rq` across all 23 shared skills vs the agent-skills copy → byte-identical.
- Every skill's `name:` frontmatter matches its directory; no non-prefixed skill dirs remain; `.claude/skills` symlinks resolve.
- Neutrality / cross-reference / tool-name gates pass; `git ls-files | grep credible-` clean.
- codex + cursor review of the PR addressed: host-neutral MCP prerequisites, README contradiction/rule fixes, dropped repo-relative doc paths.


sagarswamirao (approved):

Sha-Bang:
Added a forward-direction note to `malloy-model` § Access Control: `#(authorize)` over caller-asserted givens is a real boundary only behind a trusted tier today, and identity-bound ("secure") givens — set by a trusted tier, not caller-overridable — are the planned path to standalone authz. Steers authors to build on `given:` + `#(authorize)` as the surface that carries forward (per `docs/authorize.md` § Security model).

PR #898, opened 2026-07-17
Report configured packages and environments that failed to load

mlennie:
## What this fixes

A package or environment that fails to load is skipped rather than fatal, so the server reports `operationalState: "serving"` and simply omits it. The only signal is a log line. That has hidden real bugs repeatedly: a server can serve one of three configured packages and still look completely healthy.

## What this adds

An optional `loadErrors` array on `ServerStatus`, listing what did not load and why:

```json
{
  "operationalState": "serving",
  "loadErrors": [
    { "environment": "sales", "message": "Failed to download or mount location: ./sales-models" },
    { "environment": "examples", "package": "storefront", "message": "Package manifest ... does not exist" }
  ]
}
```

An entry with no `package` means the whole environment was skipped, which also takes down its healthy packages. An entry with a `package` means that package alone is missing while its siblings serve.

## Scope: signal only

Nothing changes about what loads or what serves. `operationalState` is deliberately untouched: the control plane maps unrecognised states to HEALTHY, the example compose healthcheck greps the raw `"operationalState":"serving"` substring, and the Playwright global setup gates on it with a 5 minute ceiling, so reporting a new state on partial load would break more than it fixed. Whether an unmountable location *should* take down healthy sibling packages is a real question and a separate change.

The field is collected at read time in `getStatus` from live state, following the existing `throttled` precedent, so there is no fourth stored status to keep in sync. It is omitted entirely when nothing failed, so a healthy server's status body is byte-for-byte what it was before.

## Notes for review

Both boot paths record through one helper. The config path runs on `--init` or an empty database; the database-restore path runs on every restart of a server with a persisted `publisher.db`. Recording in only one of them would drop failures on the path production actually takes, so they share a single funnel rather than two call sites that can drift apart.

An environment enters the live map before its database sync runs, so a throw from that tail would otherwise report a serving environment as skipped and contradict the `environments` list. Only an environment that is genuinely absent is recorded.

Messages pass through the existing `redactPgSecrets` before reaching the response body. Compiling a model resolves the package's connections, so a Postgres or DuckLake ATTACH failure surfaces here and that command embeds the connection string. Worth knowing, and called out in a comment: the helper only covers keyword-form `password=`, which is what `buildPgConnectionString` emits, so a URL-form `connectionString` supplied verbatim in config still carries its credentials through. The fix belongs in the helper rather than at this call site.

## Verification

Full gate green: typecheck, lint, prettier, 1130 unit tests, 207 integration tests. Each fix is pinned by a test that fails without it. Verified live against a running server with a healthy environment, an unmountable environment, and an environment with one bad package: both failure modes reported, siblings still serving, a fixed package cleared by reload, a healthy server omitting the field entirely, and the database-restore path reporting correctly after a restart.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,6 +31,11 @@ Poll until the server reports `serving` rather than assuming a fixed wait. From 
 curl -s http://localhost:4000/api/v0/status | jq .operationalState   # -> "serving"
 ```
 
+`serving` does not mean everything loaded. A package that fails to load is skipped, not fatal, so the
+server serves whatever did load and the package is simply absent. If data you expect is missing, check
+`curl -s http://localhost:4000/api/v0/status | jq .loadErrors`, which is absent when everything loaded
+and otherwise names each environment or package that did not load, and why.
+
 ## 2. Connect your agent
 
 Publisher exposes one MCP endpoint: `http://localhost:4040/mcp` (streamable HTTP, stateless, unauthenticated; put it behind a gateway if you expose it beyond localhost).
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -2453,6 +2453,34 @@ components:
             governor). Already-loaded packages remain serviceable while throttled;
             callers can treat a throttled server as unhealthy for new load.
             Only reported when the memory governor is enabled.
+        loadErrors:
+          type: array
+          readOnly: true
+          description:
+            Configured packages and environments that did not load, and why. A
+            load failure is not fatal to the server, so operationalState stays
+            serving and the affected packages are simply absent from
+            environments; this is the only place that difference is reported, so
+            check it before treating a serving server as complete. Present only
+            when something failed to load, so a healthy server omits the field.
+          items:
+            type: object
+            properties:
+              environment:
+                type: string
+                description:
+                  Name of the configured environment the failure belongs to.
+              package:
+                type: string
+                description:
+                  Name of the configured package that did not load. Absent when
+                  the whole environment was skipped, which is what happens when a
+                  package location cannot be mounted at all; in that case no
+                  package in the environment loads, including ones that would
+                  otherwise have succeeded.
+              message:
+                type: string
+                description: Why it did not load.
         frozenConfig:
           type: boolean
           description:
--- a/docs/api-overview.md
+++ b/docs/api-overview.md
@@ -37,7 +37,7 @@ put the server behind your own gateway before exposing it beyond localhost.
 
 | Method & path | Does |
 | --- | --- |
-| `GET /api/v0/status` | Server lifecycle (`operationalState`). |
+| `GET /api/v0/status` | Server lifecycle (`operationalState`), plus `loadErrors` for anything configured that did not load. |
 | `GET /api/v0/environments` | List environments, each with its packages. |
 | `GET /api/v0/environments/{env}/packages/{pkg}` | Package metadata (models, `explores`, `buildPlan`, …). |
 | `GET  …/packages/{pkg}/models/{path}` | A model's compiled metadata (sources, views, givens). |
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -36,6 +36,37 @@ curl -s http://localhost:4000/api/v0/environments | jq '.[].name'    # → list 
   drops. Only reported when the memory governor is enabled (see
   [configuration.md](configuration.md#operational-tuning-oom-guards)).
 
+### `serving` does not mean everything loaded
+
+A package or environment that fails to load is skipped rather than fatal, so the server still reports
+`serving` and simply serves whatever did load. Check `loadErrors` before you treat a `serving` server
+as complete:
+
+```bash
+curl -s http://localhost:4000/api/v0/status | jq .loadErrors
+```
+
+The field is absent when everything loaded. Otherwise it lists what is missing and why:
+
+```json
+[
+  {
+    "environment": "sales",
+    "message": "Failed to download or mount location: ./sales-models"
+  },
+  {
+    "environment": "examples",
+    "package": "storefront",
+    "message": "Package manifest for /publisher_data/examples/storefront does not exist"
+  }
+]
+```
+
+An entry with no `package` means the whole environment was skipped, which also takes down its healthy
+packages. An entry with a `package` means that package alone is missing and its siblings are serving.
+Either way the fix is usually the `location` in `publisher.config.json`, or a missing
+`publisher.json` in the package directory.
+
 ## Docker
 
 Two ways to run Publisher in Docker: build the image from source, or pull the pre-built image from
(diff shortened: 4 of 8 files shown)


housejester (commented):
Strong PR — the three load-bearing design calls (signal-only, read-time overlay over live state, absent-when-healthy) are each right and each explained honestly in comments, and the tests cover both boot paths including the database-restore one a restarted server actually takes. The findings below are lifecycle edges on the new maps, not design problems.

## Mental-model delta

**In one sentence:** knowledge of configured-but-absent packages moves from ephemeral log lines to a durable, queryable field on `ServerStatus` — "serving" officially stops implying "complete," and `/status` becomes the single place that difference is visible.

|---|---|---|
| Truth about a failed load | log stream only; gone once scrolled | in-memory maps (`failedEnvironments` on the store, `failedPackages` per env), surfaced at `getStatus` read time |
| Failure observability lifetime | observable exactly once (the failing load deletes the `packageStatuses` entry it would be enumerated by) | lasting record, cleared on successful load / publish of the same name |
| `operationalState` semantics | "serving," ambiguously read as "all good" | unchanged by design (existing consumers gate on it; right call) |
| Healthy-server status body | as-is | byte-for-byte identical — field absent, never `[]`; verified in code, not just claimed |

Since the field is optional and additive, JSON consumers that ignore unknown keys (the common case) are unaffected; typed SDK consumers pick it up on their next regen.

## Worth acting on

**[MED] The new signal can report ghosts: both delete paths early-return before touching the failure maps.** A *failed* load is precisely the case where the early return fires. `deletePackage` starts `if (!this.packages.get(packageName)) return;` — a boot-failed package is not in `this.packages`, so deleting it from config is a silent no-op that leaves the `failedPackages` entry behind, and `loadErrors` then reports a package that is **no longer configured**, until restart or a same-name republish. `deleteEnvironment` has the identical shape (`if (!environment) return;`) for a boot-failed environment. This inverts the field's own contract ("*configured* packages and environments that did not load") in exactly the sequence an operator is most likely to run: fail → give up → remove from config. Suggest `failedPackages.delete(packageName)` / `failedEnvironments.delete(environmentName)` ahead of the early returns, plus one test each (fail → delete → `loadErrors` gone) — the clear-on-success sites already show the pattern.

**[MED] The `redactPgSecrets` URL-form gap deserves a firmer commitment now that these strings are HTTP-bound.** The comment flags it honestly (keyword-form `password=` covered; a URL-form `connectionString` supplied verbatim in config passes through). What changes the pricing is this PR itself: it's the change that moves those messages from a log line to an HTTP response body — and the docs now actively tell users to `curl .loadErrors`. Widening the helper feels like a pre-release fast-follow rather than a someday, ideally landed before a release ships this field.

## Confirm

**[LOW] `update()` instance semantics vs. package-level failure carryover.** `addEnvironment`'s existing-env branch and `updateEnvironment` both do `this.environments.set(name, await existingEnvironment.update(...))` without touching the failure maps. The record-time guard means a live environment can't be sitting in `failedEnvironments`, so no contradiction there. But at the package level: if `update()` returns a new `Environment` instance, the old instance's `failedPackages` drops silently — whether a still-broken package resurfaces in `loadErrors` then depends on the new instance re-attempting the load. Can you confirm which way `update()` goes, and that a still-failing package reappears rather than vanishing after an environment update?

## Non-blocking

- **[LOW]** Entries carry no timestamp. Fine for the stated scope, but the first status-polling monitor built on this won't be able to distinguish "failed at boot three days ago" from "failed on the publish that just happened." A `failedAt` per entry is cheap now and annoying to retrofit.
- Things I verified rather than took on faith, all clean: the not-yet-in-`environments` guard genuinely prevents the status contradicting its own `environments` list (and the DB-sync-tail test pins it); the "both boot paths funnel through one helper" claim matches the code; `mockDbEnvironments` is reset in both setup and teardown against the shared-process leak; the stub-the-prototype-before-constructing race note is correct.

**Top three:** 1) clear the failure maps in the delete paths; 2) commit the redaction-helper widening as a pre-release fast-follow; 3) confirm `update()` carryover so a still-broken package can't vanish from `loadErrors`.

mlennie:
Thanks, this was a good catch. All three answered below; the first one was a real bug and is now fixed.

**1. Delete paths leave a ghost entry. Fixed for packages, and the environment half needs a different answer.**

You were right about the package path, and it was worse than "silent no-op": a package that fails to load is evicted from `packages` by the load catch (`environment.ts:940`), so `deletePackage` takes its early return *every time* for exactly the packages that have a `loadErrors` entry. Meanwhile `PackageController.deletePackage` goes on to call `deletePackageFromDatabase` unconditionally, so the package really is gone while `/status` keeps naming it. Fixed by clearing before the early return.

While fixing it I found the same gap on a second path: `_addPackageLocked` was the one success path that set `SERVING` without clearing, so a package that failed at boot, got fixed on disk and was re-added kept its load error for the life of the process. The lazy-load and install paths already cleared. Fixed too, and both fixes ship with a regression test that fails without them.

The **environment** half I deliberately did not mirror, because the two are not symmetric. `EnvironmentStore.deleteEnvironment` early-returns at `environment_store.ts:1104` *before* `deleteEnvironmentFromDatabase`, so unlike the package case nothing is removed from config at all. The environment is still configured, so its `loadErrors` entry is still true, and clearing it would turn a correct signal into a false negative. The real gap there is that you cannot delete a boot-failed environment at all, which I think is a separate bug and a bigger change (it needs to drop the DB row and directory for an environment that never loaded). Happy to file it, or to fold it in here if you would rather see them land together.

**2. Widen `redactPgSecrets` to the URL form. Agreed, and I would like to do it as its own PR off main.**

You are right that this PR moves those strings into a response body the docs tell people to curl, so it should not ship a release without the widening. Two reasons to split it: `pg_helpers.ts` is a shared helper that the user-facing attach path calls (`connection.ts:383/386/392`), which this PR otherwise does not touch, and a credential-redaction change deserves its own reviewable diff rather than riding along in a status-reporting PR. The release window is still open (latest published `@malloy-publisher/server` is 0.0.228, which predates this branch), so it can land before anything ships the field. If you would rather have it here as a merge gate, say so and I will move it in.

**3. Does `failedPackages` survive `Environment.update()`? Yes.**

`update()` ends with `return this` (`environment.ts:248`) and `environment_store.ts:1089` re-stores that same instance, so the map is never rebuilt and the entries persist. No code change needed. I can add a short regression test to pin it if you want it enforced rather than just true.

**On the non-blocking `failedAt`:** happy to add it, but I would rather do it in the follow-up that also settles the two items below, so the schema changes once rather than twice.

**Two follow-ups this review surfaced, both pre-existing and neither introduced here:**

- `deletePackage`'s early return also leaves `packageStatuses` and the on-disk tree behind for a package that failed *before* the lock (the `assertCanAdmitNewPackage` 503 back-pressure path). With pressure cleared the package can then resurrect in `GET /packages` with no DB row.
- The only *writer* of `failedPackages` (the `listPackages` catch at `environment.ts:655`) runs outside the package lock, so a DELETE queued behind an in-flight load can have its clear undone and leave a phantom entry. Narrow window, needs a concurrent DELETE against the runtime `addEnvironment` path, but it is real.

Both want their own tests and neither is a regression from this PR, so I have kept them out. Let me know if you would rather they block.

housejester (approved):

PR #899, opened 2026-07-17
feat(skills): publish the agent skills as their own npm package

mlennie:
## What this solves

Publisher's 28 agent skills reach an npm user as MCP prompts, and only as MCP prompts. Those prompt bodies point at 25 `reference/*.md` files that no npm consumer can open: 19 live `reference/...` pointers name files the host has no way to read. The publish root is `packages/server/`, so `skills/` sits above it and nothing there can reach the tarball. Any host that reads skill files from disk (Claude Code, the Publisher plugin, a scaffolder) gets no files at all.

A skills content edit also has to ride a joint sdk+app+server version bump. 3 of the last 10 commits touching `skills/` were pure content edits whose only change under `packages/server/src` was the regenerated bundle.

## What it does

Adds `@malloy-publisher/skills`: the 28 skills and their `reference/` directories as real files, on their own version line.

- Exports `skillsDir` and `listSkills()`, so a consumer can locate the files without hardcoding a `node_modules` layout. `listSkills()` parses frontmatter from the shipped files, so there is no generated manifest to drift. The `exports` map keeps `./package.json` and `./skills/*` reachable, without which `require.resolve('pkg/package.json')`, the usual way to find a package root, fails outright.
- `skills/` stays at the repo root and `prepack` copies it in. That is not a style preference: a symlink named in `files` packs to `package.json` alone, with exit 0 and no warning, so the copy is load-bearing.
- `check-pack` packs for real and compares the tarball against the source tree, deriving every expectation instead of hardcoding a count. An empty tarball is otherwise completely silent.
- One shared predicate drives the copy, the audit and the tests, so they cannot disagree about what may ship.

## What it does not do

It does not make MCP any fresher. `mcp/server.ts` statically imports the bundle and the bundler inlines it into `dist/server.mjs`, so newer skills still need a server build. This buys the file channel and its own semver, nothing more.

## Also in here

Replaces `skills_bundle.spec.ts`'s "at least 24 skills" floor, which sat 4 behind reality and would have passed a four-skill regression, with a regenerate-and-compare against `skills/`.

## Publishing

Publishing is a separate manual dispatch (`skills-npm.yml`), deliberately outside the sdk/app/server lockstep release, since a skill edit should not need a server release. One thing worth flagging for whoever cuts it: npm configures trusted publishing per package, on the package's own settings page, so the entry cannot exist until the package does. The first publish of this name has to be done by hand by an `@malloy-publisher` owner, with scripts enabled, and the workflow takes over from the next version. The workflow comment spells out the order.

## Verification

- Full gate green: typecheck, lint, prettier, and 1341 tests (1127 unit, 207 integration, 7 skills), 0 failures.
- Live: the built server boots and serves all 28 skills as MCP prompts, with `examples` serving 3/3 packages.
- The packed tarball installs with npm and yields 28 skills plus 25 readable `reference/` files from `node_modules` under Node 24.
- Both new gates were mutation-tested rather than assumed: dropping one skill from the bundle fails the regen-diff (the old floor passed that), and a copy that ships nothing fails the pack audit.


kylenesbit:
Pushed 6949d58, a small fix to two of the new gates.

**The problem.** `skills/credible-*/` is gitignored as a local install target (this PR adds the entry), but with a gitignored `credible-*` skill actually sitting in `skills/`, two of the new tests fail on an untouched checkout:

- `skills_bundle.spec.ts` > "is in sync with skills/" — `buildSkills()` reads every skill directory off disk, so the local credible skill makes the committed bundle look stale.
- `index.spec.ts` > "has no credible-* file committed to skills/" — the check scanned the filesystem, so it fired on the gitignored local state it exists to permit.

There was also a quieter leak behind the first one: regenerating `skills_bundle.json` on a machine with Credible installed would have baked the local skills into the committed bundle, and the sync test would then have blessed the result, since both sides included the stray.

**The fix.**

- `buildSkills()` now skips `credible-*` directories (predicate duplicated from `packages/skills/scripts/exclusions.ts` with a comment pointing there, since the server doesn't depend on that package). This fixes the sync test and closes the regeneration leak in one place.
- The "committed" guard now asks `git ls-files` instead of `readdirSync`, so only a genuinely committed stray trips it.

**Verified in both directions**, in the spirit of the PR's mutation-testing:

- With a gitignored `credible-*` skill present: both suites pass, and bundle regeneration is byte-identical to the committed bundle (still 28 skills).
- With that same skill force-staged: the committed guard still fails and names the stray file.
- Clean tree: both suites, `check-pack` (53 skill files, 58 entries), typecheck, eslint, and prettier all green.

kylenesbit (approved):

PR #900, opened 2026-07-17
fix(examples): vendor the chart library, and show a CSV table

mlennie:
Both example data apps loaded Chart.js from `cdn.jsdelivr.net`. Agent sandboxes and many corporate networks block CDNs, and the failure is easy to miss: the KPI tiles still render, because they need no chart library, while every chart and the data table stay empty. On a network that drops the packets rather than refusing them, which is what a firewall usually does, the page also hangs for the whole connect timeout before it shows you that dead dashboard.

Both examples now ship the library at `public/vendor/chart.umd.js` and load it with a relative path, so they render with no network at all.

## The CSV half

Every example read Parquet, so nothing showed that `duckdb.table()` reads a CSV just as well, and an agent working from these has no reason to know it can skip a conversion step it does not need.

`storefront` now also reads `data/regions.csv`, a 50 row state to region lookup joined from `order_items` through the customer's state, with a `sales_by_region` view that uses it. It sits alongside the Parquet rather than replacing any of it: the point is that one model reads both formats in place. It also shows up in the package's data list next to the Parquet files, which is where someone is most likely to notice it.

## Notes for review

The vendored file is `chart.js@4.5.0`'s `dist/chart.umd.js` (MIT), identical to what jsDelivr was serving, minus its trailing sourceMappingURL line since the `.map` file is not vendored. Chart.js v4 ships its UMD build already minified, so this is the same 208KB the page always downloaded, just from the package itself now.

It is committed once per example rather than shared. Only a package's own `public/` directory is web served, and a symlink pointing out of it is refused by design, so a single shared copy is not possible. Two copies also keep each example copy-pasteable on its own, which is how the docs tell you to use them.

`regions.csv` lives in `data/`, not `public/`. Everything under `public/` is served raw, so a CSV there would be a download that goes around the model's authorize and query controls.

The `malloy-html-data-apps` skill already told agents to vendor their chart library, and then excused the shipped example for loading one from a CDN. The examples now match the skill, and the same rule is in `docs/html-data-apps.md`, where a human reader will actually see it. If you copy the vendor script tag into a page in a subdirectory, note the path is relative to the page, so use `../vendor/...`.

`scripts/generate-example-data.mjs` writes the CSV alongside the Parquet, so the data stays deterministic and regenerable. Re-running the generator produces byte identical files.

## Verified

Both pages render with every off-machine request blocked. The CDN was blocked in the browser's own resolver rather than with a test hook, in both shapes: refused, and blackholed so the connection hangs. The old pages draw zero charts either way, the new ones draw all of them in 18ms.

A fresh clone with `bun run build && bun run start` serves all three packages, and clicking from the home page through to the dashboard works with the CDN blocked. Copying `examples/html-data-app` out to its own directory and running it standalone, the way the quick start describes, also still renders offline.

Full gate green: typecheck, lint, prettier, 1159 unit and 222 integration tests, and the Playwright suite at 53/53.


sagarswamirao (approved):
Reviewed. LGTM, approving.

## Vendoring the chart library is the right call

The `html-data-app` and `storefront` examples are deliberately build-free static HTML served raw by Publisher: no `package.json`, no bundler, no `node_modules` reachable by the browser. A raw HTML page has two options for a third-party lib, a CDN `<script>` tag or a vendored local file, so using the npm package directly is not possible here. The example that does have a build (`data-app`, Vite/React) already consumes its charting libs from npm and is correctly left untouched. Moving these two off the CDN is the correct fix for the sandboxed/corp-network failure described.

## Verified
- `public/vendor/chart.umd.js` is genuine Chart.js 4.5.0 UMD (208KB, MIT header intact). The small line count in the diff stat is just the minified single line plus the comment header.
- `data/regions.csv` is byte-consistent with the generator: codepoint sort order is correct at the tricky boundaries (New York before North Carolina; Washington / West Virginia / Wisconsin / Wyoming).
- No orphan join rows: customer states and `regions.csv` are both derived from the same `STATES` array, so `regions with customers.state` always resolves.
- `skills_bundle.json` is the regenerated bundle tracking the SKILL.md edit, not a hand-edit.
- SKILL.md, `docs/html-data-apps.md`, `connections.md`, `development.md`, `explorer.md`, and both READMEs are consistent with the code changes.

## Non-blocking
- The 208KB file is committed twice (once per example). The PR body already justifies it: only a package's own `public/` is web-served, symlinks out are refused, and per-example copy-pasteability is intended. Reasonable given the constraint.

PR #904, opened 2026-07-19
fix(persistence): quote bound manifest table paths for their connection's dialect

jswir:
## The bug

On **Snowflake**, every serve-time query on a `#@ persist` source failed with `Object 'DB.SCHEMA.TABLE__…' does not exist or not authorized`, even though the materialized table existed with correct data (reproduced against a live Snowflake account; also took out every dimensional-indexing job on persist sources in a production deployment).

Root cause is a write/read quoting asymmetry inside this repo: the build side CREATEs the table with `quoting.ts quoteTablePath` (per-segment quoted → case-preserved), but the manifest carries the logical unquoted name and Malloy pastes a manifest `tableName` into `FROM` **verbatim by contract** (malloy#2813: a bare name means *let the engine fold*). Snowflake folds the unquoted read to uppercase and can't resolve the case-preserved table. Postgres folds down and BigQuery doesn't fold, which is why DuckDB/Postgres/BQ setups never hit it.

## The fix

Quote at the manifest **bind** step (`Package.reloadAllModels → quoteBoundTableNames`): each entry's `tableName` is quoted with the *same* `quoteTablePath` that quoted the CREATE, for the dialect of the entry's connection — write and read share one module, so the two sides cannot drift. `connectionName` is carried from the wire manifest through `fetchManifestEntries` and the post-build auto-load so both bind paths quote.

Deliberately conservative fallbacks preserve existing behavior everywhere else:
- an entry with no `connectionName`, or whose connection can't be resolved, binds **verbatim** (bare-builder / engine-folding producers unaffected)
- a name already carrying a quote character binds verbatim (already canonical)

This is the producer-side resolution that replaced the withdrawn compiler change (malloydata/malloy#3004) — see that PR's closing discussion: quoting in the compiler would have regressed bare builders and reversed malloy#2813's verbatim contract, so the case-preserving producer (this server) now hands Malloy the quoted form.

## Tests

- `manifest_loader.spec.ts`: `connectionName` is carried through parse.
- `materialization_lifecycle.integration.spec.ts` (real DuckDB round-trip): the served **and** compiled SQL of a bound persist source now pin the **quoted** `FROM "order_summary"`.
- Verified end-to-end against a live Snowflake account: materialize → bind → query returns rows with `FROM "SCHEMA"."table__g000__hex"`, where the identical unquoted read fails.


housejester:
Reviewed the write/read quoting symmetry end-to-end — serve reload, the per-query freshness override, the chained-build in-memory manifest, the committed wire manifest, and the drop-tables path. The two name-spaces (committed=logical, in-memory=quoted) stay disjoint on all of them, and the degradation path is genuinely non-regressing. Clean, well-scoped, unusually well-tested. No correctness blockers from my pass — one cross-repo confirmation and two doc/cosmetic notes below.

## Mental-model delta

The physical name of a materialized table stops being one logical string that means the same to writer and reader, and becomes **dialect-quoted at every point it enters a `FROM`**, so the case-preserving producer and a case-folding engine (Snowflake) agree on which table is meant.

What a future dev now has to hold:
1. A manifest `tableName` reaching Malloy's `FROM` must be dialect-quoted; a *bare* name means "let the engine fold" (malloy#2813) and silently breaks case-folding engines.
2. Two name-spaces now exist: the **committed/wire** manifest (what the CP persists, what `entries[...]` returns) stays logical/unquoted; only the **in-memory** build Manifest and the **in-memory** serve entries (`freshnessEntries` / `buildManifestEntries`) are quoted. Don't cross them.
3. `connectionName` is the dialect source: absent → bind verbatim (benign, pre-change behavior); present-but-unresolvable → degrade verbatim **loudly** (metric + ERROR).
4. `isQuotedIdentifierPath` is the idempotency guard — an already-quoted name is never re-quoted.

Deliberately unchanged: non-case-folding engines resolve identically (output just gains quotes); the committed wire manifest's logical-name space; the CP contract for producers that omit `connectionName`; the drop path (still logical + `quoteTablePath`).

## Worth confirming (not blocking)

**[MED — cross-repo efficacy hinge]** The serve-*after-restart* fix only engages if the control plane actually **persists** `ManifestEntry.connectionName` into the manifest file that `fetchManifestEntries` re-reads on boot (`manifest_loader.ts:101`). The field predates this PR (added in the two-round-protocol change), and the post-build in-memory auto-load now carries it from the build's own entries — so a *fresh build* serves quoted regardless. But a **server restart** re-fetches from the CP's stored manifest; if the CP writes `connectionName` only into the inline build result and not the persisted manifest, Snowflake serve silently reverts to unquoted (no signal — a missing name is the benign path) until the next build. Can we confirm the CP persists `connectionName` on every `ManifestEntry` it writes to the manifest file?

## Minor / non-blocking

**[LOW — doc]** `ManifestEntry.connectionName` (`api-doc.yaml:4351`) — the serve-path field the primary fix reads — has no description, while the build-input `ManifestReference.connectionName` (`api-doc.yaml:4274`) got a rich one framing the field as build-seed-only ("when it seeds the build manifest"). A one-line description on `ManifestEntry.connectionName` noting the serve-side quoting use would keep the two schemas from reading as if only the build path uses it.

**[LOW — observability surface]** `getBuildManifestEntries()` now returns quoted paths, so `/status` and `/compile` surface `"schema"."order_summary"` where they previously showed `schema.order_summary`. Correct for `/compile` routing (that's the fix); for a `/status` operator diffing against the CP's logical name it's a slight surprise. No action — flagging in case anything downstream string-matches the logical form.

*Reviewed via James's review-lens (session: Bashful Bison).*

housejester (approved):

PR #905, opened 2026-07-20
feat(server): DuckDB extension-fetch policy and DuckLake catalog-format hardening

housejester:
## Summary

Tightens how the server provisions DuckDB extensions and attaches DuckLake
catalogs, with **no change to default behavior**. Full rationale and reference:
[`docs/ducklake.md`](docs/ducklake.md).

Motivation: `FORCE INSTALL … FROM community` re-downloads from the community CDN
whenever the network is reachable, so a running worker can silently drift past
the extension build verified at image-build time; there was no first-class way
to run air-gapped / pinned-image; and the DuckLake catalog-format preflight that
once produced a clear error was removed as dead code (#884), so an unattachable
catalog now fails as an opaque DuckDB error.

## Changes

- **`installAndLoadExtension`: drop `FORCE`.** Plain `INSTALL … FROM community`
  is local-first (no-ops when present), so a baked/offline-verified extension is
  never silently re-fetched.
- **New `EXTENSION_FETCH_POLICY` env var** — `on-demand` (default; unchanged) or
  `local-only` (no `INSTALL`; implicit auto-install off; locally present
  extensions still load; a missing one is a loud, actionable error). For
  air-gapped / pinned-image deployments. Unknown values fail fast at startup.
- **Every Publisher-owned DuckDB session** is pinned against implicit
  auto-install under `local-only` — environment connections *and* the
  per-package `:memory:` sandbox — so the no-network guarantee has no gap.
  DuckLake tier sessions pin regardless of policy (they need only the curated
  set); generic sessions follow the policy, so `on-demand` is unchanged. Under
  `local-only` a failed pin fails closed.
- **DuckLake catalog-format version contract.** `ducklake_version.ts` derives
  the supported range (`1.0 ≤ format ≤ engine max`) from the pinned
  `@duckdb/node-api` via an engine→max-format matrix — never an enumerated list.
  A CI check (`scripts/validate-ducklake-catalog-range.ts`, run in the Docker
  offline-LOAD job) fails the build when the pinned engine outruns the matrix.
- **Runtime range-preflight for DuckLake attaches** — a definitively
  out-of-range catalog yields a clear `UnsupportedCatalogFormatError` (→ 422)
  naming the format found, the supported range, and a migration pointer, instead
  of a deep DuckDB failure. It reads `ducklake_metadata.version` (the catalog
  *format*, not the writing client) via a plain read-only Postgres attach (a
  per-call unique temp name so a failed detach can't wedge it), and is
  **non-load-bearing**: any preflight read failure falls through to the normal
  attach. Structured `warn` logs on a rejection and on a `local-only` extension
  refusal.
- **Docs** — new [`docs/ducklake.md`](docs/ducklake.md) (attach model,
  catalog-format compatibility, extension provisioning, and the bundled-extension
  list for `local-only`); `EXTENSION_FETCH_POLICY` in `configuration.md`; a
  DuckLake subsection in `connections.md`; README index entry.

## Compatibility

No change to query results or connection behavior under the default `on-demand`
policy. Two deliberate provisioning changes on the default path, both intended
and disclosed rather than hidden:

- Community extensions (`bigquery`, `snowflake`) no longer silently self-update
  at runtime — plain `INSTALL` uses the local copy, so a baked build stops
  drifting past the CI-verified one until the extension cache rolls with an
  engine bump.
- DuckLake sessions no longer implicitly auto-fetch a referenced-but-unbaked
  extension; a DuckLake query relying on one now errors instead of silently
  fetching.

`local-only` is fully opt-in. Only the DuckLake connection type sees the
range-preflight, and it only *adds* a clearer error for an already-unattachable
catalog.

## Configuration

| Env var | Values | Default | Effect |
|---|---|---|---|
| `EXTENSION_FETCH_POLICY` | `on-demand`, `local-only` | `on-demand` | `local-only` disables all runtime extension fetching for air-gapped / pinned-image deployments |

## Testing

New unit tests: catalog-format range logic + fixtures; extension-fetch policy
(the `local-only` error path and the on-demand no-change guarantee); lazy DuckLake
attach (catalog never on the boot path); and a guard that the pinned
`@malloydata/malloy` still exports the `VirtualMap` API. Typecheck, lint,
Prettier, and the full build (with extension bake) pass; the server boots to
`serving`.


sagarswamirao:
Nice hardening overall -- dropping `FORCE` in favor of local-first `INSTALL` is the right fix for runtime drift, and the offline `--network none` LOAD smoke test is a good guard.

One design note on `EXTENSION_FETCH_POLICY`, in case it's useful to document: the two policies really serve two different acquisition paths, and it may be worth spelling that out for operators. `on-demand` is the only safe default for the published npm package -- `bake-duckdb-extensions` runs in `build`/`build:server-only`, not on `npm install`, and it writes to the HOME-based `~/.duckdb/extensions/v<ver>/<platform>` cache rather than into the package tarball, so an `npm install` / `npx` consumer starts with an empty cache and genuinely needs a first-use `INSTALL`. `local-only`, by contrast, is offline-safe only where a platform-matched bake has already populated that cache -- i.e. the Docker image (which `COPY`s the builder's cache) or a from-source `bun run build` on the target platform. Might be worth a line in `ducklake.md` steering air-gapped users to the image / from-source build as the supported offline route, since raw `npx` can't be made offline-safe without a fetch somewhere.

packages/server/src/config.ts
```diff
@@ -440,6 +440,44 @@ export const getMaxConcurrentQueries = (): number => {
...
+ *   extension network when it is not already present on disk (baked into the
+ *   image). Extensions already baked are used as-is.
+ * - `local-only`: Publisher never runs `INSTALL`, and turns DuckDB's own
+ *   implicit auto-install off on its build/serve sessions
+ *   (`autoinstall_known_extensions=false`), so no code path reaches the
+ *   network. Auto-LOAD stays on, so an extension already present on disk still
+ *   lazy-loads; a genuinely missing extension surfaces as a loud, actionable
+ *   error instead of a silent fetch. For air-gapped / pinned-image deployments.
+ *
+ * Throws at startup on an unrecognised value so a typo in a k8s manifest
```
sagarswamirao: Flagging early so it doesn't bite an air-gapped operator: the comment says an unrecognized value "Throws at startup," but `getExtensionFetchPolicy()` is only called lazily -- inside `installAndLoadExtension` / `applyExtensionSessionSettings`, both on the connection-resolve path -- and there's no call on the server boot path. So a typo like `EXTENSION_FETCH_POLICY=offline` actually boots to `serving` and only throws on the first query that resolves a DuckDB connection, which for a DuckLake-only deployment could be well after startup.

Two ways to reconcile it, either is fine: correct the wording to "throws on first connection resolve," or add a single `getExtensionFetchPolicy()` call during boot so the comment becomes true. The startup call is the more useful of the two -- someone relying on `local-only` for a no-network guarantee wants a bad value surfaced at boot, not on the first request.

packages/server/scripts/bake-duckdb-extensions.js
```diff
@@ -27,8 +27,11 @@ import { DuckDBInstance } from "@duckdb/node-api";
 // sandbox, federated-database attach, and the materialization catalog. Keep this
 // in sync with the install sites in those files.
 //
-// `community: true` mirrors the runtime's `FORCE INSTALL '<name>' FROM community`
-// (bigquery, snowflake); the rest are core extensions installed by name.
+// `community: true` marks the community-repo extensions (bigquery, snowflake);
+// the rest are core extensions installed by name. The bake deliberately uses
+// `FORCE INSTALL ... FROM community` to (re)populate the build cache from the
+// CDN at build time; the *runtime* install (connection.ts) uses a plain,
+// local-first `INSTALL` that no-ops against this baked cache.
```
sagarswamirao, line 34: One gap worth closing for the `local-only` path: the bake is best-effort -- the `main().catch` below logs a warning and never fails the build, so a CDN error or an offline build machine produces a "successful" build with a partial or empty extension cache. In CI the `--network none` LOAD smoke test catches that, but a from-source `bun run build` on a developer/air-gapped machine has no equivalent guard, so the shortfall only surfaces later as a first-query failure under `local-only`.

Since `local-only` is the mode whose whole promise is "no runtime fetch," it'd be worth making that promise verifiable: e.g. on boot under `local-only`, check that the required extensions are actually present on disk and fail fast with the same actionable message the LOAD path uses, rather than deferring to the first query. (Pairs naturally with the startup validation suggested on `config.ts`.) Optionally, having the bake fail the build when a *required* extension didn't bake would stop a partial cache from shipping silently in the first place.

sagarswamirao (approved):

packages/server/src/config.ts
housejester: Good catch — fixed. Added a `getExtensionFetchPolicy()` call on the boot path (top-level in `server.ts`, alongside the other config reads), so an unrecognised value now throws at startup rather than on the first connection resolve. Went with option (b), since an operator relying on `local-only` for a no-network guarantee wants a bad value surfaced at boot.

Verified end-to-end: `EXTENSION_FETCH_POLICY=offline` now fails the boot (`uncaughtException: Invalid value for EXTENSION_FETCH_POLICY…`, never reaches `serving`); a valid value boots and logs the resolved policy. Also corrected the getter's comment to say it's validated at startup *and* re-read on each resolve.

packages/server/scripts/bake-duckdb-extensions.js
housejester: Agree on moving the failure earlier — the boot-time policy validation I added for the `config.ts` note covers the value-validation half at startup. On the two stronger forms, a couple of tensions worth naming before we commit to them:

- A general "verify the required extensions are present at boot" has to know which extensions the *configured connections* need, which means resolving connections at boot — and that conflicts with an invariant this PR deliberately preserves: connection/catalog attach is lazy and never on the worker boot path, so an unreachable catalog can't block startup. A non-connection-resolving presence check of the baked set is feasible; the per-connection version isn't without giving up lazy attach.
- Failing the *build* when a required extension didn't bake is self-defeating for the air-gapped case: the bake needs the network, so a from-source build on an offline machine can't bake at all — hard-failing there would block exactly the offline-from-source route `local-only` is meant to support. The intended model is a best-effort bake + operator pre-provisioning, with the offline `--network none` LOAD smoke test as the CI guard.

I'd prefer to take the fuller boot-time presence check (scoped to be non-connection-resolving) as a follow-up rather than in this PR, so the lazy-attach guarantee isn't traded away. Does that split work for you?

housejester:
Thanks — added this to `docs/ducklake.md`. It now spells out that the bake writes to the HOME DuckDB extension cache (`~/.duckdb/extensions/v<version>/<platform>/`), *not* into the npm package, so an `npx` / `npm install` consumer starts with an empty cache and genuinely needs a first-use `INSTALL` (i.e. `on-demand`) — raw `npx` can't be made offline-safe. And that `local-only` is offline-safe only where a platform-matched bake has already populated that cache: the published Docker image (which copies the builder's cache) or a from-source `bun run build` on the target platform — steering air-gapped deployments to those two routes.

PR #906, opened 2026-07-20
feat(server): enforce #(authorize) across joined sources

Sha-Bang:
## Problem

`#(authorize)` gates were evaluated only against a query's **top-level run-target source**. A gate on a source reached via `join_*` was never checked, so a query against an ungated source that joins a locked one could read the locked source's data with zero enforcement. This repo's own suite documented the hole with a passing `[documented limitation]` test asserting the bypass returned data.

## Change

Walk the compiled query struct and evaluate **every reachable joined source's gate**, recursed transitively (A→B→C), with **AND-across-sources** semantics (any reachable gate failing denies; each source's own expression list stays an OR disjunction).

Key decisions:
- **Gates are read from each join field's own struct annotations**, not a `this.sources` name lookup. A name lookup would silently miss deep-transitively-imported sources (absent from `modelDef.contents`) and inline-extend joins (`referenceID` cleared) — both would stay bypassable. Reading annotations off the struct closes them.
- **The walk runs unconditionally** on the query, notebook, and `/compile` paths. The prior `compiledSource === earlySource` skip fired for exactly the common `run: joiner -> {...}` case — that skip was the root cause.
- File-level `##(authorize)` still prepends to every joined source, so the admin-override idiom works transitively. Guarded by `hasAuthorize()` so ungated models pay nothing.
- Type-filtered via `isSourceDef` so nested `RecordDef`/`RepeatedRecordDef`/`ArrayDef` (which `isJoined` also matches) are excluded; inline `duckdb.table(...)` joins carry no gate and are skipped.

This **tightens** access — queries that previously succeeded via the join-bypass now return 403. The documented `include { public/private }` mitigation becomes the enforced default.

## Residual (out of scope)

Raw `duckdb.sql(...)` reached via `/compile` remains ungated — a separate class, tracked as a follow-up.

## Tests

Flipped the `[documented limitation]` test to assert `AccessDeniedError`; added transitive A→B→C, aliased join, mixed gated+ungated (AND), inline-table-allowed, file-level admin override, and a notebook-cell case. `authorize_integration.spec.ts`: 47 pass / 0 fail. Full `packages/server` unit suite: 1133 / 0.


housejester (approved):
**Approve with comments.** This correctly closes the join-bypass, and does it the right way — reading each gate off the join field's own struct annotations rather than a `this.sources` name lookup is what makes deep-transitive and inline-extend joins enforceable at all. The unified `collectAllReachableGates` traversal is well-factored and fails closed in the right places (`gateExprsForOwnAnnotations` → `["false"]` on parse failure; malformed join-only gates deny). Coverage is genuinely thorough. A few things worth a look — none blocking.

### 1. `evaluateAuthorize` is ambient-first, so a name-colliding entry-model given can decide a *joined* source's gate — `[security]`

Every probe compiles against `this.modelMaterializer` (the entry model) and is tried ambient-first; the self-contained `bindProbeGivens` fallback only runs if the ambient probe *throws*. So when a joined source's gate references `$LEVEL` and the entry model happens to declare its own `given: LEVEL`, the ambient probe compiles and evaluates against the *entry model's* `$LEVEL` (its declaration/default) — the isolating fallback never fires because ambient didn't throw.

Scenario: entry model declares `given: LEVEL is 99`; a deep-joined source is gated `#(authorize) "$LEVEL > 3"`; a caller supplying no `LEVEL` is granted on the entry default rather than denied. Narrow (needs a name collision with a permissive default), but the gate is evaluated in the wrong namespace, and it's the one case the BLOCKING-4 tests don't reach (they all exercise the *absent*-from-ambient path). Either a note in `docs/authorize.md` or probing a joined-source gate self-contained first when its givens aren't the entry model's.

### 2. Joined-gate enforcement silently no-ops if the compiler IR shape drifts — `[robustness / latent fail-open]`

`resolveRunTargetStruct` / `collectAllReachableGates` duck-type deep into compiler internals (`prepared._query.structRef`, `.compositeResolvedSourceDef`, `.pipeline[].extendSource`, `_modelDef`, `struct.annotations.blockNotes`, `.type === "query_source"|"composite"`). The catch in `resolveRunTargetStruct` reasons correctly about the *throw* case — but not about silent drift: if a bump renames/moves `_query.structRef`, `getPreparedQuery()` still resolves, `structRef` reads `undefined`, `struct` becomes `undefined`, the joined walk returns `[]`, and execution proceeds because it never went through the read path. Joined gates quietly stop firing while the own-source gate keeps working, so it looks fine.

The integration suite is a real net for every shape it exercises (a broken `structRef` flips "denies A→B→C" to a failing test on the bump PR). The residual is a newly-introduced source shape no test covers, plus the catch-all swallowing a mismatch as "no joins." A loud invariant — when a runnable clearly has joined sources but struct resolution yields nothing, log/throw rather than treat as ungated — is cheap insurance on a security path coupled to unstable internals.

### 3. The ungated fast path is gone for every query — `[perf]`

Dropping the `hasAuthorize()` guard here and in `environment.ts` is correct (it only sees `this.sources`, so it missed deep-join gates — that was the bug). But now every query/cell/compile, including the common ungated case, runs `getPreparedQuery()` + a full struct-tree traversal. You already do a deep walk at construction (`computeAuthorizeReferencedGivenNames`) — a memoized `hasAnyReachableGate` from that same walk would restore the ungated fast path safely for all but one narrow case (a query-local `join_one` to a source that's neither in `modelDef.contents` nor reachable from any top-level source). If that case is in scope, staying unconditional is right; if not, the flag is a cheap win — worth a sentence either way. Minor, related: `assertAuthorizedForAllSources` calls `getPreparedQuery()` via both `resolveAuthorizeSourceFromRunnable` and `resolveRunTargetStruct` — fine if Malloy memoizes it, a second compile per query if not.

### Question — composite run target without `compositeResolvedSourceDef`

`collectAllReachableGates` only expands composite *members* inside the joined-field loop, so a top-level composite relies entirely on the caller pushing `compositeResolvedSourceDef`. Is that guaranteed populated for any composite that compiles/runs? If it can ever be absent, its members go ungated (fail-open) — the "denies … locked member branch" test only covers the populated path.

housejester (approved):
**Approve — all comments below are non-blocking.** Building on @housejester's approval, which correctly landed the core design (reading each gate off the join field's own struct annotations, the unified `collectAllReachableGates` traversal, failing closed in the right places). This pass focuses on the four follow-up commits pushed *after* that approval, which rework the joined-gate evaluation to isolate-first. They hold up, and the new tests exercise exactly the scenarios the original review called out. Nothing here blocks merge.

### Follow-ups resolve the earlier points

- **Ambient-first namespace collision — resolved.** `evaluateSelfContainedFirst` now denies *before* probing whenever a referenced given can't be shadowed (`decls.length !== referenced.length`). Traced the unsupplied-`$LEVEL`-with-colliding-entry-default case: the name is dropped from `decls`, the length check fails, and no ambient probe is ever reached. The ambient last-resort is only reachable once every referenced name is caller-supplied and shadowed, so no entry-model *default* can decide a joined gate. Covered by the new "denies … when no LEVEL is supplied" test.
- **Struct-shape drift at a join field — resolved.** `classifyJoinedField` logs loudly and injects a `["false"]` deny gate instead of silently `continue`-ing past a join field that isn't a walkable `SourceDef`.
- **`record`/`array`/`repeated_record` over-deny — resolved.** `isRecordOrArrayField` skips the nested-column kinds that carry `join` but aren't sources. Confirmed against Malloy 0.0.422 `malloy_types.js`: those three are exactly the `JoinBase`-carrying non-source kinds, and no real `SourceDef` type (`table`/`sql_select`/`query_source`/`composite`/…) matches the skip — so no fail-open and no residual over-deny.
- **Unconditional gate (dropped `hasAuthorize()` fast path) — documented conscious decision.** The why (the guard only sees `this.sources`, so guarding re-opens the deep-join bypass) is spelled out at both call sites; it's a cheap no-op for ungated models.

### Non-blocking observations

**1. `referencedGivenNames` only strips single-quoted string literals, but Malloy expressions also permit double-quoted ones (`shortString: DQ_STRING | SQ_STRING`).** `authorize.ts` `STRING_LITERAL_PATTERN`. A contrived gate using double-quoted Malloy literals that contain apostrophes — e.g. `#(authorize) "$A = \"foo'\" and $SECRET = \"bar'\""` — makes the single-quote regex mis-pair the two apostrophes and swallow the real `$SECRET` token → under-count → the `decls.length === referenced.length` check passes → the raw expr compiles ambiently and `$SECRET` can resolve off a colliding entry-model default. This is the one under-count direction that fails *open* (the `$NAME`-inside-`'…'` case the literal-awareness commit fixed only failed closed). It needs a very unusual gate + a colliding permissive default, so it's not a practical exposure — but since the isolation now leans on this scan, worth either widening the stripper to consume double-quoted literals too, or narrowing the comment to state DQ literals are out of scope.

**2. Top-level struct-resolution drift still degrades silently.** `resolveRunTargetStruct`: if a future Malloy IR bump renames/moves `_query.structRef`, `struct` reads `undefined`, the joined walk returns `[]`, and only `##(authorize)` still fires — with nothing logged. The field-level drift guard added here doesn't cover this top-level case. The integration suite nets every shape it exercises, so the live residual is just a newly-introduced runnable shape no test covers; a loud invariant ("runnable has joined sources but struct resolution yielded nothing → log/throw") remains cheap insurance on a security path coupled to unstable internals.

_(Review by Frantic Capybara, an automated review pass — human sign-off still recommended.)_

PR #907, opened 2026-07-20
feat(create-malloy-package): scaffold a package and local agent workspace

mlennie:
Kyle's ask from 16 July: a "create a new package" init, like create-react-app, that
creates the directory and sets up the local development environment around it.

`npm create malloy-package sales` writes the package (manifest, model, and either the
sample CSV or the file given with `--data`), registers it in `publisher.config.json`,
and wires the workspace: start and reset scripts, `.mcp.json` (or `.cursor/mcp.json`
with `--client cursor`), `AGENTS.md`, `CLAUDE.md`, a `.gitignore`, and the 28 agent
skills as real files under `.claude/skills/`.

## The rule the design follows

The tool must not claim more than it checked. Most of the work here is that rule
applied to cases where the naive version reports success over a broken result:

- **Reserved names.** 135 of them, derived from the compiler and guarded by a drift
  test rather than hand-curated. `create-malloy-package lead` used to exit 0 with a
  green check, and the server would then log `Cannot redefine 'lead'`, serve
  `packages: []`, and still report `operationalState: serving`.
- **Unservable paths.** A workspace path holding a space or an accent makes every
  model in it fail to load, because the model's relative table path is absolutized
  before DuckDB's path check upstream. `~/Documents/My Projects` and any accented
  username hit this. It now refuses up front and names the character.
- **Existing files.** `package.json`, `.mcp.json` and `publisher.config.json` are
  merged, not replaced. An unparseable manifest aborts before anything is written
  rather than being treated as empty. A symlink pointing outside the workspace is
  refused rather than written through. The run reports what it left alone and what
  it replaced.
- **`--host 127.0.0.1` in the generated start command.** Publisher defaults to
  `0.0.0.0`, which suits a deployed server, but neither the REST API nor the MCP
  endpoint has auth in front of it and this is a laptop workspace.
- **A separate `reset` script.** Publisher reads `publisher.config.json` only on a
  first boot and its persisted state wins afterwards, so a package added later is
  registered and never served, with nothing in the log to say why. `reset` carries
  `--init`; `start` deliberately does not, since `--init` also clears runtime state.

## Publishing

The package depends on `@malloy-publisher/skills` via `workspace:*`, which npm does
not rewrite at publish time. `prepublishOnly` refuses to publish a tarball still
carrying that range, so forgetting the pin fails loudly instead of shipping a version
no install can resolve.

**Both npm prerequisites are now done, so neither gates this PR any longer:**

1. `@malloy-publisher/skills` (merged in #899) is published at `0.1.0`. Its first
   publish had to be manual, with scripts enabled, or `prepack` would not have copied
   the skills in; a trusted-publisher entry can be created now that the package exists.
2. The unscoped name `create-malloy-package` is claimed, at a `0.0.0` placeholder.
   Nothing protects an unscoped name, so this had to land before the branch made the
   command public.

## Testing

285 unit tests in this package plus 14 e2e that boot a real server. The e2e suite
needs ports 4000 and 4040 free and says so rather than guessing when they are not.

The unit count includes coverage added in review for two defects found there: an
adopted-start-script bypass, where a joined runner flag in front of the server spec
(`--package=evil`, `--call=evil`, `--node-options=--require=...`) passed the
"everything in front is a runner or a flag" check and was then printed as `npm start`
and written into `AGENTS.md`; and `printable()` not escaping the Unicode bidi
overrides, which had no direct unit coverage at all.

Verification was heavier than usual because early rounds kept producing the same defect
shape: an adversarial audit, then two full verify passes, each with a PR review, a
security review, and two live agent tests (one installing from packed tarballs as a
stranger, one working inside the generated workspace with only `AGENTS.md` and the
skills as briefing). Both live testers got from an empty folder to querying their own
data. Findings fell from 4 blockers, to 3, to 2, and every finding in the last pass was
a defect in a previous round's fix rather than in the original design.

## Known follow-ups

- `--data` generates a `count()`-only model. DuckDB is already inferring the column
  types, so reading the header row and emitting dimensions and measures is the obvious
  next step, and it is the difference between one command to a row count and one
  command to a model.
- The root README does not mention this yet. The name now resolves to a `0.0.0`
  placeholder rather than a 404, so this is waiting on the first real publish.


kylenesbit (approved):
Approving. This is the most carefully reasoned code I've reviewed in this repo — the "must not claim more than it checked" rule is actually implemented rather than just asserted, and the comments earn their length by recording a specific defect and why the naive version was wrong.

All checks are green on `b121925c`, including Cross Platform on all three OSes.

## We pushed two fixes to this branch

`b121925c` is ours, not @mlennie's. **It should get a second pair of eyes before merge, since the same people who found these wrote the fix.**

**1. A joined runner flag in front of the server spec defeated `isSingleServerInvocation`.** The check accepted any pre-server token that merely *started with* a dash, reasoning it was "a runner or one of its flags". But every runner it allows takes flags that choose what actually runs, and joined into one token each begins with a dash. Verified against the real `scaffold()`:

| existing `scripts.start` | adopted? |
|---|---|
| `npx -y @malloy-publisher/server …` (baseline) | yes |
| `npx --package evil @malloy-publisher/server …` | no ✅ |
| `npx --package=evil @malloy-publisher/server …` | **yes** ❌ |
| `npx -p=evil @malloy-publisher/server …` | **yes** ❌ |
| `npx --call=evil @malloy-publisher/server …` | **yes** ❌ |
| `npx --node-options=--require=/tmp/evil.js @malloy-publisher/server …` | **yes** ❌ |

The last row is the clean one: no space, so tokenizing never splits it; `=` is not in `SHELL_CONTROL_CHARACTERS`; and it runs the *genuine* Publisher afterwards, so the ports, the loopback bind and the watch are all exactly what the tool goes on to claim. Nothing downstream looks wrong.

The harm isn't new execution — whoever can write `package.json` already owns `npm start`. It's that the tool then **vouches** for the script: prints it as `npm start`, writes it into `AGENTS.md` as the command an agent runs, and suppresses both the exposure warning and the explicit-command fallback. Same class as the `@file:/tmp/evil` spec hole already fixed two lines above.

The fix is an allowlist of boolean runner flags, mirroring what `matchesGeneratedInvocation` already does on the other side of the spec: an unknown flag declines. A declined script is left exactly as it is and the caller falls back to the explicit command, so nothing is lost by being strict.

**2. `printable()` did not escape the Unicode bidi characters.** It covered C0, DEL, C1 and the line separators, but not U+202A–202E or the isolates, which reorder the run that follows them. `preview()` is exported specifically for strings under no charset rule ("a `scripts.start` entry out of somebody else's package.json"), and those are echoed when a script is declined — so a decline reason could be made to read as a different sentence than the one composed here. Spoofing rather than execution, but a gap in an invariant this file states explicitly.

## Verification

- **+27 tests, 258 → 285.** Reverting the fix makes exactly the five bypass rows fail and nothing else, so the tests test the bug rather than passing alongside it.
- The declined script lands on reason `unrecognised-shape` — "what it runs, and what it binds, is not something this tool established" — which is accurate for this case. The fix doesn't trade a false claim for a differently false one.
- `printable()` had **no direct unit coverage at all**; it now has 18 cases including idempotency and a check that Arabic and Hebrew pass through untouched (RTL by character property is not the hazard).

## Also updated the PR description

Three blocks were stale, all corrected in the author's voice:

- Both npm prerequisites are **done**: `@malloy-publisher/skills@0.1.0` and `create-malloy-package@0.0.0` were claimed after the description was written, so neither gates this any longer.
- Test counts `151 unit + 6 e2e` → **`285 unit + 14 e2e`**, both from running the suites rather than counting `test(` calls (a grep says 11 e2e and is wrong — some are generated dynamically).
- The "full root gate green at 1490 tests" line was dropped rather than restated, since it wasn't re-run.

## Other things checked this round

`skills.ts` is the most destructive code here — `fs.rmSync(..., {recursive: true, force: true})` on every run — and it holds up: deletion targets are bounded by names read from the *bundled* payload, never user input; `targetDir` is symlink- and containment-checked before anything is written; and each target is a single component under an already-verified directory, so there's no intermediate link to traverse. Remove-then-copy is correctly justified, since `fs.cpSync` follows a symlink at the destination.

Nit, not blocking: `removed[]` names only the first five user files deleted from skill directories. Honestly labelled ("and N more"), but a user who kept notes in there loses the tail unnamed.

## Follow-ups

- **`--data` still generates a `count()`-only model.** Listed as a known follow-up; I'd promote it to the top one. DuckDB is already inferring the column types, so this is the gap between "one command to a row count" and "one command to a model" — the actual product promise.
- e2e is unverified locally (it needs ports 4000/4040 and correctly refuses to guess when something else holds them). Ubuntu CI covers it.

mlennie:
Thanks @kylenesbit, and thanks for pushing the fix rather than just filing it. I took you up on the second pair of eyes. **Your fix holds**, and the same harm turned out to be reachable around it without any flag at all. Everything below is pre-existing; none of it was introduced by `b121925c`.

## Confirming your half

Every flag in `RUNNER_BOOLEAN_FLAGS` is genuinely boolean in each runner that accepts it, checked against npm/npx 11.12.1, bun/bunx 1.3.13, pnpm and yarn 4.15.0 with a local impostor package that echoes its argv. bun and yarn fail closed on `-y`/`-q` rather than accepting them. The nearest miss is yarn's `run --require <file>`, a real preload that takes a value, and it is already declined because `--require` is not in the set. All five of your bypass rows stay declined on the combined tree.

## The road around it

The leading word was the bigger hole. `SCRIPT_RUNNERS` held ten words and the check was set membership over every pre-spec token, so position never mattered. `npm`, `pnpm`, `yarn` and `bun` are not package runners, they are multi-tool commands whose meaning comes from the word after them, and `run`/`exec`/`dlx` were in the same set as bare words. Four holes, each reproduced against the built bin:

| script | what happens |
|---|---|
| `npm run @malloy-publisher/server …` | resolves that argument against this package.json's own `scripts` first, so a `scripts` entry under that name ran arbitrary code (I wrote a file to disk to be sure) while the tool printed "boots Publisher on a loopback address" and put it in AGENTS.md. Same for `bun run`, bare `pnpm`, `yarn run`, bare `run`. |
| `npm exec -y @malloy-publisher/server …` | runs the real package but eats `--host` as npm's own config, so the server got no flags and bound `0.0.0.0` with the warning suppressed. |
| `--host<NBSP>127.0.0.1` | JS `\s` counts U+00A0, U+2000-200A, U+3000, U+FEFF; sh's IFS does not. So this reached the server as one argument, `arg === "--host"` failed, the flag was dropped, and the bind fell back to `0.0.0.0`. All six chars, each under a line calling it loopback. |
| `@malloy-publisher/server@.` and `@evil.tgz` | npm-package-arg reads a leading-`.` spec as a directory install and a tarball-extension spec as a local file install, so both run code from the workspace offline, no registry, adopted as a loopback boot. |

## The fixes (the commits above yours)

- The runner is now a one- or two-word prefix: `npx`, `bunx`, `pnpx`, plus `pnpm dlx`, `yarn dlx`, `bun x`, with a positional check. I cut too far at first and dropped `pnpm dlx`/`yarn dlx`, which are the real npx analogues and pass flags through unchanged; a review pass caught the regression and I restored them.
- All three tokenizers split on space, tab and newline only.
- The spec after `@` has to start alphanumeric and not end in a tarball extension. A differential against real `npm-package-arg` confirmed that is the complete set of its non-registry forms (directory needs a leading `.`/`/`, git/remote/alias carry `:`/`/`, all already out), so the accepted set is now a proven subset of version/range/tag.

Reverting any one of these fails exactly the tests written for it, and the control confirms our own generated script is still adopted.

## One question back on your fix

Your bypass cases land on `unrecognised-shape`, whose sentence says the script "names the Publisher server inside a longer command line". For `npx --package=evil @malloy-publisher/server …` that is not true, it is one command with a runner flag. The commits under yours added `unmodelled-shell-syntax` and `unparseable` for exactly this reason, so it may be worth routing yours to one of those rather than leaving a decline that explains itself with the wrong fact.

## One more, found while rebasing onto you

Two comments referenced `SHELL_CONTROL_CHARACTERS`, a constant these commits renamed. One was yours, one was mine. The rebase reported no conflicts and was still wrong, which felt like the PR's own theme arriving through git.

PR #908, opened 2026-07-21
feat(connections): add verify-full sslmode for proxied Postgres

Sha-Bang:
## What it does

Adds `verify-full` as a fourth TLS mode for Postgres connections reached through an SSH bastion (connection proxy). Until now the proxied ladder stopped at `verify-ca` — chain verified, hostname **not** — because the driver connects to the `127.0.0.1` tunnel endpoint and a connectionString `sslmode` can't carry a `servername`.

`@malloydata/db-postgres@0.0.423` (from malloydata/malloy#2968) adds a raw `ssl` passthrough, so full verification is now expressible: for `verify-full` we pass `ssl: { servername: <real host>, rejectUnauthorized: true }`. pg then checks the cert's SAN against the real host even though the socket lands on loopback — the tunnel host is an IP literal, so pg keeps our `servername` instead of overwriting it. Verification runs against the ambient trust store (system roots + `NODE_EXTRA_CA_CERTS`, e.g. the baked Amazon RDS roots), so there's no CA-bundle precondition like `verify-ca` has.

The other three rungs (`disable` / `no-verify` / `verify-ca`) are unchanged — they still ride the connectionString query params. A new `resolveProxiedTls` helper picks the path.

- Bump `@malloydata/*` `^0.0.422` → `^0.0.423`
- `verify-full` added to the `sslmode` enum (`api-doc.yaml`) and `PROXIED_SSLMODES`
- `resolveProxiedTls` returns either a connectionString suffix or a raw `ssl` object

Backwards compatible: existing connections and the other three modes behave exactly as before.

> Note: the control-plane's generated publisher client model needs a regen so its `SslmodeEnum` learns `VERIFY_FULL` — a separate `service`-repo change; the existing deserializer handles the value once regenerated.

## Tested

- **Unit:** new coverage for `resolveProxiedTls` (verify-full → `ssl` object with `servername`; missing-host guard; other modes → query-only) and config validation (verify-full accepted on a proxied connection, no CA-bundle required, still rejected on a direct connection). Full suite: 91 pass, 0 fail; `typecheck:server` + `typecheck:sdk` clean.
- **End-to-end:** drove the real path — `openProxy` (in-process ssh2 bastion) + `resolveProxiedTls` + `PooledPostgresConnection@0.0.423` — through an SSH tunnel to a TLS Postgres stub whose cert SAN is `db.internal.test`, CA loaded via `NODE_EXTRA_CA_CERTS`:
  - `servername` = SAN → TLS verifies and connects (fails only later at the pg layer, as expected).
  - `servername` ≠ SAN → `ERR_TLS_CERT_ALTNAME_INVALID`, annotated with the `[malloy-db-postgres]` guidance.


mlennie:
## Review: approve, with one non-blocking design note

Nice change. It's correct, well-tested, fail-closed, and backwards compatible, and the security-critical invariant it depends on holds on the exact pg version this PR ships. I traced the full mechanism rather than just the description, notes below.

### What I verified

The whole feature rests on pg keeping our explicit `servername` instead of overwriting it with the loopback socket address. Confirmed end to end against the locked version (`resolutions.pg` pins `8.7.3`, which `bun.lock` resolves and `@malloydata/db-postgres@0.0.423` still declares):

- The tunnel endpoint binds to the IP literal `127.0.0.1` (`proxy.ts`: `server.listen(0, "127.0.0.1")`), and that's what lands in the connectionString.
- pg 8.7.3 `connection.js`: `Object.assign(options, self.ssl)` copies our `{ servername, rejectUnauthorized: true }` in first, then `options.servername = host` runs only `if (net.isIP(host) === 0)` (a DNS host). `net.isIP("127.0.0.1")` is `4`, so the guard is skipped and our servername survives. This is load-bearing: if the endpoint were ever `localhost`, verification would break (fail-closed, but broken). It's hard-coded, so we're fine.
- Verification is genuinely enforced: `this.ssl` is a truthy object, so pg runs the TLS handshake even though the verify-full connectionString carries no `sslmode`, and `rejectUnauthorized: true` drives full chain + SAN-vs-servername checking.
- No bypass: `resolveProxiedTls` is the only production caller of `buildProxiedSslQuery`, and `verify-full` is intercepted before it, so it can never fall through to the `default` throw.

### Correctness / quality

- The generated `packages/server/src/api.ts` is correctly absent from the diff (gitignored and regenerated at typecheck time since #762), so the new enum value flows into the generated union automatically.
- Version bump is internally consistent: db-postgres@0.0.423 still wants `pg: 8.7.3`, matching the pin, so the pg-resolution sync from #868 didn't need re-running. All `@malloydata/*` moved together across `package.json`, both package manifests, and `bun.lock`.
- The unreachable `!host` guard in `resolveProxiedTls` is fine to keep: config validation already requires host on a proxied connection, and it's a type-narrowing guard so `servername` is never `undefined`. It's tested.
- The `connection-parameters.js` merge-guard test is the standout: it pins the real `Object.assign({}, config, parse(connectionString))` behavior, so a future pg bump that started emitting an `ssl` key from the connectionString would fail loudly instead of silently downgrading verify-full.

### Security

- Fail-closed on every axis: `rejectUnauthorized: true`, handshake forced by the truthy ssl object, and the right host is checked. No fail-open path is introduced; `no-verify` / `disable` stay operator-chosen and unchanged.
- No downgrade vector (the merge-guard test proves the bare connectionString can't clobber the ssl object).
- No injection surface: `servername` is operator config, not query-time input.
- No new secret exposure: the diff adds no logging, so the connection-object-spread log trap from #736 isn't reintroduced.

### One design note (non-blocking)

`verify-full` here verifies against Node's ambient trust store (bundled Mozilla roots + `NODE_EXTRA_CA_CERTS`), while `verify-ca` pins to the `NODE_EXTRA_CA_CERTS` bundle only (via `sslrootcert`). So this implementation's `verify-full` is stronger on hostname but broader on CA trust than its own `verify-ca`, which inverts the usual libpq expectation that verify-full is a strict superset of verify-ca. Practically low-risk (no public CA issues certs for private RDS hostnames), and it's clearly documented in the api-doc and comments. If you want true chain-pinned-plus-hostname semantics, you could pass `ca: <NODE_EXTRA_CA_CERTS contents>` into the ssl object when that env var is set. Worth a sentence here, not a blocker.

One thing I couldn't re-run locally: that `PostgresSSLConfig` in db-postgres@0.0.423 admits `servername` / `rejectUnauthorized`. Your `typecheck:server` / `typecheck:sdk` runs and the `PooledPostgresConnection@0.0.423` E2E cover that, so it's confirmed, just not from my stale checkout.

mlennie (approved):

Issue #896, opened 2026-07-16
Materialization scheduler Phase 2: CLI + full docs pass (follow-up to #888)


girishjeswani:
Phase-2 follow-ups to #888, tracked separately so that PR stays focused on the standalone scheduler + config-parity core. Sequenced after it lands.

## CLI

Add publisher CLI coverage for the materialization surface (the CLI already has a generated client to build on):

- `schedule view` / `set` / `clear` — read and edit a package's `materialization.schedule`, sharing the same publish-gate validation the server enforces.
- `materializations list` — per-package and environment-scoped, with the trigger (`ON_DEMAND` / `SCHEDULER`) column.
- Trigger a run on demand.
- Inspect a run's manifest / status.

## Agent skills (depends on the CLI)

Cost/performance optimization skills that lean on the CLI + publisher APIs — what's persisted vs not, run timings and sources built/reused from the materialization history — to recommend persist candidates, schedule/scope changes, or removals. Recommendations-only.

## Full docs pass (depends on the above)

Once the CLI and local devex are in place:

- Malloy docs site: getting-started, config/env-var reference, the publish-gate rules stated as rules, fire semantics.
- Repo README/docs: operator depth, the environment-scoped view, and an honest list of the standalone-vs-hosted divergences (e.g. tables-only local fires vs the two-phase hosted job).

Docs come last, since the content depends on how the CLI and any remaining gap-closers land.

PR #909, opened 2026-07-21
feat(cli): materialization schedule + listing CLI, tuning skill, and docs

girishjeswani:
## Summary

Phase-2 follow-up to the standalone materialization scheduler (#888) — the CLI, an optimization skill, and operator docs from #896.

### CLI (`malloy-pub`)

The CLI already had per-package `list`/`get`/`materialize`/`stop`/`delete` for materializations. This fills the gaps:

- **`schedule view` / `set <cron>` / `clear`** — read and edit a package's `materialization.schedule`. `set` also sets `scope: version` (a schedule requires it), and the server's publish-gate validation (5-field UTC cron grammar, scope/freshness rules) is enforced server-side and surfaced as-is — an invalid cron exits non-zero.
- **Environment-scoped listing** — `list materialization --environment <env>` (no `--package`) lists every package's runs via the `/environments/{env}/packages/materializations` endpoint, labeled by **Package**.
- **Trigger column** — both list views now show `ON_DEMAND` vs `SCHEDULER`.

Under the hood: the client `updatePackage` wrapper now carries `scope` + `materialization`, and a `listEnvironmentMaterializations` wrapper was added. No `api-doc.yaml` change — the endpoints already exist; the CLI regenerates its client at build.

### Agent skill

`skills/malloy-materialization-tuning` — a Publisher-specific skill that reads the materialization history via the CLI (`list`/`get`: per-run `durationMs`, `sourcesBuilt`/`sourcesReused`, trigger) and recommends what to persist, what to stop persisting, and how to schedule/scope it. **Recommendations-only** — the local counterpart to the platform's usage-driven optimization. Symlinked into `.claude/skills/`.

### Docs

- `docs/materialization.md` — persist annotation, the publish-gate rules stated as rules, on-demand + scheduled builds, fire semantics, the CLI, and an honest **standalone-vs-hosted** divergence list (tables-only vs two-phase, per-version tables, physical-table GC).
- Three scheduler env vars documented in `docs/configuration.md`.
- Linked from the docs index and the README.

## Testing

- CLI unit + integration green: schedule set/view/clear roundtrip, invalid-cron rejection (exit 1), and env-scoped listing with the Package/Trigger columns all exercised against a real server.
- `typecheck` + `prettier` clean on changed files. No product names anywhere.

> Note: 3 pre-existing CLI unit failures (`getEnvironment`/`getPackage`/`getConnection` "fetch and print JSON") exist on `main` independent of this branch — not touched here.

## Notes

- Addresses the CLI + optimization-skill + repo-docs portions of #896. The remaining item in #896 is the public Malloy **docs-site** getting-started page, which lives in a separate repo.


mlennie:
@girishjeswani Reviewed end to end, including live: I ran a real server off this branch against the persist-test fixture and exercised every claim. Schedule set/view/clear roundtrip works, an invalid cron exits 1 with the publish-gate message surfaced verbatim, the env-scoped listing shows Package + Trigger with `ON_DEMAND`, and every field the tuning skill documents (`metadata.durationMs` / `sourcesBuilt` / `sourcesReused` / `trigger`, `entries[*].sourceName` / `physicalTableName` / `realization`) is real and populated in the `get` JSON. The "3 pre-existing CLI unit failures" claim reproduces exactly on main (47 pass / 3 fail, precisely those three), and the branch adds 6 passing unit tests plus 3 integration tests (15/0 total). Server suites green on the head and on a local merge with current main. The bundle-sync spec proves `skills_bundle.json` matches `skills/`, the `.claude/skills` symlink resolves, zero product names anywhere including the regenerated bundle, and all three scheduler env vars in `configuration.md` match `config.ts` exactly (names, defaults, min 1000, positive-int guard, off by default). DCO on all 4 commits. This is a strong PR. Two things I would fix before merge, both small; the rest is optional.

**1. `schedule clear` (and `set`) silently destroy a freshness policy.** Both commands send a fresh `materialization` object, and the server replaces the stored object wholesale when the body carries one. Live repro on this branch: package with `{"freshness": {"window": "24h"}}` and no schedule; `malloy-pub schedule clear` succeeds and leaves `{"schedule": null}`, freshness gone. The SDK card can't hit this (its clear button only renders when a schedule exists, and schedule+freshness can't coexist), but the CLI has no such guard, and "clear the schedule" on a package with nothing to clear deleting an unrelated policy is a trap. Suggest: `clear` no-ops with a message when no schedule is set (it already fetches the package first, so it's one check), and `set` prints a warning when it is replacing an existing freshness policy rather than dropping it silently.

**2. The tuning skill's Step 4 removal flow breaks serving.** Step 4 says rebuild first, then `delete materialization <id> --drop-tables` to clean up the removed table. But `dropTables` drops every physical table in that run's manifest, and auto-run assigns stable table names and carries unchanged sources forward, so an old run's manifest names the same tables the newest manifest still serves. Live repro on this branch: materialize twice (run 2 reuses, both manifests reference `order_summary`), `delete <run1> --drop-tables`, then a query against the source fails with `Catalog Error: Table with name order_summary does not exist!` while run 2 sits at `MANIFEST_FILE_READY`. Fix is wording plus order: in the skill, delete the old run with `--drop-tables` first and then `materialize --wait` (which rebuilds everything still persisted, ending in the desired state), and add one sentence warning that `--drop-tables` drops every table in that run's manifest, including tables newer manifests still reference. `docs/materialization.md`'s CLI section could carry the same one-line warning. A server-side guard (refuse or warn when an entry is referenced by a newer manifest) feels like a worthwhile follow-up issue rather than something for this PR.

Smaller items, your call whether pre-merge or follow-up:

- **Scope stranding + wrong pointer:** `schedule set` flips the package to `scope: version` and `clear` leaves it, which the docstring acknowledges by pointing at `update package`, but the CLI's `update package` only forwards location/description, so there is no CLI way back to `scope: package`. Either add a `--scope` option to `update package` or fix the docstring to say the way back is a direct PATCH / `publisher.json` edit.
- **Empty cron passes the gate:** `schedule set ""` exits 0, persists `"schedule": ""`, and flips scope to version (live-verified). Every server rule is truthiness-guarded so the empty string skips them all. A one-line non-empty check in the CLI closes it (a server-side fix for the truthiness guards would be a nice follow-up).
- **Docs:** the env-scoped listing section says the view "adds a Package column", but it also drops the Error column relative to the per-package view, so a failed run's message is invisible env-wide; worth saying (or adding the column). And in the standalone-vs-hosted list, "its connection and canonical SQL (its `#@ persist name`)" reads as if `name=` is the identity; it is the physical table name, while identity is the sourceEntityId content address. Line 25 of the same doc has it right.
- **Skill wording:** the env-wide listing is newest-first with a package label, not "grouped by package"; an agent expecting contiguous per-package blocks would misread interleaved rows.
- **skills/README.md:** the new skill isn't added to the shared-vs-Publisher-specific lists that drive the upstream sync. It names `malloy-pub` throughout, so it belongs in the Publisher-specific list.
- **CLI has no CI gate:** root typecheck/lint/test all omit the cli package and no workflow builds it, so the "regenerates its client at build" story is only verified on dev machines. Chaining `typecheck:cli` (after a generate step) or adding a small CI job would catch generated-symbol drift like the new `listEnvironmentMaterializations` dependency; fine as a follow-up issue.
- **Nit:** a couple of new CLI output strings still carry em-dashes ("(none — publish / on-demand only)") after the de-em-dash commit, and the wholesale table re-pad in README/configuration.md inflates the diff about 3x over the real change (I verified with `git diff -w` that it is content-neutral).

Security review: nothing blocking on either surface. The new endpoints go through the generated axios client with path templating, error output doesn't expose anything beyond the server's own 400 bodies, and the skill is recommendations-first with an explicit approval gate before any mutating step.

girishjeswani:
Thanks for the exhaustive review @mlennie — the live repros were especially useful. Addressed in `a33a7761` (CLI) and `c307d5a5` (skill/docs); Summary:

**Blockers**

1. **`schedule clear`/`set` destroying a freshness policy** — `clear` now no-ops with a message when no schedule is set (so it never PATCHes, and a freshness-only package's policy is left intact), and `set` warns before replacing an existing freshness policy. Added unit tests for both (no-op clear never calls `updatePackage`; set-over-freshness still PATCHes) plus the existing live roundtrip.
2. **Tuning skill Step 4 removal flow breaking serving** — reordered to **drop the old run `--drop-tables` first, then `materialize --wait`**, with a warning (in the skill and in `docs/materialization.md`'s CLI section) that `--drop-tables` drops every table in that run's manifest, including tables a newer manifest still serves.

**Smaller items (all in-PR)**

- **Empty cron** — `schedule set` now rejects an empty/whitespace cron client-side (unit + integration test, exit 1).
- **Scope pointer** — `clear`'s docstring now points at `publisher.json` / a direct PATCH (not `update package`, which can't change scope).
- **Env-scoped list** — added the **Error** column so a failed run's message isn't invisible env-wide; docs updated (leading Package column; newest-first interleaved, not grouped).
- **Identity wording** — corrected to say a table's identity is the content-addressed `sourceEntityId`, not the `#@ persist name` (physical table name).
- **skills/README.md** — added `malloy-materialization-tuning` to the Publisher-specific (never-synced) list.
- **Em-dashes** — removed from the remaining CLI output/help strings.

**Filed as follow-ups (not this PR, per your call)**

- #916 — server-side guard so `--drop-tables` won't drop a table a newer manifest still references.
- #917 — reject an empty-string schedule at the publish gate (close the truthiness-guard gap server-side).
- #918 — CI gate that builds/typechecks the `cli` package to catch generated-client drift.

Left the README/`configuration.md` table re-pad as-is (content-neutral per your `git diff -w`).

packages/cli/src/commands/schedule.ts
```diff
@@ -0,0 +1,98 @@
...
+ * policy on a freshness-only package. Scope is left as version; to return a
+ * package to `scope: package`, edit `scope` in its publisher.json (or PATCH the
+ * package directly); the CLI does not expose a scope switch.
+ */
+export async function clearSchedule(
+  client: PublisherClient,
+  environmentName: string,
+  packageName: string,
+): Promise<void> {
+  const current = await client.getPackage(environmentName, packageName);
```
Sha-Bang, line 92: `!current?.materialization?.schedule` is truthiness-guarded, so a persisted empty string (`""`) — the same hole as #917 — would no-op here and leave junk in place.

Prefer treating "has a schedule to clear" as `schedule != null` (and optionally clear `""` / whitespace as junk), so this can clean up a stuck empty schedule even before the server-side publish-gate fix lands. Low urgency given client `set` now rejects empty.

Sha-Bang (commented):
CLI / skill / docs look solid after the freshness-wipe, empty-cron, and drop-tables follow-ups — leaving the README hero video to the in-progress fix.

A couple of non-blocking should-fixes:

1. **`schedule clear` truthiness guard** (inline) — treat presence as `schedule != null` so a stuck `""` can still be cleared (pairs with #917).
2. **Thin test gaps** — worth adding when convenient, not merge-blocking:
   - integration: `clear` on a freshness-only package never PATCHes (unit covers it; live path was the original bug)
   - env-scoped list asserts the Error column header
   - `clear` when already cleared prints the no-op message and exits 0
3. **Table re-pad** in README / docs indexes — content-neutral per `git diff -w`, but it inflates the review surface; fine to leave.

Not blocking merge once the hero video lands.

packages/cli/src/commands/schedule.ts
girishjeswani: Fixed in 7d468834. `clear` now treats presence as `schedule != null`, so any set value clears — including a stuck `""` or whitespace — while an absent schedule (`null`) still no-ops so a freshness-only package's policy is never touched. That lets it clean up the empty-string state even before the server-side publish-gate fix (#917) lands.

Tests: a unit test that a stuck `""` schedule clears (PATCHes `{schedule: null}`), plus integration for the live no-op-never-PATCHes path (`clear` on an already-cleared package, exit 0). Also added the two other thin-gap assertions you flagged: the env-scoped listing asserts the `Error` column header, and `clear` on an already-cleared package prints the no-op message and exits 0.

girishjeswani:
Addressed @Sha-Bang's follow-up in `7d468834` (CLI only):

- **`schedule clear` truthiness guard** — now treats a set schedule as `schedule != null` instead of truthiness, so a package stuck with `schedule: ""` (the same hole as #917) can be cleared rather than no-op'd. An absent schedule (`null`) still no-ops, so a freshness-only package's policy is never wiped.
- **Thin test gaps** — added the three you called out: a unit test that a stuck `""` schedule clears (PATCHes `{schedule: null}`); an integration test for the live no-op-never-PATCHes path (`clear` on an already-cleared package prints the no-op message and exits 0); and an assertion that the env-scoped listing carries the `Error` column header.

Left the README / docs table re-pad as-is per your note (content-neutral by `git diff -w`).

Verification: CLI unit 8/8 schedule tests pass (the 3 failing suite tests are the documented pre-existing env/package/connection JSON-print ones); CLI integration 17/17; prettier + `tsc --noEmit` clean.

Sha-Bang (approved):

girishjeswani:
Pasting the video URL here so it becomes public

https://github.com/user-attachments/assets/376a809d-8016-41a7-9464-a5634ea0589d

PR #910, opened 2026-07-21
docs: point people at npm create @malloy-publisher/malloy-package

mlennie:
The scaffolder exists but nothing tells anyone it does: zero mentions in the README, and zero across all 28 skills. Someone who wants to explore a CSV of their own has no way to discover the one command that sets that up.

The README gains a section between the quick start and the agent section, which is where a reader who has just seen the bundled examples asks how to use their own data. `malloy-getting-started` gains the same pointer, for the case where an agent has the skills but there is no workspace to work in.

Both show the `npm create` and `npx` forms, because they parse options differently and getting it wrong is a dead end rather than an error a reader can reason about: `npm create` reads anything before a `--` as one of its own options, so the separator is required, while `npx` forwards flags untouched and counts a `--` as an extra argument. Measured on npm 11.12.1 rather than assumed.

The skills bundle is regenerated, since it is a build-time asset generated from skills/ and a spec fails when the two drift.

Do not merge this before the `create-malloy-package` name is claimed on npm. It is unscoped and unregistered today, so publishing an instruction to run `npx create-malloy-package` would point readers at a name anyone can take.


skills/malloy-getting-started/SKILL.md
```diff
@@ -13,6 +13,8 @@ At minimum you need `malloy_getContext`, `malloy_executeQuery`, and `malloy_sear
 
 If none of the tools are there, either the server is not running or your client connected before it was. Start the server (`npx @malloy-publisher/server --port 4000`, or `bun run build && bun run start` from a clone) and wait until `curl -s http://localhost:4000/api/v0/status` reports `operationalState: serving`.
 
+If there is no Publisher workspace here at all, and the user wants to explore a CSV or Parquet file of their own rather than the bundled examples, `npm create malloy-package <name>` scaffolds one: the package and a starter model over their data, registered so the server actually serves it, plus the start script, the MCP config and these skills. Seed it from their file with `npm create malloy-package <name> -- --data ./their-file.csv`, keeping the `--`, which is how `npm create` passes options through. Because it writes a `.mcp.json` that did not exist when the client connected, the user has to restart or reconnect once before these tools appear, and their client will ask them to approve the new project-scoped server the first time.
```
kylenesbit: I don't think this needs to be specific to CSV or Parquet files.  In fact, it'd be nice for the skills to create a package and  request CSV / parquet files or credentials to setup a database configuration.

README.md
```diff
@@ -36,6 +36,32 @@ all DuckDB-backed, no credentials required. Give the server a moment to report `
 curl -s http://localhost:4000/api/v0/status | jq .operationalState   # → "serving"
 ```
 
+## Start from your own data
+
+The command above serves the bundled examples. To serve a CSV or Parquet file of your own, scaffold a
```
kylenesbit: Same as a above -- CSV / parquet files of connection credentials.

mlennie:
@kylenesbit Good call, reworded both spots to say "your own data" rather than "a CSV or Parquet file" (2ec1ec6a).

To be precise about what ships today, since it shapes how far the docs can go: the scaffolder's `--data` seeds the starter model from a local file only (CSV, Parquet, or Excel, all via DuckDB). It does not set up a database connection. But a package is just Malloy, so its model can point at any connection the config defines, so I kept the accurate `--data` file list and added a line that the same workspace serves a warehouse when you point the model at a Postgres/BigQuery/etc. connection. That broadens the framing without claiming a credentials flow the tool does not have yet.

Your second point, having the scaffold offer to set up a database configuration and having the skills request either data files or credentials, is a real feature I would like to build, but it is scaffolder plus skills work rather than a docs change. I will file it so it does not get lost. Happy to sketch it if you want to shape the flow.

One heads-up on this branch: I rebased it onto current main (it was 17 behind) and regenerated the skills bundle, which is a build artifact the spec fails on when it drifts, so your original review may show as outdated. The content you commented on is unchanged apart from the CSV/Parquet rewording above.

jswir (approved):
Approving on content — the section lands in the right place and the `--` explanation is genuinely useful. I verified the headline command end to end rather than reading it: on npm 10.8.2, `npm create @malloy-publisher/malloy-package sales` resolves, scaffolds, and prints scoped hints (the published `0.0.1` dist already carries #926's strings). The `--data` list matches `scaffold.ts:2028` (`.csv`, `.parquet`, `.xlsx`). Bundle regen is minimal and correct: one line, two occurrences of the scoped command, zero unscoped.

Five things to fix before merge, none of them in the prose itself.

**1. Title and body are stale against the diff.** Title says `npm create malloy-package`; the body says "Do not merge this before the `create-malloy-package` name is claimed on npm. It is unscoped and unregistered today." The diff uses the scoped command throughout, and the registry says that gate is already met: `@malloy-publisher/create-malloy-package@0.0.1` is published and is `latest`, while the unscoped `create-malloy-package` was **unpublished** on 2026-07-24T17:26:52Z. As written a reviewer will think this is blocked when it isn't, and that title lands in squash-merge history describing a command the PR doesn't document.

**2. Show one command form, not two.** This is the one I'd actually push on. The two lines differ in *two* places at once (`npm create` + `.../malloy-package` vs `npx` + `.../create-malloy-package`), and that cross-wires in practice — we watched someone who knows this codebase well produce the invalid third form, `npx create @malloy-publisher/malloy-package`, on their first retry. The failure mode isn't an error: npm installs and executes an unrelated registry package named `create`. (That package is harmless — 23 lines of `console.log`, no filesystem access — but that's luck, not design.) Lead with `npm create`, demote the `npx` form to a footnote for the flags-without-`--` case.

**3. Add `mkdir <name> && cd <name>`.** `packages/create-malloy-package/README.md:12` opens with `mkdir my-data && cd my-data`, but this section and the skill paragraph both omit it — and this is the copy a newcomer reads first. Without it, the natural move is to run it in whatever directory you're already in. Someone did exactly that in a directory holding 46 unrelated projects; the scaffolder writes `publisher.config.json`, `package.json`, `.mcp.json`, `AGENTS.md`, `CLAUDE.md`, `.claude/`, and later `publisher_data/` at that level, and because `CLAUDE.md` discovery walks up, every agent session beneath that directory then inherits the workspace config.

**4. `# npx does not, and rejects it`** — "it" parses as the flag before it parses as the `--`. Something like `# npx takes flags as-is; a -- here becomes a stray argument` matches the prose below it.

**5.** Body says "zero across all 28 skills"; there are 30 in the tree now.

**Ordering:** this can technically merge alone since npm already resolves the scoped name, but doing so leaves the README pointing at a scoped package while `packages/create-malloy-package/package.json` still says unscoped. Merge #926 first, or together.

mlennie:
Thanks @jswir, this was a genuinely useful pass. All five are in, and two of them changed my mind rather than just the text.

The `mkdir` one I had considered and talked myself out of, reasoning that the section names no directory so the reader stays put. You are right that staying put is the harm, and the CLAUDE.md-discovery-walks-up consequence is the part I had not thought through. It is in both the README and the skill paragraph now.

On the two command forms: agreed, and the failure is worse than a cross-wire, because `npx create @malloy-publisher/malloy-package` installs and runs a real unrelated package called `create` rather than erroring. `npm create` now leads and `npx` is a closing footnote naming the difference.

On the `--` wording I went to check rather than reword, and the tool's own help text is slightly off. Running the published 0.0.1 bin, `sales -- --data ./orders.csv` reports `got 3`, not 4, so the `--` is consumed and it is the flags after it that land as strays. Both README sentences now describe that. The same imprecise sentence lives in `create-malloy-package`'s own `excessArgumentsHint`, worth a follow-up there rather than here.

Title and body are rewritten, including the stale do-not-merge gate and the 28-versus-30 count.

Three things beyond your list, found while verifying the fixes. The section is called "Start from your own data" but the command it leads with quietly ships the scaffolder's own sample orders, so it now says so. `--data` actually produces a *smaller* model than the bare form, since the scaffolder does not read your columns, so you get a row count and an overview rather than the sample model's measures and charts; that surprised me and is now stated. And the section never named `npm start`, which matters more than it sounds: the generated start script pins the server version, binds loopback, and uses `--watch-env`, where the quickstart command above the section does none of those three, so a reader who never sees `npm start` edits a model on the copy-not-watch path where their changes silently do nothing.

Agreed on ordering, this goes in with or after #926.

PR #911, opened 2026-07-21
feat(server): DuckDB/DuckLake materialization tier for `#@ persist storage=`

housejester:
## Summary

Adds a **DuckDB/DuckLake materialization tier** for `#@ persist` sources: a source can be materialized into a *registered DuckDB or DuckLake connection* instead of its own warehouse, and served back from that materialized table — cross-dialect, with no model change. **Off by default**, behind a deployment kill switch that ships dark.

Walkthrough with runnable examples: [`docs/persist-storage-tutorial.md`](docs/persist-storage-tutorial.md).

This is a large diff, but fwiw 75% of it is tests.

## What's new

- **`#@ persist storage=<connection>`** materializes a source into that connection (a DuckDB/DuckLake destination) via native per-engine query-passthrough (`postgres_query` / `bigquery_query` / `snowflake_query`) on a disposable, **per-build-isolated** session — the source warehouse computes, only results cross into the store. Absent, or the reserved `storage=source`, is the unchanged in-warehouse path. The connection name `source` is reserved.
- **`PERSIST_STORAGE_MODE`** deployment switch (`off` default | `write-only` | `on`). A kill switch: `off` is a byte-identical no-op; moving it *down* never fails a loaded package (a `storage=` source reverts to serving live and surfaces a package warning). Validated at boot. See [`docs/configuration.md`](docs/configuration.md).
- **Serve from storage** (`on`): a query against a materialized source is served from the stored table via a transient **virtual-source transform** — the source's base is rebound to a typed virtual source declaring the post-build `DESCRIBE` schema, with its dimensions, measures, materialized-target joins, and views re-declared over the stored columns. Anything not reproducible **fails the shape compile and falls back to serving live** (full → drop views → drop views+joins → base-only), so turning it on can never make a query wrong.
- **Connection type `ducklake`** (catalog + `bucketUrl` storage) — see [`docs/connections.md`](docs/connections.md).

## Naming & lifecycle

- **Physical tables are named by `name=` verbatim.** The auto-run (standalone) server names a `storage=` table by its `#@ persist name=` value (or the source name) verbatim — exactly as the in-warehouse path does — and a rebuild atomically replaces it in place (DuckLake's catalog swap is transactional). No hashed suffix, no coexisting generations, no operator convenience view: the publisher is a pure mechanism, not a naming authority.
  - *Note for anyone reading the commit history:* content-addressed names (`<logical>__m<hash>`), coexisting generations and a convenience view were added in `687ed697` and then **reverted** in `46eb340f` / `8579f118`. Verbatim naming is the shipped behavior; the intermediate commits are not. The revert keeps a `storage=` materialization behaving exactly like an in-warehouse one — same table name, same in-place replace — so the only thing `storage=` changes is which warehouse the table lands in.
  - *Hashes that do exist, and are not this:* the **in-warehouse** build stages into `<name>_<12 chars of sourceEntityId>` and `ALTER TABLE ... RENAME`s it onto the bare name, so that suffix is transient and never served; `materializedTableId` is a manifest identifier, not a table; and the manifest is **keyed** by the content-addressed `sourceEntityId`, whose entry then records the plain physical name. None of these put a hash on the table a query reads.
- **A caller that orchestrates builds owns physical naming.** The build trusts `buildInstructions.physicalTableName` verbatim; the serve path resolves whatever physical name the manifest records. So generation management (immutable generations, safe schema evolution, rollback) is the caller's responsibility on the orchestrated build path (`buildInstructions` + `manifestLocation`), not the auto-run server's.
- **Reclaim is destination-aware and records-driven.** Deleting a materialization with `?dropTables=true` drops the stored table through a build-scoped **read-write** attach (the serve attach is read-only), and only for names the publisher recorded building — never a catalog scan. A table still referenced by another retained record is left alone, so dropping a superseded record cannot take out a table a live binding is reading. Reverting a source to live is in-place: the binding is dropped without a restart.

## Serve-binding contract (host/control-plane consumption)

Serving a `storage=` source across a fleet needs the host to tell each replica *which physical table backs each materialized source* — a cross-connection serve binding, distinct from the same-connection `tableName` substitution the in-warehouse path already uses. Both are carried in the one manifest the publisher already fetches from a package's `manifestLocation`:

- The fetched manifest is **tier-split** at bind: an entry naming a `storageConnectionName` (carrying its captured `schema` + `sourceName`) becomes a **serve binding applied to the already-compiled models — no recompile**; an entry without one stays a same-connection `tableName` substitution (which recompiles, as before). A storage entry never enters the tableName manifest.
- **A storage-only refresh costs no recompile.** The host rewrites the manifest and hits the existing manifest-rebind `PATCH` per generation; convergence is eventually-consistent and safe (a not-yet-rebound replica serves the prior generation's table, never an error).
- Entries are keyed by the build's content `sourceEntityId` (= the serve handle): a freshness refresh keeps the handle and swaps only the table path; a schema-changing generation gets a new handle. (This is the one hard cross-producer contract.)
- When a `manifestLocation` is bound, the **host is authoritative**: the publisher's local-store rebind is skipped, so an orchestrated deployment's local build records can't clobber the host's bindings with a staler generation.

## Multi-tenant build isolation

Each build/GC session runs on its **own in-memory DuckDB instance** (a unique working directory yields a private instance while the primary stays `:memory:`, so a fresh DuckLake catalog still auto-initializes). Without this, the DuckDB client library pools `:memory:` instances by a share key that excludes the connection name, so build sessions with identical config would be pooled onto one shared instance and collide on their transient `ATTACH` aliases — a cross-tenant read (source) or write (read-write destination) when two environments have same-named, different-credential connections. A regression test fails loud if a library change ever re-pools them.

## Chained sources

- A `storage=` source that reads another `storage=` source in the same destination is built by **reading the parent's materialized table** (rolled up in DuckDB) — reusing the parent's work, consistent-by-construction. If it can't (a parent field that isn't a stored column, a live join, or a cross-destination parent) it falls back to recomputing the upstream from raw, or is refused under `strictUpstreams`. Reported by `publisher_storage_chained_build_total{outcome}`.

## Safety / eligibility

A `storage=` source is refused at build time (HTTP 422 / failed build) when it can't be safely frozen-and-served:

- an **unbound free parameter** (a per-query template, no single relation to freeze);
- a **given reference** — a security refusal: a given-filtered table materialized once and served to everyone would leak filtered rows (row-level access binds per query);
- a **non-DuckDB-portable** served shape (checked post-build against the captured schema).

A source protected by `#(authorize)` should also not be materialized — the served virtual shape carries no gate to evaluate. That refusal is intended to land alongside the upstream transitive-`#(authorize)` enforcement it will reuse (#906); until then, serve authorize-gated sources live.

**Within-package collision check:** two *distinct* persist sources (different `sourceEntityId`) that resolve to the same physical target (`name=`/fallback + destination) would clobber each other. This is surfaced as a warning at load, and — behind `PERSIST_COLLISION_ENFORCE` — rejected at publish (staged so a pre-existing latent collision doesn't dark an already-published package). Cross-version and cross-package collisions are out of a single package-version's view and remain the host's responsibility.

## Observability

Under the `publisher` meter: `publisher_storage_serve_routing_total{outcome}` (the serve hit-rate KPI), `publisher_storage_chained_build_total{outcome}`, build/GC/eligibility counters, and a `served_from=storage` attribute on `malloy_model_query_duration` (absent otherwise, so an `off` deployment's histogram is unchanged). `storageServeBindings` is exposed on package status.

## Testing

Unit + integration coverage for the build session, per-build instance isolation, serve transform, routing, eligibility, collision validation, manifest tier-split binding, chaining, and the kill switch — validated end-to-end against a live DuckLake (Postgres source → DuckLake): fresh-catalog init, repeated builds, serve-from-storage, freshness-flip, and the manifest-driven serve binding. No Malloy-core change; `#904` FROM-bind quoting is applied at every bind site.

## Known limitations (intentional, called out)

- **Compile-time fallback only.** Once a query is routed, a run-time storage error surfaces to the caller (no silent live retry) — so the store is a hard dependency for routed queries under `on`. Two reviewers have asked for this to honor a binding's `freshnessFallback=live`; it is a follow-up, not a behavior change in this PR.
- **Nested/STRUCT/ARRAY columns collapse to `json` in the serve shape**, so a query that *traverses* a nested field is not reproducible over the stored table and falls back to serving live. Round-tripping the built table is covered; the traversal fallback is pinned by scenario.
- **A persisted `sql_select` is built but not read on the in-warehouse path.** Malloy substitutes a manifest table for a `query_source` but not for an `sql_select` (`getStructSourceSQL` looks up the manifest only in the `query_source` case), so a colocated `#@ persist` on an `sql_select` produces a table nothing queries. The `storage=` tier is unaffected — it routes through the virtual-source transform rather than manifest substitution. Both behaviors are pinned by scenario.
- **An `extend` of a persisted source inherits its `#@ persist` annotation**, which is by design in Malloy — but the extension then inherits the parent's `name=` too, so both resolve to the same physical target. `#@ -persist` opts the extension out.
- **Rolling back** below this version after building with `storage=`: delete the package's materializations first (an older build would bind a storage entry as a same-connection table it can't resolve).
- **Production-representative measurement** (warehouse → object-store DuckLake, at volume) is a follow-up; correctness is proven end-to-end, perf/cost characterization is not yet.
- The `#(authorize)` refusal is a follow-up gated on #906.


packages/server/src/service/model.ts
```diff
@@ -1956,7 +2316,8 @@ export class Model {
             rowLimit,
             givens: querySurfaceGivens,
             abortSignal,
-            buildManifest,
+            buildManifest: effectiveBuildManifest,
+            virtualMap: serveVirtualMap,
          });
       } catch (error) {
```
sagarswamirao, line 2328: **Runtime storage failure hard-errors even when the binding declares a live fallback.** Once a query is routed to the storage tier, `runnable.run(...)` executes against the stored table and a missing/corrupt table is caught here and re-thrown to the caller (MalloyError as-is, else a 400). The compile-time path has the full -> drop-views -> base-only -> live ladder, and the freshness path honors a per-binding `fallback` (`live`/`fail`) -- but a *runtime* storage error consults neither; it always surfaces.

For an orchestrating deployment that treats the store as a performance tier over a still-live source, a transient missing-table (a not-yet-converged rebind, an over-eager GC) becomes user-facing errors rather than a transparent live serve. Would it make sense for the runtime storage-error path to honor the same `freshnessFallback=live` the staleness path already does (drop the binding, re-run live), while `fallback=fail` keeps today's surfacing? That keeps a security-gated source failing closed but lets a perf-tier binding degrade gracefully.

packages/server/src/service/materialization_serve_transform.ts
```diff
@@ -0,0 +1,779 @@
...
+      .toUpperCase();
+
+   // Temporal-with-time first (so "TIMESTAMP WITH TIME ZONE" doesn't fall through).
+   if (base === "DATE") return "date";
+   if (
+      base.startsWith("TIMESTAMP") ||
+      base === "DATETIME" ||
+      base === "TIMESTAMPTZ"
+   ) {
+      return "timestamp";
```
sagarswamirao, line 220: **TIMESTAMPTZ is served under whatever timezone the DuckDB serve session defaults to.** `TIMESTAMPTZ` / `TIMESTAMP WITH TIME ZONE` maps to Malloy `timestamp` here, but nothing pins `SET TimeZone` on the build or serve session, so DuckDB renders the stored instant to a wall-clock using the process/host timezone. The instant is preserved at rest, but the rendered value can differ from a live query against the source warehouse, and can shift run-to-run if the serving host's timezone changes.

Suggest pinning `SET TimeZone='UTC'` on both the build and serve sessions (or carrying an explicit tz) so the rendered wall-clock is deterministic and matches live. Worth pairing with the sub-day-precision follow-up scenario 23 already notes.

packages/server/src/service/materialization_serve_transform.ts
```diff
@@ -0,0 +1,779 @@
...
+   // Arrays / nested collections carry as json (lossy but safe) for now.
+   if (/\[\s*\]/.test(raw)) {
+      recordServeShapeTypeFallback("array");
+      logger.warn(
+         "Mapping DuckDB array/collection type to json for serve shape",
+         {
+            duckdbType: raw,
+         },
+      );
+      return "json";
```
sagarswamirao, line 205: **Nested/STRUCT/ARRAY columns collapse to `json` in the serve shape, and the nested-traversal scenario has no executable coverage.** Any array or struct type maps to `json` here (safe and lossy). Scenario 30 (`nested-data-roundtrip`) asserts that a `nest:`ed `by_region` column "round-tripped through DuckLake with its data intact" and is queryable via `by_region.region` -- but the scenario is a `.md` description, not a test, and the serve unit specs only exercise scalar columns and a join-built `by_region` view, never `.field` traversal over a `json`-declared virtual column.

If a `::json` virtual column doesn't support record-field traversal the way a real repeated-record does, that query would fail the serve-shape compile and fall back to live (safe) -- but the scenario's stated contract wouldn't actually hold. Worth an e2e unit test that materializes a LIST-of-STRUCT, binds it, and runs `by_region.region` through the serve shape; if it doesn't hold, soften the scenario to "carried opaquely as json."

packages/server/src/service/package.ts
```diff
@@ -732,6 +809,47 @@ export class Package {
...
+    * An empty map clears the binding (reverts to serving live).
+    */
+   public bindColocatedServeManifest(entries: FreshnessManifest): void {
+      this.recordManifestBinding(entries);
+   }
+
+   public bindStorageServeBindings(
+      entries: Record<string, ManifestEntry>,
+   ): void {
+      this.storageServeBindings = deriveServeBindings(entries);
```
sagarswamirao, line 842: **Host-supplied serve bindings aren't re-checked for eligibility.** `bindStorageServeBindings` derives bindings and pushes them straight onto the models with no `assertMaterializationEligible` pass. For the publisher's own builds that's fine -- the build-time gate already refused ineligible sources. But this is the documented injectable seam where an orchestrating host can supply bindings directly; a binding for a `given`-filtered or authorize-gated source injected here would route to the frozen table with no gate to evaluate, which is exactly the row-level-access leak the build gate exists to prevent.

Since the serve virtual base carries no `#(authorize)`/`given` to enforce, could the serve-side binding path assert eligibility on host-supplied bindings (or assert an explicit "host bindings are pre-validated" contract) so the seam can't become an enforcement bypass?

packages/server/src/service/materialization_build_session.ts
```diff
@@ -0,0 +1,572 @@
...
+      // attach alias) — an unqualified name lands in the build session's own
+      // (private, throwaway) in-memory catalog, not the DuckLake/DuckDB store,
+      // so the rows would be captured for the schema and then vanish with the
+      // session. DuckLake's
+      // catalog swap is transactional, so CREATE OR REPLACE is atomic — no
+      // staging/rename dance needed on this path.
+      const target = quoteTablePath(
+         `${destinationName}.${physicalTableName}`,
+         "duckdb",
+      );
```
sagarswamirao, line 243: **The CTAS target re-quotes the table path while the serve path passes an already-quoted name through -- an author-quoted `name=` diverges write-vs-read.** The build target uses `quoteTablePath` (unconditional per-segment quote), but the serve side resolves via `quoteManifestTablePath`, which passes an already-quoted path through unchanged. For an author who writes `#@ persist name="My Table" storage=...`, the build creates a table whose name includes the quote characters, while the serve resolves the unquoted-inner form -- so the serve never finds the table, silently falls back to live, and the build is wasted.

DuckDB's case-insensitive matching hides the case-folding half of this, but the quote-character mismatch bites regardless. Using `quoteManifestTablePath` for the CTAS target (matching the serve side and the colocated path) keeps write and read byte-identical -- this is the `virtual`-render-site quoting verify that was flagged as not-yet-closed.

sagarswamirao:
**The `mz-storage` hammer scenarios don't appear to be wired into CI** (no reference under `.github/`, no hammer test script), including the security ones -- user-SQL-cannot-mutate-lake, endpoint-readonly, cannot-build-from-lake, and the visibility-leak guards. The unit specs cover the highest-value cases (real-compiler eligibility including the transitive-through-join walk, passthrough single-quote escaping, and the per-build DuckDB-instance isolation regression test), so the core is protected -- but the end-to-end security proofs rely on a manual run and won't catch a regression automatically. Worth a CI job that runs the `tags: security` scenarios, or porting the key ones into the integration suite.

sagarswamirao:
Two lifecycle hardening notes:

1. In `createIsolatedBuildSession`, `mkdtempSync` runs before `new DuckDBConnection`; if the constructor throws synchronously, the disposer that owns `rmSync` is never returned, so the temp working dir leaks on disk. Wrapping the two so a construction throw removes the dir closes it. (Latent -- I couldn't confirm the exact call shape here has a synchronous-throw path, so it's structural rather than a confirmed leak today.)

2. On a partial multi-source / chained build (a parent table is built, then a later source fails), the run is marked failed with no committed manifest, so the table-drop GC -- which reads the committed manifest -- can't reclaim the already-built parent table. Best-effort-dropping tables built earlier in the run on the failure path (mirroring the single-table gate-refusal drop) would avoid the orphan. For DuckLake this is at-rest data plus Parquet files, so it's worth closing before the tier carries real volume.

sagarswamirao:
One other thing we discussed that's worth highlighting here: Will the publisher support materializing and persisting a single version of a package to different locations?

If yes, does that mean we would run one job grouped by materialization type, or would there be different phases? Also, how would data retrieval work in that case? Would the publisher switch between one of the multiple connections configured at query time?

If no, would it be simpler to add `storage=<connectionName>` to the `publisher.json` so users don't have to specify it for each source repeatedly?

sagarswamirao:
The tutorial looks great: https://github.com/malloydata/publisher/blob/de4c8ebc0a071367fc3060980fa5c8b9ca40e2ad/docs/persist-storage-tutorial.md.

Once the design is finalized, could we add an integration test in one of the follow-up PRs that covers the workflow described in the documentation?

jswir:
A few non-blocking findings from a review pass (verified against `fb6aeee5`). The blockers are already handled — the Tier-3 credential redaction landed in `718cd5ea`, and the given/RLAC rebind gap is tracked on @sagarswamirao's `package.ts` thread — so these are should-fix / nits only.

**should-fix**

- **`config.ts:544` — `PERSIST_COLLISION_ENFORCE` fails open.** `getPersistCollisionEnforce` does an ad-hoc `raw.trim().toLowerCase() === "true"`, bypassing the file's own `parseBoolEnv` (`:192`, accepts `1/true/yes/on`, throws on garbage) that other flags use (`:233`, `:304`) and that the sibling `getPersistStorageMode` (`:516`) loud-fails with. So `PERSIST_COLLISION_ENFORCE=1` or `=yes` silently resolves to `false` — the publish-time collision check an operator believes they hardened stays warn-only. One-liner: `parseBoolEnv("PERSIST_COLLISION_ENFORCE") ?? false`.

- **`materialization_service.ts:1409` — chained (Tier-3) catch conflates infra failure with shape-ineligibility.** The redaction fix (`718cd5ea`) went in, but the catch is still unconditional: a genuine attach/CTAS infra error (auth/network/quota) from `buildDownstreamIntoStorage` is metered `inline_fallback` — same bucket as a legitimate "parent refinement not carried" — and triggers a redundant recompute-from-raw that, if the outage is destination-side, fails the same way again. `MaterializationEligibilityError` (`errors.ts`) already exists; throwing it only from the model-compile-failure site and rethrowing the rest as a distinct infra failure would separate the two signals.

**nits**

- **`connection_config.ts:274` — `ducklake` skips eager shape validation.** No `case "ducklake"` in `validateConnectionShape`, so the required `storage.bucketUrl` is only checked lazily at first ATTACH (`connection.ts:609`) — unlike `motherduck` (same duckdb family), validated at boot. A `case "ducklake"` checking `storage?.bucketUrl` would fail loud at config-load like its siblings.

- **`materialization_eligibility.ts:235` — second authorize-reachability walk.** `referencesAuthorize`/`walkForAuthorize` parallels `model.ts`'s `collectAllReachableGates`. The build-time context justifies a separate walk and the generic fail-closed form is arguably more robust, but the two won't inherit each other's fixes — worth a one-line note that they're intentionally distinct.

jswir (approved):

kylenesbit:
Nice work — the tier reuses the persist machinery cleanly (destination-scoped skip-if-unchanged, shared `evaluateManifestFreshness`, the `off` → serve-live safety). Two findings plus one forward-looking design question for the authors' radar.

### 1. Runtime storage errors don't honor `freshnessFallback=live` (should-fix)

There's an asymmetry in how the freshness `fallback` is respected. The compile-time ladder and the `freshServeBindings` gate both honor `live` / `stale_ok` / `fail` — but once a query is *routed* to storage, a runtime missing/corrupt-table error from `runnable.run()` in `Model.getQueryResults` is caught and re-thrown (MalloyError as-is, else a 400). It never consults the binding's `fallback`, so a binding that explicitly declared `fallback=live` still hard-errors on a transient missing table (a not-yet-converged rebind, an over-eager GC) instead of degrading to a live serve.

This is the same gap Sagar raised on the `model.ts` thread, and it's currently documented as a known limitation ("compile-time fallback only"). Given that `freshnessFallback` is a first-class part of the freshness contract, it's worth closing: on a runtime storage error, if the routed binding's `fallback` is `live`, drop the binding and re-run live; keep `fail` (and security-gated sources) surfacing as they do today. That makes the runtime path honor exactly the config the gate already reads, and keeps a perf-tier binding degrading gracefully while a fail-closed one still fails closed.

### 2. PR description is stale vs. the final code on naming (doc nit)

The "Naming & lifecycle" section says tables are named by `name=` verbatim with "no hashed suffix, no coexisting generations, no operator convenience view." That was true earlier, but `687ed69` (content-address physical tables so generations coexist) and `095d10b` (destination-aware GC) re-introduced content-addressed names (`<logical>__m<hash>`), the convenience view, and superseded-generation GC. Worth reconciling the description with the final behavior before merge, since the naming/lifecycle contract is what downstream readers will rely on.

### 3. Forward-looking: split build vs. serve connections (not this PR)

We'd like to make sure a follow-up direction is on your radar so the abstractions here don't foreclose it: **feature parity with Looker PDTs** — run the materialization build on one connection (e.g. an XLARGE warehouse) while serving on another (e.g. MEDIUM), where both connections point at the **same underlying data source** but differ only in credentials / non-data-impacting configuration (warehouse size, role, session params).

The foundation this PR lays is most of the way there: build-compute and serve are already decoupled (`buildSourceIntoStorage` takes `sourceConnection` and `destinationConnection` separately), connections resolve by name, and `applyConnectionFingerprint` already lets two distinct connections declare the same data identity — so a table built on one connection can content-address identically to the one served on another (or, cleaner: content-address on the serve connection only and treat the build connection as a pure execution detail).

The key requirement to design for now: **this has to generalize beyond DuckDB/DuckLake.** Today `STORAGE_DESTINATION_TYPES` is `ducklake`/`duckdb` and the passthrough is `bigquery`/`snowflake`/`postgres`. The build-vs-serve split we care about includes cases with no DuckDB in the picture at all — e.g. Snowflake→Snowflake, same account/db/schema, XLARGE build connection vs MEDIUM serve connection. So the "which connection executes the build" concept should be an engine-neutral connection role, verified to target the same data source, rather than something shaped around the DuckDB store. GC/drop and content-addressing should anchor on the *owning* (serve) connection; the build connection owns nothing at rest.

**Naming question for you:** with a build-connection role likely coming, does it make sense to rename `storage=` to `connection=`, or keep them distinct? Our lean is to **keep them distinct**, because they're different roles:

- `storage=` today = *where the materialized table lands and is served from* (a destination store, possibly a different engine, served via the virtual-source transform).
- the build-connection concept = *which connection executes the build compute* (same data source as serve, different creds/warehouse), independent of where the result lands.

A single `connection=` would conflate "where it lives" with "what builds it," and the two paths behave differently (cross-engine passthrough + virtual-source serve vs. same-DB `tableName` substitution). But since `storage=` will lock in vocabulary once it ships, it's worth settling the naming holistically now — including whether a third key (e.g. `build=` / a package-level `materialization.buildConnection`) is the right home for the build-connection selector. Curious how you're thinking about it.

kylenesbit (approved):
Approving. The `storage=` tier is a faithful, consistent extension of the existing persist functionality — it reuses the build orchestration, content-addressing, destination-scoped skip-if-unchanged, and (importantly) the same `evaluateManifestFreshness` gate, so freshness and schedule config carry through both tiers identically. Kill-switch and manifest tier-split behaviors respect the existing host-authoritative / never-hard-fail posture.

Two non-blocking follow-ups noted in my comment: (1) runtime storage errors should honor `freshnessFallback=live` (compile-time fallback only today), and (2) reconcile the stale "Naming & lifecycle" section with the content-addressing/GC that actually landed. Also raised a forward-looking design question about splitting build vs. serve connections (Looker-PDT parity) and the `storage=` vs `connection=` naming — for a future PR, not this one.

housejester:
Thanks all. Everything raised is either handled here or queued for an immediate follow-up PR; nothing looks like it should
hold this one.

@kylenesbit — the "Naming & lifecycle" section is actually still accurate, and I should
have made the history easier to read. Content-addressed names and the convenience view were in here, but then were reverted. Big reason was to keep the `storage=` mz behaving exactly like an in-warehouse one, just landing in a different
warehouse and dialect. Keeping the naming identical means everything an operator already knows about
`#@ persist` transfers, and the only question `storage=` answers is *which* warehouse.

### Already landed on this branch

- **Write-vs-read quoting divergence** (`materialization_build_session.ts:243`): confirmed,
  and worse than described — an author-quoted `name=` produced a hard 400 at run time, past
  the compile-time fallback, rather than a silent live serve. The CTAS target now uses the
  same quoting authority as the serve side (`quoteManifestTablePath`, #904), so write and
  read agree. The colocated path has the same divergence and is pinned as a known-red
  scenario rather than fixed blind.
- **Orphaned tables on a part-way failed build**: a failed run now reclaims the tables it
  created before the failure. Scoped to the orchestrated path, where names are host-assigned
  and unique by construction — a shared-name reclaim needs the cross-package/cross-environment
  collision check that doesn't exist yet, and widening it before that lands could drop a
  table another package is serving.
- **Nested/STRUCT coverage**: the traversal case now has executable coverage and the
  fallback-to-live behavior is documented in the tutorial.

### Going into the follow-up PR

Work is already done locally and needs its own review rather than a ride-along on this
approval:

- **`PERSIST_COLLISION_ENFORCE` fails open** (`config.ts:544`) — agreed, and the one-liner
  is right. The sibling `PERSIST_STORAGE_MODE` parse had a related gap (untrimmed,
  case-sensitive) that's already fixed on the branch; this one belongs with it.
- **Host-supplied serve bindings aren't eligibility-checked** (`package.ts:842`) — confirmed,
  `bindStorageServeBindings` runs no eligibility pass. The fix is to assert per binding and
  *skip* the ineligible one, degrading that source to live rather than failing the bind.
- **`createIsolatedBuildSession` temp-dir leak** (`materialization_build_session.ts`) —
  structural, as you said; `mkdtempSync` precedes the constructor.
- **Chained catch conflates infra failure with shape-ineligibility**
  (`materialization_service.ts:1409`) — good catch. Throwing `MaterializationEligibilityError`
  only from the compile-failure site and rethrowing infra errors separately fixes both the
  metering and the pointless recompute-from-raw against a destination that's down.
- **`ducklake` skips eager shape validation** (`connection_config.ts:274`) — a `case`
  checking `storage?.bucketUrl` to match `motherduck`.
- **Runtime storage errors honoring `freshnessFallback=live`** - two votes moves it off the
  "documented limitation" list. It's a follow-up rather than part of this PR because the
  routed-query error path has to distinguish a *missing table* (degrade to live) from a
  *security-gated* source and from `fallback=fail` (both must keep surfacing), and that
  deserves its own tests rather than being folded into a merge.

Plus a set of memory fixes found while soak-testing the tier, which are the main reason the
follow-up wants separate review.

### Deferred, with reasoning
- **The second authorize-reachability walk** (`materialization_eligibility.ts:235`) — keeping
  both, adding the note that they're intentionally distinct. The build-time walk is
  fail-closed and generic by design; making it inherit the serve-time walk's shape would
  couple two things that want to evolve separately.
- **CI for the `mz-storage` scenarios** — agreed this is the real gap, and it's now much
  more tractable: the suite is one command, self-contained, and runs in ~50s (it needs
  Docker for a throwaway Postgres). Wiring it is its own PR.

### On the build-vs-serve connection split

@kylenesbit - noted. engine-neutrality was definitely on my mind. i have some thoughts beyond/in addition to the mechanical 'port it to support X as a storage destination as well'. 

On naming: agreed on keeping them distinct, and the way I'd frame it is that these are
three independent axes that happen to be adjacent in the annotation:

- **`name=` — the logical / alias axis.** What the materialized table is called. Already
  live and load-bearing on the in-warehouse path, and unchanged by this PR.
- **`storage=` — the store / tier axis.** Which registered store the table lands in and is
  served from. It denotes a store, not a schema, and it is the only axis this PR adds.
- **a build-connection selector (or connection profile) — the execution axis.** Which connection runs the build
  compute, against the same underlying data as serve.

Collapsing any two of those into `connection=` (or `destination=` even, which was the initial stab), makes the annotation ambiguous, and the paths genuinely differ underneath: cross-engine passthrough
plus a virtual-source serve, against same-DB `tableName` substitution. Given three axes, the
third key wants its own home, but not sure where yet (on annotation seems too fixed/baked, but maybe it cascades like freshness, etc).

packages/server/src/service/model.ts
housejester: will follow up (see comment at bottom)

packages/server/src/service/materialization_serve_transform.ts
housejester: will follow up (see comment at bottom)

packages/server/src/service/package.ts
housejester: will follow-up.

packages/server/src/service/materialization_serve_transform.ts
housejester: will follow up.

PR #912, opened 2026-07-22
feat(materialization): incremental persist (refresh="incremental")

girishjeswani:
## Summary

Realizes the `refresh="incremental"` value of the persist `refresh` knob on the publisher build
path. Today a persisted source is always fully rebuilt (staging + atomic rename) on every run — fine
for most sources, but prohibitive when a per-row cost dominates the build (an `ML.GENERATE_TEXT`
classification, an embedding call, an expensive UDF): every refresh re-pays that cost for rows that
have not changed. `refresh="incremental"` makes such a source **build once, then update in place**.

## Behavior

- **`refresh="incremental"`** (on `#@ persist`) selects the incremental path. Unset / `"full"` keeps
  the default full-rebuild path, unchanged.
- **First run** (target absent): `CREATE TABLE <target> AS (<source SQL>)` — the warehouse infers the
  schema; DDL is never derived from Malloy intrinsic types.
- **Subsequent runs**: `MERGE INTO <target> USING (<source SQL>) ON <primary_key>`, `WHEN MATCHED
  THEN UPDATE` + `WHEN NOT MATCHED THEN INSERT`. In place (no staging), and idempotent under retry
  (a re-run is a no-op on already-merged rows, unlike a raw `INSERT`).
- The merge key is the source's Malloy **`primary_key`** — no new persist-level `unique_key` knob.
  An incremental source with no primary key is a `BadRequestError` at build time.

### Compile-safe self-reference construct

The reserved CTE **`__credible_incremental_keys`** lets a source refer to "what I have already
materialized" without a bootstrap problem — the dbt `is_incremental()` + `{{ this }}` analog, with no
compiler change. The author writes it with an empty body (matches nothing → compiles, and seeds
everything on the first run); on an incremental run the publisher rewrites the body to `SELECT
<primary_key> FROM <target>`. Because the filter sits **inside** the source SQL, an expensive step
runs only for new rows. The construct is optional (omit it → upsert every returned row).

### Force full rebuild (`forceFullRebuild`)

A new run flag (dbt `--full-refresh` analog) rebuilds every incremental source through the full
staging + rename path this run. Deliberately distinct from `forceRefresh`: the scheduler sets
`forceRefresh` on every run, so overloading it would silently turn scheduled incremental refreshes
back into full rebuilds.

## Backward compatibility

Additive and off by default. The current full-rebuild body is extracted **verbatim** as
`buildFullSource`; `refresh` unset dispatches straight to it. `getSQL()` is unaffected by the
annotation, so content-addressing and reuse-rebind are identical for full sources. The reuse-skip
exemption and the new validation rules fire only for incremental sources.

## Validation & dialects

- Publish gate: `refresh` must be `full` | `incremental`; incremental + freshness `fallback: "live"`
  is rejected (a live serve returns the delta, not the dataset).
- The generated MERGE uses an **unqualified** `SET` target column (`SET col = S.col`) for
  portability: BigQuery and Postgres 15+ reject a target-alias-qualified column there; Snowflake
  accepts unqualified. Supported on BigQuery / Snowflake / Postgres 15+; DuckDB persistence is
  unsupported generally, and Postgres < 15 has no `MERGE`.

## Testing

- New `materialization_incremental.spec.ts`: `refreshMode`, the SQL builders
  (`hydrateIncrementalKeysCte` incl. nested-paren balancing, `buildMergeSQL`), and the
  `buildOneSource` dispatch (full unchanged / first-run CTAS seed / MERGE / `forceFullRebuild` /
  missing-pk error / CTE hydration).
- Policy Rules for `refresh` and incremental+`live` in `persistence_policy.spec.ts` (also proves the
  compiler carries `refresh` through to the wire plan).
- `forceFullRebuild` validation in the controller spec; run-metadata shape updated.
- `tsc --noEmit` clean; all affected suites green.

## Docs

- `docs/materialization-incremental.md` — the annotation surface, the self-reference construct, build
  behavior, the escape hatch, validation, dialects, and back-compat.

## Notes

- Based on `main` and independent of the CLI schedule work (#909) — it needs only the scheduler
  service already merged in #888. For bars-style hosted use, pair with `scope: package` + hosted
  `freshness` (not the version-scoped cron).
- This is the publisher-side realization of the `refresh` roadmap item; it is intentionally narrower
  than the control-plane spec's fuller Phase E1 design (version-private zero-copy clone,
  `unique_key` + event-time watermark) and merges in place on the primary key. The two can converge
  later.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4069,6 +4069,15 @@ components:
           default: false
           description:
             Build a new table even when a source's sourceEntityId is unchanged.
+        forceFullRebuild:
+          type: boolean
+          default: false
+          description: >-
+            Escape hatch (dbt --full-refresh analog): rebuild every incremental
+            (refresh=incremental) persist source from scratch via the full
+            staging+rename path this run, instead of MERGE-ing new rows. Use
+            after a schema or logic migration. Independent of forceRefresh, which
+            only bypasses content-hash reuse for full-rebuild sources.
         sourceNames:
           type: array
           items:
--- a/docs/materialization-incremental.md
+++ b/docs/materialization-incremental.md
@@ -1,0 +1,137 @@
+# Incremental materialization (`refresh="incremental"`)
+
+By default a persisted Malloy source is **fully rebuilt** on every materialization run — the
+publisher stages a new table with `CREATE TABLE … AS (<source SQL>)` and atomically renames it into
+place. That is the right behavior for most sources, but it is prohibitively expensive when the
+source contains a per-row cost that dominates the build: an `ML.GENERATE_TEXT` classification, an
+embedding call, an expensive UDF. Reprocessing every row on every refresh burns that cost on rows
+that have not changed.
+
+`refresh="incremental"` makes a source **build once, then update in place**: the first run seeds the
+whole table, and every subsequent run `MERGE`s only the rows the source produces into the existing
+table. Combined with the compile-safe self-reference construct below, the source can filter itself
+down to *new* rows before the expensive step runs, so the per-row cost is paid once per row for its
+lifetime.
+
+This is the publisher-side realization of the `refresh` knob documented in the control-plane
+persistence spec (`ms2data/service/docs/persistence.md`, §3). It is intentionally a smaller design
+than the spec's roadmapped Phase E1 (which envisions version-private, zero-copy clones keyed on a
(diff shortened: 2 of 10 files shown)


girishjeswani:
Closing to replace with a Credible-free version — the reserved CTE and docs referenced 'credible', which does not belong in the standalone OSS publisher. Will reopen a clean PR.

PR #914, opened 2026-07-22
feat(server): fast, transparent first boot for GitHub-hosted packages

mlennie:
The first `npx @malloy-publisher/server` boot spent most of its time cloning the full history of this repo just to copy out the three example packages, and did it silently. This PR makes that boot fast and observable.

**Shallow clone.** Package downloads now clone with `--depth 1`. The clone's git history was always discarded (packages are copied out of the temp clone, which is then deleted), so the history fetch was pure wasted transfer. Measured on the bundled default config against real GitHub: end-to-end first boot drops from ~20s to ~5s.

**Download progress.** The boot log now says which packages are being fetched for which environment, prints throttled clone progress (stage changes plus 25% steps), and a `Mounted package "x" (i/N)` line per package, so a first boot no longer looks hung.

**Readiness line.** When initialization completes and the server reports `serving`, it emits one stable line: `PUBLISHER_READY url=http://localhost:4000 environments=1 packages=3 loadErrors=0`. Scripts and agents can wait for the token instead of polling; `/api/v0/status` remains the canonical check, and the counts carry the "serving does not mean everything loaded" caveat inline (a package that failed to mount is in `loadErrors=`, not `packages=`). If initialization throws, `PUBLISHER_INIT_FAILED` is emitted instead, with the message run through `redactPgSecrets`, so a waiting script fails fast on a broken config. Counts come from load-free state via a new read-only `Environment.getPackageStatuses()` accessor, never via `listPackages()`, which would compile every package at boot and delay the signal.

Tests: a real `file://` shallow-clone proof (fails without the `--depth 1`), readiness-line specs for the healthy, failed-package, skipped-environment, and failed-boot shapes, URL derivation edge cases (port override, `0.0.0.0`/`::` and explicit IPv6 hosts, `--port 0`), and progress-reporter throttle and never-throw specs. README and AGENTS.md document the contract and correct the stale first-boot time claim.

Verified live: the default GitHub config boots to `PUBLISHER_READY` in ~5s vs ~20s on 0.0.229; a restart (database-restore path) and a runtime `POST .../packages` add (also shallow now) both verified; a bad-package boot prints `loadErrors=1`, matching `/api/v0/status`.


Sha-Bang:
**Blocker:** Windows CI fails on the new hermetic shallow-clone test (`clonePackageRepository > produces a depth-1 clone of the source's tip`).

Cause is the test harness, not the production clone change: it sets `GIT_CONFIG_GLOBAL` / `GIT_CONFIG_SYSTEM` to `os.devNull` (`\\.\nul` on Windows), and Git for Windows errors with `fatal: unable to access '//./nul': Invalid argument`. Use a temp empty config file (or another Windows-safe path) instead of `os.devNull`.

Everything else looks good once that lands.

mlennie:
@Sha-Bang Good catch, thanks. Fixed in 9115b7f2: the harness now writes an empty temp config file and points GIT_CONFIG_GLOBAL/GIT_CONFIG_SYSTEM at it instead of os.devNull, which isolates the spec from the runner's git config on every platform. Full local gate is green (1223 unit / 225 integration / 7 skills); windows-latest will confirm on this push.

Sha-Bang (approved):

PR #915, opened 2026-07-22
fix(server): redact URL-form connection strings in redactPgSecrets

mlennie:
## What

`redactPgSecrets` covered only keyword-form `password=...`. A URL-form `connectionString` (`postgres://user:pass@host/db`) supplied verbatim in config passed through untouched, so the cleartext password could reach log lines and the redaction-guarded response paths that already funnel through this helper. This widens the helper to also redact URI userinfo passwords, keeping the username visible (as the keyword pass keeps `user=`).

This is the pre-release fast-follow @housejester asked for on #898: keyword-form was covered, the URL form was not, and #898 is what moved these strings from a log line into an HTTP response body.

## Why the password reaches those strings

DuckDB's postgres extension embeds the full DSN in ATTACH failure messages. Verified live against DuckDB 1.5.2: a failing `ATTACH 'postgres://alice:secret@host/db' (TYPE postgres)` produces `IO Error: Unable to connect to Postgres at "postgres://alice:secret@host/db": ...`, password in the clear. The keyword form embeds `password=secret`, which the old regex already caught; the URL form was the gap.

## How

Three passes in `redactPgSecrets`, ordering load-bearing (URI passes run before the keyword pass, otherwise the keyword pass's `\S+` eats a URI's `@host/db` tail when a password contains the text `password=`):

1. Any-scheme URI userinfo `scheme://user:pass@`. The password class is greedy to the last `@` before a `/`, matching how WHATWG URL and libpq split userinfo, so a literal `@` in a password redacts fully; bounding at `/` keeps path-`@` URLs (`https://h:443/@scope/pkg`) and comma-joined URI lists from collapsing. The username class allows an empty username (`postgres://:pw@h`) and a literal `@`, so Azure's `user@servername` form (`postgres://myadmin@srv:pw@host`) redacts rather than passing through; it excludes `?#` so a passwordless URI with `:...@` in its query is not mangled.
2. A postgres-only mop-up for the one shape pass 1 fails open on: a raw `/` inside the password (RFC-invalid, but user connection strings flow through verbatim). Scheme-restricted so it can never touch an `https` URL.
3. The original keyword-form pass, unchanged.

Usernames stay visible. Over-redaction is the accepted safe direction (for example `?password=a&sslmode=b` also loses the `&sslmode=b` tail).

## Verification

30 unit cases in `pg_helpers.spec.ts` (up from 3), including percent-encoded and literal-`@` passwords, Azure `user@servername` usernames, IPv6 hosts, comma-joined and whitespace-separated multi-URI messages, path-`@` and query-`@` non-secrets left intact, email prose untouched, a scheme abutting a word char, and the exact DuckDB error shape above. Reverting the fix fails the URL-form cases, so the coverage is real. Full root gate green (typecheck, lint, prettier, unit + integration + skills, 0 failures). Live check: a DuckLake connection with a URL-form catalog `connectionString`, including the Azure `user@servername` form, now logs `postgres://myadmin@myserver:***@...` through the real helper on attach, with no cleartext password anywhere in the output.

## Scope

Redaction helper plus its spec, plus a one-line update to the `environment.ts` comment that requested this widening. Related gaps found while verifying are left as follow-ups rather than widening this PR: the connection-test endpoint returns its error message unredacted, and `buildPgConnectionString` emits discrete-field values unquoted so a whitespace password truncates under the keyword pass (and fails to connect). Both live in `connection.ts` and are separate changes.


housejester (approved):
Approving — this is a careful, well-tested hardening of `redactPgSecrets` that strictly improves on `main` (a mixed-version fleet only ever redacts *more*), and all three call sites use the result display-only, so the over-redact bias can't break anything functional. The common valid-DSN path is airtight against the regexes.

Two non-blocking residual-leak notes on malformed input, neither gating:

**1. Pass-2 comment slightly overstates its guarantee.** The comment says pass 2 mops up "a raw `/` in the password" for postgres URIs. That holds *only when no `@` precedes the `/`*. With both — `postgres://u:p@a/b@h/d` — pass 1 backtracks to the password-internal `@`, writes `postgres://u:***@a/b@h/d`, and that inserted `***@` then blocks pass 2's first-`@` mop-up, so the `a/b` fragment survives:

```
in       : postgres://u:p@a/b@h/d
after p1 : postgres://u:***@a/b@h/d
after p2 : postgres://u:***@a/b@h/d
```

Requires a doubly-RFC-invalid password (raw `@` *and* raw `/`, both unencoded), so real-world exposure is low. Suggest either tightening the comment to note pass 2 recovers raw-`/` passwords only when no `@` precedes the `/`, or (if you'd rather close it) decoupling the pg mop-up from pass 1's output.

**2. Non-postgres scheme + raw `/` password leaks.** Pass 1 is any-scheme but the raw-`/` mop-up is postgres-only, so `mysql://u:pa/ss@h/d` passes through with the password intact. This is inside your documented "pass 2 is scheme-restricted" decision and non-pg schemes are incidental to a pg helper — accepted scope, just flagging the any-scheme pass 1 sets up an expectation the mop-up doesn't meet for that one shape.

Both are strictly narrower than what `main` leaks today, so ship it. Nice work on the 30-case coverage and the monotonic-not-idempotent property test.

— Plucky Numbat

mlennie:
@housejester Thanks for the careful read. Merging as approved.

On 1: agreed, and nice catch on the mechanism. With both a raw @ and a raw / in the password, pass 1's inserted ***@ satisfies the mop-up's first-@ before it can reach the raw /, so the post-@ fragment survives. I have a small follow-up PR queued in the same area (routing testConnectionConfig's errorMessage through redactPgSecrets), so I'll tighten the pass-2 comment there rather than re-spin this one on a doubly-invalid input.

On 2: yes, that's the documented scheme-restriction decision. A pg helper covering non-pg schemes with RFC-invalid raw-/ passwords felt like the wrong trade against mangling https URLs; happy to revisit if a real non-pg DSN ever shows up in these error paths.

PR #920, opened 2026-07-22
docs(malloy-review): reword rubric line to avoid secret-scanner false positive

ndrcndn:
## What

Reword one word in `skills/malloy-review/reference/rubric-correctness.md` (line 114)
so it no longer trips secret scanners.

The construct list read `PIVOT/conditional aggregation`. Secret-scanning tools
(gitleaks `generic-api-key`, and similar keyword+entropy heuristics) misread the
slash token `PIVOT/conditional` — preceded by the keyword `access` — as an API key,
producing a false positive on this documentation prose.

Change: `PIVOT/conditional aggregation` → `PIVOT and conditional aggregation`.
Identical meaning; the scanner no longer matches.

## Why

Downstream projects vendor this skill file byte-identical and run secret scanners in
CI, where this line surfaces as a recurring false positive that can only be silenced
with a scanner allowlist. A one-word reword upstream removes the false positive at the
source for every consumer.


kylenesbit (approved):
Thank you!

PR #922, opened 2026-07-23
feat(server): read .xlsx package data files in place

mlennie:
Makes an Excel file in a package readable with no conversion step, the same way parquet and csv already are.

## What it does

- Bakes DuckDB's core `excel` extension so `.xlsx` reads work offline and in Docker (the docker smoke test now asserts it loads with `--network none`). Online, npx users already worked via DuckDB's autoinstall on first use.
- Lists `.xlsx` files as embedded package databases (schema and row count) alongside parquet and csv. `duckdb.table('data/x.xlsx')` already resolves through DuckDB's replacement scan, so no Malloy change was needed. For a specific sheet or other read options, a SQL source works: `duckdb.sql("SELECT * FROM read_xlsx('data/x.xlsx', sheet = 'Q2')")`.
- Teaches the agent skills to read spreadsheets in place instead of converting them, including the `read_xlsx` sheet and type-inference quirks.

## Resilience

Widening the data listing to `.xlsx` means Excel's owner file (`~$name.xlsx`, written whenever a workbook is open) and any partial or corrupt spreadsheet would reach the schema probe. Left alone, one such file would fail the whole package load. The listing now skips `~$` owner files before probing and omits a file whose probe throws with a warning, rather than dropping the package. A model that actually references a bad file still fails loudly via the worker compile path.

## Testing

Full root gate green (typecheck, lint, prettier, unit + integration + skills). Verified end to end against a real server: a package with a `~$` owner file, a corrupt `.xlsx`, and valid spreadsheets serves the valid data and lists only the good files, under both the default and `local-only` extension-fetch policies. The new probe unit test skips rather than fails on a runner that cannot reach the extension CDN.


jswir:
## Peer review

The `.xlsx` support, extension baking, and offline CI assertion are clean and follow existing precedent (`.csv` was added the same way; tests use the real DuckDB engine with a committed fixture and pin concrete values). The main concern is the error-resilience change in `readDatabases`: it's broader than the xlsx feature it rides in with — it silently changes fail-loud behavior for pre-existing `.parquet`/`.csv` too, collapses distinct failure signals into one silent branch, and bypasses the codebase's existing `warnings` surface. A secondary design question: `excel` is the only baked extension that skips the `installAndLoadExtension` mechanism established by #905.

### Important

**1. `packages/server/src/service/package.ts:1331-1361` — the per-file `catch` is not xlsx-scoped and collapses distinct failure signals.**

Three problems bundled here, all newly introduced:

- **It reverts fail-loud for `.parquet`/`.csv`, not just `.xlsx`.** Pre-PR, `readDatabases` was a bare `Promise.all` whose rejection propagated through `package.ts:397-399` into `Package.create`'s outer catch and surfaced in `status.loadErrors` (the comment at `package.ts:363` shows the worker isolation was written specifically *not* to re-map `readDatabases`'s own errors — those were meant to bubble). A corrupt `.parquet` now silently vanishes from the listing instead of failing the load. This is a behavior change to already-shipped formats, not called out in the PR description, `api-doc.yaml`, or tests.
- **It collapses corrupt-file vs systemic/transient signals.** A genuinely bad file (permanent, evidence-carrying) and a transient/systemic failure (shared-connection fault, `excel` unavailable under `EXTENSION_FETCH_POLICY=local-only`) funnel into the same "warn and omit" branch. Under a shared connection (`package.ts:1330` resolves one `conn` for all probes), a single connection-level fault could present as several valid files silently disappearing at once.
- **It uses only `logger.warn`, bypassing the house `warnings` surface.** The codebase already has `ApiPackage.warnings` (`package.ts:60`, assembled at `636`) for exactly "non-fatal but worth surfacing" (renderTagWarnings, exploreWarnings). A dropped database file is the same category but gets server-log-only treatment, so `GET …/packages/{pkg}/databases` (the discovery surface an agent hits *before* writing a model) shows fewer databases with zero signal why, and the list is cached until reload (`package.ts:179`).

Recommendation: keep the resilience (legitimate for corrupt spreadsheets), but narrow the catch to the actual read, distinguish a bad-file result from a systemic error (let systemic errors fail the load as before), and push skipped files into `metadata.warnings` so discovery consumers see them.

**2. `packages/server/scripts/bake-duckdb-extensions.js:42` + `packages/server/src/service/connection.ts` — `excel` bypasses the `installAndLoadExtension` mechanism established by #905.**

Every other baked extension (`httpfs`, `aws`, `azure`, `postgres`, `ducklake`, `bigquery`, `snowflake`) goes through `installAndLoadExtension`, whose docstring states the invariant that DuckDB's implicit auto-load is "never load-bearing." For `excel` it now *is* load-bearing — `getDatabaseInfo` runs `duckdb.table('…xlsx')` with no install site. Consequence under `local-only` with an unbaked `excel`: every other extension throws the crafted actionable message ("Bake the extension… or set EXTENSION_FETCH_POLICY=on-demand", `connection.ts:186-200`), but xlsx instead hits finding 1's generic catch and disappears with a generic warning. Design question, not a hard blocker — either route `excel` through `installAndLoadExtension` for policy/error parity, or update the `connection.ts` docstring and `docs/connections.md` to document `excel` as a deliberate, explained exception (the bake-script comment notes it, but the invariant's own docstring doesn't).

### Nit

**1. `packages/server/src/service/package.ts:1347-1349` — the safety claim is untested.** The comment states as fact that a model referencing a silently-dropped file "still fails the package load loudly via the worker compile path." It's plausible (worker compile runs in parallel via `package.ts:397-399`), but no test exercises the interaction. This claim is the keystone that makes skip-don't-fail safe; add an integration test where a model references a file the listing dropped, or soften to "should."

**2. `docs/connections.md` — the `duckdb.table('data/budget.xlsx')` example references a fixture that doesn't exist** in `examples/storefront/data/` (only `.parquet`/`.csv` are there). Illustrative-only; minor consistency drift.

**3. PR hygiene** — the resilience change (finding 1) is a larger-blast-radius change to pre-existing formats riding inside the xlsx feature. At minimum call out the `.parquet`/`.csv` contract change explicitly in the PR description and cover a broken `.parquet`/`.csv` case in tests (currently only `broken.xlsx` is tested).

### Considered and dismissed

- *Security — new complex parser (zip+OOXML) over attacker bytes, no resource bounds:* not a new trust boundary. Anyone who can drop an `.xlsx` can already drop a `.malloy` running arbitrary SQL; package authors are trusted, and no resource bounds exist for `.parquet`/`.csv` either.
- *Case-sensitive `.endsWith(".xlsx")` (misses `.XLSX`):* pre-existing for `.parquet`/`.csv`; the PR extends the same gap rather than introducing it.
- *`~$` filter shape:* verified it's an allowlist (`.parquet`/`.csv`/`.xlsx`) with `~$` as an additional exclusion, not a blocklist boundary — correct shape.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

jswir:
## Runtime finding: xlsx files with spaces in the filename are silently dropped from the listing

Ran this branch against a real server (`bun packages/server/src/server.ts`, default `on-demand` fetch policy, `excel` autoinstalled) with a package containing a real-world spreadsheet named `Workshop_June12_Attendee list.xlsx`, plus controls. Result matrix:

| File | Result |
|---|---|
| `attendees_nospace.xlsx` (byte-identical copy, no space) | ✅ Listed, **35 rows**, columns detected |
| `Workshop_June12_Attendee list.xlsx` (space in name) | ❌ **Silently omitted** from `GET …/databases`; `status.loadErrors: none` |
| `broken.xlsx` (corrupt) | ✅ Correctly skipped (resilience works as intended) |
| `~$Workshop….xlsx` (Excel owner file) | ✅ Correctly filtered before probing |

### Mechanism

`getDatabaseInfo` (the probe this PR extends to `.xlsx`) builds `source: temp is duckdb.table('<abs path>/Workshop_June12_Attendee list.xlsx')`. Malloy's `duckdb.table()` rejects a path containing a space:

```
Invalid DuckDB table path: "…/Workshop_June12_Attendee list.xlsx" — expected an identifier
path, a quoted identifier path, a single-quoted literal ('foo.csv'), or a simple file/URL path
(some file/URL paths require quoting). … use a SQL block instead: connection.sql("""SELECT * FROM …""").
```

The new `try/catch` in `readDatabases` then swallows that compile error into a `logger.warn` and omits the file. So the API returns `loadErrors: none` and a listing that just doesn't contain the spreadsheet — a **valid, readable file is indistinguishable from a corrupt one, with no user-facing signal.** This is the "collapsed failure signals + no `warnings` surface" concern from the review made concrete: the failure here isn't corruption, it's an unquoted-path bug, and it hits the single most common Excel naming convention (spaces in the name).

### Confirmed it's the space, not the bytes

- Byte-identical copy renamed without the space → lists fine (35 rows).
- A model using `duckdb.table('data/Workshop_June12_Attendee list.xlsx')` → fails to compile and takes the **whole package** down via `loadErrors` (this does confirm the "a model referencing a bad file fails loudly via the worker compile path" claim in the `readDatabases` comment — but it also means the pattern the docs/skills recommend is unusable for spaced filenames).
- Workaround verified working end-to-end: `duckdb.sql("""SELECT * FROM read_xlsx('data/Workshop_June12_Attendee list.xlsx')""")` compiles and queries → `row_count = 35`, generated SQL `SELECT COUNT(1) FROM (SELECT * FROM read_xlsx('data/Workshop_June12_Attendee list.xlsx'))`. The docs mention `read_xlsx` only for sheet selection, not as the required form for spaced names.

The existing tests don't catch this because the fixture is `database.xlsx` (no space).

### Suggested fixes

1. **Quote the path in the probe.** `getDatabaseInfo` now handles files whose names it doesn't control, so it should read via a quoted `read_xlsx(...)` SQL block (or otherwise escape the path) rather than bare `duckdb.table('…')`. This is also the whole-bug-class fix — the same unquoted-path failure would hit any `.parquet`/`.csv` with a space today.
2. **Surface skipped files via `metadata.warnings`**, so a dropped file is visible in the API rather than silent in the server log.
3. **Document spaced filenames** in the skill/docs — `duckdb.table()` breaks on them; `read_xlsx` SQL source is the workaround.
4. **Add a spaced-filename fixture** to `package.spec.ts`.

<sub>🤖 Generated with [Claude Code](https://claude.com/claude-code)</sub>

kylenesbit (approved):

PR #923, opened 2026-07-23
feat(server): semantic malloy_getContext retrieval via an embedding provider

mlennie:
Setting `EMBEDDING_API_KEY` switches `malloy_getContext`'s question retrieval (tier 4) from lexical lunr/BM25 to embedding cosine similarity. This closes the known token-gap misses the lexical baseline was built to measure: "departure delay" never matched a field named `dep_delay`, and a question phrased in words the model never uses ("refund percentage" for `return_rate`) came back empty. Without the key, behavior is unchanged, byte for byte.

## How it works

- A minimal OpenAI-compatible `/embeddings` client over global fetch, no new dependencies. `EMBEDDING_MODEL` defaults to `text-embedding-3-small`; `EMBEDDING_API_BASE` points at any compatible endpoint (Ollama, vLLM), so everything can stay on-machine; `EMBEDDING_DIMENSIONS` is optional and omitted from requests when unset. The key must be set explicitly: an ambient `OPENAI_API_KEY` is deliberately not read, so having one exported never silently sends model metadata to an external endpoint.
- Each package's entities (the same curated set the tool already indexes: sources, views, named queries, dimensions, measures) are embedded as humanized name plus annotation text and cached in an `entity_embeddings` table in `publisher.db`, keyed by content hash, so only new or changed entities re-embed, across restarts too. Searches run in DuckDB with `list_cosine_similarity` (brute force; at the entity counts one Publisher serves, a vectorized scan beats carrying a vector-index extension), with a 0.20 similarity floor so an off-topic question still returns an empty result.
- The first question against a package kicks off the embedding sync in the background and answers lexically; once the sync lands, later questions are semantic. Provider failures fall back to lexical with a 60s cool-down; packages above 5,000 entities stay lexical. When the provider is configured the payload carries `retrieval: "semantic" | "lexical"` and semantic entities a `score`; unconfigured payloads are unchanged.
- Concurrency is handled with a per-package mutex, a globally unique generation counter, and orphan guards, so package reloads, deletions, and provider changes (including a backend that changes its dimensionality behind the same model name) never serve stale or empty results marked as semantic. Deleting a package or environment cleans up its cached vectors.
- `get_context_eval.ts` becomes a lexical-vs-semantic A/B: cases cover the bundled `examples` environment (including token-gap cases) so it runs against a default-config server out of the box, skips absent environments, and labels every case with the server's retrieval mode. On the bundled storefront package the lexical baseline scores 3/5 with both gap cases returning nothing; the semantic run closes them.

## Security notes

- The key travels only in the Authorization header. Auth-failure response bodies are dropped from error messages (they can reflect key material) and literal key occurrences are scrubbed from other error excerpts; `apiKey`/`api_key` join the log-redaction key list.
- What leaves the machine when enabled: entity names, their annotation text, and query strings. Model source, data, and results are never sent. Documented in `docs/configuration.md`.

## Testing

- Unit specs for the provider (batching, truncation, error hygiene, config gating) and the index (hash-diff sync, upserts across model switches, the dimensionality heal including the mixed-dims state, purge backoff, cross-instance memo invalidation, torn-write visibility, deletion cleanup, the non-blocking heal). Concurrency pins were mutation-checked: reverting a fix makes a test fail.
- An integration suite boots the real server against a deterministic in-test embedding endpoint and covers the cold-start flip to semantic, warm-cache behavior, `sourceName` narrowing, and the marked lexical fallback when the endpoint fails. Runs on the Windows/macOS/Linux CI matrix.
- Verified live: eval A/B against a running server (lexical 3/5 baseline, semantic closes a gap case), cache persistence across restarts (zero re-embeds on reboot), and the no-key boot serving a payload identical to current main.

Docs: env vars and a "Semantic retrieval for malloy_getContext" section in `docs/configuration.md`, plus `docs/ai-agents.md` and `README.docker.md` updates.


jswir (commented):
## Peer review — semantic `malloy_getContext`

Strong, carefully-engineered PR — the concurrency machinery holds up (I ran the unit suite against real DuckDB: green), failure signals stay distinct, SQL is parameter-bound, key redaction is sound, and the unconfigured payload is genuinely byte-identical. I also booted it live on an isolated instance; it serves and ranks correctly. The feedback below is what I'd want addressed, not a rejection.

### Blockers (pre-merge)

**1. Access-control metadata is embedded and sent to the provider.**
[`get_context_tool.ts:79-92`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/mcp/tools/get_context_tool.ts#L79-L92) + [`embedding_index.ts:86-89`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/mcp/tools/embedding_index.ts#L86-L89). `docText` falls back to *all* annotation lines when there's no `#(doc)`, so an entity's `#(authorize)` predicates get embedded. Confirmed live: booting the PR against a local logging endpoint and querying `governed-analytics`, the server sent this input for `orders_secured`:

```
"orders secured: #(authorize) \"$ROLE = 'admin'\" #(authorize) \"$TENANT = 'acme' or $TENANT = 'globex' or $TENANT = 'initech'\""
```

plus an internal `#(malloy) drillable ...` line from another entity. With the default `EMBEDDING_API_BASE` that goes to OpenAI. It's disclosed in `docs/configuration.md`, but an operator turning this on for retrieval quality won't read that parenthetical as "my row-level-security rules and tenant names leave the box." Suggest embedding `#(doc)` text only, or explicitly excluding predicate-bearing annotation kinds (`#(authorize)` etc.) from the fallback.

**2. Config never validates at boot.**
[`config.ts:107-151`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/config.ts#L107-L151). The docstring promises malformed companion vars "surface loudly in the log at startup," but `getEmbeddingConfig()` is only called on the tool path — its three siblings (`getExtensionFetchPolicy`, `getMemoryGovernorConfig`, `getMaterializationSchedulerConfig`) are all called in `server.ts` (lines 208/209/237). So a bad `EMBEDDING_API_BASE`/`EMBEDDING_DIMENSIONS` boots clean and only surfaces as a `logger.warn` on the first query that reaches tier 4 — or never. This is the same issue #905 caught on `getExtensionFetchPolicy` and fixed by adding a boot-path call. Add a `getEmbeddingConfig()` call on the `server.ts` boot path (also resolves the per-call re-derivation).

### Should-fix (real, narrower trigger — resolve now or file as fast-follows with behavior documented)

**3. Global failure cooldown over-degrades healthy packages.**
[`embedding_index.ts:139,167-169`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/mcp/tools/embedding_index.ts#L163-L169). Every other piece of state here is scoped per package, but `providerFailureAtMs` is process-wide and `inCooldown()` is checked at the top of `trySemanticSearch` for all packages. A single-query timeout or a dims-mismatch backoff on package A (line 658) forces package B — healthy, correctly cached — to lexical for 60s. Scope the cooldown per-package for the timeout/mismatch triggers; keep it global only for a genuine connection-level failure.

**4. Dims-mismatch heal oscillates instead of terminating.**
[`embedding_index.ts:651-696`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/mcp/tools/embedding_index.ts#L651-L696). The backoff branch (657-669) calls `markProviderFailure()` but never advances `meta.lastPurgeAtMs` — only the purge branch does (693). So for a durably-inconsistent provider (e.g. a mid-upgrade replica set serving mixed dims under one model name), after each 60s cooldown clears the guard is false again and it purges + full-re-embeds the whole package — indefinitely, just throttled to ~once/60s. The code's own comment (651-655) says the backoff exists so it won't "loop full re-embeds indefinitely," but it still does. Advance `lastPurgeAtMs` in the backoff branch to a real terminal state, or document oscillation as accepted. The test at `embedding_index.spec.ts:961` only covers one cycle.

### Nits

- **Vectors shape-validated but not numeric-validated** — [`embedding_provider.ts:150-164`](https://github.com/malloydata/publisher/blob/313b03ddd6c9b64608d05c61541df6b9162ed7c9/packages/server/src/service/embedding_provider.ts#L150-L164). A provider returning `[1, null, "x"]` passes and is stored, then silently dropped later with no error surfaced. Add a `Number.isFinite` check.
- **Malformed-config throw path untested at the tool level** — `get_context_tool.ts:86-97`. Every semantic tool-spec test uses `_setEmbeddingProviderForTests`, which bypasses real env parsing, so nothing exercises this catch block.
- **`kind|source|name` entity key hand-rebuilt in three places** — `embedding_index.ts:100-101`, `get_context_tool.ts:472-483` and `:170`. Extract one shared helper.
- **`apiKey`/`api_key` added to `SENSITIVE_KEY_NAMES`** (`logger.ts:110-111`) with no locking test — #878's reviewer asked for exactly that (a reflective test asserting each credential field is redacted).

### Design notes (not defects — recommendations)

- **Lazy indexing:** keep the lexical *floor* — dropping it would make a paid, data-egressing external dependency mandatory on the hot path and break the opt-in / byte-identical default. But the "first question per package silently answers lexical, later ones are semantic" behavior is a surprising warm-up. Consider triggering the sync on **package load/reload** (still background, still lexical fallback) instead of on first query, so answers are semantic from question #1.
- **Local model is a strong prod config:** `EMBEDDING_API_BASE` already accepts any OpenAI-compatible local server (Ollama/vLLM/TEI), and the few-hundred-entity scale is trivial locally. Pointing at a local model **eliminates blocker 1's egress entirely** — worth having the docs lead with local rather than OpenAI-as-default.

*Reviewed with a multi-agent pre-submission pass; findings verified by reading the code and a live isolated run (feature enabled against a local logging endpoint).*

kylenesbit:
## Review — semantic `malloy_getContext`

I checked out the branch and verified the three properties I care most about, rather than reading the description alone:

- **Embeddings really do persist in DuckDB.** `entity_embeddings` lives in the durable `publisher.db` handle (not the `:memory:` analytical connection), and it's created from `initializeSchema` with `CREATE TABLE IF NOT EXISTS`, so an existing database picks it up on upgrade without `--init`. I wrote a throwaway spec that syncs a package, closes the DuckDB file, resets all in-process state, reopens the same file with a fresh `Package` instance, and queries again: rows survive, the re-sync logs `embedded: 0`, exactly one embedding call follows (the query vector), and the ranking is identical. Excluding `dims` from the hash diff is the right call for providers that ignore the `dimensions` parameter.
- **Ranking is pure cosine similarity, no reranking.** One DuckDB query — `list_cosine_similarity`, a 0.20 floor, `ORDER BY score DESC, entity_name`, `LIMIT`. No second stage, no cross-encoder, no fusion with the lunr scores; semantic and lexical are strictly either/or. The `entity_name` in the ORDER BY is only a determinism tiebreak.
- **No generated descriptions.** The client only ever calls `POST <base>/embeddings`; there is no completion or summarization anywhere in the branch. `doc` is passed through verbatim from `#(doc)` via the pre-existing `docText()`, exactly as the lexical path already did. The only additive result field is the numeric `score`.

The 24 unit specs pass against real DuckDB on my machine.

Approving, but I'd like these four fixed before merge. They're the same items @jswir raised; I verified each independently against the code and agree with all of them.

### 1. `#(authorize)` predicates are embedded and sent to the provider

`docText()` falls back to joining **all** annotation lines when an entity has no `#(doc)`, so row-level-security predicates and tenant names become embedding input and, with the default `EMBEDDING_API_BASE`, leave the box for OpenAI. This is the serious one. Since nothing on this path needs the fallback, embed `#(doc)` text only and drop the fallback for embedding purposes (the response `doc` field can keep whatever behavior it has today — that's a separate decision).

### 2. Embedding config never validates at boot

`server.ts` calls `getExtensionFetchPolicy()` and `getMemoryGovernorConfig()` on the boot path but never `getEmbeddingConfig()`, so a malformed `EMBEDDING_API_BASE` or `EMBEDDING_DIMENSIONS` boots clean and only surfaces as a warning on the first query that reaches tier 4 — contradicting the function's own docstring, and the same issue #905 caught on `getExtensionFetchPolicy`. Add the boot-path call.

### 3. The provider cooldown is global while everything else is per-package

`providerFailureAtMs` is module-level and `inCooldown()` gates `trySemanticSearch` for every package, so one package's query timeout or dims-mismatch backoff degrades every other healthy, correctly-cached package for 60s. Scope it per package for the timeout and mismatch triggers; keep it global only for a genuine connection-level failure.

### 4. The dims-mismatch heal oscillates instead of terminating

The backoff branch calls `markProviderFailure()` but never advances `meta.lastPurgeAtMs` — only the purge branch does. For a durably inconsistent provider (mid-upgrade replicas serving mixed dims under one model name), each expiring cooldown clears the guard and it purges and full-re-embeds the whole package again, indefinitely, just throttled to roughly once a minute. The comment above the branch says this exists so it won't loop full re-embeds, but it still does. Either advance `lastPurgeAtMs` in the backoff branch to reach a real terminal state, or document the oscillation as accepted.

### Non-blocking, worth a follow-up

- **The on-disk cache is warm but unused on the first question after a restart.** The synced memo lives in an in-process `WeakMap<Package, SyncState>`, so the first call per package returns `unavailable: "indexing"` and answers lexically even when every row is present and current — I confirmed this in the restart spec above. Running the hash diff first and searching inline when nothing needs embedding would close it, as would @jswir's suggestion of syncing at package load/reload.
- **Vectors aren't numeric-validated** — `embedding_provider.ts` checks shape but not contents, so `[1, null, "x"]` is stored and misbehaves silently later. A `Number.isFinite` pass is a few lines.
- **`MIN_SIMILARITY = 0.2` is hardcoded** and tuned for `text-embedding-3-small`. The docs recommend pointing at a local Ollama or vLLM model, and cosine distributions differ across model families, so this should probably be an env var — otherwise a local model returns either everything or nothing.
- **Query vectors aren't cached**, so the same question asked twice pays two round trips. A small LRU on the query string would be nearly free.

kylenesbit (approved):
Approving. Verified persistence in DuckDB across a real restart (zero re-embeds), pure cosine ranking with no rerank stage, and no generated descriptions; unit specs pass against real DuckDB locally.

Please land the four pre-merge items in https://github.com/malloydata/publisher/pull/923#issuecomment-5084667505 before merging — the `#(authorize)` annotation egress is the one I'd treat as non-negotiable.

mlennie:
Thanks both, this was a genuinely useful review. All four pre-merge items are addressed, plus the nits. Pushed on top and rebased onto current main.

## Blockers

**1. #(authorize) predicates no longer embedded.** This was the right catch. The leak was `docText` falling back to all annotation lines when an entity had no `#(doc)`, so the fallback string (which could carry `#(authorize)` predicates and `#(malloy)` internals) was both shown in the response and used as embedding input. The embedding path now uses a new `#(doc)`-only extractor (`docOnlyText`, no fallback) via a separate `embedDoc` field; `embeddingText` reads `embedDoc`, so a governed, undocumented entity embeds only its humanized name. The response `doc` field keeps its prior behavior, which is a separate concern as you both noted. Pinned by a unit test that asserts an `#(authorize)`-only entity's embedding text is just the name and contains none of the predicate, tenant names, or the word "authorize".

**2. Config validates at boot.** `server.ts` now calls `getEmbeddingConfig()` on the boot path alongside `getExtensionFetchPolicy()` / `getMemoryGovernorConfig()`, so a malformed `EMBEDDING_API_BASE` / `EMBEDDING_DIMENSIONS` throws at startup instead of surfacing as a warn on the first tier-4 query. It logs the posture the server booted with (model + host only, never the key).

**3. Cooldown is per-package.** `providerFailureAtMs` moved from a module global onto per-package `PackageSyncMeta`, so one package's query timeout or dims-mismatch no longer degrades every other healthy, cached package for the window. If the endpoint is genuinely down, each package cools itself on its own first failed probe, which at the entity counts a single Publisher serves is negligible. I went fully per-package rather than keeping a global fast-path for connection-level failures, because distinguishing "connection-level" from "this-package-timed-out" reliably means fragile error-string classification, and pure per-package is strictly more correct (no false cross-package degradation) at a cost of at most one extra probe per package per window. Pinned by a test that fails against the old global cooldown (package A fails, package B still reaches semantic).

**4. Heal oscillation throttled.** Worth flagging that my first attempt here was wrong and a pre-submission adversarial pass caught it: advancing `lastPurgeAtMs` in the backoff branch was a no-op, because the backoff arms the cooldown at the same instant and both used the same 60s window, and the cooldown gate blocks every heal-reaching call for that whole window, so the first call after it cleared found the purge guard also expired and re-purged anyway. The real fix is a separate purge-suppression window (`HEAL_PURGE_SUPPRESSION_MS`, 10x the cooldown) that outlasts the cooldown, so a call arriving after the cooldown clears still lands inside suppression and backs off. The result is a durably dims-inconsistent provider (mixed dims under one model name) re-embeds at most once per suppression window, lexical between, rather than once per 60s. I kept it throttled rather than fully terminal on purpose: a "never re-purge" would strand the cache on the old dims if the provider later settles on a genuinely new one, so this way it self-recovers within the window. The timing windows are now injectable, so the cooldown-vs-suppression relationship is deterministically tested (one test drives real expiry to prove a post-cooldown mismatch backs off, mutation-checked by co-anchoring the guard; another proves self-recovery once the window elapses).

## Nits

All four done: `Number.isFinite` numeric validation of returned vectors (so `[1, null, "x"]` is rejected, not silently stored); the `kind|source|name` key extracted to one shared `entityRowKey`; a tool-level test exercising the malformed-config fallback with real env parsing (not the test override); and a reflective `SENSITIVE_KEY_NAMES` redaction test that iterates the list and pins `apiKey`/`api_key`.

## Design notes

On the ones I did not change, and why:

- Sync on package load vs first query: agreed the "first question is lexical, later ones semantic" warm-up is a little surprising, and Kyle's related point that the on-disk cache is warm-but-unused on the first question after a restart is a real rough edge. I'm treating it as a follow-up rather than folding it in here: triggering the sync at load/reload (still background, still lexical fallback) is a clean, self-contained change, and it pairs naturally with the hybrid lexical+semantic follow-up. Happy to do it as the next small PR.
- `MIN_SIMILARITY` as an env var: agreed, and it is on the follow-up list. It is calibrated for `text-embedding-3-small`, and a local model's cosine distribution differs, so it should be tunable before the docs lean harder on local models. I am keeping it hardcoded for this PR because the right default is exactly what the pending real-model eval is meant to establish, and I do not want to ship a knob before I know its calibration.
- Lead docs with local rather than OpenAI-as-default: reasonable, and now that blocker 1 stops predicate egress the remaining egress is entity names + `#(doc)` text + queries. I will fold a docs pass into the same follow-up that adds the load-time sync and the similarity env var, so the local-first framing and the knob land together.
- Query-vector LRU: noted as a cheap follow-up.

One thing worth stating plainly: this is verified correct and safe-when-off (opt-in, no-key path byte-identical), but I have not yet validated that it retrieves better with a real embedding model or a real agent, only against a stub. I would rather land it as opt-in infrastructure now and run that real-model / real-agent A/B before anyone promotes it as a retrieval-quality win. Full gate is green on the rebase (unit 1546, integration 232).

PR #926, opened 2026-07-24
chore(create-malloy-package): scope as @malloy-publisher/create-malloy-package

mlennie:
Renames the scaffolder's npm package to `@malloy-publisher/create-malloy-package`, matching the org's `@malloy-publisher/<name>` pattern (agreed on Slack with Sagar and James). npm names are immutable, so this ships under the scoped name and the old unscoped `create-malloy-package` gets deprecated with a pointer here.

## What changes, and what deliberately does not

Only the npm `name` field and the user-facing invocation strings change. The package directory stays `packages/create-malloy-package`, and the bin stays `create-malloy-package`, so `npx @malloy-publisher/create-malloy-package` still exposes a `create-malloy-package` binary and `check-pack`'s bin assertion is unchanged.

The command becomes `npm create @malloy-publisher/malloy-package` (verified empirically: `npm create @scope/foo` resolves to `@scope/create-foo`), or `npx @malloy-publisher/create-malloy-package`. The scaffolder's own decline and excess-argument hints and the package README now print the scoped command. The tool still refers to itself as `create-malloy-package` (its bin name) in bug-report messages, which stays accurate.

## Tradeoff worth stating

The scoped command is longer than the create-react-app-style `npm create malloy-package` this package was originally named for, which was the reason it was unscoped. The team chose naming consistency across the org's packages over the shorter command.

## Publishing

`publishConfig` already carries `access: public`, so the scoped first publish is public. That first publish is manual by an npm owner (npm will not accept a trusted-publisher entry until the package exists), the same as the unscoped one was; the workflow comment now says so. The public docs (README pointer, docs.malloydata.dev) get the scoped command once the scoped package is published.

320 package tests pass, gate clean.


sagarswamirao:
@mlennie, a few pointers:

* The workflow currently publishes on pushes to `main` or when a PR is created with changes to either the workflow itself or the package path. We should remove these. It should be run on demand only.
* The workflow is currently decoupled from the Create Release workflow. We should use the Create Release workflow instead, where a single workflow triggers all releases. This is the same pattern that `malloydata/*` follows. It also makes it easier to see the code diffs between GitHub Releases.
* The workflow should use OIDC (you've wired it up correctly).

mlennie:
Thanks @sagarswamirao Taking these in order:

1. Dispatch-only: done (12ada99f). Removed the push and pull_request triggers, leaving workflow_dispatch. The package's tests still run in build.yml on every PR, so nothing is lost there; check_pack now runs only ahead of a dispatched publish, which is also where the prepublishOnly hook re-runs it.

2. Wiring it into the Create Release workflow: agreed, and I want to do that. It is a bit more than this PR, so I would land it as a follow-up rather than expand the rename. There is one decision I need your call on before I build it, and it is a one-way door because of how npm trusted publishing works. release.yml today writes ONE shared version into sdk/app/server. Do you want create-malloy-package (and skills) to share that number, so a server release republishes them even when nothing in them changed, or keep their own version lines with the release workflow triggering them and the job no-opping when the version is unchanged? Also worth knowing: npm allows one trusted publisher per package and matches the CALLING workflow's filename, so once these are triggered from release.yml, each package's trusted-publisher entry has to name release.yml, not its own workflow file. I have not set up the trusted-publisher entry for the scoped create-malloy-package yet (the first publish was manual), so there is nothing to reconfigure, we just point it at release.yml from the start.

3. OIDC: thanks, glad that part is right.

The scoped package is live and the old unscoped name is unpublished (we were inside npm's 72-hour window). So which way on the versioning and I will wire up the release integration.

jswir (approved):
Approving the change itself — the rename is complete and the reasoning about what deliberately *doesn't* move (directory, bin, `check-pack`'s bin assertion) is right. I checked completeness with `git grep` for `npm create malloy-package`, `npx create-malloy-package`, and `npm init malloy-package` across the branch: zero hits. Every remaining `create-malloy-package` is a directory path or a script name, which is what you said would stay. I also confirmed `npm create @scope/foo` → `@scope/create-foo` by running the real command.

Two things to fix, one of which will break the publish path on first use.

**1. `0.0.1` is already published, so the dispatch job cannot succeed.** The registry has `@malloy-publisher/create-malloy-package@0.0.1`, published 2026-07-24T17:26:14Z, `dist-tags.latest = 0.0.1`. `package.json` still says `0.0.1`, and the "Read version" step takes it straight from there with no check against the registry — so the first `workflow_dispatch` dies with `EPUBLISHCONFLICT`. Bump to `0.0.2` here, or before anyone dispatches. Worth adding a "does this version already exist" guard next to the existing (and good) "Verify the pinned skills version is on npm" step, since that one already establishes the pattern of asking the registry before an immutable version is minted.

Relatedly, the body describes the manual first publish in the future tense ("That first publish is manual by an npm owner") when it already happened three days ago. The PR is landing what's on npm — which is fine, and I verified the published tarball is clean (skills dep rewritten to `^0.1.0`, no `workspace:*`, bin `create-malloy-package`) — but it should read that way.

**2. "The old unscoped `create-malloy-package` is deprecated and points here" is not true.** The registry says `404 Unpublished on 2026-07-24T17:26:52.034Z` — unpublished, 38 seconds after the scoped publish. An unpublished package can't carry a deprecation message, so `npm create malloy-package` yields a bare E404 with no pointer to the new name. Please fix the comment at `create-malloy-package-npm.yml:185` and the matching sentence in the body; as written it tells the next reader a redirect exists when none does.

Two follow-ons worth a decision: the unscoped name is now free for anyone to claim (npm's 24-hour republish block expired on the 25th), and anyone who saw the old command gets a dead end. Publishing a minimal placeholder under the unscoped name that prints the scoped command would close both and actually deliver the pointer this comment claims.

**3. Dispatch-only drops `check-pack` from PR CI.** The body's claim that tests still run on PRs is accurate — they come via `build.yml` → `cross-platform-tests.yml`, untouched. But `check-pack` now appears nowhere except this dispatch job and `prepublishOnly`; `Verify the published tarball` is absent from this PR's check list and present on #910/#934. Publishing is still gated (this job runs `check-pack --publish-ready`, and the comment about `--ignore-scripts` bypassing `prepublishOnly` is why that matters), so nothing ships unguarded — but a change that breaks packability now surfaces only when someone tries to release. If you want the PR-time signal back, a `check-pack` step in `build.yml` is better than restoring the push/PR triggers.

**4.** Nit: `check-pack.ts:246` now runs to ~90 chars, past the surrounding wrap. Prettier won't reflow comments so lint stays green; it's just ragged.

mlennie:
Thanks @jswir, this was a genuinely useful read. Pushed e4c1de8c. Taking them in order, with one note up front: I am not editing the PR body, so the two corrections you asked for there are recorded in this comment instead. The body still describes the first publish in the future tense and still says the unscoped name is deprecated and points here. Both are wrong, and the accurate version is below.

1. Version. Bumped to 0.0.2, in package.json and in the one matching workspace version line in bun.lock. That lockfile edit is deliberately by hand rather than a re-resolve: a full `bun install` after the rename is what re-hoisted Malloy into a 0.0.422/0.0.423 split and turned this branch's CI red earlier, so I changed the single line and validated with `bun install --frozen-lockfile`, which reports no changes and leaves the lockfile byte-identical.

Guard added where you suggested, next to the skills check, as "Verify this version is not already published". Two design points worth stating because neither is obvious from the diff. It reads the package name out of package.json rather than hardcoding it, since `npm publish` publishes whatever package.json says and a hardcoded name would survive a rename and quietly check a package nobody is publishing; given this PR is itself a rename, that is not hypothetical. And it fails closed when the registry cannot be reached, rather than treating "no answer" as "free", because the step emits a `::notice` asserting the version is not on npm and it should not make that claim without having actually asked.

One thing worth recording because it contradicted my first assumption: `npm view pkg@version` exits non-zero with an E404 for BOTH "no such version" and "no such package", not 0-with-empty-output. So the check is "npm view succeeded" for taken, and "failed with an E404" for free. I ran all three branches for real before shipping it: 0.0.1 blocks, 0.0.2 passes, and a bogus registry blocks.

Correction to the body, since I am not editing it: the first publish is not future work. It happened on 2026-07-24, 0.0.1 is on npm, and this PR lands what is already published.

2. The false claim. Fixed at create-malloy-package-npm.yml, and verified against the registry before I rewrote it rather than taking it on trust: `npm view create-malloy-package` returns `404 Unpublished on 2026-07-24T17:26:52.034Z`, which is 38 seconds after the scoped 0.0.1 went up at 17:26:14Z. The comment now says it was unpublished rather than deprecated, that an unpublished name cannot carry a deprecation message so `npm create malloy-package` returns a bare E404 that points nowhere, and that the name is consequently free for anyone to claim. That last part is the correction to the body as well: there is no redirect, and anyone following an older doc gets a dead end rather than a pointer.

While rewriting that paragraph I also caught myself doing the same thing you flagged. The original text said to add the trusted-publisher entry after the first publish, and my first pass turned that into "the entry was added". npm does not expose trusted-publisher config to `npm view`, so I cannot confirm it from here, and asserting it would have been another sentence that reads true and is not. It now reads as a precondition to confirm before the first dispatch rather than as history.

3. check-pack at PR time. Agreed, and agreed on the shape: a check-pack step in build.yml is better than restoring the push and pull_request triggers, which is what we just removed for Sagar. Leaving it as a follow-up rather than widening this PR.

4. The wrap nit at check-pack.ts:246. Leaving it, and it rides along with the follow-up above. For what it is worth I checked my own new lines against the file rather than assuming: the longest are 133 and 91 characters, against this workflow's own pre-existing longest of 167, 136, 136, 128 and 116, and a `::error` workflow command cannot be wrapped anyway.

One thing that is in this PR and was not in your review, because the decision came out of it. This also bumps SERVER_VERSION from 0.0.231 to 0.0.232, which is the xlsx skew you raised on #934. Putting it there would not have fixed anyone: users run the PUBLISHED scaffolder with the pin baked into its dist, so only a scaffolder release reaches them, and bumping it inside a two-line skills release would have looked like a fix while the broken path stayed live. Worth adding that the skew is already live rather than something #934 introduces: `.xlsx` acceptance entered scaffold.ts in #907, which is exactly what published as 0.0.1, so today `npm create @malloy-publisher/malloy-package sales -- --data budget.xlsx` already scaffolds a workspace whose pinned server cannot read the file it was seeded from. So the bump lands here, and the thing that actually fixes it for users is publishing scaffolder 0.0.2 after this merges.

The comment above that constant asks whoever bumps it to confirm the new version still honours `--host`, which I did rather than assumed: pulled the published 0.0.232 tarball and confirmed `--host` is still parsed and documented, along with `--server_root`, `--config`, `--watch-env`, `--port` and `--init`. That matters because the failure mode is silent, a generated workspace would bind 0.0.0.0 and serve an unauthenticated REST and MCP endpoint while its own package.json and AGENTS.md still claimed 127.0.0.1. I also ran the built scaffolder and read what it produced: the generated start script is `npx -y @malloy-publisher/server@0.0.232 --server_root . --config ./publisher.config.json --host 127.0.0.1 --watch-env default`, and every `--host` in the generated workspace is `--host 127.0.0.1`, three of them, with no 0.0.0.0 anywhere.

On your two follow-ons: agreed on both, and I think the placeholder is the right answer. It closes the squat window and it is the only thing that would actually deliver the pointer the old comment claimed. It needs an npm publish from an owner rather than a code change, so it is not something this PR can carry, but it is the item with a clock on it: npm's 24-hour republish block expired on the 25th, so the unscoped name is claimable by anyone right now, and that exposure exists whether or not any of these PRs merge.

Verification, since a few of these changes are ones tests cannot see: full gate green on the current head, 2132 pass and 0 fail across the server unit and integration suites, skills and create-malloy-package, run the way CI runs them as two separate `--serial` invocations rather than one `bun run test`.

PR #929, opened 2026-07-26
storage= tier: review follow-ups, a serve-binding security fix, and 8.0→0.70MB per build

housejester:
## Summary

Review follow-ups from #911, plus the memory work that soak-testing the tier turned up. **Every should-fix raised on that PR, and every finding from the review round on this one, is closed here.** Off by default still: `PERSIST_STORAGE_MODE` is unchanged, and nothing here alters behavior when the tier is off.

Two things are worth a reviewer's attention before the rest: a **row-level access bypass** (first section) and a set of **memory fixes** that take a build from 8.0MB to 0.70MB of RSS.

## Security: a host-supplied binding could bypass row-level access control

`bindStorageServeBindings` filtered a host's manifest for *structural* completeness only — a sourceName, a destination, a table, a schema — and re-applied none of the eligibility gate the build enforces. A binding naming a `given`-referencing source was therefore honoured, and a given binds per query for row-level access control, so one table built once was served to every caller.

Confirmed rather than theorised: querying the filtered source with `REGION=US` returned **both** regions from the stored table.

The split the fix rests on: a host is authoritative about **which table backs a source** — it owns generations and rollout. It cannot be authoritative about **whether a source may be served frozen at all**, because that is decided by compiling the model, which the host does not do. And a manifest can name such a source without anyone being careless: a source that was given-free when built acquires a `given` on the next model edit, and the old manifest still points at a real table until convergence catches up.

Eligibility is now decided once at compile — where the compiled sources exist — and consulted at bind, so the tier-split still skips the recompile it exists to avoid. A refused binding is **dropped, not fatal**: that source serves live, and the rest of the manifest applies.

Crucially it is established **positively**: the gate carries the eligible source names, not just the refusals. Reading a missing refusal as consent admitted two further states, neither of which needs anyone to forge anything:

- **The plan compute failed.** It does live schema RPCs, so a warehouse blip during load is the ordinary way — and the package still loads and serves, with nothing examined at all.
- **The source never reached the plan.** Malloy admits only query-shaped sources as build roots, so `#@ persist` on a filtered pass-through (`X is <table> extend { where … }`, which stays type `table`) is dropped — and that is precisely the row-level-access shape. Also confirmed rather than theorised: scenario `host-binding-of-unplanned-source` returned every region before the fix.

Scoped deliberately: `#(authorize)` was never exposed (it is evaluated on the original model surface before routing is chosen), and a non-portable shape already self-corrects through the fallback ladder. Givens were the gap.

## Memory: 8.0MB → 0.70MB of RSS per build

Found by soaking the tier rather than by reading it. Measured in a container built from this repo's `Dockerfile`, driving builds over REST and reading `/proc`:

| | per build |
|---|---|
| before | **8.0MB** |
| after, 250-build soak | **0.70MB** |

Three causes, in order of size:

- **The build-time servability gate opened its own DuckDB instance per build** (~5.9MB). Its own comment already noted it needs no isolation — no attach, no credentials — so isolating it was uniformity with the build session, which *does* need it. One process-wide session recovers 94% of what deleting the gate entirely would, and is slightly faster.
- **Auto-run recompiled the whole package after every materialization** (~1.1MB), even to bind an empty manifest. `bindManifest` already had the right guard; it now applies to `reloadAllModelsForPackage` too. Not reached on the orchestrated path, which leaves distribution to the caller.
- **A build session's `close()` only refcount-decremented its DuckDB instance**, stranding 2 idle Postgres connections per build — freed only at process exit. Growth tracked build *throughput*, so one package on a schedule could exhaust a worker's connections by itself. `db-duckdb` cannot fix this generally (0.0.389 disconnected unconditionally and was reverted in malloydata/malloy#2793 because the language server holds weak_ptrs), but a build session is server-owned, single-use and never escapes the function that made it.

The residual is a linear slope close to what DuckDB's own create/attach/read/write primitives cost in the same image, and it reproduces with no publisher code in the stack.

## The run-time live fallback

`freshnessFallback=live` says the tier is an optimisation, not a dependency, so a store that fails **under** a routed query degrades to serving live — the same answer the compile-time fallback ladder already gives. Without it the store became a hard dependency the moment a query routed: a not-yet-converged rebind or an over-eager GC surfaced as a user-facing error for a source explicitly marked safe to serve live.

Narrow by construction: only when the query routed, only when every binding **that produced the serve shape** says `live`, never for a client error or an abort. The decision reads the shape's own bindings rather than the package's, because bindings are pushed package-wide and `freshnessFallback` is per entry — so a stale sibling the freshness gate already dropped cannot veto a source that is serving. The retry re-supplies the real givens, since the storage path suppresses them and the live source may filter on them, and it re-derives the row cap from the live shape.

An answer produced this way is recorded as `served_from=live_fallback`, not as a storage hit: the hit rate is the tier's headline KPI and must not rise while the tier is broken. `publisher_storage_serve_routing_total{outcome=runtime_live_fallback}` is the signal that this is happening.

## Other review follow-ups

- **An infra failure is no longer confused with a shape limit** in the chained build. Recompute-from-raw writes to the same destination, so retrying an outage there fails identically — and metering it as `inline_fallback` filed outages beside legitimate shape misses.
- **`PERSIST_COLLISION_ENFORCE` no longer fails open** — `=1`/`=yes` silently resolved false, leaving warn-only exactly the hardening an operator thought they had enabled — and it is now read at boot, so a malformed value fails the server rather than a later publish request.
- **A `ducklake` connection's shape is validated at config load**, not at its first build.
- **A part-way failed build reclaims the tables it created**, scoped to the orchestrated path where names are host-assigned and unique.
- **The metric labels are documented as emitted** — `runtime_live_fallback`, `infra_failure`, and `served_from=live_fallback` were missing from the tutorial and release notes.
- Plus the temp-dir leak on a constructor throw, and a note on why the build-time authorize walk is intentionally separate from the serve-time one.

## Test harness

The `hammer/` scenario suite grew the coverage for the above and had its execution model reworked. The change that matters: **every scenario now gets its own database, catalog, environment and publisher**, booted from a config naming only its own packages. Previously a scenario was read against a server that had already loaded sixty other scenarios' packages, which quietly made every scenario a half-truth.

- 64 pass / 3 known-red, ~50s for the suite, and **`known-red` is now a first-class status**: reported separately, excluded from the exit code, and *failing the run if it starts passing* — so a fix cannot land while the scenario silently keeps claiming the bug.
- `hammer/README.md` now states what a scenario is for (a rule the Publisher commits to, not an observation about current behavior); the step vocabulary moved to `hammer/GRAMMAR.md`. 22 scenarios were retitled to state the rule they prove.
- `## Manifest` entries take per-entry attributes, because a host stamps a manifest one entry at a time: `unplanned` for a source the build plan does not contain, and `fallback=`/`asof=`/`fresh=` for a mixed freshness set, which `## Bind` (one value for all entries) could not express.

Wiring the suite into CI is deliberately **not** in this PR — it needs a Docker-capable runner and invites its own questions, and blocking the security fix behind that discussion seemed like the wrong trade. It is the natural next PR.

## Testing

`1508` unit, zero failures; integration unchanged. Hammer 64 pass / 3 known-red, exit 0.

Every fix above has a test or a scenario, and **each was verified to fail without its fix** — including the two bypasses, which were reproduced end to end before being closed. The unit tests are the ones that matter for `main`, since hammer is not yet wired into CI: `package.spec.ts` covers the bind-time eligibility filter in all three states, and `model.spec.ts` covers the run-time fallback (row cap, error mapping, metric label, and shape-vs-package binding scope).


kylenesbit (approved):
Approving. The security analysis is the strongest part of this PR and I don't want the fix waiting on anything below. The framing is right: a host is authoritative about *which table* backs a source but cannot be authoritative about *whether* a source may be served frozen at all, because that is decided by compiling the model. Deciding eligibility once at compile and consulting it at bind keeps the recompile the tier-split exists to avoid. And scenario 65 tests the real thing rather than a proxy — a genuine eligible table, a forged binding pointed at it, `REGION=US`.

Four things I'd like fixed before merge. All four are in code this PR adds, and the first is the one that matters.

## 1. The new eligibility gate fails open when the build plan can't be computed

`ineligibleSources` starts as `{}` and is only assigned inside a best-effort try/catch (`package.ts` 559-575). `computePackageBuildPlan` recompiles the models, so it does schema RPCs against the source warehouse — and a warehouse blip during boot is the ordinary way that fails, not an exotic one. When it throws, the package still loads and serves, but `ineligibleSources` stays empty, so `bindStorageServeBindings` finds no reason for any binding and honours all of them. That is the bypass this PR fixes, reachable with nobody forging anything.

It is also the same shape as the `PERSIST_COLLISION_ENFORCE` bug three commits later: failing open in the one direction the check exists to prevent.

The fix fits the design already stated in the comment — refused bindings are dropped, not fatal. Make the field tri-state, leave it `undefined` when the plan compute failed, and refuse every storage serve binding in that state with a warning. Failing closed costs a package serving live until the next successful load, which is a performance regression rather than an outage.

## 2. Neither the security fix nor the runtime fallback has a unit test

The server specs touched are `config.spec.ts`, `connection_config.spec.ts`, `materialization_build_session.spec.ts` and `materialization_service.spec.ts`. `package.spec.ts`, `model.spec.ts` and `build_plan.spec.ts` are untouched, so the bind-time eligibility filter and the run-time live fallback are covered only by scenarios 65 and 66 — and this PR deliberately leaves hammer out of CI. Once this merges, nothing running on `main` protects the row-level-access fix from being refactored away.

I agree with not blocking the security fix behind the Docker-runner discussion. That argument doesn't extend to unit-testable pure logic: a test in `package.spec.ts` asserting that `bindStorageServeBindings` drops a binding whose source is in `ineligibleSources` and keeps its neighbour is a handful of lines, and it's the one I'd insist on.

## 3. The live-fallback retry can answer with an empty result set and a 200

`rowLimit` is initialized to `0` and assigned from `getPreparedResult`, which the code deliberately placed inside the try. If the storage failure surfaces at prepare rather than at run, `rowLimit` is still `0` when the retry runs (`model.ts` 2358-2363). The DuckDB connector treats that as a hard cap — `if ((rowLimit !== undefined && index >= rowLimit) || ...) return;` fires on the first row — so the caller gets a successful, empty answer instead of an error.

Prepare is reachable on the storage path for the case this feature exists for: a host-supplied binding whose table path is missing or malformed resolves through `virtualMap` at prepare. `resolveModelQueryRowLimit` never returns 0, so recomputing it on the degrade path (or flooring to `getDefaultQueryRowLimit()`) closes it.

## 4. A failure in the retry escapes all the error handling

The `liveRunnable.run` call sits inside the catch with nothing around it, and `getQueryResults` has no outer try. So a second failure — plausible, since a broad outage takes the source warehouse down alongside the store — skips the `runtime-given-*` mapping, the `MalloyError` rethrow, the `Query execution failed:` wrapping, and the `status: "error"` histogram record. `internalErrorToHttpError` falls through to `httpError(500, error.message)`, so a case that was a clean 400 becomes a 500 with no error metric behind it. Wrapping the retry so a second failure lands in the existing `else` branch handles both.

---

Separately, and not blocking: the description's bullet about the CTAS target and the serve path quoting a table the same way is already on `main` (from #904 — `quoteManifestTablePath` and the comment explaining the choice both predate this branch), and no commit here touches quoting.

The commit-level rationale is unusually good throughout. The cross-environment data-loss reasoning behind gating the reclaim to orchestrated builds, and making `known-red` fail the run when it starts passing, are both things I'd have asked for and wouldn't have thought to.

packages/server/src/service/package.ts
```diff
@@ -836,10 +842,44 @@ export class Package {
...
+    * always correct because the tier is a performance tier. The rest bind.
+    */
    public bindStorageServeBindings(
       entries: Record<string, ManifestEntry>,
    ): void {
-      this.storageServeBindings = deriveServeBindings(entries);
+      const derived = deriveServeBindings(entries);
+      const allowed = derived.filter((binding) => {
+         const reason = this.ineligibleSources[binding.sourceName];
+         if (!reason) return true;
```
sagarswamirao: Flagging this early so it doesn't bite downstream: the eligibility lookup fails open on its own initialization failure.

`ineligibleSources` defaults to `{}` (`package.ts:166`) and its only writer is inside the best-effort catch at `package.ts:559-575`. `computePackageBuildPlan` recompiles the models, which the comment right above it notes does live schema RPCs against the source warehouse — so an unreachable warehouse or a schema-fetch timeout during load throws, the warn fires, and the package still loads and serves. At that point this filter reads an empty map, `reason` is `undefined` for every binding, and every host-vouched binding is admitted, including a `given`-referencing source. `model.ts:2311` then sets `effectiveGivens = undefined` for the routed query, so the frozen table is served with no filter to re-impose.

That's the bypass this method closes, reachable without anyone forging a manifest — just a warehouse blip in the window between the worker compile and this second one.

Worth noting the rest of this surface is consistently fail-closed: `referencesGiven` returns `true` when introspection fails (`materialization_eligibility.ts:170-172`), and `projectToPublicColumns` refuses when the public surface is unreadable (`build_plan.ts:174-184`). This default is the outlier.

The fix fits the design already stated in the docstring — refused bindings are dropped, not fatal. Making the field tri-state (leave it `undefined` when the plan compute failed, and treat that state as refuse-all here, with a warning) costs a package serving live until the next successful load, which is a performance regression rather than an outage. Could you make that change?

packages/server/src/service/package.ts
```diff
@@ -836,10 +842,44 @@ export class Package {
...
+    * Refused bindings are DROPPED, not fatal: that source serves live, which is
+    * always correct because the tier is a performance tier. The rest bind.
+    */
    public bindStorageServeBindings(
       entries: Record<string, ManifestEntry>,
    ): void {
-      this.storageServeBindings = deriveServeBindings(entries);
+      const derived = deriveServeBindings(entries);
+      const allowed = derived.filter((binding) => {
+         const reason = this.ineligibleSources[binding.sourceName];
```
sagarswamirao: A second gap in the same lookup, and this one I haven't seen raised: a `#@ persist` source that Malloy drops from the build plan is admitted, and it happens to be the row-level-access shape.

`collectIneligibleSources` (`build_plan.ts:668-680`) iterates only `compiled.sources`, so absence from that record reads as "eligible" here. But `detectDroppedPersistSources` (`build_plan.ts:437-466`) exists precisely because some sources carry `#@ persist` and never enter `buildPlan.sources` — and its own comment names the shape: "a filtered pass-through `X is <table> extend { where … }`, which stays type `table`; only query-shaped sources are treated as roots."

That is the canonical given-filtered source: `source: orders is duckdb.table('orders') extend { where: tenant = $tenant }` with `#@ persist storage=lake`. So the one shape the plan silently drops is also the shape whose refusal matters most. Those names land in `droppedPersistSources`, which is surfaced as an operator warning but never consulted here.

A standalone publisher doesn't reach it — the build path calls `assertMaterializationEligible` on the same source and lets it propagate, so no local manifest entry is produced. The exposure is the host-supplied manifest path (`environment.ts:1523`), which is the threat model this docstring invokes: a host that built the table under an older compiler, or that names the source directly, gets it admitted.

Consulting `droppedPersistSources` alongside `ineligibleSources` when filtering would close it. Same root cause is worth a thought more generally: any source absent from `compiled.sources` is currently treated as eligible, so the safer default here may be to require a positive eligibility result rather than infer it from a missing refusal.

packages/server/src/service/model.ts
```diff
@@ -2326,65 +2330,115 @@ export class Model {
...
+               },
+            );
+            recordStorageServeRouting("runtime_live_fallback");
+            queryResults = await liveRunnable!.run({
+               rowLimit,
+               givens: querySurfaceGivens,
+               abortSignal,
+               buildManifest,
+            });
+            executionTime = performance.now() - startTime;
```
sagarswamirao: A query answered live ends up recorded as a storage hit, because `serveVirtualMap` is never cleared before this fall-through — it's still the truthy map from line 2190.

So the success histogram at line 2472 records `status: "success"` with `served_from: "storage"` for a query served from the live warehouse, and `queryResults.connectionName` on that same data point will be the live connection, contradicting the label.

`materialization_metrics.ts:322` calls this hit rate the headline KPI of the tier, and the doc for `runtime_live_fallback` says the fallback is meant to be "invisible in the hit rate alone." As written it's worse than invisible: it counts as a hit, so the KPI rises while the tier is broken. The counter at line 2369 doesn't correct the histogram attribute.

Clearing `serveVirtualMap` here before falling through (or labelling it `live_fallback`) fixes it — nothing after line 2331 reads it, so clearing looks safe.

packages/server/src/service/model.ts
```diff
@@ -2326,65 +2330,115 @@ export class Model {
...
+         //    unfiltered rows.
+         const canDegradeToLive =
+            !!serveVirtualMap &&
+            !!liveRunnable &&
+            !abortSignal?.aborted &&
+            !String((error as { code?: string })?.code ?? "").startsWith(
+               "runtime-given-",
+            ) &&
+            this.serveBindings.length > 0 &&
+            this.serveBindings.every((b) => b.freshnessFallback === "live");
```
sagarswamirao: This guard reads the whole package's bindings rather than the ones this query routed through, so I don't think the narrowing documented just above holds.

`Package.pushStorageServeBindingsToModels` (`package.ts:887-891`) pushes the entire package's `storageServeBindings` onto every model, and `freshnessFallback` is per-entry, so mixed values across one package are the normal case. `this.serveBindings` here is the raw field — not the fresh subset, and not filtered to the sources this query touched.

Concretely: a query touching only source A with `fallback=live` is denied its fallback because sibling B carries `stale_ok` or is unset (common before the field is stamped, and `undefined !== "live"`). That's the opposite of the stated intent, and it means the feature stays dark on any package with a mixed or partially-stamped binding set.

The `fail` direction happens to be safe, but only because a `fail` binding stays in this array even when `freshServeBindings` drops it from the shape — it's the superset saving it, not the check.

Returning the binding set that produced the shape from `loadServeShapeQuery` alongside `runnable`/`virtualMap`, and running the `every()` over that, would make the comment true. Worth noting scenario 66 uses a single binding, so it can't catch this — a two-binding case with differing `freshnessFallback`, querying only the `live` one, would fail today.

packages/server/src/config.ts
```diff
@@ -541,10 +541,12 @@ export const getPersistStorageMode = (): PersistStorageMode => {
...
  */
-export const getPersistCollisionEnforce = (): boolean => {
-   const raw = process.env.PERSIST_COLLISION_ENFORCE;
-   return raw !== undefined && raw.trim().toLowerCase() === "true";
-};
+export const getPersistCollisionEnforce = (): boolean =>
+   // parseBoolEnv, not an ad-hoc === "true": an operator who writes `1` or `yes`
+   // has asked for the check to block, and an ad-hoc compare would silently leave
+   // it warn-only — the flag failing open in exactly the direction it exists to
+   // prevent. A typo throws at startup, like every other flag here.
```
sagarswamirao, line 548: Small one: this comment says a typo throws at startup, but this getter isn't read at startup.

`server.ts` imports `getPersistStorageMode`, `getMemoryGovernorConfig` and `getMaterializationSchedulerConfig` and evaluates all three at module scope; `getPersistCollisionEnforce` isn't imported there at all. Its only caller is `Package.formatPersistenceCollisionRejections` (`package.ts:1203`), reached from the publish path — so an unrecognized value throws on a publish request rather than at boot.

The precedent is right there: `server.ts:168-172` is a bare `getPersistStorageMode()` call added specifically to surface a malformed value at startup instead of lazily per-request. The same one-liner here would make the comment accurate.

Separately, worth a release note: `PERSIST_COLLISION_ENFORCE=1` resolves false (warn-only) today and true (publish-blocking) after this change, and the flag isn't documented in `docs/configuration.md`.

docs/persist-storage-tutorial.md
```diff
@@ -403,10 +403,12 @@ is a brief window where the running server's serve binding still describes the
...
 differ. A query that reaches a **newly added** field fails the serve-shape
 compile and falls back to serving live (safe). A query over a **removed** field
-still compiles against the stale binding and then errors at run against the new
-table — there is no run-time fallback (see "Compile-time fallback only" in the
-release notes). So a column-removing rebuild can surface transient query errors
-until the binding refreshes; it does not return wrong data. Explicit generation
+still compiles against the stale binding and then fails at run against the new
+table. What happens next depends on the binding's `fallback`: `live` recomputes
+the source from the warehouse and answers correctly, while `fail` (and the
+`stale_ok` default) surface the error. So a column-removing rebuild can surface
```
sagarswamirao, line 409: While this file is open — the metric label enumerations further down (lines 682 and 684) are stale after this PR.

`publisher_storage_serve_routing_total{outcome=storage|live_fallback}` gains `runtime_live_fallback`, and `publisher_storage_chained_build_total{outcome=parent_reuse|inline_fallback|strict_refused}` gains `infra_failure`.

The `runtime_live_fallback` omission is the one that matters operationally — `materialization_metrics.ts:317-321` describes it as the interesting signal precisely because it means the tier is broken while queries still succeed, which the hit rate alone won't show. Right now the only doc telling an operator what to watch doesn't mention it. `RELEASE_NOTES.md:26` enumerates the metrics without either new label too.

sagarswamirao (approved):
Re-reviewed at `c75e3eaf`. All six points I raised are addressed, and the eligibility fix is better than what I suggested — inverting to a positive allowlist (`SourceEligibility { eligible, refused }`) closes the general class rather than the two instances I found, and the three distinct refusal reasons make each state legible in the logs. Scenarios 67 and 68 confirming both were live before the fix is the right way to have settled it.

The fallback commit is a clean job on all four defects: `failQuery` routing the original and the retry through one mapping, and `servedFrom` as an explicit three-state rather than inferred from `serveVirtualMap`, both read better than the minimal fix would have. The `package.spec.ts` cases exercise the real method and would fail if the gate regressed.

Two non-blocking observations, neither worth holding the merge:

**The CTAS→DESCRIBE window is still open.** `buildSourceIntoStorage` issues `CREATE OR REPLACE TABLE` and then `describeTable` (`materialization_build_session.ts:294-300`). If the DESCRIBE throws after the CTAS commits, `buildOneSource` propagates before returning an entry, so the table never enters `builtThisRun` and the new reclaim cannot see it — a committed table with no manifest entry and nothing that will ever name it. It's the same "unreachable forever" class the reclaim was written for, and the serve-shape gate next door already covers its equivalent window with a targeted drop (`materialization_service.ts:1624-1663`). Narrow enough to leave, worth a follow-up.

**The reclaim has no unit test.** It's the one path here that issues a DROP against a customer warehouse, and its correctness rests on three guards — orchestrated-only, created-this-run-only, and the still-referenced check. I traced all three and they hold, and scenario 64 covers the happy path plus the still-referenced inverse, but that's hammer-only and hammer isn't in CI yet. A test asserting a carried-forward entry is never dropped would be cheap insurance for a destructive operation.

Approving.

PR #933, opened 2026-07-27
feat(server): per-query metadata for backend attribution

housejester:
> Rebased onto `main` now that #932 (`chore(deps): upgrade Malloy to 0.0.426`) has merged — this needs the `RunSQLOptions.queryMetadata` it brought in.
>
> **Note for reviewers:** `packages/server/src/api.ts` is generated and gitignored, so a fresh checkout needs `bun run generate-api-types` before `typecheck` (CI does this for you).

## Problem

Every statement Publisher sends arrives at the backend indistinguishable from every other one. A warehouse bill cannot separate interactive traffic from materialization builds, or one package's cost from another's, and there is no way to find the backend query behind a slow API call. `malloydata/malloy#3019` added the channel — `RunSQLOptions.queryMetadata`, a flat bag of string properties each connector attaches per query (Snowflake `QUERY_TAG`, BigQuery job labels, a leading SQL comment elsewhere). Nothing declares or fills that bag.

## What this adds

A `QueryMetadata` bag declarable at five levels, most specific wins **per property**:

| Level | Declared as |
|---|---|
| Connection | `queryMetadata` on the connection config (plus `queryMetadataEnforced` — see below) |
| Package | `materialization.queryMetadata` in `publisher.json` |
| Model file | `## materialization.queryMetadata.<name>="<value>"` |
| Persist source | `#@ persist queryMetadata.<name>="<value>"` |
| Request | `queryMetadata` + `queryClass` on a query or SQL request |

On top of whatever is declared, Publisher attaches its own context — `class` (`interactive`/`materialize`/`index`/`ops`), `environment`, `package`, `model`, `source`, `trigger`, `run_id`, `query_id` — so attribution needs no modeling work from a package author. Context wins over a declared property of the same name: a caller must not be able to label its own query as a build and corrupt exactly the attribution this exists to produce. Every context field except `query_id` is a property of the unit of work rather than of the individual call, so a repeated query produces the same values (see the correlation section for what `query_id` costs).

`version` is reserved but not yet reachable: the only path that supplies it is a query request's `versionId`, which the route rejects with 501 until versioned packages land. It keeps its slot in the shed order so wiring it later is a one-line change.

**Ships dark.** `PUBLISHER_QUERY_METADATA` defaults to `off`, which attaches nothing at all, including caller-supplied properties; set it to `on` to enable attribution. See the section on the default below.

Attached on: model queries, notebook cells and MCP `malloy_executeQuery` (`interactive`); every statement of a materialization build (`materialize` — the staging drop, the CTAS, the swap, the rename, and the cleanup drop after a failure, so one build reads as one unit of work); raw connection SQL and the drops that retire a materialization's tables (`ops`). A storage-routed query that degrades to live under `freshnessFallback=live` re-resolves against the live connection and tags the retry, so the statement that produced the rows is the one carrying the bag.

## Properties a deployment owns

`Connection.queryMetadataEnforced` is the same map at a precedence no declaration can reach: above every author layer, below the server's context, and shed only after every declared property.

It exists because the connection default is the *least* specific layer, which is the wrong home for the property a multi-tenant host is billed by — a package author's `#@ persist queryMetadata.tenant=…` outranks it, and under the 20-property cap it is the first thing shed. So the tenant label was both forgeable by the tenant and the first casualty of a crowded bag.

The trust split it rests on is the **deployment's**, not this code's. Publisher does not authenticate a connection write — REST and MCP are unauthenticated by design — so "operator writes connections, authors write packages" is real where a gateway or control plane enforces it and absent on a bare Publisher, where the same caller can do both. The layer is worth using in the first case, which is what it was built for; the docs say so in those terms rather than implying an authorization check exists here.

Not a deployment-level env bag, deliberately: one Publisher serves many tenants, so the value has to come from the connection. The publisher still invents no tenant-specific keys — it just stops forcing a host's identity into a layer its own users outrank. Validated at both boundaries like `queryMetadata` (400 on a write, warning on config load) and counted against the same budget.

## Correlating an API call with a backend query

A query response carries `queryCorrelationId` — the id Publisher minted for that query and attached as its `query_id` property; over MCP the same id is `_query_id` in the `malloy_executeQuery` envelope. The same value sits in `QUERY_HISTORY.query_tag` (inside the JSON tag), `INFORMATION_SCHEMA.JOBS.labels`, or the statement text, so a caller can pull the warehouse-side row for its own API call: cost, duration, plan. This is the replacement for a read-back of the backend's native query id, which upstream deliberately did not ship — and it covers Postgres, DuckDB and MySQL, which have no native id to read back. Both fields are null/absent until a deployment turns the feature on.

Minted server-side rather than taken from the request, because it is the platform's join key: a caller-supplied value can be omitted, reused across calls, or collide with another caller's, and then the response has nothing meaningful to hand back. `query_id` is reserved like the rest of the context; a caller that wants its own id declares it under its own property name. Builds need no response field — every statement of a build already carries `run_id`, which is the materialization id the create call returned.

The id is minted for **every** backend, not only the ones with a native tag mechanism. On a backend that carries the bag as a comment, that makes each statement's text unique and so bypasses an exact-text result cache. Deliberate: a correlation id you can only sometimes rely on is worth less than the cache hit it costs, and leaving the feature off is the way out for a deployment that disagrees. Every other context property is stable per unit of work and has no such effect.

Scope of that cost, since it drives the default discussion below: Snowflake's `QUERY_TAG` and BigQuery's job labels are out of band, so the statement text is unchanged and their caches are untouched. Of the comment-carrying backends, Databricks is the one with an exact-text result cache that bites — Postgres and DuckDB have nothing to break, and MySQL 8 dropped its query cache.

## Keeping the bag legal, without breaking queries

Malloy validates the bag when it issues the statement and **throws** on one it cannot render, so a bag assembled one character too long would fail a customer's query. `service/query_metadata.ts` therefore never throws — it sanitizes, truncates and drops, metering every drop:

- **20 properties.** Core enforces this cap too, by throwing (`QUERY_METADATA_MAX_PROPERTIES`); the contribution here is arriving under it. Author properties are dropped before context: losing a caller's `team` label degrades attribution, losing `class` breaks it.
- **Snowflake's 2000-char tag**, which core does not cover. `db-snowflake` clamps by slicing `JSON.stringify(bag)`, which truncates mid-JSON and leaves an unparseable `QUERY_TAG` — silently unqueryable in `QUERY_HISTORY`, which is the whole point of tagging. So whole properties come out here instead, on the serialized size.

Boundaries with a human behind them reject rather than clamp — a per-request bag and a connection update both come back as a 400. Where nobody is waiting it is reported instead: a package/model/source declaration becomes a package warning, and a connection default in a loaded config becomes a log warning, because a tag must never be the reason an environment refuses to come up. Two things that are legal but will not do what they say are reported the same way: a name BigQuery would silently drop (its label grammar needs a leading letter; other backends keep the name), and a declaration with no room left for the server's context.

Core's `validateQueryMetadata` is not exported from `@malloydata/malloy` (only the `QueryMetadata` type is), so the contract is mirrored here in one place with the upstream pin in a comment. Exporting it upstream would remove the mirror; happy to send that separately.

## `scope` moves into `materialization`

`scope` was the only build-behavioral knob left at the manifest root while `schedule`, `freshness` and now `queryMetadata` all live in the `materialization` block. It moves in, and:

- the root form keeps working, because scope rides the published artifact — a package published before the move must keep parsing until it is republished. It is deprecated only when it is the *sole* home: whenever the server rewrites a manifest it now writes **both** homes in sync, so a package edited through the API still resolves correctly on an older Publisher that reads only the root, and warning about that would blame an operator for something the server did;
- declaring both homes with **different** values throws rather than guessing: scope decides whether an artifact is version-owned, and picking wrong could reuse a table across versions that was never meant to be shared;
- the model-file freshness read moves under the same envelope, keeping the bare `## freshness.*` form as a **per-property** fallback, so a file with one knob in each home resolves both.

## A note on where the advisory findings go

They deliberately do **not** go through `Package.persistencePolicyWarnings`. That method is a publish **rejection** gate (`formatPublishRejections`) and a non-empty result also disarms the standalone scheduler — so routing a deprecation or a mistyped tag through it would fail the publish of every package still using a root-level `scope` and silently stop its cron. They surface on the package's operator `warnings` array instead, alongside the render-tag and `storage=` findings.

`materialization_config_validation.ts` owns those rules. The scope/schedule/freshness coherence rules that genuinely reject still live in `persistencePolicyWarnings`; moving them into the same module is a mechanical follow-up with no behavior change, kept out of this diff.

## Identity, and what is not tagged

Query metadata is observability only: excluded from the connection digest upstream and from `sourceEntityId` here, so changing a property never re-addresses a source or orphans its table.

That holds by construction — `sourceEntityId` is `makeBuildId(connectionDigest, getSQL())`, `queryMetadata` exists only on `RunSQLOptions` (the dispatch path, never the compile options `getSQL` takes), and every connector's `getDigest()` enumerates connection-shape fields — but the failure mode is bad enough (every source re-addresses, every built table orphaned at rest) that `query_metadata_identity.spec.ts` now pins it end to end: a real package compiled with tags at all three model-side layers produces a byte-identical entity id to the same package without them. The test was verified against an injected leak, so it fails if one ever appears.

Three honest gaps:

- **Schema fetches are untagged.** The bag is per-call and `fetchSchema` takes no `RunSQLOptions`, so package-load schema work carries no attribution. Worth knowing before treating a tagged bill as complete.
- **`POST …/sqlTemporaryTable` is untagged.** It issues a real `CREATE TEMPORARY TABLE AS` through a driver call that takes no `RunSQLOptions`.
- **A `storage=` build's destination statements are untagged.** The DuckDB/DuckLake session path issues its own SQL outside the connector's `runSQL(sql, options)` shape; the warehouse-side read is where the cost is, and that is covered.

## The default: dark for a release, and the open question

`PUBLISHER_QUERY_METADATA` now defaults to **`off`**, following `PERSIST_STORAGE_MODE`'s precedent. This is the rare change that touches every statement the server sends, and the risk that decides it is upstream rather than here: core throws at dispatch on a bag it cannot render, and the contract it validates against is mirrored in `service/query_metadata.ts` from a pinned version. Clamping, shedding and never-throwing all sit downstream of that mirror staying accurate, so a tightened upstream limit would surface as failing customer queries on a path nobody opted into. A release behind the switch buys field validation first.

Still open, and worth maintainer input — @kylenesbit argued for **context on, correlation id opt-in** instead, on the grounds that `query_id` is the only property that is not byte-stable and therefore the only one that costs anything. That is right as far as caching goes (and see the scoping above: it is really only Databricks). Two things I would want settled with it:

- It is not reachable by configuration today. The switch is a single `on|off` and the mint is unconditional at the three request boundaries, so the split needs either a third mode or a conditional mint — plausibly keyed on whether the backend attaches out of band rather than on an env var.
- It does not address the drift risk above, which applies to any non-empty bag, context included. The better answer to that one is probably a conformance test that dispatches the widest bag the clamp can emit through a real connector (core's `validateQueryMetadata` is not exported, so the connector is the way in). With that in place I would not argue against on-by-default.

Happy to land either shape.

## Verification

- **0 failures** — 1720 server unit, 232 server integration, 7 skills, 362 create-malloy-package
- `typecheck`, `lint`, `prettier --check`, `validate-duckdb-sync`, `validate-pg-sync` all clean
- Review passes over this branch. Round 1 turned up three blockers, all fixed in the third commit: a typo'd kill switch that would have 500'd every query, a `PATCH { scope }` that authored a manifest the loader would later refuse (and the matching rollback hole), and notebook cells that were not actually tagged. Round 2 found the connection layer had no declaration boundary while two sentences claimed it did (400 on the update API, warn at config load). Round 3 fixed the two remaining holes in the "every statement" claim — MCP queries carried no connection or enforced layer, and the `freshnessFallback=live` retry ran untagged while the response still returned the failed attempt's id — plus five places a property went missing with nothing to point at (`runContext.runId` unvalidated, a non-string manifest value dropped before the validator could name it, a `materialization` PATCH silently untagging a package's builds, `__proto__` vanishing into the prototype setter, and the fail-open connection lookups logging nothing), and a docs pass over four claims that outran the code. Details in the commit messages.
- New tests cover the contract mirror and every clamp/drop path, the four-layer per-property resolution (including the envelope-vs-bare fallback), the scope two-home resolution and conflict, the advisory warnings, that all five statements of a build carry the resolved bag, that the MCP path resolves the enforced layer, that the live-fallback retry is tagged against the live connection with a truthful id, and that query metadata never moves `sourceEntityId` — plus that an untagged build is byte-identical to before

Docs: `docs/query-metadata.md` (new), plus the `materialization.md` policy section and the two materialization skills updated for the `scope` move.


kylenesbit (approved):
Reviewed against the design this implements ([ms2data/service#5663](https://github.com/ms2data/service/pull/5663)). The drift from that doc — the flat bag instead of dialect-namespaced blocks, `queryCorrelationId` instead of `QueryExecutionMetadata`, the context property renames — is all forced by what Malloy 0.0.426 actually shipped or is a deliberate improvement, and it's disclosed. That reads as the design needing an amendment, not this PR needing a change.

Approving on the strength of the resolve module and its tests. Three things I'd want fixed before this attaches metadata in anger, though — each one is a case where the code doesn't do what the surrounding documentation says it does, and two are in the threat model the commit messages themselves invoke.

Verified locally on `ff7d93a`: `typecheck:server` clean, 1703 server unit tests pass (0 failures). Note `packages/server/src/api.ts` isn't committed, so `bun run generate-api-types:server` is needed before typecheck on a fresh checkout.

---

### 1. The reserved context namespace isn't actually reserved

`api-doc.yaml`, `docs/query-metadata.md`, and `d11d4f8` all state that a caller cannot overwrite the server's context — *"a caller must not be able to label its own query as a build"*, *"`query_id` is reserved like the rest of the context"*. But that's enforced by **overwriting**, and `queryContextProperties` only emits fields that are set:

```ts
// service/query_metadata.ts
const put = (name: string, value: string | undefined) => {
   if (value !== undefined && value !== "") out[name] = value;
};
```

So a reserved name the server has no value for isn't reserved at all — it's just an author property. On an interactive model query the server sets `class`, `environment`, `package`, `model`, `query_id`, but never `source`, `trigger`, or `run_id`, and sets `version` only when the request carried a `versionId`.

Calling `mergeQueryMetadata` with `request: { trigger: "scheduler", run_id: "someone-elses-build", source: "order_rollup", version: "v99" }` and the interactive path's context:

```json
{ "trigger": "scheduler", "run_id": "someone-elses-build", "source": "order_rollup",
  "version": "v99", "model": "orders.malloy", "environment": "prod",
  "class": "interactive", "package": "sales", "query_id": "server-minted" }
```

Only `class` and `query_id` held. That matters because `docs/query-metadata.md` tells a reader to group a build by `run_id` (*"every statement of one build carries `run_id`, which is the materialization id the create call already returned"*), so any caller can attribute its interactive spend to a build's run, or make its query read as scheduler-triggered.

`query_id` is safe today only by accident — every request-bearing path happens to mint one. `correlationId` is optional in `ModelQueryMetadataInput` precisely so a path can skip it, and the notebook-cell path already does; add a request layer there later and the join key becomes forgeable too.

Suggested fix: strip every name in the reserved set from the author layers before merging, rather than relying on the overwrite. `CONTEXT_SHED_ORDER` is already the canonical list of those nine names, and already has a parity test pinning it against `queryContextProperties`.

### 2. The `freshnessFallback=live` retry issues an untagged statement, and still returns a correlation id

`service/model.ts` — the storage attempt passes the resolved bag, the live retry doesn't:

```ts
queryResults = await liveRunnable!.run({
   rowLimit,
   givens: querySurfaceGivens,
   abortSignal,
   buildManifest,
});
```

When a storage-served query degrades to live, the statement that actually produces the rows carries no bag at all: no `class`, no `package`, and no `queryMetadataEnforced` properties — the layer whose whole justification in `cb28cbe` is that nothing can drop it.

Meanwhile the return is unconditional:

```ts
// The id from the bag that was actually attached, not the one supplied:
// a bag shed under budget pressure sheds context last, but a caller
// should be told the truth about what it can look up.
queryCorrelationId: appliedQueryMetadata?.query_id ?? null,
```

So the caller gets an id that appears nowhere in the backend's query history — exactly what that comment says it's avoiding. The live prepared result is already fetched a few lines above for the row limit, so re-resolving against its `connectionName` and passing the bag should cover it; if resolution yields nothing, null the id.

### 3. MCP `malloy_executeQuery` gets only the bare context

Both call sites in `mcp/tools/execute_query_tool.ts` invoke `model.getQueryResults(...)` with no `queryMetadataInput`, so MCP queries carry `class`, `package`, `model` — but no `environment`, no connection default, and **no `queryMetadataEnforced`**.

This is the same omission `d4e8c01` fixed for notebook cells (*"they issue real backend SQL on an interactive path and were carrying no attribution at all, including `class` and `package`"*), and MCP is this repo's headline query surface — so a multi-tenant host's enforced tenant label goes missing on the traffic most likely to be interactive. `environmentName` is already in scope there; the connection lookup can be lifted from `model.controller.ts`. Omitting the correlation id is fine given there's no response field for it.

---

### Non-blocking, worth a follow-up issue

- **A materialization PATCH still wipes the package's `queryMetadata`.** `d4e8c01` fixed this in the CLI, but the hazard is server-side and remains for every other caller. In `service/environment.ts` the incoming block replaces the on-disk one wholesale, and I confirmed that PATCHing `materialization: { schedule: "0 6 * * *" }` onto a package declaring `{ scope: "version", queryMetadata: { team: "finance" } }` writes back `{"schedule":"0 6 * * *","scope":"version"}` — `scope` is preserved, `queryMetadata` is gone. The new `materialization_schedule_surface.spec.ts` test covers the `scope` half of that exact scenario. The precedent for the fix is right there in the same function; since the wire schema is `oneOf QueryMetadata | null`, omit-to-preserve and explicit `null`-to-clear would also let the CLI drop its two workarounds.

- **A conflicting `scope` fails the *load*, not just the publish.** `resolvePackageScope` throws, and `readPackageMetadata` calls it during load, so the package doesn't serve at all — while `docs/materialization.md` describes the posture as "at **package load** (warn — the package still serves)" and lists the conflict rejection under it. Throwing is defensible for something this load-bearing (and `parsePackageScope` already threw on an invalid value), so probably a docs fix.

- **The advisory warnings mislabel inherited properties.** `materializationConfigWarnings` checks each source's *resolved* bag and labels every finding `#@ persist queryMetadata:`, so one bad property in `publisher.json` yields a correct package-level warning plus one per source pointing the author at an annotation that doesn't contain it. Reporting at both levels is reasonable; the source-level label should just say it was inherited.

- **No pinning test for the identity exclusion** (design §5.7 asked for one). It holds by construction — `computeSourceEntityId` is `source.makeBuildId(connectionDigests[...], source.getSQL())` and neither input reads annotations — but it's the invariant with the worst failure mode, and cheap to pin.

- **`resolveQueryMetadata` is only tested against the fake tag** in `materialization_test_fixtures.ts`, and `tagQueryMetadataLayer` has three silent-empty exits. I ran all six declaration forms from `docs/query-metadata.md` through the real `@malloydata/malloy-tag` parser and they all resolve correctly, so no live bug — but one test through the real parser would keep this from silently becoming a no-op, which is precisely the "silently dropped" failure the module exists to prevent.

- Small wording thing in the PR body: core *does* enforce the 20-property cap (`QUERY_METADATA_MAX_PROPERTIES = 20`, and `validateQueryMetadata` throws on it). The contribution here is enforcing it by *dropping* so core never throws, plus the serialized-size budget, which core genuinely lacks.

On the open question about the default: since only `query_id` breaks exact-text result caching and the other eight context properties are byte-stable per unit of work, a third mode (context on, correlation id off) fits the actual trade-off better than flipping the whole feature dark — and design §10 already settled injection as always-on.

PR #935, opened 2026-07-28
feat: Malloy-tag dashboards, and one interactivity engine behind them and notebooks

kylenesbit:
## What this is

`dashboards/*.malloy` becomes a first-class Publisher surface: a dashboard is Malloy plus render tags, discovered at load, described by a manifest over REST, and rendered by a `<Dashboard>` the SDK exports so it works from any React app rather than only the Console.

The larger change is underneath. Notebooks and dashboards had been two implementations of the same idea, and they are now one. Control rows, URL-addressable given state, Apply batching, and `# drill` navigation are a single set of primitives both surfaces consume, so the notebook's Parameters panel gained the enriched controls and drill for free and the legacy Filters panel is gone. Alongside that, `givens` means one thing across the API, Pages are Data Apps, the examples are consolidated into `storefront`, and the docs and home page describe what Publisher actually does now.

[docs/malloyyo-dashboards-design.md](docs/malloyyo-dashboards-design.md) is the design record — read it first. [docs/dashboards.md](docs/dashboards.md) is the user-facing guide, and [docs/choosing-a-surface.md](docs/choosing-a-surface.md) is the when-to-use-which.

**This PR is a handoff, not a merge candidate.** It is 212 files. The section below proposes ten smaller PRs in dependency order.

## Suggested split

Ordered so each is independently reviewable, and the first six are renaming, docs, the domain model, and foundation.

**1. Renames.** Three mechanical renames, no behavior change: Pages → Data Apps (`/pages` → `/dataApps` with no alias, `PageViewer` → `DataAppViewer`, `pageEmbed` → `dataAppEmbed`, the `html-pages-*` fixtures and the `html_pages` integration directory), Publisher App → Console (`docs/publisher-app.md` → `docs/console.md`), and the "Governed Reports" UI label → "Notebooks". Splittable into three if the REST break wants its own review.

**2. Docs and the design doc.** Lands `docs/malloyyo-dashboards-design.md` first so it is the map for PRs 5–9, plus `docs/security-posture.md`, the docs index, and the standalone fixes (a broken `givens.md` anchor, `development.md`'s stale standalone-recipe claim). Includes the `api-doc.yaml` description rewrite — the resource hierarchy was missing dashboards, data apps, and materializations and named a manifest file that does not exist — which is prose only, no schema change.

**3. Model-annotation ownership.** A standalone correctness fix worth merging on its own: `modelAnnotations` folds in everything a file imports, so an imported model's `##` tags were setting the title and autorun for every notebook that imported it. Adds `ownModelNotes` and uses it wherever a tag describes *this* document, keeping full lineage for `authorize`. Must precede anything that reads tags.

**4. Givens vocabulary and API contract.** `givens` had come to mean four things at once. It now always means a collection of `Given` declarations; values keyed by given name are `GivenValues` (`EncodedGivenValues` in the URL form), declared starting values are `startingGivens`, and a bare list of names is `givenNames`. Also folds the control contract onto `Given` (retiring `DashboardGivenSpec` and `DashboardSuggest`), renames package-warning `target` → `subject` (it meant the opposite of a `# drill` target), and maps `NotImplementedError` to 501. Crosses the generated-client boundary, so it needs `bun run generate-api-types`; the Python client is generated at build time from the spec, so it follows automatically. The dashboard-specific schemas are deliberately held back to PR 6 so they can be born correctly named.

**5. Server: tag reading and given control specs.** `service/motly.ts` (MOTLY route filtering, doc-comment text) and `readGivenControlSpec`, with the compiler-contract probes. Valuable alone: it is what lets a notebook's Parameters panel read `# label`, `# control`, and `# suggest` off the declaration.

**6. Server: dashboard discovery, manifest, lint, REST.** `service/dashboard.ts`, the controller, the two read endpoints, the dashboard schemas, load-time lint warnings, `drill_probe.spec.ts`, and the fixtures and integration suite. No UI, so the REST surface is testable on its own.

**7. SDK: shared given and drill primitives, notebook adopts them.** `useGivensState`, the given codec and filter helpers, the enriched controls, and `components/drill/`. The notebook moves onto all of it — same panel, URL state, Apply, `## autorun=false`, drill in cells — and the legacy DimensionFilter Filters panel and `useGivensForm` are removed. Breaking for SDK consumers, and the riskiest review in the set.

**8. Dashboard rendering.** `components/Dashboard/`, the app's `DashboardPage` and `NotebookPage` and routing, the package-page Dashboards section, `docs/dashboards.md`, `docs/choosing-a-surface.md`, and the Playwright spec with the `workers: 1` config change.

**9. Examples and skills.** `storefront` absorbs the HTML data app and grows `dashboards/`, `givens.malloy`, `data_app.malloy`, and `public/app/`; `examples/html-data-app` is deleted; the guided-tour notebook is rewritten; `examples/data-app` gains `PackageDashboard` as the standalone-React proof; `skills/malloy-dashboards` lands. One gate: `malloy-html-data-apps` and `malloy-html-data-app-runtime` are upstream-shared skills, so their 11 changed lines need a PR to `ms2data/agent-skills` first, then a verbatim copy.

**10. Console polish.** The `MALLOY_ACCENT` palette and per-content-type icon tints, the home page rewrite and `docLinks`, Add Connection button parity, data-app chart axis readability and drillable axis ticks, and refreshed screenshots.

## Two bugs the final review turned up

Worth calling out because both were invisible to the typechecker.

**Starting values were silently dropped.** The `startingGivens` rename reached `api-doc.yaml` and the client but not the server's notebook response, which still emitted `givens`. A blanket `as ApiRawNotebook` cast on that object literal meant TypeScript accepted the stale name. Fixed, with the two fields the control row depends on now bound to the schema's names so the cast cannot swallow the next rename. **For review:** that cast exists because the response also carries `type`, `modelPath`, `modelInfo`, and `queries`, which `RawNotebook` does not declare — a pre-existing spec gap that PR 4 is the natural place to close.

**`useGivensState` was losing the values it had just adopted.** The hook copied the manifest's starting values into state with a render-phase update. React can restart a render attempt and drop queued updates, so the prune effect ran against the pre-adoption state and wiped them — controls rendered blank and every cell ran twice, once with the given and once bare. It now derives `draft` and `applied` by reading through to `initial`, keeping only the user's edits in state, tagged with the starting point they were made against. No effect syncs props to state, so both symptoms are gone by construction.

Also: Playwright ran spec *files* in parallel against one shared, mutable server, so a spec reading the package list could catch another spec's environment write and find `storefront` missing. That accounted for four flaky failures across four specs; the config now pins `workers: 1`.

## Test plan

Run from a clean checkout of this branch.

- [ ] `bun install && bun run build`
- [ ] `cd packages/server && bun test src/` — 1,636 pass
- [ ] `cd packages/server && bun test tests/integration/dashboards tests/integration/query_givens tests/integration/sdk_givens tests/integration/authorize tests/integration/html_data_apps tests/integration/legacy_routes` — 72 pass
- [ ] `cd packages/sdk && bun test` — 119 pass
- [ ] `npx tsc --noEmit` in `packages/server`, `packages/sdk`, `packages/app` — clean
- [ ] `cd packages/app && npx playwright test` — 88 pass. The two `environment-connections` specs need the stub BigQuery connection `app-playwright.yml` injects into `publisher.config.json`, so they pass only in CI.
- [ ] Manually: `bun run start:init`, open a `storefront` dashboard, change a control, drill a cell, and confirm the URL round-trips as a shareable link.

Known pre-existing failure, untouched by this branch: `tests/integration/mcp/mcp_transport.integration.spec.ts` and `mcp_execute_query_tool.integration.spec.ts` use `describe.serial`, a Vitest API `bun test` does not have, so they fail when the whole `tests/` tree is run under bun.

## Notes for whoever splits this

- Every commit needs a DCO sign-off.
- `useGivensState` has no unit coverage, because the SDK has no React test harness — the bug above was caught only by Playwright. Adding `@testing-library/react` is worth its own small PR before PR 7 rather than inside it.
- `RELEASE_NOTES.md` keeps several `[Unreleased]` sections below the released `[0.0.208]` heading. That predates this branch; the new sections follow the existing layout rather than reshuffling it.


jswir:
Read this end to end against the branch (built clean, OpenAPI regen included; I didn't run the suites). Agreed on handoff-not-merge, and the ten-way split is the right shape — renames and vocabulary first so the SDK change lands against settled names. The design doc is the best artifact here: "one engine, two document types" is a real constraint rather than a slogan, the Malloy assumptions are pinned against the compiler instead of assumed, and cutting the JSX sandbox is better reasoning than most decisions to ship one.

Findings below are grouped by **where in the staging they belong**, since that's the actionable axis.

---

## Fix inside slice 7, before the shared layer is load-bearing

Both live in `useGivensState` — the primitive both surfaces now depend on, and the one hook with no unit coverage.

**1. Reset re-applies the document's starting values, and runs every query twice.** `clearAll` empties the edit map, which is reported outward as an empty parameter set; the host clears the query string; `initial` recomputes back to `# artifact { givens {…} }`; the new key discards the edit. Controls snap back to where they started, and the starting values get written into the URL as explicit params.

Repro on the bundled example: open `storefront/dashboards/regions` (opens at `REGION=West`), press **Reset**. Tiles run unfiltered, then run again at West. On a notebook that's every cell twice — the same double-run the branch fixed for the mount path, reintroduced through Reset. Clearing an individual control is fine; it commits `""` rather than deleting the key.

**2. Writing parameter state to the URL discards every other query parameter.** `DashboardPage` and `NotebookPage` both read `givens` from the whole search string and write it back with `setSearchParams(next, { replace: true })`, so the first control change strips anything that wasn't a given. `paramCodec` is explicitly written to tolerate an unrelated param — *"a tracking tag, say — cannot break the dashboard"* — and the host then deletes it. Merge into the existing params instead of replacing.

**On the test harness:** the note about adding `@testing-library/react` before slice 7 rather than inside it is right, and both of the above are in the hook it would cover. The state bug you caught was found only by Playwright; these two would be as well.

---

## Fix while the code is open

**3. A dashboard with no `title=` renders its title twice.** The manifest sets `title: artifact.title ?? doc ?? name` and `description: doc` from the same `#"` comment (pinned by unit test, so intentional at that layer), and `DashboardHeader` renders both. `dashboards/regions.malloy` uses only a doc comment, so its page reads "Sales by region" on two consecutive lines. `getNotebookListing` has the same shape. One-line fix in the header: suppress the description when it equals the title.

**4. `#(description="…")` is a malformed annotation prefix, and the compiler now says so.** Malloy takes a route from the sigil to the first whitespace, so this parses as route `(description="Narrow` and the editor flags all five declarations in the new `examples/storefront/givens.malloy`.

Not introduced here — it's on `main` in `docs/givens.md`, `docs/row-level-access.md` and `examples/governed-analytics/orders.malloy` — but this PR is where it should die, for two reasons:

- It's the one control tag left behind. The whole move here was lifting `label`, `control`, `range_min/max` and `suggest` into a server-derived `GivenControlSpec`, on the stated grounds that a client shouldn't have to re-parse them. `description` is still re-parsed client-side, by regex over raw annotation strings in `GivenInput.annotationHelperText`.
- It's an unpinned compiler assumption, in a repo that pins its others (`drill_probe`, `exports_probe`) precisely so an upgrade fails loudly. It currently survives only because the malformed route happens to contain a letter, so `isReservedRoute` lets it onto the wire.

Suggested shape: add `description` to `GivenControlSpec`, read it from a well-formed `# description="…"`, have the control read `given.description`. **Careful with the obvious shortcut** — switching to `#(doc) …` leaves `annotationHelperText` matching on `startsWith("#(")`, finding no `description=`, and rendering `doc) Narrow to one product category…` as the helper text.

---

## Ordering constraints for the split

**5. Slice 10 links to docs that slices 2 and 8 provide.** `docLinks.ts` points at `blob/main/docs/choosing-a-surface.md` and `dashboards.md`, both new here. Fine on the branch as a whole; merge 10 first and the home page ships two live 404s. Either constrain the order, or point `REPO_DOCS` at the released tag rather than `main` so a link can't outrun its file. Worth an eleventh line in the notes.

The upstream-skills gate on slice 9 is already noted and looks right.

---

## Worth a look, lower confidence

**6. A single-query dashboard that renders as a plain chart may be drawn 20,000px tall.** The single-query form passes `WHOLE_PAGE_HEIGHT` as the tile's `maxHeight` and no `chartHeight`; when the renderer reports `kind === "chart"` the policy resolves to `min(maxHeight, chartHeight ?? maxHeight)`. Every bundled dashboard carries `# dashboard`, so `kind` is `"dashboard"` and the path is never exercised — a one-chart dashboard (top-level `# line_chart`) would hit it. Reasoned from the code path, not observed; worth thirty seconds in a browser.

**7. A failed suggest query is indistinguishable from an empty one.** `useSuggestOptions` generates `run: <source> -> { group_by: <dim> }` with no `limit`, so a high-cardinality dimension meets the row cap and errors, and nothing surfaces — the dropdown just reads as having no options. `freeSolo` softens it. A cap on the suggest query plus a visible error state would both help.

**8. A notebook's Parameters panel scrolls away from the cells it governs.** A control there re-runs every cell, but on a long notebook a reader looking at a chart can't see what it's filtered to, or change it without losing the chart. A dashboard never has this problem — control row and grid are on one screen — so this is a per-document-type refinement, which the guiding principle explicitly carves out. `position: sticky` on the panel is enough; I prototyped it to check, and the scroll container is `MainPage`'s `overflow: auto` with no `overflow: hidden` in between. Guard on there being controls to pin, since `GivensPanel` returns null with no givens. If the pinned strip reads as heavy, `layout="bar"` already exists.

---

## Two smaller notes

**9. The second controls implementation isn't actually gone.** The design doc says the legacy panel and the hook behind it "were deleted, not left beside the shared one." `useGivensForm` was — but `components/filter/DimensionFilter.tsx`, the three `useDimensionFilters*` hooks and their README sections remain SDK public surface with no in-repo consumer, and the server's `filter_params` path is untouched. Either finish the removal or narrow the claim; as written a reader goes looking for code that's still there.

Related, and worth confirming with real users before slice 7 lands: removing the panel means a notebook whose sources carry `#(filter)` annotations silently loses its filter UI. The release notes are straight about it and the migration path to `given:` is right — just want to be sure nobody's mid-migration.

**10. One Playwright assertion asserts a proxy rather than the thing.** The "lists them only once" check is a page-wide `getByRole("button", { name: /dashboards\/overview\.malloy/ }).toHaveCount(0)` standing in for "the row isn't under Semantic Models." It can't scope, because `PackageSection` is an anonymous `Box` with no landmark or name — so any future change that legitimately mentions that path elsewhere on the page fails a test whose behaviour is still correct. Giving `PackageSection` a `component="section" aria-label={title}` makes each section a named landmark (a real a11y win — six sections a screen reader can jump between) and lets the assertion scope to the list it means.

I hit this concretely: the Dashboards row's secondary label is the slug, which is just the title again in lower case, where Notebooks and Data Apps both show their `path`. Switching it to `dashboards/overview.malloy` — one file is one dashboard, so it's unambiguous, and it's what a reader needs to go edit the thing — breaks that assertion for the wrong reason. Small, but it belongs with slice 8 rather than as a drive-by.

---

## What holds up well

- The two Malloy behaviours the design leans on are pinned against the real compiler, not a re-implementation: a dimension's annotations reaching the result's output field, and per-file given bindability. If an upgrade broke either, drill would stop firing everywhere at once and those tests are what would say so.
- The filter-syntax encoder is a single tested module, and the follow-up commit fixing the drifted hand-rolled copy in the example is the right instinct — a value carrying a backslash used to end its own quoting early.
- The lint refuses to guess. Only tile forms that can be *proved* wrong are reported, and a `to=self` is called dead only when no model in the package declares its given.
- The Playwright flake was diagnosed to mechanism — spec *files* racing one shared mutable server — rather than retried away, with the reasoning in the config.
- `ownModelNotes` (slice 3) stands entirely on its own and should go in early regardless of the rest: an imported model's `##` tags were setting title and autorun for every notebook importing it.

mlennie:
Reviewed the branch independently (built clean, full gate run, and drove both surfaces live). Agreed on handoff-not-merge and on the ten-way split. James has already covered this thoroughly, so rather than restate his findings I have grouped mine the same way, by slice, and flagged only what is new, what corrects something, or where I saw a different symptom than he did.

**Gate on the branch head:** typecheck 0, `bun test src/` 1,636 pass, integration 259 pass across 32 files, prettier clean apart from the six git-ignored generated files. The generated clients need no regeneration: all four targets are git-ignored, `openapi-spec-validator` passes on this spec, and `openapi-python-client generate` succeeds with no warnings. One note on the test plan: the repo's CLAUDE.md says the Python pre-commit hook's `git diff --exit-code` catches spec drift, and it cannot, because the output is git-ignored. The CI regen job is the real gate.

## The one that changes an assessment: the filter encoder does not work

This is the finding I would most want checked before slice 7, because it contradicts "the filter-syntax encoder is a single tested module" sitting under what holds up well. The module is well-structured and well-tested. Its central mechanism is not real.

`encodeFilterList` quotes values by wrapping them in `"…"`. **Malloy's string-filter grammar has no double-quote quoting; backslash is its only escape.** So the quoting is not merely inert, it corrupts. Feeding the encoder's own output straight into `@malloydata/malloy-filter`:

```
encodeFilterList(["Ben & Jerry, Inc"]) -> '"Ben & Jerry, Inc"'
  parsed: {operator:'=', values:['"Ben & Jerry', 'Inc"']}      <- TWO values, plus stray quote chars
```

`filterValue.ts`'s own docstring says a comma "has to be quoted or `Ben & Jerry, Inc` would filter for two brands". It quotes, and it still filters for two brands. Which also means `68c48c27` is correct reasoning built on a mechanism that does not exist, so I would not treat that one as closed either.

The wider problem is that unescaped operator characters in ordinary data become operators. Measured end to end against `data_app.malloy`, counting rows on `scoped_orders`:

```
CATEGORY=''           -> 25356    (no filter)
CATEGORY='Outerwear'  ->  2535    (correct)
CATEGORY='-Outerwear' -> 22821    (= 25356 - 2535: NEGATED)
CATEGORY='%'          -> 25356    (filter bypassed entirely)
CATEGORY='null'       ->     0    (the null operator)
```

Reachable three ways: a drilled cell value, a picked suggest option, or a hand-edited URL. And it surfaces with no signal at all: loading a notebook at `?CATEGORY=-Outerwear` renders the control showing `-Outerwear` as though it were an ordinary selected category while the cells run the negation.

Fair framing on severity: this is not privilege escalation, because `docs/security-posture.md` is explicit that a given is whatever the caller sends, so anyone who could poison a value could also POST `REGION=%` directly. What it is, is a reader silently getting rows outside the scope they asked for, driven by data they do not control, which defeats the row-scoping pattern `docs/givens.md` advertises and `governed-analytics` ships.

Fix shape: backslash-escape `\` `,` and a leading `-`, escape `%` and `_` so a literal stays literal, escape the bare `null` / `empty` keywords, and give `decodeFilterList` the matching inverse. Both copies (the SDK and `examples/storefront/public/app/format.js`).

Two consequences of the same root cause, both proved by running them:

- **The flagship example's "All" option returns zero rows.** `controls.js` sets `all.value = ""`, `encodeFilterValue("")` yields the two-character string `""`, and Malloy matches that literally. Pick a category, go back to "All", every tile empties, and it is in the shareable URL as `?CATEGORY=%22%22`.
- **`decodeFilterList` is order-dependent.** `["Nike","Ben & Jerry, Inc"]` round-trips to `["Nike"," Ben & Jerry, Inc"]`, because the `", "` join puts the space outside the quotes and `wasQuoted` suppresses the trim. `filterValue.spec.ts` misses it purely by argument ordering: every case is single-element or puts the quoted value first. Reverse that one pair and it fails.

## Slice 7, in addition to James's two

**`#(filter)` is not deprecated on the notebook surface, it is broken, and the release notes say otherwise.** This sharpens James's note 9. `Notebook.tsx:216` now passes a hardcoded `undefined` where `origin/main` passed `filterParams`, and the server is not inert: `filter.ts:254` still throws `FilterValidationError` when a `required` filter has no value. So for a model with a required `#(filter)`, every cell fails server-side with no UI able to satisfy it, and because the per-cell catch only `console.error`s, the reader gets a blank cell with no message. That catch is byte-identical to `origin/main`, so the swallowing is not new; what changed is that reaching it went from rare to routine for those models. `RELEASE_NOTES.md:243` says "the server side is unchanged and still honours the deprecation window; only the UI panel is removed", and `:270` says such models "will continue to work unchanged". Neither holds for required filters, and for the rest "unchanged" means "unfilterable". No shipped example uses `#(filter)`, so no gate catches it.

**An undocumented breaking change to a published component's props.** `onNavigate` went from `(to: string, event?: React.MouseEvent)` to `(to: string, event?: NavigationClick)`, a four-key `Pick`. Under `strictFunctionTypes` that is contravariant, so an existing host handler stops compiling:

```
error TS2322: Type 'OldProp' is not assignable to type 'NewProp'.
  Type 'NavigationClick' is missing the following properties from type
  'MouseEvent<Element, MouseEvent>': altKey, buttons, clientX, clientY, and 25 more.
```

`RELEASE_NOTES.md:245` enumerates the other SDK breaks carefully and this one is missing. One line fixes it.

**Reset: I saw a different symptom than James did, so it is worth checking both.** On `storefront/dashboards/regions`, changing Region to Northeast, pressing Apply, then pressing Reset:

```
open        : control="West"      kpi=$488,718.06  url=?REGION=West
applied NE  : control="Northeast" kpi=$408,550.88  url=?REGION=Northeast
after Reset : control="Northeast" kpi=$408,550.88  url=?REGION=Northeast
```

Reset is completely inert: no URL change, no re-render, and zero queries fired. I did not reproduce the run-unfiltered-then-run-at-West double-run he describes, and I tried it both with and without Apply first. Either way the button does not do what it says, and the two observations may share a cause in `clearAll` reporting an empty edit map outward.

**The in-flight run guard was replaced by supersede-without-cancel.** `origin/main:Notebook.tsx:493` deliberately refused to start a run while one was executing, with a comment explaining why. The new effect supersedes instead, and the generation guard discards *results* without cancelling requests: there is no `AbortController` anywhere in the notebook path. So three quick control changes on a 12-cell notebook queue three full runs, four concurrent each, all but the last thrown away. The stale-result correctness is genuinely better than before; the load ceiling that came with the old design is gone and was not replaced.

**Confirming James on the orphaned second implementation**, with the concrete state: `useDimensionFilters` and friends are still exported from `hooks/index.ts`, and the only remaining consumer is another orphaned hook in the same dead cluster. Zero product consumers.

## Slice 6: one omission produces three findings

`discoverDashboards` is the only listing path in `package.ts` that never consults `exploreSet()`. On a package with `explores` set and `queryableSources: "declared"`, where the dashboard file is not listed, `GET /dashboards/<slug>` returns the full manifest, query name, given names, types, defaults and internal suggest-query names, while every one of those names then 404s. Notebooks are uncurated in *both* directions; dashboards took "always listed" without "always queryable", so a curated package gets the worst of both. I verified it by staging two copies of `governed-analytics` differing only in `publisher.json`.

The same path also serves the compile error for a dashboard file that failed to compile, and compile errors name tables, columns and connections. And because it lists any `dashboards/*.malloy` that failed to compile regardless of tag, a syntax error in `dashboards/_shared.malloy` produces a phantom dashboard that vanishes once fixed, which `api-doc.yaml` says cannot happen ("a file in that directory with no artifact tag is a shared include and is not listed"). A load-time lint when a dashboard's `entryFile` is not in `explores` under `"declared"` would close the first, and `lintDashboard` already has what it needs.

Worth saying `docs/dashboards.md:214` documents the curation requirement well. `skills/malloy-dashboards/SKILL.md` mentions `explores` zero times, though, and that is the surface an agent acts on: it would follow the skill on a curated package, get zero lint warnings, see the dashboard listed with its title, and 404 on open.

**Also slice 6: a `## artifact` tag with no `tiles=` vanishes silently.** The composite branch is gated on `composite?.tiles?.length`, and `lintUndiscoveredDashboard` only fires on tags that fail to *parse*. `## artifact { title="Overview" }` parses fine, so it produces no dashboard and no finding. The docstring calls that outcome "the least debuggable outcome in the whole feature", which is right, so the parses-but-incomplete case probably wants the same treatment.

Smaller, same slice: `api-doc.yaml`'s `DashboardManifest` leaves all seven shared fields undescribed where `Dashboard` describes each, and nothing in the spec says `path` is the model you POST to, which is the load-bearing fact behind having no run endpoint. Only `docs/dashboards.md:386` says it.

## Slice 8

**Confirming James's duplicate title, with the manifest values.** `GET …/dashboards/regions` returns `title: 'Sales by region'` and `description: 'Sales by region'`, and the page renders both on consecutive lines. Verified in a browser.

**The drill affordance is mouse-only.** `markDrillableCells` adds a class and nothing else: no `tabindex`, no `role`, no key handler. The whole affordance is `cursor: pointer` plus a hover colour and underline, and the comment at `RenderedResult.tsx:308` says the hover state is "the only thing that tells a reader a cell navigates", with the resting state left plain on purpose so a blue column does not compete with the data. Sound visual reasoning, but a keyboard user cannot discover or activate a drill, and a touch user gets no affordance either. Since a drill is a navigation action, a focusable element with a key handler seems worth it, and it would give the hover style a `:focus-visible` to attach to. This dovetails with James's note 10 about `PackageSection` landmarks.

**`# dashboard { columns=N }` is never validated.** The lint checks only the composite `dashboard_columns`, and three of the four shipped storefront dashboards use the unvalidated spelling, so `columns="wide"` is silently dropped with no finding.

## Slice 9

Beyond the "All" option above:

- **Any unknown query parameter breaks the data app.** `app.js` runs `readUrl(); render();` before `await loadGivenContracts()`, and on that first pass `contracts` is empty so every parameter is sent as a given. Live on `/environments/examples/packages/storefront/?utm_source=slack`: four queries 400 and the page shows `givens: unknown given 'utm_source'. Model surfaces [CATEGORY, BRAND, REGION, SINCE, MIN_SALE]`. Contracts land, the second `readUrl()` drops the key, but no `render()` follows, so the boxes stay until the user touches a filter. Worth being clear this is the example's problem and not the engine's: `paramsToGivens` drops undeclared names on purpose and its docstring says exactly why ("an unrelated query parameter … a tracking tag, say — cannot break the dashboard"). Verified the asymmetry live: the same parameter on the notebook route renders all 43 cells with zero failed queries. Same class as the other hand-rolled drifts.
- The global `#errors` box is append-only, so with an error active each filter change stacks another identical box until reload.
- `format.js`'s `readAtLeast` is unanchored and does not require `>=`, so `<= 500` reads as 500 and re-encodes as `>= 500`, the opposite filter. The SDK's `decodeAtLeast` anchors and returns undefined.
- `controls.js` extracts `#(description=…)` with a stricter regex than `GivenInput`, so any other annotation shape shows a hint in the Console and nothing in the page.
- `drill.js` does not check that a drill's given is declared, where the SDK warns and does nothing.
- **The five shipped example dashboards are exercised by nothing.** The new 627-line `package-dashboards.spec.ts` runs against `fixtures/dashboards-test`, and the integration suite does not cover them because a per-model compile failure is non-fatal. They do work today, I drove all four, but they could break with every gate staying green.

## Slice 2, and one correction to James

- **`docs/givens.md` documents `string[]` as a given type in three places**, including a widget table promising "multi-value autocomplete with chip removal". It does not compile (`error syntax-error : unexpected ']'`), and this PR's own new `given.ts:150` says why: Malloy's grammar emits only scalar parameter types for given declarations. All three lines predate the branch, so this is a pre-existing doc bug the PR stands next to rather than introduces, but `given.ts:150` is new and says the opposite. Related: the `array<…>` branches in `GivenInput.tsx:439` and `paramCodec` are unreachable for the same reason and could be deleted rather than maintained.
- **Correction to James's note 7.** The suggest query does have no `limit:`, but it does not error on a high-cardinality dimension: `resolveModelQueryRowLimit` applies `DEFAULT_QUERY_ROW_LIMIT = 1000`, well below the 100,000 cap, so the option list is **silently truncated to 1000 in arbitrary order** rather than failing. That makes the missing error state slightly less urgent and the missing `limit:` slightly more so, since the truncation is invisible either way. His conclusion (cap the query, surface an error) is still the right fix.
- **Narrowing his note 2.** Confirmed that both hosts replace the whole query string, and merging is the right fix. Worth knowing the blast radius is smaller than it looks: queries are unaffected because `paramsToGivens` filters to declared names, so what is lost is the unrelated parameter itself, not the dashboard.
- `docs/discovery-and-access.md:72` quotes `governed-analytics/publisher.json` as a two-entry `explores` list; this PR makes it three by adding the dashboard, and that doc is not in the diff, so it now misquotes the example it links.
- `docs/dashboards.md:357`'s load-check list omits four findings the code emits (the malformed-tag one, the `.jsx`/`.tsx` component warning, bare `# drill` with no `to=`, and "tile filters by a given this file does not import"). `SKILL.md:234` names 4 of about 10; its operative instruction is still correct since "fix all of them" attaches to the package warnings rather than the list, so that one is a completeness gap.
- `RELEASE_NOTES.md:124` says starting values arrive as `givens` on `RawNotebook` and `DashboardManifest`; the field is `startingGivens`, and line 235 records that rename in the same document.
- On James's note 4: agreed, and his reasoning about `description` being the one control tag left client-side is the better argument. Adding only the measurement: this PR introduces 6 new multi-word instances (5 in `givens.malloy`, 1 in `orders.malloy`), with 3 pre-existing, and every compile of every model in `storefront` carries them because `givens.malloy` is imported everywhere.

## Cross-cutting, worth an eleventh line in the notes

**`examples/data-app` is covered by no gate at all.** It is not in root `workspaces` (`["packages/*"]`), not in the `typecheck` or `lint` chains, absent from `.github/workflows` and the Makefile, and its own `package.json` has no lint script. So the 119 new lines of `PackageDashboard.tsx` are never typechecked, linted or built by CI. They are correct, I checked by hand, but the standalone-React proof is the one artifact with no automated coverage. Also `typecheck:cli` and `lint:cli` are defined but unchained, which is pre-existing. In that file, `Promise.reject(res.statusText)` can reject with `""`, which is falsy, so the error Alert never renders and the user gets the heading and blurb over blank space, with no loading or empty state either.

## Separately: the `index.malloy` follow-on

I am picking up the `index.malloy` convention from `docs/malloyyo-dashboards-design.md:681` next, so flagging the one place I plan to depart from that spec, since it is yours.

The doc says to default `explores` to `["index.malloy"]` **and** `queryableSources` to `"declared"` when a package has an `index.malloy` and no explicit `explores`. I plan to do the first and not the second. `queryableSources` already defaults to `"declared"`, and its enforcement is a 404 deliberately indistinguishable from "does not exist", so defaulting it off the presence of a filename means an existing package that merely happens to contain an `index.malloy` silently loses query access to every other source by name, with no config edit and no actionable error. Malloyyo can be absolute here because every project is greenfield and `--develop` is the escape hatch; we have packages in the wild and no mode split. So the convention would curate listings, and closing the query boundary would stay an explicit act. Explicit `explores` wins either way, and a disagreement between the two gets the load-time warning your doc specifies. If you intended the stricter reading, say so and I will build it that way with a migration note.

The `discoverDashboards` finding above is the same question wearing a different hat, so whatever we settle for `index.malloy` should answer it for dashboards too.

mlennie:
### The ten-way split is complete: all slices are merged into `main`

Closing this out. Every slice proposed above landed as its own reviewable PR, plus a handful the reviews turned up along the way. Nothing from this branch is outstanding, so the conflicts now showing here are against work that has already been merged by other means and are not worth resolving.

The split held up. Each slice was independently reviewable, the renames and the vocabulary went first so the SDK change landed against settled names, and the design doc arrived early enough to be the map for the slices that followed.

**Slice 1 — renames**
- *refactor: Pages become Data Apps, and the App becomes the Console* — https://github.com/malloydata/publisher/pull/972
    The three mechanical renames, no behaviour change: `PageViewer` → `DataAppViewer`, `pageEmbed` → `dataAppEmbed`, `docs/publisher-app.md` → `docs/console.md`, and the "Governed Reports" label → "Notebooks".
- *feat!: rename the /pages endpoint to /data-apps* — https://github.com/malloydata/publisher/pull/973
    The REST break split out for its own review, as the plan suggested it might want to be.

**Slice 2 — docs and the design record**
- *docs: land the dashboards design doc and the security posture* — https://github.com/malloydata/publisher/pull/969
    `docs/malloyyo-dashboards-design.md` first, so it was the map for slices 5 through 9, plus `docs/security-posture.md`, the docs index, and the `api-doc.yaml` description rewrite.
- *docs: stop the dashboards design doc claiming what has landed* — https://github.com/malloydata/publisher/pull/984
    A follow-up: the design doc described the end state in the present tense while most of it was still unmerged, which read as a status claim rather than a design.

**Slice 3 — model-annotation ownership**
- *fix(server): read a document's own model annotations, not its imports'* — https://github.com/malloydata/publisher/pull/967
    The standalone correctness fix, merged early as recommended: an imported model's `##` tags had been setting the title and autorun for every notebook that imported it. Adds `ownModelNotes` while keeping full lineage for `authorize`.

**Slice 4 — givens vocabulary and API contract**
- *refactor(api): give givens one meaning across the spec* — https://github.com/malloydata/publisher/pull/976
    `givens` had come to mean four things. Now it is always a collection of `Given` declarations, with `GivenValues`, `EncodedGivenValues`, `startingGivens` and `givenNames` for the others. Folds the control contract onto `Given`, retires `DashboardGivenSpec` and `DashboardSuggest`, and renames the package-warning `target` → `subject`.

**Slice 5 — server: tag reading and given control specs**
- *feat(server): read a given's control contract off its own tags* — https://github.com/malloydata/publisher/pull/981
    `readGivenControlSpec` and the MOTLY route filtering, with the compiler-contract probes. This is what lets a notebook's Parameters panel read `# label`, `# control` and `# suggest` off the declaration.

**Slice 6 — server: dashboard discovery, manifest, lint, REST**
- *feat(server): discover, lint and serve package dashboards* — https://github.com/malloydata/publisher/pull/987
    `service/dashboard.ts`, the controller, the two read endpoints, the dashboard schemas, load-time lint warnings, and the fixtures and integration suite. No UI, so the REST surface was testable on its own.

**Slice 7 — SDK: shared given and drill primitives**
- *test(sdk): add a DOM test harness, and fix two defects it surfaced* — https://github.com/malloydata/publisher/pull/978
    Landed first, as the notes recommended, rather than inside slice 7. It immediately caught two defects in the hook the harness exists to cover.
- *feat(sdk): shared given and drill controls, and a notebook built on them* — https://github.com/malloydata/publisher/pull/985
    `useGivensState`, the given codec and filter helpers, the enriched controls, `components/drill/`, and the notebook moving onto all of it. The riskiest review in the set, taken with the harness already in place.
- *test(sdk): pin a given's date to UTC, and run the suite east and west of it* — https://github.com/malloydata/publisher/pull/1019
    A date given had been passing only in the author's timezone.

**Slice 8 — dashboard rendering**
- *feat(sdk,app): render a dashboard, and route to it from the Console* — https://github.com/malloydata/publisher/pull/1043
    `components/Dashboard/`, `DashboardPage`, the routing, the package-page Dashboards section, `docs/dashboards.md`, `docs/choosing-a-surface.md`, and the Playwright spec.

**Slice 9 — examples and skills (split in three, and gated on an upstream PR first)**
- *docs(skills): stop the two data-app skills naming a shipped example* — https://github.com/malloydata/publisher/pull/1035
    Slice 9a. The upstream-shared skills gate the plan flagged: `malloy-html-data-apps` and `malloy-html-data-app-runtime` are shared with `ms2data/agent-skills`, so the change went there first (merged as `d4f066b`) and came back as a verbatim copy, exactly as the notes required.
- *examples(storefront): drive the data app from givens, and let Malloy escape them* — https://github.com/malloydata/publisher/pull/1027
    Slice 9b. storefront absorbing the HTML data app, plus `dashboards/`, `givens.malloy`, `data_app.malloy` and `public/app/`. The "let Malloy escape them" half is the filter-encoder fix described below.
- *feat(skills): a dashboards authoring skill* — https://github.com/malloydata/publisher/pull/1020
    Slice 9c. `skills/malloy-dashboards`.

**Slice 10 — Console polish**
- *feat(sdk): Console polish — say what this server does, and give each content type its own signal* — https://github.com/malloydata/publisher/pull/1026
    The accent palette and per-content-type icon tints, the home page rewrite and `docLinks`, Add Connection parity, and data-app chart axis readability. Landed after slices 2 and 8 so its doc links could not outrun the files they point at, which was one of the ordering constraints raised in review.
- *fix(sdk,app): three follow-ups from the #1026 review* — https://github.com/malloydata/publisher/pull/1051
    The deferred items from that review, split out rather than folded in.

### One loose end, and it is not from this branch

- *fix(sdk,app): restore a loading-gate comment lost in a squash* — https://github.com/malloydata/publisher/pull/1058
    Still open. A review comment from the #1043 discussion was dropped when slice 8 was squashed; this puts it back. Comment-only, zero behaviour change. It is the only slice-derived PR not yet merged, so noting it rather than letting the summary above read as "everything is in".

### What the two reviews changed

Both reviews are worth reading against the merged result rather than the branch, because several findings changed the shape of what shipped rather than just fixing a line.

The one that changed an assessment was the filter encoder. It quoted values with `"…"`, and Malloy's string-filter grammar has no double-quote quoting, so the quoting did not merely fail to help, it corrupted: `encodeFilterList(["Ben & Jerry, Inc"])` parsed back as two values. Unescaped operator characters in ordinary data became operators, so a drilled value of `-Outerwear` ran a negation while the control displayed it as an ordinary selection, and `%` bypassed the filter entirely. That is the fix in the second half of #1027, applied to both the SDK and the hand-rolled copy in the example.

Also worth recording: `#(description="…")` was a malformed annotation prefix that the compiler had started flagging; the second controls implementation was still present as SDK public surface after the design doc said it had been deleted; `#(filter)` on the notebook surface was broken rather than deprecated, with the release notes saying otherwise; `discoverDashboards` was the only listing path that never consulted `exploreSet()`, so a curated package could list a dashboard whose every query then 404'd; and a `## artifact` tag with no `tiles=` vanished with no dashboard and no lint finding.

PR #936, opened 2026-07-28
feat(connections): support METADATA_SCHEMA for DuckLake catalogs

housejester:
Lets several independent DuckLake catalogs share one catalog database, by giving each its own schema for its `ducklake_*` metadata tables.

**Ships as a no-op.** `catalog.metadataSchema` is optional and absent means DuckLake's default — the catalog connection's default schema. Every existing config emits exactly the command it emitted before.

## 1. `METADATA_SCHEMA` support

`catalog.metadataSchema` on a DuckLake connection is threaded into the ATTACH as `METADATA_SCHEMA '<schema>'`. Validated at config load to `^[A-Za-z_][A-Za-z0-9_]*$`, and to being a string: the value comes from untyped JSON and `RegExp.test()` coerces, so `true` would otherwise pass the pattern as `"true"` and fail inside `escapeSQL` at the first attach — the runtime failure the load-time check exists to prevent.

**The preflight is the part worth reviewing.** `preflightDuckLakeCatalogFormat` range-checks the catalog's recorded format so an unsupported catalog fails as a clean 422 rather than a deep DuckDB error. It reads `ducklake_metadata` — which moves when the metadata schema moves — and it **fails soft**, logging and returning. So passing `METADATA_SCHEMA` without also qualifying that read would have silently disabled format checking for exactly the catalogs using the new option, with nothing in the logs to say so. The read is now `<db>."<schema>".ducklake_metadata`.

Its test poisons the recorded version to `9999.0` and asserts both that `UnsupportedCatalogFormatError` is raised and that the message quotes `9999.0`, so it discriminates rather than passing off a sibling test's leftover catalog.

The "no catalog found in this schema" warning is deliberately hedged: a fresh named schema legitimately has no `ducklake_metadata` until the first read-write attach creates it, so the message says that is expected, and `does not exist` can equally mean a missing database or role.

## 2. Instance-isolation guard (tests only)

`db-duckdb` pools instances on a share key that excludes the connection name, while the DuckLake attach is `ATTACH OR REPLACE … AS <connectionName>`. Two same-named connections sharing an instance would therefore clobber one another. That does not happen today because `databasePath` and `workingDirectory` differ per connection and per environment, and both are in the share key — but nothing pinned it.

`duckdb_instance_isolation.spec.ts` pins it behaviourally. **Ordering in that test is load-bearing** and commented as such: constructing both connections before using either lets the two `init()` promises race past the cache miss so each gets a private instance, which is false isolation. Initialising sequentially and then forcing a shared working directory reproduces the clobber (`a=2, b=2`), which is what shows the test can fail. Each connection creates and populates its own attached store, so nothing reopens a file another connection still holds — Windows enforces that lock where POSIX advisory locking tolerates it.

It deliberately does not assert the converse (that same key ⇒ same instance): that is `db-duckdb`'s pooling policy, not this repo's contract.

## 3. Logging

The three per-attach diagnostics moved to `debug`. `redactPgSecrets` removes the password and URI userinfo, but a `DATA_PATH` is not a secret it knows about, so logging the assembled ATTACH at info disclosed the bucket URL and whatever its prefix encodes to every reader of the logs. The mode and the outcome stay at info.

## Docs

`docs/ducklake.md` gains a co-tenancy section, and two corrections it made visible:

- **One data path per catalog.** Separating metadata does not separate data, and `ducklake_delete_orphaned_files` works from `DATA_PATH` — aimed at one of two catalogs sharing a prefix it treats the other's Parquet as orphaned, irreversibly. Publisher never deletes by prefix (its sweeps enumerate through the attached catalog), but the tooling around it does.
- **Both attach modes.** The docs claimed Publisher only ever attaches read-only, which a `#@ persist storage=` build has not matched since it landed. Serving attaches read-only; a build attaches read-write on a transient session; neither sets `AUTOMATIC_MIGRATION`.
- **Version skew.** Configuration validation ignores unknown keys, so an older server accepts `metadataSchema` and silently attaches the default schema — reading the wrong catalog, and writing into it, with no error either way if that schema holds a catalog.

The separation this gives is called out as **organizational, not access-control**: what separates two catalogs is which schema a configuration names, so a role that can reach a sibling schema could name it. Per-schema roles, or a database per catalog, are what make it a boundary.

## Testing

- 1585 unit tests pass; typecheck and Prettier clean.
- Live DuckLake integration tests against real Postgres cover the accept/reject boundary for `metadataSchema`, the schema-qualified preflight (fails without the qualification), a mixed-case schema, and that two catalogs in one database stay mutually invisible.
- The non-string validator case fails without its `typeof` check, verified by removing the check.


kylenesbit (approved):
Reviewed against the tier design ([`docs/ducklake-materialization-tier.md`](https://github.com/ms2data/service/blob/dev/docs/ducklake-materialization-tier.md)) and the sibling persistence docs. Every constraint it names is honored: C3's derived range rather than a resurrected `SUPPORTED_CATALOG_VERSIONS` list, `AUTOMATIC_MIGRATION` never set in either mode, and C4's fail-soft preflight — which is really the insight here, since threading the schema is what keeps the range check from silently protecting nothing for exactly the catalogs using the new option. The attach-mode correction also closes a real divergence: §5's owner × role table has said read-only-to-serve / read-write-to-build since 2026-07-15 and `docs/ducklake.md` had claimed read-only-always since `storage=` landed.

Verified on `4c368d9`: `typecheck:server` clean, 1585 server unit tests pass (0 failures).

Two things, one of which is a fix on the *service* side rather than here.

---

### The one place the two documents disagree

This PR's docs say, twice and correctly, that `metadataSchema` is *"an organizational boundary, not an access-control one"* — what separates two catalogs is which schema a configuration names, so any config whose catalog role can reach a sibling schema can name it and read that catalog's metadata, and therefore its data-file paths.

The tier design asserts the opposite shape without flagging it:

- **§3 Tenancy** lists *"shared Postgres, per-org database/schema separation, quotas per org, encryption at rest"* as the v1 tenancy properties.
- **§9 Q3** specifies the spike's arrangement concretely: one `ducklake` logical database plus *"a dedicated `ducklake` Postgres user scoped to that one database."*

One role, one database, per-org schemas is precisely the configuration this PR says is not a boundary — every org's catalog metadata is reachable by the same user. (The data boundary Q3 *does* claim — *"grants don't reach the `controlplane` database — workers still cannot read CP tables"* — is a different boundary and remains true.)

So the design needs either per-schema roles added to the D1b provisioning item, or §3's tenancy bullet reworded to stop implying schema separation is tenant isolation. Worth settling before D2, because the spike would otherwise bake in the weaker arrangement and read as though the tenancy question were closed. The docs in this PR are the ones I'd keep; the design is what should change.

### The one code suggestion

`OVERRIDE_DATA_PATH true` is still unconditional in the attach, and it interacts with the new co-tenancy guidance in a way nothing detects. The docs now tell operators to give each catalog its own `bucketUrl` prefix and explain the irreversible consequence of not doing so (`ducklake_delete_orphaned_files` treating a co-tenant's Parquet as orphaned). But the likely mistake is copying a connection config and changing only `metadataSchema` — and with `OVERRIDE_DATA_PATH` the configured path silently wins over whatever the catalog recorded, so that misconfiguration produces no error anywhere.

The preflight is already attached to the catalog Postgres read-only and already reading `ducklake_metadata`, and the recorded `data_path` is a row in that same table. Reading it in the same query and warning when it disagrees with the configured `bucketUrl` would turn the documented rule into a detected one at no extra round-trip.

One caveat that argues for landing this with D2 rather than in this PR: per the design, `OVERRIDE_DATA_PATH` exists because of a real path-computation mismatch that D2 hasn't verified away (§5: *"expected unnecessary for catalogs we provision … but that's verified in D2, not assumed"*). So it has to be a warning scoped to a genuine divergence rather than a prefix or relative-path difference, or it will fire on every attach.

PR #937, opened 2026-07-28
ci(release): trigger the skills and create-malloy-package publishes from release.yml

mlennie:
One release now publishes every package. `release.yml` gains a `publish-packages` job that, for `@malloy-publisher/skills` and `@malloy-publisher/create-malloy-package`, reads the version from `main`, asks npm whether it is already published, skips it if so, and otherwise dispatches that package's own publish workflow and waits for the version to land on the registry. This is the follow-up promised in the comment added to `create-malloy-package-npm.yml` in #926.

## No npm settings change is required, and none should be made

Worth stating up front, because the obvious implementation would have required one. npm allows a single trusted publisher per package and matches it against the **top-level calling workflow's filename**, not the reusable workflow that actually contains the publish command. So if `release.yml` called `skills-npm.yml` with `workflow_call`, npm would see `release.yml`, and both packages' trusted-publisher entries would have to be re-pointed at it. That is a one-way door: from that moment, dispatching either workflow on its own could no longer publish, and it would fail with a bare `ENEEDAUTH` after every check and both registry round-trips had already passed. Being able to publish a skill edit without cutting a server release is the reason those packages were split out in the first place.

Dispatching through the Actions API avoids all of it. A dispatched run is its own top-level run, so its OIDC claim still names its own file, both existing entries stay correct and untouched, and both entry points keep working. `workflow_dispatch` is one of the two events that still create a run when triggered with the default `GITHUB_TOKEN`, so this needs no PAT and no new secret, only `actions: write` on the one job.

## On "the same pattern used in the malloydata repository"

Checking that before building it: `malloydata/malloy`'s `release.yaml` has no `workflow_call` and no `uses: ./` anywhere. It is a single inline job that loops over the workspace packages and publishes each one, and its `.github/workflows/CONTEXT.md` records a deliberate decision against splitting publishing across workflow files. It also documents the same npm rule this PR is working around, in as many words: "Matching is on the top-level caller workflow filename, not a reusable workflow it calls."

So the intent behind the ask is right and this PR implements it: one top-level workflow is the single place a release starts, and it publishes everything. The literal mechanism is not what that repo does, and adopting it here would cost a capability that repo never had to give up, because all of its packages are version-locked together and registered against the one file.

## Versioning

This delivers the trigger half of the ask and deliberately not the versioning half, because aligning the numbers is no longer reachable. `skills` is published at 0.1.x and `create-malloy-package` at 0.0.x, against sdk/app/server's 0.0.23x, and npm versions are immutable, so there is no path back to a shared number. Each package therefore publishes at the version its own `package.json` declares. That is also why an unchanged package has to skip cleanly rather than fail: otherwise every server release would go red whenever a package happened not to change. `malloydata/malloy`'s publish loop skips already-published versions for the same reason.

The gate lives in `release.yml` rather than in the children, which keeps this small: the children keep their existing behaviour verbatim, and `create-malloy-package-npm.yml` needs no functional change at all. Its deliberate hard failure on an already-published version is right for a human dispatch and simply never fires on the release path, because the release does not dispatch a version that is already on npm.

## Also in here

- `skills-npm.yml` gains the `github.ref == 'refs/heads/main'` guard its sibling already had, plus an already-published precheck so a forgotten version bump fails in seconds instead of at `npm publish` with `EPUBLISHCONFLICT` several minutes in. Both run before the install, and the version reaches the script through `env:` rather than being interpolated into the shell body of a job that holds `id-token: write`. The same two hardenings are applied to `create-malloy-package-npm.yml`, which had the identical patterns.
- A prerelease release skips both packages and says so. Their own versions carry no hyphen, so their publish workflows would otherwise tag them `latest` from a prerelease run. `gh-release` already skips prereleases for the same reason. The check is inside the script rather than a job-level `if:` deliberately: a skipped job writes no annotation and no summary yet still reports success, which would have made a green release that published nothing invisible.
- The job checks out `main`, never the release branch. `prepare` bumps only sdk/app/server, so a release branch's copies of these two packages are identical to `main`, and both children guard on `main`. Dispatched on a release branch their publish jobs would skip, and a skipped job reports success, so the release would go green having published nothing.
- The two dispatches are serialized, because `create-malloy-package` pins its skills dependency from the monorepo and then asks the registry whether that range resolves. In parallel it loses the race, which is exactly what happened when both were dispatched by hand.
- It waits on the registry rather than on the child run's conclusion, since a child whose publish job was skipped reports success. Asking "is this version on npm" also has no correlation race with a concurrent manual dispatch.
- `publish-packages` is deliberately not in `gh-release`'s `needs`. These packages are independent of the sdk/app/server artifacts, and `prepare` walks the version forward on a re-run rather than resuming, so gating the tag on them would turn a failed package publish into a partial release that cannot be retried at the same version. A failure here still reddens the run.
- A new `.github/workflows/CONTEXT.md`, modelled on the one in `malloydata/malloy`, records the publishing rules that are not visible in the YAML. Publisher had no markdown anywhere describing how npm publishing works, and the one-publisher-per-package rule is the kind of thing that is expensive to rediscover.

## Verification

No workflow was dispatched and no npm setting was touched.

`actionlint` 1.7.12 with `shellcheck` 0.11.0, compared against an `origin/main` baseline: 25 findings on each side, identical sets, zero new. The new shell was also checked directly, with a planted-defect canary to confirm the analysis was actually reaching it.

The `publish-packages` script was extracted from the workflow and run against the live registry with the dispatch stubbed, over eight scenarios: both versions already published gives zero dispatches and a green skip, which is what the next release will do since `main` and npm agree today; dispatched but never appearing fails loudly with a link; dispatched and appearing succeeds and then correctly skips the second package; `main` moving mid-release fails before dispatching anything; a sustained registry outage stops the wait only after three consecutive errors; a single transient blip does not stop it and the publish still completes; alternating errors and 404s never trip the abort, proving the counter resets; and a prerelease run publishes nothing while announcing why.

The fail-open case was checked rather than reasoned about, since it is the worst thing this job could do: a failing `gh api` and an empty-but-successful one both abort before any dispatch. A third case demonstrates the `local x=$(...)` form that would have failed open, which the code avoids by declaring the variable separately.

Two claims that looked right and were not, both caught by checking the source: indenting registry output does not stop the runner parsing a `::` workflow command, because it trims leading whitespace first (hence the `[npm] ` prefix); and `gh api repos/{repo}/commits/main --jq .sha` was confirmed against the public API to return the same value `git rev-parse HEAD` produces.

Moving the precheck ahead of the install puts it after `setup-node`'s `registry-url`, which writes an `.npmrc` containing an auth placeholder, so that was checked rather than assumed: with `NODE_AUTH_TOKEN` unset, a public read returns the version and a missing version returns `E404`, not `ENEEDAUTH`.

Root gate green: typecheck, lint, prettier, 2152 tests passing. The change touches no TypeScript.

## Follow-ups, not in this PR

Binding each trusted-publisher entry to a GitHub Environment restricted to `main` is the control that actually survives someone pushing a branch whose copy of a workflow drops the ref guard, since npm matches repo plus filename and never a ref. It is orthogonal to this change and needs the `environment:` key to land in the workflow before the name is set on npm, never after. It also needs branch restrictions only and no required reviewers, since a pending approval would stall the release's wait. Also worth doing later: making `release.yml` re-runnable at the same version, which `malloydata/malloy` already has and Publisher would benefit from more, and applying the same `env:` treatment to the two remaining interpolations in `create-malloy-package-npm.yml`.


sagarswamirao:
Reviewed. No correctness bugs found, and the core decision -- dispatching the children through the Actions API rather than `workflow_call` -- is right. Approving with nits. One finding below is worth addressing before this lands.

## On the existing release/npm wiring

This follows the established pattern rather than diverging from it. `npm-sdk.yml` and `docker-image.yml` are `workflow_call`-only, invoked from `release.yml` with an explicit `ref`; the two new packages deliberately are not, and the stated reason holds up. I checked it against `malloydata/malloy` directly: its `.github/workflows/CONTEXT.md` states the same npm rule verbatim ("Matching is on the top-level caller workflow filename, not a reusable workflow it calls"), and its root `CONTEXT.md` gives the same reason for keeping build inline. So the "same pattern as the malloydata repository" intent is honored; the literal mechanism differs because that repo is Lerna fixed-mode and version-locked and these two packages are not.

One imprecision in the PR description: that repo's `release.yaml` is three sequential jobs (`pull_and_build`, `precheck`, `npm-release`), not "a single inline job". The publish loop itself is one job. Minor, but the description asserts it as fact.

## package.json fields

I chased down which fields a prior review forced onto sdk/app/server: it was `repository`, and both new packages already have it. Verified against the live registry -- `@malloy-publisher/skills@0.1.1` and `@malloy-publisher/create-malloy-package@0.0.2` both publish with a SLSA v1 provenance attestation and `_npmUser = GitHub Actions`, `trustedPublisher: {id: github}`. Provenance already works with `repository.url` alone.

`homepage`, `bugs`, `author`, and `publishConfig.provenance` are set by none of the five packages -- npm synthesizes homepage/bugs from `repository.url`. Do not add `publishConfig.provenance: true`: it is a no-op under trusted publishing and errors on a local publish. Both new packages are in fact richer than sdk/app/server, which never had `license`, `keywords`, or `files`.

Two optional nits:

- `packages/skills/package.json` has `repository` without a `directory` field, while `packages/create-malloy-package` sets `"directory": "packages/create-malloy-package"`. Adding it to skills would make the two consistent.
- Neither `check-pack.ts` asserts manifest metadata. A future new package missing `repository` would fail at `npm publish` with a 422 rather than in `check_pack` on the PR, which is the wrong side of an irreversible step. One assertion in each script would move it.

## The one substantive finding: the sha guard's window is sized by the first package's publish latency

`EXPECTED_SHA` is read once, before both `publish_pkg` calls. `live_sha` is then re-read per package, but the second package's check runs only after the first package's full dispatch-and-wait, which is budgeted at `POLL_BUDGET_SECONDS=1500`.

So on any release where `skills` actually publishes, `create-malloy-package` is compared against a sha that is up to 25 minutes old. Any unrelated merge to `main` inside that window fails the second package, with an error reading "main moved ... so `<spec>` may not be the version `<wf>` would publish" -- when in the common case nothing under `packages/create-malloy-package` changed at all. `main` moved once during the time I spent reviewing this.

The intent is sound, and the asymmetry is real: `skills` publishes content from the repo-root `skills/` tree, so a commit that leaves its version untouched can still change what ships under it. But applying one job-start sha to a package whose check happens 25 minutes later makes a spurious failure likely on a busy day, and because the failure stops the sequence there is no partial-success path. Two options: re-read the baseline per package, or scope each package's comparison to the paths it actually publishes from.

## Smaller items

**Four failure paths return non-zero without a `$GITHUB_STEP_SUMMARY` line** -- "could not ask npm", the unreadable-sha branch, "main moved", and "npm stopped answering". All four emit a `::error` annotation, so nothing is invisible, but this sits against the script's own rationale for doing the prerelease check inline ("a skipped job writes no summary and no annotation ... completely invisible"), and the "main moved" message explicitly directs the operator to the job summary, which the failing package contributes nothing to. One `>> "$GITHUB_STEP_SUMMARY"` per path closes it.

**A failing `gh workflow run` aborts with no diagnostic.** It is a bare command under `set -e`, so it correctly aborts rather than polling 25 minutes for a run that was never created -- that behavior is right. But it is the only exit from `publish_pkg` with no purpose-built error, so it surfaces as an unexplained step failure with only the preceding `::notice` and `gh`'s stderr to go on.

**The same three rules are now stated in four places each.** The dispatch-not-`workflow_call` rule appears in `CONTEXT.md`, `release.yml`, `skills-npm.yml`, and `create-malloy-package-npm.yml` -- roughly 43 lines across four copies. Same shape for the guardrail-not-a-boundary paragraph (three near-verbatim copies; compare the two children's versions directly) and the `--ignore-scripts` warning (three). The new `publish-packages` job is 119 comment lines to 87 code lines.

The cost is drift, not tokens: four copies of a rule npm can change unilaterally means a future correction lands in one or two of them, and the dispatch rule is both the most-copied and the most likely to need updating. Now that `CONTEXT.md` exists and declares itself the home for the rules that bite, collapsing the YAML copies to a pointer would cut roughly 60 lines without losing a fact. Keep the step-local comments that are genuinely local -- the through-`env` injection note is a good example -- and keep the poll-budget reasoning in `release.yml`, which explains why the budget is sized against the whole child run and why it is a wall-clock deadline rather than an attempt count. That reasoning appears nowhere else and is not recoverable from the code.

**Reviewer attribution and PR numbers in checked-in YAML.** `create-malloy-package-npm.yml` carries a named reviewer and a PR number in a comment; that predates this PR, but the surrounding sentence was rewritten here, so it is in scope. `release.yml` newly adds a review-response narrative -- "This delivers half of what the review asked for ... The ask was to call the new npm workflow from here". The technical content behind it (fixed-mode versioning upstream, versions already on separate lines, npm immutability) is worth keeping; the framing is addressed to a reviewer rather than to the next person reading the file, and git history already carries it.

**Unsourced external claims.** `CONTEXT.md` states the two load-bearing npm rules as bare fact, one of them as "npm's own docs describe this" with no link. Both are corroborated by the malloydata repo's own CONTEXT.md, which also records the concrete command that establishes them and the 409 a second registration returns -- but that corroboration is not cited here. A rule that cannot be checked from this repo and drives a one-way-door design decision should carry a source.

## Verified clean

Worth recording what I checked and found correct, since several of these are easy to get wrong:

- `set -e` genuinely serializes the two calls: a non-zero return from the first aborts before the second is dispatched, so the skills-before-scaffolder ordering invariant is actually enforced, not just documented.
- Each `publish_pkg` call computes a fresh deadline from the then-current `SECONDS`, so the second package is not shortchanged, and `timeout-minutes: 60` correctly covers 2 x 25m plus checkout and setup.
- The `local`-without-command-substitution discipline is applied consistently across all three substitutions, so a failing `node -p` does abort rather than masking. The one `local` with an assignment is a literal string, no substitution.
- The `errors` counter resets on an E404 and only trips on consecutive failures, as the comment claims.
- `shellcheck` at `-S style` on the extracted script: zero findings.
- `RELEASE_NOTES.md` is accurate on both halves of its behavior-change note. `create-malloy-package-npm.yml` genuinely already carried the `refs/heads/main` guard before this PR; only `skills-npm.yml` gains it. And the change is strictly limited to the `workflow_dispatch` path, since `publish` was already gated on `event_name == 'workflow_dispatch'` -- push and pull_request behavior is unchanged.
- `CONTEXT.md` carries no hardcoded version numbers, so the usual staleness trap does not apply. Its claim that `main` does not carry the released sdk/app/server version is independently confirmed: `main` is well behind the registry.
- No em-dashes, smart quotes, or emoji anywhere in the diff; all five changed files are pure ASCII. Note that the pre-existing `gh-release` job uses emoji in its release-note output, which is established style for that output -- the new code correctly did not add more, and should not be "aligned" toward it either way.

sagarswamirao (approved):

mlennie:
@sagarswamirao Thanks, this was a genuinely useful review. Everything is applied except two items I declined and one I deferred, all called out below. Rebased onto current main as a signed merge, so DCO is green again.

## This diff has changed materially since you approved it

Flagging that up front rather than leaving you to find it. Beyond the items you raised, six further defects turned up in my own fix, two of which would publish an unverified version. If any of what follows changes your view, say so and I will hold it rather than merge on the existing approval.

## Your substantive finding, and why I took your second option rather than your first

You were right about the window, and the fix you listed first would not have worked.

`EXPECTED_SHA` comes from `git rev-parse HEAD` on the job's checkout. That checkout is pinned for the life of the job and nothing in `publish-packages` commits or pulls, so re-reading it per package returns the identical sha both times. Moving the line would have looked like a fix and changed nothing. I had already written that version before I noticed, which is the main reason I am spelling it out: it is the kind of change that gets made once and then believed.

So the comparison is now scoped to the paths each package's published content is built from, and those turned out not to be simply where each package lives. skills packs the repo-root `skills/` tree, and its `dist/` is emitted by `tsc` whose version bun resolves from the root lockfile, so it watches `skills/`, `packages/skills/`, `bun.lock` and the root `package.json`. create-malloy-package ships `dist/` and `templates/` from its own directory plus a dependency range its publish job bakes in by reading `packages/skills/package.json` at dispatch time, so that one file is in its list even though no skills content enters its tarball; its bun version is pinned in its own watched workflow file and `bun build --packages external` puts nothing from node_modules in the bundle, so it needs no lockfile entry. Each list also carries its own child workflow file.

Every branch of the comparison fails closed, because the asymmetry is severe: a wrong "proceed" publishes an immutable version whose content nobody verified, while a wrong "stop" costs a re-run.

Two details from checking my own work, both of which would have been invisible in review. The compare request is deliberately not `--paginate`: that endpoint paginates its `commits` array while `files` comes back whole and capped, so paginating repeats the filenames once per commit page, which on a wide enough window inflates the count and trips the cap check, reintroducing the same false failure you were pointing at. And it emits `previous_filename` as well as `filename`, because a rename is reported under its new path only, so moving a file out of a watched directory would otherwise change the tarball unnoticed.

## Six more defects, found after your review, all in my own fix

Two of them fail open, and both contradicted comments I had confidently written saying the opposite.

**The compare is three-dot, so merge-base relative, and nothing asserted that main was actually ahead.** If main is rewritten or rewound during the wait, the file list omits the commits that mattered, or comes back empty, and the guard reads that as "nothing relevant changed" and publishes. It now requires `.status == "ahead"` and aborts on anything else.

**`npm view` can exit 0 with an empty stdout, and I was reading that as "published".** This repo already knows: `check-pack.ts` says in as many words that an empty stdout is a no rather than a yes, and your own pinned-skills check in `create-malloy-package-npm.yml` guards `[ -z "$RESOLVED" ]`. `release.yml` was the one caller that did not. In the poll loop it was the worse of the two: an empty answer read as success would have printed "published", written that to the job summary and returned 0, so a release could end green asserting a publish that never happened.

Also fixed: three bare command substitutions that killed the step with no annotation and no summary line, which is the same silent-failure class you flagged and which I had claimed to have fixed; `node -p` printing the string "undefined" with exit 0 for a missing field, which builds the spec `undefined@0.1.3`, and the package named `undefined` exists on npm so that can resolve as "already published"; and a directory entry never matching the directory's own path. The path list now asserts each entry's kind against the checkout, so dropping a trailing slash fails the release instead of silently disabling a guard.

## The smaller items

All four bare failure paths now write a `$GITHUB_STEP_SUMMARY` line, including the one whose error text tells the operator to go and read the summary. That was the one that bothered me most once you pointed at it.

A failed `gh workflow run` now says what happened and where to dispatch by hand. You were right that aborting there is the correct behaviour; it was just the only exit from that function with no explanation of its own.

`packages/skills` declares `repository.directory`, matching its sibling.

## The three judgment calls

On the duplication, I took your suggestion but not all the way. Each child keeps the imperative locally, since someone editing that file needs to hit "do not convert this to `workflow_call`" without opening another file, and points at `CONTEXT.md` for the reasoning, which is where a correction belongs if npm ever changes the rule. `release.yml` keeps the poll-budget reasoning, per your note that it exists nowhere else and is not recoverable from the code.

On attribution, the reviewer name, the PR number and the paragraph addressed to a reviewer are gone from the YAML. The technical content behind them stays, rewritten for the next person reading the file.

On sourcing, both npm rules in `CONTEXT.md` now cite npm's trusted-publishing docs and the malloydata/malloy `CONTEXT.md` you pointed me at, including the 409 it records for a second registration. I added a line saying these are npm's behaviour rather than ours and can change unilaterally, since everything downstream in that file rests on them.

## Declined and deferred

No `publishConfig.provenance`, `homepage`, `bugs` or `author`, on your own analysis: a no-op under trusted publishing, and synthesized from `repository.url` respectively. Your verification that both packages already publish with a SLSA attestation on `repository.url` alone settled it.

The manifest assertions in the two `check-pack.ts` scripts are deferred rather than declined. You are right that a 422 at publish is the wrong side of an irreversible step, but no current package has the problem, so I would rather it landed as its own change.

Separately, and not fixed here: `create-malloy-package-npm.yml`'s `sed -i` step interpolates a `package.json` version into a sed program, in a job holding `id-token: write`. GNU sed's `e` command executes shell, so a crafted version string is a code-execution path in the publishing job. It predates this branch and needs repo write to reach, so it is not an escalation, but it deserves its own change and I did not want to bury it in this one.

## The description is wrong, and I am leaving it that way

You were right: malloydata's `release.yaml` is three sequential jobs (`pull_and_build`, `precheck`, `npm-release`), not "a single inline job". Only the publish loop is one job.

Correcting it here rather than editing the description, so the record of what was claimed stays intact. Treat this paragraph as the correction.

## What was verified, and what still is not

Typecheck, lint and prettier clean, 2404 tests passing. All three workflows parse, and `shellcheck -S style` on the extracted `publish-packages` script is clean, which I ran because you did.

Beyond that, the step's actual script is extracted from the YAML and executed against stubbed `gh` and `npm` over 22 scenarios, and the `--jq` program is extracted and run against 11 fixtures. Together they assert exit code, which workflows were actually dispatched, that no non-zero exit is silent, and that no summary line claims a publish that did not happen. Reverting each substantive fix in turn breaks one of those suites, so the coverage is not decorative. Three of the fixes initially showed as uncovered, which is the argument for mutation-testing rather than trusting a green suite.

Two gaps I would rather name than paper over. No test in this repo covers these workflow files, and the stubs are mine, so none of this proves that real `gh` and real `npm` behave as stubbed: the first real dispatch is still the first end-to-end proof, and the recovery path if it aborts is the one documented in `CONTEXT.md`. And `gh` evaluates `--jq` with gojq internally, so the "missing files array fails closed" guarantee is measured against stock `jq` and only argued for the real invoker.

PR #938, opened 2026-07-29
fix(mcp): make tool results legible, and return flat rows with a limit signal

jswir:
Every `malloy_*` tool built its MCP content blocks by hand. Four problems, all fixed here by a shared `mcp/tool_response.ts` that all five tools now use.

## 1. `resource.type` was never a real field

The resource block set `type: "application/json"`. The MCP `ResourceContents` schema is:

```ts
z.object({ uri, mimeType?, _meta? }).passthrough()
```

`type` is not in it. `.passthrough()` accepted the field and every client ignored it, so the payload went out untyped. It is now `mimeType`.

## 2. Errors were invisible to a text-only client

Error results carried the `{error, suggestions}` payload **only** inside the embedded resource block. A client that renders text blocks for an `isError` result and nothing else shows the agent no message at all.

Errors now emit a text block alongside the resource. It carries the message *and* the suggestions — the suggestions are the part that makes an error actionable (which source was denied, which view to look for, that a reload is needed), and a client that cannot read the resource has no other way to reach them.

The resource block stays at **index 0**, so every existing reader of `content[0].resource.text` is unaffected. This matches the precedent already in `execute_query_tool.ts`, which appended its render-warning text block second.

## 3. `malloy_compile`'s diagnostics path was the one error left invisible

`malloy_compile` fails in two ways. A thrown error (unknown environment, authorize denial) goes through the catch and is covered by #2. But a **failed compile** — `result.status === "error"`, the ordinary case of a bad Malloy snippet, and by far the most common way the tool fails — returned `isError: true` with a resource block and nothing else. It was exactly the failure mode #2 exists to close, in the tool where it matters most.

An earlier revision of this branch argued that case was already covered, on the grounds that "the diagnostics themselves are the message." They are only the message to a client that parses the resource. Fixed: the error-severity diagnostics are now rendered as text by `formatDiagnosticsText`.

It deliberately does **not** route through `jsonToolError`. That would replace the documented `{status, diagnostics}` payload with `{error, suggestions}` and break every parsing client. The resource block keeps its shape; only a text block is added. `COMPILE_DESCRIPTION` documents it.

## 4. `malloy_getContext` errors carried no remediation

`get_context_tool.ts` built its own error payload from a per-site expression:

```ts
const message = error instanceof Error ? error.message : "Unknown error";
```

Two consequences. Its errors carried **no `suggestions`** where its three sibling tools all do, so an unknown package name arrived as a bare string with nothing to act on. And a non-`Error` throwable produced that literal `"Unknown error"` text.

Now routed through `classifyToolError` / `jsonToolError` like the siblings, which homes each error class to real remediation (not-found for a missing package, back-pressure for a shed load, internal for our own bugs) and stringifies a non-`Error` throwable instead of inventing a message. The `results: []` contract every tier answers with is preserved via `extraPayload` — a parameter that until now had **no caller at all**, despite a doc comment describing precisely this use.

## Scope note: the "Unknown error" report, revisited

Session notes reported `malloy_compile` "crashing instead of returning a diagnostic," returning a bare `{"error": "Unknown error"}` for three inputs. An earlier revision of this PR called that unrelated. **Half of that conclusion was wrong, and the fix is now in this PR.**

What held up: all three inputs were reproduced against a real `environment.compileSource`, and none crashed.

| Input | code | message |
|---|---|---|
| `extend { view: overview is ... }` colliding with an existing view | `definition-name-conflict` | `Cannot redefine 'overview'` |
| `avg(price).round(2)` | `syntax-error` | `something is missing before 'round'` |
| aggregate in a row-level `where:` | `aggregate-in-where` | ``Aggregate expressions are not allowed in `where:`; use `having:` `` |

What did not hold up was the inference drawn from it. That retest established only that `compileSource` returns clean diagnostics; it never traced them through `registerCompileTool`, which then dropped them into a resource block with no text (#3). A clean diagnostic that no client can render is indistinguishable, from the agent's side, from a crash — which is how it got reported as one.

The supporting argument was backwards too. It read: the reported payload had no `suggestions` key, every error this server emits has one, therefore it did not originate here. In fact the compile diagnostics payload is `{status, diagnostics}` and has **never** had a `suggestions` key, and `malloy_getContext`'s errors did not have one either until #4 in this PR. The missing key was evidence *for* this server, not against it.

Mechanism, stated plainly: a text-only client renders an `isError` result, finds no text block, and reports its own placeholder. Both #3 and #4 close it.

## Verification

- **1592** server unit tests pass, 0 fail (1582 before; +10 new)
- **14** MCP integration tests pass, 0 fail — verified identical on a second worktree at the pre-fix commit
- lint clean on `src/**`; typecheck clean apart from two pre-existing errors in `src/service/package.ts`, confirmed identical on an unmodified tree
- grepped every `isError` construction in `src/mcp/`: all now route through `jsonToolError` (which always attaches text) or the compile path (which now does). The string `"Unknown error"` is no longer produced anywhere in MCP source — it survives only in comments explaining the bug

Tests added, by what they pin rather than that they fired:

- `tool_response.spec.ts`: `mimeType` present and `type` gone, resource-at-index-0, `formatErrorText` renders message + suggestions
- `server.protocol.spec.ts`: `expectToolErrorPayload` requires a text block containing the error, and now covers `malloy_getContext` as well as compile and reload. A separate case drives `malloy_compile`'s **diagnostics** path over the real in-memory transport — the layer a `.passthrough()` schema cannot catch a wrong field name at, and the only place the dropped-text-block bug was observable
- `compile_tool.spec.ts`: exact text-block wording, plus a rendering matrix — multiple errors, warnings excluded, missing location, `line 0` (a real position, not a falsy one), character defaulting, and the defensive no-error-severity fallback. Also that a clean compile stays resource-only
- `get_context_tool.spec.ts`: not-found classification for an unknown package, a non-`Error` throwable, and suggestions + text block on both pre-existing error tiers
- `mcp_execute_query_tool.integration.spec.ts`: `resource.mimeType` (6 assertions), and now the text block on the syntax-error case — the busiest tool's only real-transport error coverage, which previously only had the field renamed

## Not fixed here

`tests/integration/mcp/mcp_get_context_semantic.integration.spec.ts` throws `beforeAll() expects a function as the second argument` at line 78 during load, so **the whole file never runs** — it reports as "1 error" beside the 14 passes rather than as a failure. Pre-existing on `main`, unrelated to this change, and left for its own PR. Flagging it because it means the semantic-retrieval integration coverage is currently dead.

## Shape

12 files, +663 / -231. The production change is smaller than that suggests: `execute_query_tool.ts` alone is **-75** (its two duplicated 25-line result blocks collapse into one helper call), `reload_package_tool.ts` and `docs_search_tool.ts` shrink too, and the bulk of the additions are `tool_response.ts` plus tests.


jswir:
Folded PR 3 (the result envelope) into this branch, since it rewrites the same `execute_query_tool.ts` return sites this PR touched. Retitled accordingly — this is no longer a narrow content-block fix.

## ⚠️ Breaking: `malloy_executeQuery`'s default result shape

```jsonc
// before
{"schema":{"fields":[...]},"data":{"array_value":[{"record_value":{"wine_count":{"kind":"number_cell","number_value":150930}}}]}}

// now
{"rows":[{"wine_count":150930}],"row_count":1,"query_row_limit":1000,
 "limit_hit":false,"truncated_for_size":false}
```

`verbose: true` returns the old shape, for callers inspecting render annotations or cell types.

## The correctness half, which is the real reason

`resolveModelQueryRowLimit` pushes `DEFAULT_QUERY_ROW_LIMIT` (**1000**) into the SQL for any query carrying no `limit:` of its own. 1000 is far below `PUBLISHER_MAX_QUERY_ROWS` (100k), so `assertWithinModelResponseLimits` raises nothing:

> `run: wines -> { select: * }` over 150,000 rows returned 1000 rows, with **no indication anything was left behind.**

An agent then reports statistics on a silent 0.7% sample. `Model` now returns the `rowLimit` it applied, and the envelope compares it against the row count:

```
row_count: 1000 | query_row_limit: 1000 | limit_hit: true
warning: Returned exactly 1000 rows, the row limit applied to this query, so there are
         probably more. This is not a complete result: add an explicit limit, aggregate,
         or filter rather than reporting these rows as the whole set.
```

Costs no second query and no `COUNT(*)`. It's a **bound, not a total** — the server can't know the true count because the database applied the cap — and the wording says so.

**Three things could shorten a result and they were indistinguishable:**

| | before | now |
|---|---|---|
| Query row cap (1000, pushed into SQL) | **silent** | `limit_hit` |
| Hard ceiling (100k rows / 50MB) | 413, already loud | unchanged |
| Client per-result size limit | unhandled | `truncated_for_size` + warning |

## The token half

`{"wine_count":{"kind":"number_cell","number_value":150930}}` carries the same information as `{"wine_count":150930}` at several times the size, on every row of every query. `rows` is the existing `compactResult` — already computed by `getQueryResults` and discarded by this tool — so it's also the exact shape `Publisher.query()` hands an in-package data app. Previewing a query over MCP now shows the shape the render code will receive, instead of one that taught `row.avg_price.number_value`.

## Boundaries held

- **REST and `Publisher.query()` untouched.** `compactJson` still returns a bare array; that's a shipped browser contract with SDK consumers.
- **The size cap lives in the MCP layer**, not `query.controller`, because it's tuned to a client's context budget rather than being a property of the query.
- `bigIntReplacer` moves to `src/json_utils.ts`. Not tidying: `compactResult` is raw driver output and DuckDB returns `count()` as a BigInt, so `JSON.stringify` throws on the most common query anyone writes. There's a regression test.
- The two call modes now share one post-run path instead of two copies that had already drifted.

## A test caught a real bug

The first draft appended the warning *after* fitting rows to the size budget — so the truncation meant to bring the payload under the cap left it back over. Fitting now reserves the widest warning the call can produce. That assertion (`serializeEnvelope(e).length <= budget`) is what found it.

## Verification

1604 unit tests, 15 MCP integration tests, lint and typecheck clean.

## Still open

Credible's `execute_query` reshapes Publisher's verbose result itself in Python (`_compact_result_with_meta`). Converging it on this envelope — and deleting that duplicate flattening — is a separate PR in that repo.

mlennie:
Reviewed by running it against a real server over MCP rather than reading the diff, and the engineering here is sound: all four content-block bugs are real, and the silent-truncation bug is worse than the description makes it sound.

I found five things worth fixing and, rather than hand you a list two days before the demo, I have pushed fixes for four of them straight to this branch. **Full disclosure below, with a sha per change and how to drop any of them.** The fifth is the PR description, which only you can edit.

## What I pushed

Four commits on top of `b80e14c5`, all DCO-signed, fast-forward so nothing of yours was rewritten:

| sha | change |
|---|---|
| `9ab68f34` | keep 64-bit integer values exact past 2^53 |
| `d9c08fb0` | do not report a too-large row as an empty result |
| `8136b94c` | only report `_limit_hit` when the silent default capped the query |
| `d6ece839` | declare `queryRowLimit` on `QueryResult` in the OpenAPI spec |

To drop any one: `git revert <sha>`. To drop all four: `git revert d6ece839 8136b94c d9c08fb0 9ab68f34`. Every one is independent and none touches your six commits. Push back on any of it, especially the first, which is the most opinionated.

### `9ab68f34` flat rows lost 64-bit precision, and it was a regression

This is the one I would have flagged as a blocker. `compactResult` is raw driver output, so a DuckDB BIGINT column arrives as a JS BigInt, and `bigIntReplacer` rendered it with `Number()`. Past 2^53 that rounds silently:

```
driver value      : 9007199254740993n      1234567890123456789n
old MCP rows      : 9007199254740992       1234567890123456800
_meta.schema said : {"kind":"number_type","subtype":"bigint"}
```

The payload declared 64-bit and shipped a rounded double. An agent that reads such an id and then filters or joins on it gets a wrong or empty answer with no error anywhere, which is exactly the failure class this batch exists to kill.

It is specifically an MCP regression rather than a pre-existing wart. The `Malloy.Result` envelope this tool used to return carried bigint twice, a lossy `number_value` and an exact `string_value` (`API.util.wrapResult` attaches the latter whenever the subtype is bigint), so the exact digits were on the wire. Flat rows kept only the lossy half.

Two judgement calls, both worth your eye:

- **I fixed the shared replacer in `json_utils.ts`, not a private one for `query_envelope.ts`.** A private one was the narrower change, but it would have broken the thing this PR is justified on: I verified that MCP `rows` is byte-identical to what `Publisher.query()` returns, and fixing only MCP would have made the two diverge precisely on the values that matter. Byte-identity re-verified live afterwards.
- **The conversion is range-conditional, not "always stringify".** `order_items.id` is a BIGINT column, so always-string would turn `id: 100001` into `id: "100001"` for every data app. As written, values inside the safe range are untouched and only values a JSON number cannot represent become strings, which is what Malloy itself does for the same case. The cost is that a column's JSON type varies with magnitude; that is documented in the code and is the part I would most expect you to want to argue about.

### `d9c08fb0` a too-large row was reported as an empty result

`fitToBudget`'s binary search can land on zero rows, and it then said "Showing 0 of N rows; the rest were dropped to fit the result size limit". That reads as an empty result set, so an agent reports "no rows matched" for a row that was simply too big to send. Reachable rather than theoretical: the hard ceiling upstream is `maxBytes` (50MB), so one row with a large text or JSON column sails through `assertWithinModelResponseLimits` and lands here.

Worth flagging one trap I hit fixing it: the new wording is longer than the old one, so swapping it in after the search would have reintroduced the exact overflow your `1a57eafc` ordering exists to prevent. The search now measures against whichever wording is longer. There is also a case dropping rows cannot fix, when `_meta` alone exceeds the limit; the docstring claimed the finished payload always stays under the cap, and now says what actually happens.

### `8136b94c` `_limit_hit` fired on every deliberately-limited query

`resolveModelQueryRowLimit` folds the query's own `limit:`/`top:` into the same `rowLimit`, so equality also held whenever an author limited the query on purpose. Against the bundled storefront example that is three of the eight top-level views. Live, before:

```
top_products (top: 10)            rows=10   cap=10   _limit_hit=True
top_customers (top: 10)           rows=10   cap=10   _limit_hit=True
explicit limit: 10 on 25k rows    rows=10   cap=10   _limit_hit=True
```

That matters because of the Contract rule you wrote alongside it: "check `_limit_hit` before reporting any total, count, or 'top N'. True means the row cap cut the result off". For a `top: 10` view that is backwards, since a top-10 view is the complete answer to "top 10", and the warning then advised adding an explicit limit to a query that already had one.

I took the `_limit_source` route rather than gating the flag, so the signal is kept and the contract rule is what gets fixed. The information was already in hand: `resolveModelQueryRowLimit` is passed the query's own limit, so the call site knows which of the two produced the cap. After:

```
top_products / top_customers / explicit limit:   _limit_hit=False  _limit_source=query
order_items select * (25,356 rows)               _limit_hit=True   _limit_source=server_default
products (200 rows)                              _limit_hit=False
```

**One case I deliberately left alone:** `customers` holds exactly 1000 rows and gets the 1000-row default, so it still reports `_limit_hit: true` for a complete result. Under the default cap the server genuinely cannot tell "exactly 1000 exist" from "1000 returned, more behind", and warning is the safe direction. Flagging it so it is a known limit rather than a surprise.

### `d6ece839` `queryRowLimit` was not in the OpenAPI spec

`query.controller.ts` returns it on `QueryResult` but `api-doc.yaml` never declared it, so it was missing from the generated server, SDK, CLI and Python clients. The pre-existing `as ApiQuery` cast suppressed the excess-property check, so typecheck stayed green and nothing flagged the drift. Since the stated rationale is that Credible cannot compute `_limit_hit` without it, and Credible reaches it through a generated client, it seemed worth closing. `queryRowLimitSource` is declared alongside it for the same reason. Regenerating produces only those fields; the generated outputs are gitignored so only the spec is committed.

## The one left for you: the description

It no longer matches the code, and this is the part I cannot fix. It describes `{rows, row_count, query_row_limit, limit_hit, truncated_for_size}` with `verbose: true` restoring the old shape. At the tip the fields are `_query_row_limit` / `_limit_hit` / `_rows_truncated` / `_total_rows` / `_returned_rows`, there is no `row_count` at all, and `verbose` is gone. I confirmed over the wire: the declared params are `environmentName, packageName, modelPath, query, sourceName, queryName, filterParams, givens`.

The stale text is `340c3aff`'s BREAKING notice, which is the natural thing for a reader to quote, and it did mislead the brief I was reviewing from before I ran it. Worth restating against the tip and saying plainly that there is no opt-out back to the old shape, since that makes this more breaking than it currently reads, not less. `_limit_source` needs a line too.

## What I verified, before and after

All four content-block bugs are real. I booted #942's branch as a baseline since its only change is a description string:

| bug | before | after |
|---|---|---|
| resource typed with a `type` field | `mimeType` absent, `type: application/json` | `mimeType: application/json` |
| errors invisible to a text-only client | no text block | text block with message and suggestions |
| compile diagnostics returned no text | no text block | `Compile failed with 1 error ...` |
| getContext errors had no remediation | `{"error": ..., "results": []}` | classified error plus suggestions |

Two you did not claim but should get credit for: the old getContext error leaked the server's absolute filesystem path, now `Resource not found: examples/...`; and the old code held the literal `error instanceof Error ? error.message : "Unknown error"` that produced the bare "Unknown error" in your friction report.

The correctness bug, on our own example: `run: order_items -> { select: * }` returned 1000 of 25,356 rows, 3.9%, with no signal, in a 1,207,727-character payload. Now capped at 89,943, 13.4x smaller, with the markers and a warning.

The size-budget ordering fix is load-bearing, not decorative: I reimplemented the buggy ordering at a 3,000-character budget and it lands at 3,164, over by 164, against 2,956 for yours.

Gate after my four commits: typecheck and lint exit 0, `prettier:check` clean, 1622 server unit tests pass (6 added), MCP integration spec 10/10.

## Landing

Merging cannot reach Friday's demo: no workflow publishes on push to main, `release.yml` is `workflow_dispatch` only and is the sole caller of `npm-sdk.yml` and `docker-image.yml`. The break is also MCP-only, since `packages/app/src` and `packages/sdk/src` reference neither `malloy_executeQuery` nor `/mcp`, and the REST change is additive.

On ordering with #942: `git merge-tree` finds one conflict, confined to the final paragraph of `COMPILE_DESCRIPTION`, where you append a sentence about the new text block and #942 inserts one about wrapper line offsets. Both are wanted, so the resolution is keep both. #942 is going in first, so this branch picks up that one-paragraph rebase.

mlennie (approved):
Approving. Noting for the record that four of these eleven commits are mine rather than James's: `9ab68f34`, `d9c08fb0`, `8136b94c`, `d6ece839`, pushed to this branch and disclosed in detail above, plus `d28b164a` merging main in to clear the out-of-date gate and resolve the one-paragraph `compile_tool.ts` conflict with #942. James's six are untouched. Happy for any of mine to be reverted independently if he disagrees; `git revert d6ece839 8136b94c d9c08fb0 9ab68f34` drops all four.

The substance is his and it is good: the four content-block bugs are real, and the silent 1000-row truncation was returning 3.9% of a table with no signal at all. Verified by running it against a real server over MCP, before and after, not by reading the diff.

One thing to do at merge rather than now: **the squash commit message needs editing before it lands.** The PR description still describes `verbose: true` and the field names `row_count` / `truncated_for_size`, none of which exist at the tip, and that text is what ends up in `main`'s history permanently. It also needs a line for `_limit_source`. Happy to hand over corrected text.

PR #939, opened 2026-07-29
docs: three Malloy facts that sent agents down wrong paths (incl. JSON support)

jswir:
Three gaps in the shared gotchas skills, each taken from a real session where an agent got stuck, and each **verified by compiling the failing and working forms against a real package** rather than reasoned about.

## 1. Method syntax vs scalar functions (`malloy-gotchas-queries`)

The skill already documents that `.sum()` on a joined path is the *required* form. That makes the omission actively misleading: an agent generalises the rule and writes `avg(price).round(2)`. Compiled results:

| Form | Result |
|---|---|
| `avg(price).round(2)` | ✗ `something is missing before 'round'` |
| `price.avg().round(2)` | ✗ `something is missing before 'round'` |
| `avg(price).floor()` | ✗ `something is missing before 'floor'` |
| `avg_price.round(2)` | ✗ `Cannot call function round(number, number) with source` |
| `price.round(2)` | ✗ `Cannot call function round(number, number) with source` |
| `round(avg(price), 2)` | ✓ |
| `round(avg_price, 2)` | ✓ |
| `price.avg()` | ✓ |

Two distinct rules, both now documented: **nothing chains onto a function call's result**, and **scalar functions have no method form at all**. The parse error names `round` without saying it is unsupported in that position, so it reads like a typo somewhere else, which is what cost the reporting session several isolation round-trips.

## 2. Redefinition through `extend` (`malloy-gotchas-modeling`)

The file covered `Cannot redefine` only for query-based source *columns*, and had no `view:` mention anywhere. So extending a source to replace one of its views reads as legal. It is not:

```malloy
// "Cannot redefine 'overview'" when sales already declares view: overview
source: wines is sales extend { view: overview is { aggregate: record_count } }
```

An extension adds to the parent's namespace; it does not override it.

## 3. JSON is loadable (`malloy-gotchas-modeling`)

An agent concluded only CSV worked and **shelled out to `python3` to read a `.json` file**. That inference was reasonable, because four independent places said so:

- this skill's Excel section enumerated `.csv`/`.parquet`/`.xlsx` and omitted JSON
- `README.md:73` says "CSV, Parquet, or Excel `.xlsx`"
- `docs/packages.md:62` names only `.csv` and `.parquet`
- `create-malloy-package`'s `--data` validator **rejects** `.json`

Verified working: `duckdb.table('x.json')`, `duckdb.table('x.ndjson')`, `read_json_auto` in a SQL source, and a `::number` cast on a string column. This PR fixes the skill; the README, docs, and scaffolder validator are fixed separately so this stays reviewable.

## Sync

Both files are **shared** with `ms2data/agent-skills` (`skills/README.md:16`). The identical text is going up there in a companion PR, verified byte-for-byte, so the next manual `cp` is a no-op rather than a conflict.

`skills_bundle.json` regenerated (`skills_bundle.spec.ts` enforces it); 11 bundle tests pass, no em-dashes.

## Pre-existing drift found while syncing

`malloy-gotchas-modeling`'s entire **Excel Files** section exists in Publisher and not upstream. Untouched here since it is not mine to move, and it carries a Publisher-specific parenthetical, but it should be reconciled.


jswir:
Added a second commit after review feedback: **a gotchas entry is only read after something breaks, and the reported agent never broke anything.** It read the format lists, concluded JSON was unsupported, and reached for `python3` without ever trying. So the fact also has to appear where formats are stated affirmatively.

Two places describe what a *model* can read, and both omitted JSON:

- **`skills/malloy-modeling/SKILL.md:27`** (pre-flight checklist, read before writing any code). Item 9 was `.xlsx`-specific; it now covers `.csv`, `.parquet`, `.json`, `.ndjson`, `.xlsx`, and explicitly says not to inspect a file with python or jq instead of querying it.
- **`docs/packages.md`** "Where the data comes from", which named only `.csv` and `.parquet`.

**`README.md`'s `--data` line is deliberately unchanged.** The scaffolder's validator (`scaffold.ts:2021-2035`) still rejects `.json`, so documenting it as accepted would make the README wrong. That line moves together with the validator fix.

`malloy-modeling` is Publisher-local (`skills/README.md:17`), so no upstream counterpart is needed and ms2data/agent-skills#72 is unaffected.

mlennie (approved):

PR #940, opened 2026-07-29
feat(scaffolder): accept JSON for --data, and name the files it skipped

jswir:
Two ways the scaffolder taught agents something false about the data it can model. Both come from one session, where an agent pointed at a folder of three wine datasets, got one, and concluded JSON was unsupported.

## 1. `--data` rejected `.json`

DuckDB reads JSON in place exactly like CSV, and the generated `duckdb.table('<path>')` template needs **no change at all** to handle it, so the refusal was arbitrary. It was also the loudest of four signals telling agents only CSV worked, which is how one ended up shelling out to `python3` to read a `.json` file rather than querying it.

`.json` and `.ndjson` now scaffold and produce a working model. Verified by running the scaffolder for real:

```
$ create-malloy-package wine --data ./reviews.json
✓ Created package wine in ./wine

source: wine is duckdb.table('data/reviews.json') extend {
  measure: record_count is count()
  ...
```

## 2. `--data` silently ignored everything else in the directory

It takes exactly one file and never said so at the point it mattered, so pointing it at an export folder modelled one and left the omission invisible. Reproducing the original scenario:

```
$ create-malloy-package wine --data ./exports/winemag-130k.csv
✓ Created package wine in ./wine
  Not included: winemag-130k.json, winemag-150k.csv. --data takes one file;
  copy any others into wine/data/ and add a source for each.
```

This does not change what gets scaffolded, only what the run tells you. Capped at five names so a large directory does not flood the output.

## Notes

- The accepted extensions are now **one list** (`LOADABLE_DATA_EXTENSIONS`) used both to validate `--data` and to find siblings, rather than a literal repeated per call site.
- **`README.md`'s `--data` line moves here, with the validator**, rather than in #939's docs commit. Documenting JSON as accepted while the CLI still refused it would have been wrong in the other direction.
- `findSiblingDataFiles` swallows an unreadable directory rather than failing an otherwise successful scaffold, and runs filenames through the existing `printable()` so a crafted name cannot inject escapes into the terminal.

## Tests

327 pass, lint and typecheck clean. New coverage: `.json` and `.ndjson` scaffold to a correct model path; the rejection message lists the new extensions; siblings are sorted, exclude the chosen file, and ignore non-loadable ones; no `siblingDataFiles` key when the file is alone; the output caps at five with "and N more"; and no line at all when there were no siblings.

## Follow-up, not in this PR

The same session's bigger onboarding problem is still open: "Next steps" prints `npm start` with no indication it is a long-running process, then two `curl` checks below it that return nothing until it is running. That is the next PR, along with the MCP/skills registration remedies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,8 @@ against your package, in watch mode, so edits to the model take effect as you sa
 
 Run bare like that, the package comes with a small sample dataset, so there is something to query
 before you have wired up anything of your own. To start from one of your own files instead, pass
-`--data` (CSV, Parquet, or Excel `.xlsx`):
+`--data` (CSV, Parquet, JSON, newline-delimited JSON, or Excel `.xlsx` - DuckDB reads all of them
+in place, so nothing needs converting first):
 
 ```bash
 npm create @malloy-publisher/malloy-package sales -- --data ./orders.csv
--- a/packages/create-malloy-package/README.md
+++ b/packages/create-malloy-package/README.md
@@ -305,12 +305,12 @@ npx @malloy-publisher/create-malloy-package sales --data mydata.csv
 - `name` (positional): the package name. Omit it to only set up the agent workspace in
   the current directory (write the MCP connection, agent instructions, and skills)
   without scaffolding a package.
-- `--data <file>`: seed the package from your own CSV, Parquet, JSON, NDJSON, or XLSX file instead of
-  the built-in sample. The file is copied into the package and the starter model points
-  at it. DuckDB reads all three formats in place; an Excel file is read as its first
-  sheet. It seeds a new package, so it requires a package name: it cannot be combined
-  with the setup-only mode above, and passing it without a name is an error rather than
-  a silently ignored flag.
+- `--data <file>`: seed the package from your own CSV, Parquet, JSON, NDJSON, or XLSX
+  file instead of the built-in sample. The file is copied into the package and the
+  starter model points at it. DuckDB reads all of them in place, so nothing needs
+  converting first; an Excel file is read as its first sheet. It seeds a new package,
+  so it requires a package name: it cannot be combined with the setup-only mode above,
+  and passing it without a name is an error rather than a silently ignored flag.
 - `--client <claude-code|cursor>`: which agent client to wire up. Defaults to
   `claude-code`. `AGENTS.md` and the skills in `.claude/skills/` are written for every
   client; the MCP config file (`.mcp.json` for Claude Code, `.cursor/mcp.json` for
--- a/packages/create-malloy-package/README.md
+++ b/packages/create-malloy-package/README.md
@@ -305,7 +305,7 @@ npx @malloy-publisher/create-malloy-package sales --data mydata.csv
 - `name` (positional): the package name. Omit it to only set up the agent workspace in
   the current directory (write the MCP connection, agent instructions, and skills)
   without scaffolding a package.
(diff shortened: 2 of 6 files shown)


mlennie (approved):
Approving. I checked the headline claim by running it rather than trusting the extension list, since `duckdb.table()` depends on DuckDB's replacement scan rather than on the scaffolder: both `.json` and `.ndjson` compile and execute in place through Publisher's own stack (`@malloydata/db-duckdb`, `@duckdb/node-api` 1.5.3-r.2), not just the CLI. Also worth noting for #939, a JSON-stored number comes back `VARCHAR`, which is exactly why the `::number` cast note over there is right.

The sibling-naming code holds up under a look for escaping holes. Exclusion compares raw basenames before `printable()` is applied, which is the correct order (escaping first would break the match for any name containing a control character), sorting is on raw names with display-escaping last, and `siblingDataFiles` has exactly one consumer. I tried to construct a filename that reaches stdout unescaped and could not.

Three things happened to this branch that aren't yours, all called out here so none of them are silent:

- `2864082c` wraps `src/scaffold.spec.ts:379`, which was 84 columns and failing `lint_format`. I confirmed the failure was real on your original tip rather than a no-op fix.
- `e9b22cb7` fixes `packages/create-malloy-package/README.md`, which still said DuckDB reads "all three formats" in place directly under a bullet this PR had just grown to five. It now matches the root README wording your own diff introduced.
- I rebased the branch with `--signoff` to clear DCO, so your commit is now `eff08a94` and shows me as committer. Author is still you, and the tree is byte-identical to what you pushed.

Revert any of them if you'd rather own them.

On DCO more broadly, since it was blocking the whole batch rather than just this PR: none of the commits across all seven branches carried a `Signed-off-by`, so I signed all fifteen and pushed. All seven are green on it now and nothing is needed from you. The tips moved, so reset rather than pull if you have any of them checked out:

```
git fetch origin && git checkout <branch> && git reset --hard origin/<branch>
```

Worth knowing why that worked, because I assumed it wouldn't: `--signoff` signs with the committer identity, not the author's, so these come out authored by you and signed by me. I expected the DCO app to reject that as a mismatch and it doesn't, which means a rebase can unblock someone else's batch. I confirmed it on one branch before touching the rest.

Two nits, take or leave. `findSiblingDataFiles` calls `readdirSync` without `withFileTypes`, so a directory named something like `exports.csv` gets reported under "Not included:" and the advice to copy it into `data/` doesn't apply to it. And `ScaffoldResult.siblingDataFiles` holds `printable()`-escaped names, so an exported interface is display-shaped rather than real filenames; harmless for the single consumer, but a future programmatic caller would reasonably expect to `path.join()` them, so it's worth a word in the doc comment.

One heads-up that isn't about this PR: #940 and #941 conflict with each other, despite the batch being described as independent. Both add a test at the same anchor in `packages/create-malloy-package/src/index.spec.ts`. It's a keep-both resolution, but whoever merges second will hit it.

Last thing, and it's on me rather than you: I have an unpushed branch that touches five of the same files (it pins `@latest` on the scaffolder command and stops the `.xlsx` starter model reading one garbage row). I retrialled the merge against your current tip and it comes out at two conflicts, both in `scaffold.ts` and its spec, both additive, with 351 tests passing afterwards. Yours should land first and I'll take the rebase. Worth knowing that my `isSpreadsheet()` ends up narrower than your `LOADABLE_DATA_EXTENSIONS` rather than duplicating it, which is a better shape than either branch had alone.

PR #941, opened 2026-07-29
fix(onboarding): name the two failures that leave an agent with no tools

jswir:
Both of these come from one reported session, and both were invisible when they happened.

## 1. `npm start` holds the terminal, and nothing said so

Read top to bottom, "Next steps" printed a boot command that looks like a step you move past, then two `curl` checks that look like the next commands to run:

```
Next steps:
  npm start   start Publisher (web UI + MCP) with the package in watch mode
  http://localhost:4000   explore in the browser

Check it is ready:
  curl -s http://localhost:4000/api/v0/status
```

Running only the curls is exactly what happened. They answered **nothing** — `curl -s` is silent on a refused connection — so the failure had no text to read at all. Both facts are now stated above the checks they invalidate:

```
Check it is ready:
  In a second terminal: the boot command above keeps running until you
  stop it. Until the server is up these print nothing at all, not an
  error, because -s silences curl's connection failure.
  curl -s http://localhost:4000/api/v0/status
```

## 2. The MCP config and skills need the session to have *started* here

A project's `.mcp.json` and `.claude/skills/` are only discovered by an agent session that started in that directory. The reported session was rooted one directory **up**, so it saw neither, and nothing anywhere reports that. It also looks identical to the already-documented "I started the server myself" case while needing a different fix, which is why it took three rounds to diagnose.

**The two symptoms do not share a fix with each other either**, and that is the part this PR is careful about:

| Symptom | Cause | Fix |
| --- | --- | --- |
| `malloy` listed in `/mcp` but disconnected | client connected before the server existed | Reconnect |
| `malloy` not listed at all | session started outside the directory | relaunch from here, **or** `-s user` |
| tools present, no skills auto-invoked | same cause | relaunch from here only |

`claude mcp add ... -s user` fixes the second and **not** the third: the MCP server list is fixed at session boot so a user-scoped registration frees it from the directory, while skills are rescanned as the working directory changes and just need the session rooted here. The output hands over that command and says so, rather than implying it clears both:

```
Connect an agent:
  MCP endpoint http://localhost:4040/mcp
  .mcp.json points at it, so Claude Code offers it.
  An agent that starts the server itself can't reconnect MCP in that
  session; ask the user to reconnect it. See AGENTS.md.
  Both .mcp.json and .claude/skills are only picked up by an
  agent session that STARTED in this directory. Launch your agent from
  here. If you cannot, register the server so the directory stops
  mattering (skills still need the session rooted here):
    claude mcp add --transport http malloy http://localhost:4040/mcp -s user
```

## Where it lands

The four places an agent or a user actually reads:

- **the scaffolder's printed output** (`src/index.ts`)
- **the AGENTS.md it generates** (`templates/AGENTS.md`, using the existing `{{mcpPort}}` variable; verified it renders rather than emitting the literal)
- **this repo's own `AGENTS.md`**, as an explicit third case with the symptom-to-fix table above
- **`README.md`**, whose `claude mcp add` line omitted `-s user` entirely and so recommended precisely the project-scoped registration that fails this way

## Tests

325 pass, lint and typecheck clean. Five new assertions on real rendered output: the terminal warning exists **and precedes** the commands it is about (a warning printed after them is not a warning), the silent-failure note, the started-here warning naming both `.mcp.json` and `.claude/skills`, the `-s user` command rendered at a non-default port, and that the output does not claim the registration also fixes skills.

Verified by running the scaffolder for real, not only through the assertions.


mlennie:
This is a real fix for the thing that cost you the most time, and the symptom-to-fix table is the part I'd most want to keep. One correction before it lands, because the PR contradicts itself on the claim everything else hangs off.

**The MCP half is accurate.** `~/.claude.json` matches your description exactly: project entries carry `mcpServers` (local scope) plus `enabledMcpjsonServers`/`disabledMcpjsonServers` for the `.mcp.json` trust decisions, and user scope lives at the top level. Default `claude mcp add` scope is local, so it is tied to the directory you ran it in, and `-s user` does free it. The README paragraph and row 2 of the table are right, and adding `-s user` to the quickstart is a good catch.

**The skills half says two incompatible things.** `AGENTS.md:52` has:

> `.claude/skills/` [is] only discovered from the directory the agent session started in. A session launched from a parent directory sees neither.

while the table's third row and `templates/AGENTS.md:177-178` have:

> skills are rescanned as the working directory changes

Those cannot both be true, and the second one is the correct one. I have it first-hand from a session yesterday: it started in the workspace root, and the moment work moved into a subdirectory it picked up all 30 skills with no restart, announced by a "New skills discovered" notice.

Your own `NOTES.md` says the same thing, and says it better than the PR does:

```
the moment the shell's cwd changed and the next tool call happened, the harness
rescanned and picked up all 30 malloy-* skills automatically, no /exit/relaunch
required. So MCP and skills have different discovery mechanics [...] This
asymmetry itself is worth knowing.
```

So the prose has drifted from your own source report. The fix is to bring `AGENTS.md:52` in line with the table rather than the other way round, since the table is what a reader acts on. It also makes the PR's argument stronger, not weaker: the two symptoms looking identical while needing different fixes is exactly why it took three rounds to diagnose, and that story only works if the fixes really are different.

Worth being precise about which of the three the `-s user` command actually solves, too. It frees MCP from the directory; it does nothing for skills, and the current text is careful about that in the scaffolder output but looser in AGENTS.md.

Nothing in the diff is failing: prettier is clean over the changed tree, all five new tests pass, and both typecheck risks are fine (`mcpConfigPath` and `mcpPort` are already on `ScaffoldResult`, and `scaffold.ts:1768` already passes `mcpPort`, so the new `{{mcpPort}}` renders). The red check is **DCO**, and it is not just this PR: every commit across all seven branches in the batch is missing a `Signed-off-by` trailer, so all seven are blocked on the same thing. One fix per branch:

```
git rebase --signoff origin/main && git push --force-with-lease
```

Safe to run on `scaffolder-json-and-siblings` too, which now carries two signed commits of mine on top of yours: I checked, and `--signoff` leaves an already-signed commit with one trailer rather than doubling it.

The five new assertions are well aimed, particularly the one checking the warning precedes the commands it is about. That is the kind of thing that rots silently otherwise.

Two heads-ups. This conflicts with #940: you both add a test at the same anchor in `index.spec.ts`, keep-both resolution, but it will surprise whoever merges second. And I have an unpushed branch touching four of the same files, which merges against yours with zero conflicts, so nothing needed from you there.

mlennie (approved):
Approving, with the correction from my earlier comment pushed as `1b755532` rather than left for you. It is a fast-forward, so your commit is untouched.

The contradiction turned out to be in three places, not one, and the third is the one I would not have found by reading:

- `AGENTS.md:52` had `.mcp.json` and `.claude/skills/` both discovered only from the session's start directory, against the table eight lines below.
- `templates/AGENTS.md` carried both halves in a single paragraph: skills "are rescanned as the working directory changes, so they need the session actually rooted here". Those two clauses cancel each other out.
- The printed success output in `index.ts` said "Both `<mcpConfig>` and `.claude/skills` are only picked up by an agent session that STARTED in this directory", which is the version a user actually sees.

All three now say the same thing: `.mcp.json` is fixed to the start directory, `.claude/skills` is rescanned as the working directory changes and needs nothing.

The reason I pushed this rather than flagging it is the fourth site. `test("does not claim the registration also fixes skills")` asserted the output contains `"skills still need the session rooted here"`. The intent is right, and it is the assertion I would have wanted too: do not overclaim what `-s user` fixes. But it pins a statement that is not true, so correcting the docs later would have required deleting a test, and a test enforcing a wrong claim reads as coverage from every angle. It now asserts the true mechanism and `not.toContain`s the old phrasing, so the claim cannot drift back silently.

Your `NOTES.md` is what settled it, and it is sharper than the PR text:

```
the moment the shell's cwd changed and the next tool call happened, the harness
rescanned and picked up all 30 malloy-* skills automatically, no /exit/relaunch
required. So MCP and skills have different discovery mechanics [...] This
asymmetry itself is worth knowing.
```

I saw the same thing independently during a run yesterday: session started at the workspace root, work moved into a subdirectory, all 30 skills picked up with a "New skills discovered" notice and no restart.

This makes the PR's own argument stronger rather than weaker. Two symptoms that look identical needing two different fixes is exactly why it took three rounds to diagnose, and that only holds if the fixes really do differ. Revert the commit if you would rather restate it yourself; you have the original session context and I would not argue with your wording.

Suite is 331 pass on the package. The one failure is `MALLOY_RESERVED drift against the installed @malloydata/malloy`, which fails identically on `origin/main` under the same borrowed `node_modules`, so it is environmental rather than anything here.

The five new assertions are well aimed, particularly the one checking the warning precedes the commands it is about. That is the kind of thing that rots silently otherwise.

PR #942, opened 2026-07-29
docs(mcp): tell malloy_compile callers how to check part of a source

jswir:
`malloy_compile` **appends** the submitted source to the model, so it has to stand on its own as top-level Malloy. The two most natural ways to validate an edit both fail, and both messages name the symptom rather than the cause:

```malloy
view: by_pts is { group_by: points }   ->  no viable alternative at input 'view:'
source: sales is base extend { ... }   ->  Cannot redefine 'sales'
```

Nothing in the tool description or the docs said so, or said what to do instead. Two forms work:

```malloy
query: check is sales -> { group_by: points }              // no wrapper needed
source: check is sales extend { measure: m is max(price) } // fragment in its real namespace
```

Both compile against the real source, so inherited measures resolve and `private:` fields stay hidden exactly as they do in place.

## This started as a parameter, and shouldn't have been one

The first draft added an `extendSource` param that generated the wrapper. Testing what an agent can already do **today, with no server change**, killed it:

| Approach | Result |
|---|---|
| Resubmit the whole edited source | ❌ `Cannot redefine 'sales'` |
| Write the `extend` wrapper by hand | ✅ works |
| Copy the source under a new name | ✅ works |
| Top-level `query:` instead of a `view:` | ✅ works, **no wrapper at all** |

The parameter added **zero capability** — it emitted exactly the string in row 2, which an agent can type. It would have been permanent public surface duplicating a technique it doesn't even do best: the top-level `query:` form is simpler and needs no wrapper. The real gap was that nothing taught any of this.

So the production diff is **7 lines**: a "Checking part of a source" section in `COMPILE_DESCRIPTION`, mirrored in `docs/ai-agents.md`.

## The test drives the real compiler

`compile_fragment_techniques.spec.ts` runs against a real `Environment` and package, not a mock, because only Malloy decides what compiles and a hand-written checker would drift from it and could pass on the exact case it exists to catch. Advice that lives only in prose rots silently; this is what stops it. It pins:

- both documented failures, by their real messages
- both working forms
- that inherited measures resolve through each form
- that a `private:` field stays hidden in the wrapper **exactly as it does in an in-place extension** (the fidelity claim the description makes)
- the redefinition caveat, plus the rename workaround

## Verification

1581 server unit tests pass, lint and typecheck clean (typecheck has two pre-existing errors in `src/service/package.ts`, confirmed identical on a clean tree).

## Note for the reviewer

This touches `compile_tool.ts`'s description constant only, so it should not conflict with #938, which changes that file's return statements.

### DIFF
--- a/docs/ai-agents.md
+++ b/docs/ai-agents.md
@@ -26,6 +26,20 @@ The server listens at `http://localhost:4040/mcp` (set the port with `--mcp_port
 ### Authoring tools
 
 - `malloy_compile`: compile Malloy source against a model and return structured diagnostics (`severity`, `message`, `line`, `character`) without running a query, so an agent can validate a change while authoring instead of firing a throwaway query. Positions are 0-based and relative to the model file with the submitted source appended to it.
+
+  Because the source is *appended*, it has to stand on its own as top-level Malloy, and the two most natural ways to check an edit both fail for reasons that have nothing to do with the edit. A bare `view:` / `dimension:` / `measure:` is not a top-level statement (`no viable alternative at input 'view:'`), and resubmitting the source you are editing collides with the model's own copy (`Cannot redefine '<name>'`). To check part of a source, either send the view body as a top-level query:
+
+  ```malloy
+  query: check is orders -> { group_by: status, aggregate: revenue }
+  ```
+
+  or wrap the fragment in a throwaway extension, which puts it in the namespace the real edit will live in:
+
+  ```malloy
+  source: check is orders extend { measure: aov is revenue / nullif(order_count, 0) }
+  ```
+
+  Both compile against the real source, so inherited measures resolve and `private:` fields stay hidden exactly as they would in place. An extension *adds* to a source's namespace rather than overriding it, so a fragment reusing an existing view name reports `Cannot redefine` too; rename it for the check.
 - `malloy_reloadPackage`: recompile a package from its on-disk content so a source or view added after boot becomes queryable by name, without restarting the server. This is the other half of the authoring loop: validate with `malloy_compile`, save, reload, then query. A reload that fails to compile leaves the package's files alone and keeps serving the previously compiled model, returning the compile errors.
 
 ### Skills as MCP prompts
--- a/packages/server/src/mcp/tools/compile_tool.ts
+++ b/packages/server/src/mcp/tools/compile_tool.ts
@@ -51,8 +51,13 @@ const COMPILE_DESCRIPTION = `Compile-check Malloy source against a model and ret
 - source (required): the Malloy text to validate.
 - includeSql (optional): also return the generated SQL when the source ends in a runnable query. The query is still not executed and no data is scanned.
 
+## Checking part of a source
+Your source is APPENDED to the model, so it has to stand on its own as top-level Malloy. Two things follow, and both have error messages that name the symptom rather than the cause:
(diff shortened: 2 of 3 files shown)


mlennie:
Approve with one addition. I ran all four documented forms through `malloy_compile` over MCP against `examples/storefront` rather than reading them, and every claim holds exactly as written. A second pass then found one interaction worth a sentence in the new section, which is the only thing keeping this from a plain approve.

Both documented failures fail, with the messages the description promises:

| form | result |
|---|---|
| bare `view:` | `no viable alternative at input 'view:'` |
| bare `dimension:` | `no viable alternative at input 'dimension:'` |
| bare `measure:` | `no viable alternative at input 'measure:'` |
| resubmitting the source being edited | `Cannot redefine 'order_items'` |

Both documented working forms work, including the two fidelity claims. `query: check is order_items -> { aggregate: order_count }` and `source: check is order_items extend { measure: aov is total_sales / nullif(order_count, 0) }` both compile clean, so inherited measures do resolve through the wrapper. The caveat holds too: `extend { view: by_category is ... }` reports `Cannot redefine 'by_category'`, and renaming to `by_category__check` compiles.

`compile_fragment_techniques.spec.ts` passes 8/8. Pinning this against the real compiler through `env.compileSource` rather than a hand-written checker is the right call, given nothing else enforces the advice.

**The one thing I would add: `includeSql: true` breaks both of the forms you recommend.** The tool description documents `includeSql` a few lines above the new section, and combining it with either recommended form produces an error-severity diagnostic that blames Malloy:

```
query: check is order_items -> { aggregate: order_count }        + includeSql
  -> error: Internal compiler error (likely a Malloy bug - please file an issue): Model has no queries.
source: check is order_items extend { measure: m2 is count() }   + includeSql
  -> error: Internal compiler error (likely a Malloy bug - please file an issue): Model has no queries.
run: order_items -> { aggregate: order_count }                   + includeSql
  -> success, sql returned
```

That makes sense once you see it (neither form leaves a runnable query at the end of the model, which is exactly what `includeSql` documents itself as needing), but an agent reading this section will not predict it, and the message actively tells them to go file a bug against Malloy. One clause in the new section saying `includeSql` only applies to the `run:` form would close it. This is pre-existing behavior rather than something the PR introduces, but the PR is what will send people into it.

One nit, take it or leave it: line 70 asserts only `expect(errors.join(" ")).toContain("view:")` for the bare-`view:` case. The description promises the specific `no viable alternative at input 'view:'`, and the compiler really does produce it, so the assertion could pin that instead. As written it would still pass if the message degraded to something much less useful that happened to mention `view:`.

One thing worth flagging on the description's "should not conflict with #938" note: it does conflict, though trivially. `git merge-tree` against `origin/mcp-error-legibility` reports one conflict, in `packages/server/src/mcp/tools/compile_tool.ts`, and it is confined to the final paragraph of `COMPILE_DESCRIPTION` under `## Response`. You add "Any wrapper you add counts toward that offset too." mid-paragraph; #938 appends a sentence about the new text block to the end of the same paragraph. The two are about different things and both are wanted, so the resolution is just to keep both sentences. Your larger `## Checking part of a source` section merges cleanly on its own.

My suggestion is that this one lands first. It is three files, changes no runtime behavior, and is fully verified, whereas #938 is carrying review findings. Whoever lands second resolves the same two-sentence paragraph either way.

mlennie (approved):

PR #946, opened 2026-07-29
fix(create-malloy-package): pin @latest, and stop a spreadsheet scaffolding to a silently wrong model

mlennie:
Two failures a user cannot see, both found by running the scaffold-to-agent journey end to end against the published packages rather than by checking versions.

## The command in our own docs hands people an old scaffolder

`npm create @malloy-publisher/malloy-package <name>` resolves an unversioned name through npm's npx cache, and any cached copy satisfies it, so on a machine that has run the command before npm never asks the registry. The user silently gets whatever version they first installed. Today that is a scaffolder pinning an older server and an older copy of the skills, and the scaffolder reports success either way.

This is not theoretical. Two fresh directories, minutes apart, on the same registry: the bare command produced a scaffolder that pinned server 0.0.231 and installed 29 skills, while `@latest` produced 0.0.2, server 0.0.232 and 30 skills. It went unnoticed because every end-to-end check we have ever run used `@latest` or an isolated cache, which is exactly the command that works.

So `@latest` is now pinned everywhere the command is written down, including the four strings the tool itself prints, and each place says in one sentence why the version is there, because a bare `@latest` reads as noise and gets tidied away. That fixes the command as written but not the command as remembered, and not the next release, so the scaffolder also checks its own version against the registry once it has finished writing and says something when it is behind.

That check is the first network call this package has ever made, so it is built to be unnoticeable when it fails: one unauthenticated GET of a public document, sending nothing about the user or their files, bounded and fail-open on every path, and `CREATE_MALLOY_PACKAGE_NO_UPDATE_CHECK=1` skips it entirely. Both READMEs document it. Note it deliberately uses `node:https` rather than `fetch`: `AbortSignal.timeout` bounds the promise and not the socket, so against an address that drops packets the process sat for nine seconds after all output with nothing left to do. Measured, and recorded in the code so it does not get simplified back.

## A real spreadsheet scaffolds to a model that reads one garbage row

A `.xlsx` is only a table when its header is the first row, and a report export usually has a title or a banner above it. DuckDB then takes that row as the column names and stops at the first blank line, so a workbook with thousands of rows loads as a handful, or as one, and nothing reports a problem: the package loads, `load_errors` is 0, and the query returns 200 with the wrong number. On the test export used here the count is 1 where the truth is 1,500.

The scaffolder never opens the data file, and giving it a spreadsheet parser to find out is a much larger change, so this does not detect the case. It removes the silence. A `.xlsx` now gets its own starter model whose comment states the failure, tells the reader to run `overview` and compare the row count, and carries a two-step remedy: a probe that shows the sheet as it really is, then the `read_xlsx` form with the sheet, the range and a filter for the decoration rows. The success output repeats the check and points at the model file. A `.csv` is untouched, because there the read either parses or fails loudly and the same warning would be noise.

## Also

The generated `AGENTS.md` told the reader that the ports and watch mode all come from flags in the generated start command. That command carries only `--watch-env`; 4000 and 4040 are Publisher's own defaults and are never named in it. An agent following the instruction went looking for flags that were never written.

## How this was verified

Everything about the spreadsheet advice was run against a live server rather than reasoned about, which is the only reason it is correct. Four rounds of review found the advice itself wrong in ways that read perfectly well, and every one of them was found by running it, not reading it:

- The first remedy returned 99,995 rows for a 1,500 row sheet, because an explicit range reads every row in it including the blank spacers, subtotals, total and footnotes. It needs a filter as well as a range.
- `stop_at_empty` looks like the obvious repair and truncates 220 rows into the data.
- The probe step could not be run as written at all: Publisher refuses raw SQL in an ad-hoc query, so it had to become a temporary model source.
- The cast example did not parse. `x::number.sum()` is a syntax error, and the correct `sum(x::number)` then throws at query time on a real sheet's `N/A`, so the conversion belongs in the SQL layer with `try_cast`.
- The remedy told you to re-run the check without telling you to reload first, or to confirm the reload compiled. A failed reload leaves the previous model serving, so the query answers 200 with the old number and the edit looks inert. Observed live: reload 424, next query 200 with the stale count.
- The sheet-discovery trick named the wrong sheet on the very workbook it was written for, because DuckDB's suggestion is chosen by spelling. There is no sheet-listing function at all, so that paragraph now says to take the name off the tab in Excel.

The version check needed the same treatment. `AbortSignal.timeout` bounds the promise and not the socket, so the first version left the process sitting for nine seconds after all output on a packet-dropping network. Moving to `node:https` fixed that and introduced a worse one, because a socket timeout is an inactivity timer that a trickling server resets forever, leaving the promise unsettled indefinitely. It now carries both a socket timeout and a total deadline, plus a cancel the failure path uses so an errored run does not sit waiting for an answer nobody will read. Measured: blackhole settle 1503ms and exit 1504ms, trickle bounded, failure path 0.24s to 0.05s.

Each of those is asserted in the tests with the measured numbers in the comments, because the failure mode is a later edit quietly dropping one. Final state on the test export: naive read 1 row, remedy 1,500 rows, matching independently computed truth.

The full gate passes: typecheck, lint, prettier, and the suites (server 1573 + 232, skills 7, create-malloy-package 346, hammer 20). That package gained 19 tests. Note the suite runs under `bun` while the shipped bin runs under `node`, so every claim about the network path was also measured by hand under node.

## Overlap with #940, now merged

#940 merged while this was in flight, so this is rebased onto current `main` rather than routed around it. Two conflicts, in `scaffold.ts` and its spec, both places where we had each added something at the same anchor. Both sides kept.

One thing worth a look in the resolution: `isSpreadsheet()` is deliberately narrower than #940's `LOADABLE_DATA_EXTENSIONS` rather than derived from it. Every extension on that list loads in place, but only `.xlsx` loads in place and can still be silently wrong, so only `.xlsx` earns the extra warning. The doc comment on `isSpreadsheet` said the validator rejected everything but csv/parquet/xlsx, which #940 made false, so it now states the narrowing and why instead.

## Follow-ups, not in this PR

- The scaffolder still cannot detect a bad spreadsheet, only warn about the possibility. Detecting it means reading the file, which pairs naturally with the deferred follow-up to emit real dimensions and measures from the inferred column types instead of a bare `count()`: one header probe would serve both.
- `malloy-gotchas-modeling` documents six xlsx quirks but not that a default read stops at the first blank row, which is the mechanism this whole change is about. Worth adding to that skill.
- `tests/harness/mcp_test_setup.ts` computes a fixed E2E port from `MCP_PORT || 4040`, which collides with anything already on 4042/4043 locally. Its sibling `rest_e2e.ts` binds an OS-assigned port instead, which is the better pattern.


jswir:
Reviewed this, and rather than leave a list I pushed the mechanical parts to the branch. **Four commits, all independent — revert any of them freely, none is load-bearing for the others.** Rebased onto your `88de783f`, so nothing of yours moved.

But read the last section first, because it questions the premise rather than the code.

### Merge conflict

`skills_bundle.json` only, and it is generated, so I regenerated rather than hand-merging the two JSON lines. Result carries both sides: #944's 25 reference prompts from `main` plus this branch's `@latest`. Same name set as main's copy, differing in exactly one entry (`malloy-getting-started`).

### `@latest` has no test (`577ca8df` → cherry-picked)

The pins are the fix here, and each carries a sentence saying why. But the failure mode was never someone arguing with the sentence, it is someone straightening a README and dropping four characters they read as noise. You gave the xlsx template's `WHERE` exactly this protection for exactly this reason; the printed commands did not get it.

`pinned_command.spec.ts` scans src, templates, scripts, both READMEs and the getting-started skill. Three lines name the bare form on purpose (the `registry_check` docstring, the `EUNSUPPORTEDPROTOCOL` note, and one workflow comment), so they are an explicit allowlist with a third test that fails when an entry stops matching. A blanket ban fails on the sentences doing the warning — the same trap your `::number.sum()` assertion hit.

Mutation-checked: dropping `@latest` from the root README fails and names `README.md:61`.

### The version check ran in CI (`be75471a`)

`CREATE_MALLOY_PACKAGE_NO_UPDATE_CHECK` was the only way off and nothing set it, so `create-malloy-package-npm.yml`'s smoke step made a live registry call on every run for advice nobody would read. `CI` and `NO_UPDATE_NOTIFIER` now skip it too — the second is what npm's own update-notifier honours, and someone who has already turned these off globally shouldn't have to learn a third spelling.

`false` and `0` mean off, because `CI=false` is a value real builds set. This does change `CREATE_MALLOY_PACKAGE_NO_UPDATE_CHECK=0`, which used to disable and now doesn't; the READMEs document `=1`.

Worth knowing: **adding `CI` would have turned the suite red on every runner.** The `fetchLatestVersion` tests would have stopped making requests and gone green for the wrong reason — the same shape of hole you're fixing in #948. They now clear all three names per test and restore them after. Confirmed both ways.

Two things in `getBody` while I was there. Body size counted `String#length`, which is UTF-16 code units, so a non-ASCII body was measured at up to half its real size and the cap let through twice what it names. And the over-cap path now settles itself rather than relying on what `destroy()` emits — it does settle on its own and fast, but by a different route per runtime (node: res `error` at 12ms; bun: res `end` at 12ms), and neither is promised. I had first assumed that path waited out the full deadline; measured, it doesn't, and the comment says what was measured.

The new test discriminates on the cap itself, not that line — remove `MAX_BODY_BYTES` and it goes from ~20ms to the deadline. Its 3s bound sits under bun's 5s per-test timeout so the assertion reports the number instead of the runner reporting "timed out".

### AGENTS.md still named three of four flags (`fe729102`)

The sentence you replaced was wrong for reading as exhaustive. Its replacement reads the same way and is still short one: `--host 127.0.0.1` is in the command and not in the list — the one flag with a consequence the agent can't undo, and the file spends a section on it further down.

Enumerating at all can't work, though: `{{startCommand}}` is `npm start` whenever the workspace already had a Publisher start script, and then the flags are that script's, not the four this scaffolder writes. No "the command carries X" is true in both cases. It's now a list of what each flag decides and what its absence means, framed as something to go read. No test asserts this prose; I checked before rewriting.

---

### The part worth your attention: I could not reproduce the staleness

Two comments in this package, ~200 lines apart, make opposite claims about the same npx mechanism:

- `scaffold.ts:258` (pre-existing) — *"`npx -y @malloy-publisher/server` with no version re-resolves `latest` on every start, forever… a workspace generated months ago boots whatever was published this morning."* This is the stated reason the generated start command pins an exact server version.
- `registry_check.ts:4` (this PR) — *"any cached copy satisfies it, so npm never asks the registry. The user gets whatever version they first installed, forever."*

So I tested it, on npm 10.8.2 against the live registry.

The mechanism is real and more specific than the comment says. Each distinct spec string gets its own `_npx` slot, and the slot records what `latest` resolved to **as a frozen semver range**:

```json
{ "dependencies": { "@malloy-publisher/create-malloy-package": "^0.0.2" } }
```

On `0.0.x`, `^0.0.2` means *exactly* 0.0.2 — which is the same caret-pins-on-0.0.x fact `scaffold.ts:270` already relies on. Offline, that slot is served without touching the registry.

**Online, though, it upgrades.** I planted 0.0.1 in the bare slot *and* rewrote its range to `^0.0.1`, then ran the bare command with the registry holding 0.0.2. It returned **0.0.2**. npm re-resolved.

I couldn't complete the equivalent downgrade test through `npm create` rather than `npx` — that path takes its own slot and I ran out of room — so treat this as tested on the `npx` path only.

Your 0.0.231-vs-0.0.232 observation is obviously real, so something produced it; I just can't get the stated mechanism to be the thing. Possibly a different npm, or the offline/`--prefer-offline` path, or the two runs straddled a publish.

None of that argues against the pins — they're harmless and correct under every path where the slot *is* reused. But two things follow:

1. The two comments should be reconciled, whichever way it lands. Right now the package tells opposite stories, and the `scaffold.ts` one is load-bearing for the `--host` pinning argument.
2. **I deliberately did not do the thing I'd otherwise have suggested.** `npx @malloy-publisher/server` is unpinned in ten doc sites (`README.md:27`, `docs/deployment.md:10`, `docs/configuration.md:17,20`, `docs/ai-agents.md:41,146`, `AGENTS.md:26`, `skills/malloy-getting-started/SKILL.md:14`, `skills/malloy-html-data-apps/SKILL.md:141`). Sweeping `@latest` across those rests entirely on the mechanism above, and I'm not confident enough in it to touch ten more files. Your call once the claim is settled.

### Verified

Every new test mutation-checked — I broke what each one guards and confirmed it goes red: `@latest` dropped from the README; `MAX_BODY_BYTES` removed (fails at 3629ms vs 1500); opt-out reverted to plain truthiness; env-clearing guard removed under `CI=true` (2 fail, which is what would have hit CI).

`create-malloy-package` 359/0, skills 7/0, server unit 1638/0, typecheck and eslint clean, `bunx prettier './packages/**/*.{ts,tsx}' --check` clean. Server *integration* fails on `beforeAll() expects a function as the second argument` and a network connect — both pre-existing and environmental, untouched by this.

mlennie:
Thanks @jswir, and all four are keepers. Taking the last section first, because you were right to question it and the answer is that we were both measuring correctly and describing different things.

**The two comments are both true. They describe different npm subcommands, and neither said so.** That is the whole contradiction. Measured on npm 11.12.1 against the live registry, in an isolated cache (`npm_config_cache` pointed at a scratch dir seeded from the real stale slot, so nothing touched the shared one):

`npm create @malloy-publisher/malloy-package` records its slot spec as `@malloy-publisher/create-malloy-package@*` with the range frozen at `^0.0.1`, and served 0.0.1: server pin 0.0.231, 29 skills. I ran it twice, so it is sticky, not a one-off. `npm create @malloy-publisher/malloy-package@latest` in that same cache got 0.0.2, server 0.0.232, 30 skills. And `npx @malloy-publisher/create-malloy-package` bare records the spec verbatim with no `@*` and re-resolved to 0.0.2 over a slot that was already holding 0.0.1.

So `npm create` rewrites a bare spec to `@*` before caching, and `@*` is satisfied by whatever range the slot froze, which on 0.0.x is an exact version, so no request is made. `npx` records the bare name as given and re-resolves it. Same cache directory, opposite behaviour, which is exactly why the scaffolder needs `@latest` in its documented command and the generated start script pins the server exactly. Your npx result is right; it just does not carry over to the `npm create` path, and that is the path every doc uses. Reconciled in 3cef566e, where each comment now names its subcommand and points at the other.

**The straddled-publish theory is ruled out twice over.** `npm view` puts 0.0.2 at 2026-07-28T18:46:17Z, and the 07-28 run was a side-by-side pair minutes apart where the bare form returned 0.0.1, `@latest` returned 0.0.2, and `npm view` reported 0.0.2 in the same session. Better than that, the two `_npx` slot hashes recorded that day are still on this machine, `284615ec773232c2` holding 0.0.1 with spec `@*` and `d2b146a1566d76e7` holding 0.0.2 with spec `@latest`, and they still reproduce today. The claim no longer rests on one observation.

**On the ten unpinned `npx @malloy-publisher/server` doc sites: I would leave them.** They are on the subcommand that already re-resolves, so `@latest` is a no-op there, and re-resolving is the hazard the `SERVER_VERSION` pin exists to prevent rather than one worth spreading. You were right not to touch ten files on an unsettled claim, and I think the settled claim says do not touch them at all.

**On your four commits.** The `pinned_command.spec.ts` allowlist-with-a-third-test shape is better than what I would have written; the mutation check naming `README.md:61` is the part that makes it worth having. The `CI` and `NO_UPDATE_NOTIFIER` opt-outs are plainly right, and the note that adding `CI` would have turned the suite green for the wrong reason is the same hole class as #948, so thank you for catching that it needed the per-test clearing rather than just the guard. The `String#length` versus byte-length bug in `getBody` is a real one I would not have found. And your `AGENTS.md` rewrite is correct that enumerating cannot work when `{{startCommand}}` may be someone else's `npm start`; framing it as what each flag decides is the right shape.

One thing to know about the branch since you looked: I put the two additions from the integration run through five review rounds against a booted server, which found and fixed a few things worth naming. The trust note was not host-gated, so a `--client cursor` run was being told to accept a Claude Code dialog that does not exist. The REST compile form came out entirely, because on a live server it answers 200 whether or not the source compiled and `{"source": ""}` answers `status: "success"` for a model path that does not exist, so the check I had documented would bless a typo; reload is the one whose failure is a status code, and its 424 and 503 are measured. Your merge also brought #941 in, which fired the cursor invariant on its `claude mcp add -s user` block, so that is now behind a `registrationNote` placeholder, empty for Cursor, with the directory-scope fact left ungated since it holds for both hosts.

Suite is 362/0 with the whole set, and the branch is now current with main.

One practical note from bringing it current, in case you hit it too: GitHub's "Rebase branch" fails on this branch with "Rebase conflict between head and base", and it will keep failing. Replaying 21 commits onto a moved main means replaying the ones that already carry main's changes through your merge, which collide with it. A merge is clean instead, `merge-tree` gives zero conflicts, so I merged main in locally with `--signoff`. Worth doing it that way rather than with the green button, since the button's merge commit is unsigned and the DCO check goes red. I also checked the one interaction the merge creates: #948 tightened `skills_bundle.spec.ts` and this branch regenerates `skills_bundle.json`, and that spec passes 11/11 against the regenerated bundle.

jswir (approved):

PR #950, opened 2026-07-30
Storage destinations: a disjoint set from connections

housejester:
A `#@ persist … storage=<name>` source is materialized into, and served from, a
warehouse other than the one its model reads. That warehouse had to be a
registered **connection**, which put it in the namespace models resolve names
from: a published model file or a notebook cell could name it and read it
directly. This adds a second list, `storageDestinations`, that only the
build and serve paths resolve.

NOTE: The alternative was a flag on the connection itself: a boolean, or a mode saying "this one is a materialization destination, not a queryable warehouse"... and it remains a defensible choice: one list, one lifecycle, one persistence path, and a much smaller diff than this one. What decided against it is where the filtering would have to live. A destination has to be invisible to GET /connections, to GET /connections/{name}, to the connection SQL endpoints, and to the connection lookup a model compiles against (basically, it should not be queryable via the various paths supported), and that last one has no per-connection round trip to hook: the package lookup hands any name but duckdb straight to the environment config, so keeping a flagged connection out of a model's scope means assembling one Malloy config for models and another for build and serve. That is this same split, derived a layer later and expressed as a filter every consumer has to remember rather than as a set it cannot reach. The cost of forgetting is what's asymmetric: with a flag, a consumer that iterates connections without checking it gets the destination for free (possibly inappropriate for the use case) — including consumers not written yet, which is the part review can't cover — whereas with two lists the omission is structural, because a destination isn't in the list being iterated. A flag fails open; a disjoint set fails closed. Worth re-litigating if the duplicated lifecycle here turns out to cost more to maintain than the filter would have.

`PERSIST_STORAGE_MODE` still defaults to `off` — no behaviour changes for a
deployment that isn't using the storage tier.

## Declaring one

`storageDestinations` sits beside `connections`, in
`publisher.config.json` or on the environment create/update body:

```jsonc
{
  "name": "examples",
  "connections": [ /* the warehouses your models query */ ],
  "storageDestinations": [
    {
      "name": "lake",
      "type": "ducklake",
      "ducklakeConnection": {
        "catalog": { "postgresConnection": { /* … */ } },
        "storage": { "bucketUrl": "s3://bucket/prefix" }
      }
    }
  ]
}
```

An entry has the same shape as a `ducklake` connection, so its validation applies
unchanged. `type` is restricted to `ducklake` — anything else admitted here would
be a second, unaudited way to define a warehouse.

## What a destination now is

- **Not resolvable from a model.** `connection.table(...)`, `connection.sql(...)`,
  notebook cells and query text all resolve against `connections`, so a model
  naming a destination doesn't compile and the package carrying it doesn't load.
- **Not addressable over the API.** Absent from `GET /connections` and
  `GET /connections/{name}` (404, as for a name that was never registered), and
  the connection SQL endpoints can't reach it. The status endpoint reports each
  destination's `name` and `type` so an operator can confirm what a server picked
  up; the config, which holds warehouse credentials, is never returned.
- **Independently named.** A connection and a destination may share a name and
  mean two different warehouses. Only `duckdb` is refused, since the per-package
  sandbox claims it.
- **Persisted.** Destinations have their own table in `publisher.db` and survive a
  restart like connections do. The config file seeds an environment that has none
  stored.
- **Resolved strictly.** A build naming a destination that isn't configured fails
  with 422 instead of falling through to a connection of that name — that
  fallback would materialize into a warehouse nobody chose.
- **Attached lazily, as DuckLake connections already are.** Nothing reaches a
  destination on the startup path, so an unreachable catalog can't take a boot
  with it; an unusable destination surfaces at its first build.

Destination DuckDB files also move under a `.destinations/` root, because an
instance is pooled by database path and the pool key excludes the connection
name: same root plus a shared name would have meant one shared instance and a
clobbered `ATTACH`.

## Compatibility

`storage=<name>` resolves only through `storageDestinations`, so a lake
registered as a connection no longer builds. The tier is gated behind
`PERSIST_STORAGE_MODE=off` and hasn't shipped in a release, so this is the config
shape for a feature that isn't live yet rather than a migration.

`docs/connections.md` and the `storage=` tutorial said a DuckLake connection
doubled as a destination; both are corrected, and the changed tutorial commands
were re-run against a real server.

## Testing

Unit and integration suites cover the split in both directions — a model, a
notebook cell and a query all fail to resolve a destination, while a binding onto
that same destination resolves on the serve path, asserted together on one
environment so neither half can pass alone.

The `hammer/` scenarios exercise it end to end against real Postgres and
DuckLake: a new scenario materializes into a destination and serves a routed
query from it while two packages that merely name it are refused. Two existing
security scenarios are updated where the split supersedes what they were
guarding — user SQL can no longer *name* the lake rather than merely being unable
to mutate it, and the SQL endpoint can no longer reach it at all.


kylenesbit (approved):
Approving. The disjoint-set choice is right, and I checked that the boundary actually holds rather than taking the description's word for it: model compiles and notebook/ad-hoc query text resolve through `Package.malloyConfig` (connections only), the connection endpoints and `sqlQuery` resolve through `apiConnections`, and the serve shape gets its own config narrowed by `restrictMalloyConfigToConnections` to just the destinations its own bindings name. All three `Package.create` sites get `attachDestinationServeConfig`, and both writes to the `models` map (initial create and `reloadAllModels`) re-push it, so there is no load path that ends up without it. The reasoning in the description for two lists over a flag is the correct call for the reason given — the package lookup hands any non-`duckdb` name straight to the environment config, so a flag would have forced two assembled configs anyway.

Three things worth fixing, one of them confirmed by running it.

## 1. An environment with no destinations never assembles its destination config

`setStorageDestinations` early-returns when the new list equals the previous one. In the constructor both sides are `[]` for the overwhelmingly common no-destination case, so `rebuildDestinationMalloyConfig()` never runs and `destinationMalloyConfig` stays `undefined` — contradicting the `!` and its "Always assigned: the constructor sets the destination list unconditionally" comment.

Both consequences reproduce against a real `Environment`:

- `closeAllConnections()` logs `error: Error closing storage destinations for environment <env> {"error":{}}` for every environment that has no destinations. That is an error-level line on every shutdown of every deployment not using the storage tier, which is the opposite of "no behaviour changes for a deployment that isn't using the storage tier" — and the serialized `{"error":{}}` gives an operator nothing to act on.
- `getStorageDestinationMalloyConfig()` throws `TypeError: undefined is not an object (evaluating 'this.destinationMalloyConfig.malloyConfig')`. Reachable on the serve path whenever bindings exist but the list is empty, which is precisely the boot where the stored destinations could not be read — the case `markStorageDestinationsUnknown` exists for. The live-fallback `catch` swallows it and logs at `debug`, so it degrades quietly instead of surfacing the intended `ConnectionNotFoundError`.

Gating the no-op on the config having been built fixes both:

```ts
if (this.destinationMalloyConfig && storageDestinationsEqual(previous, this.destinations)) {
   return;
}
```

## 2. The missing-environment-directory restore branch can permanently prune stored destination rows

In the boot restore loop, the branch for an environment whose path no longer exists passes the config's destinations through unconditionally (`environment_store.ts` ~652):

```ts
storageDestinations: environmentConfig.storageDestinations,
```

`getProcessedPublisherConfig` normalizes that to `[]` for a config that declares none, and `[] !== undefined` reads as an explicit clear, which marks the list authoritative. `addEnvironment` then calls `addEnvironmentToDatabase` → `syncStorageDestinations`, which prunes every row the list does not hold. So an environment whose directory is briefly absent at boot loses every API-registered destination permanently — the failure `abd11c2` was written to close, arriving by the config path instead of the read-error path.

The two sibling call sites (lines 539 and 1510) both guard with `...(x.length ? {…} : {})` for exactly this "empty reads as clear" hazard; the same guard belongs here (or a `markStorageDestinationsUnknown`). Connections in that branch survive because `addConnections` never prunes, so the loss falls only on destinations.

## 3. Smaller notes

The legacy `POST /projects` alias in `server-old.ts` calls `environmentStore.addEnvironment(req.body)` without the `processStorageDestinationsOrThrow` gate that `POST /environments` gained in `3875274`, so through that route a destination the server cannot read still returns 200 with the entry quietly absent — the outcome that commit set out to eliminate on create.

The rename to `storageDestinations` left a few "materialization destination" stragglers. Only one is reader-facing: `docs/persist-storage-tutorial.md:21` still introduces the concept as "a **materialization destination**". The rest are comments (`server.ts:955`, `server-old.ts:119`, `DuckDBRepository.ts:151`, `environment_store.spec.ts:255`).

---

Also checked and clean: `tsc --noEmit` on the server passes, the committed `api.ts` matches `api-doc.yaml` after regeneration, and the 191 tests across the five new/changed spec files pass. Scenario 69 is a good shape for this — both halves of the split asserted on one environment, with `cites:` pinning the refusal reason so it cannot pass for the wrong one — and the `lake_probe` addition to scenario 09 strengthens rather than weakens it, since it shows the read-only attach still holds where a name does resolve.

PR #951, opened 2026-07-30
docs(skills): correct the Excel serial-date claim and the bare .xlsx read, add stop_at_empty and stddev

mlennie:
## What this fixes

Four things in `malloy-gotchas-modeling`, found by meeting the shipped skills the way an agent meets them rather than by reading them. These reach agents three ways (files on disk, `skills_bundle.json`, and MCP prompts since #944), so a wrong line lands three times.

**1. The Excel serial-date format was invented.** The section said `all_varchar = true` returns date cells as Excel serial strings "for example `'46027.0'`". Measured against a real export: 1,459 bare integers like `'44929'`, none carrying a decimal point. This is not cosmetic. An agent wrote its serial-date detector as `regexp_matches("Order Date", '^[0-9]+$')`, which the documented format would have failed on for every one of those rows, sending them through `try_strptime` to NULL and silently dropping its largest finding. What is actually true is narrower and depends on the writing tool: a sheet Excel wrote gives `'44929'`, DuckDB's own xlsx writer gives `'44927.0'`, and a cell carrying a time of day gives `'44929.5'`. So the section no longer tells you to pattern-match for an integer, it tells you to `try_cast`, which accepts all three and nulls a cell that was stored as text.

**2. It labelled a failing pattern `// RIGHT:`.** A bare `duckdb.table('data/budget.xlsx')` was labelled RIGHT with the claim that `duckdb.table('file.xlsx')` "just works". On a messy workbook that call returns one row: DuckDB takes the lone title cell in A1 as the only column name, and the default `stop_at_empty` halts at the blank line under the banner. The never-convert message is right and stays, with the reason now spelled out, but the unconditional promise does not. The RIGHT label is conditional on a sheet that is a plain table, the section says to check the row count and the inferred column types before building on the read, and the quirk list gains the title-row case.

**3. Two missing quirks.** With no `range`, `stop_at_empty` defaults to true and the read stops at the first blank row, which on a real sheet is usually a spacer between blocks rather than the end of the data. And a `range` then reads every cell inside it, so an overshot bound pads with all-null rows and columns while a short bound drops data with no error at all, which makes the row filter the thing that ends the read rather than tidying-up.

**4. `stddev` works and nothing said so.** The section that is emphatic about which statistical functions do not work never mentioned it. It is a native Malloy aggregate rather than a raw-SQL escape, so it compiles inline and as a `measure:`, which is exactly what that section is about, and it is the sample standard deviation. Dispersion is what an agent reaches for to answer whether one category is clearly worse, and no skill in either repo mentioned it. The closing "search the docs first" checklist said statistical functions do not compile as measures, so it now names `stddev` as the exception.

## What grew, and why

The corrections obliged some guidance that was not in the original ask, and it is worth being explicit about it. Telling a reader to pass a `range` obliges the section to say what a wrong bound costs. Telling them a money column arrives as text obliges it to say how to convert one, and there the honest answer turned out not to be a recipe: five successive regex forms for handling accounting negatives were each measurably wrong, including one that turned `1234 (est)` into `-1234`. So the section gives the plain-case expression, names precisely what it cannot do, and tells the reader to sample the column's shapes first and to check the result against the sheet's own total row read as raw text, because a wrong sign survives a row count, survives `select: *`, and cancels out if you parse both sides the same broken way.

This also fixes the same unconditional `.xlsx` promise in `malloy-modeling`, which is the skill an agent reads first, so the two no longer disagree.

**The tradeoff, stated plainly:** the Excel section goes from 228 words and 6 bullets to 1,504 words and 12 bullets. Everything in it is a correction that was measured, and I would not cut further without dropping one, but it is now the longest section in the file. A reasonable follow-up is that the messy-workbook material belongs in its own skill or in the public docs rather than in a gotchas list, and I am happy to move it if you would rather.

## How it was verified

Every example was compiled and run, not eyeballed. The `read_xlsx` form in the fence returns 1,500 rows on the workbook these numbers came from, both through the standalone compiler and through a package served by a running Publisher, where the bare call returns 1. The fence text as served over MCP is byte-identical to the file and compiles to the same result. Every number in the section was re-derived from the raw sheet XML as well as through the extension, so a quirk in `read_xlsx` could not make a wrong figure look right twice.

`skills_bundle.json` was regenerated rather than hand-edited: a fresh regeneration is byte-identical to the committed file, it still reports 55 entries (30 skills plus 25 reference files), and exactly two entry bodies differ from `main` with nothing added, removed or reordered.

Gate: typecheck, lint, unit (1638), skills (7) and integration (232) all pass. The six `prettier:check` failures are the standing git-ignored generated files that CI never sees.

## Notes for review

`malloy-gotchas-modeling` is mirrored in `ms2data/agent-skills`, so this has a twin PR there. That repo has no Excel section at all, so only the `stddev` paragraph and the checklist line are mirrored, and those two hunks are byte-identical across the repos. `malloy-gotchas-queries` was byte-identical between the repos before this change and still is.

This regenerates `skills_bundle.json` and so does #946, so it should merge after it to avoid regenerating twice.


jswir (approved):
Generally LGTM, but I might I could imagine the xlsx stuff going into an xlsx specific skill?

mlennie:
@jswir Agreed, and I'd rather do it as a follow-up than fold it in here.

Why this PR is shaped the way it is: the four things it corrects are live wrong claims that reach agents three ways (files on disk, `skills_bundle.json`, and MCP prompts since #944), and one of them nearly cost a real analysis its biggest finding, so I wanted the correction landable on its own.

But you're right about the shape. That section went from 228 words to about 1,500 and it is now roughly twice the length of the next-longest section in the file, and it has stopped being a list of gotchas: it reads as a pipeline, which is profile the top rows, read with a range, check the row count, convert the text columns, verify against the sheet's own total. That is a workflow, and workflows belong in their own skill.

A split also fixes something this PR cannot. The Excel material is publisher-only, because `ms2data/agent-skills` has no Excel section at all, so today the divergence is a section sitting inside a file that is otherwise byte-identical across the two repos. A separate `malloy-xlsx` skill turns that into a whole file one repo has and the other does not, which is a much cleaner thing to keep in sync. It would also let the frontmatter description say "spreadsheets", which `malloy-gotchas-modeling`'s does not, so retrieval currently under-serves this material even when it is correct.

If that works for you I'll open the split as its own PR. Two things I'd decide in it rather than now: whether the new skill is publisher-only or mirrored upstream, and whether the genuinely one-line facts stay behind as gotchas (only the first sheet is read by default, every number is a double, headers with spaces need backticks) with just the pipeline moving out.

PR #955, opened 2026-07-31
feat(server): write .mcp.json on startup so npx no longer needs claude mcp add

mlennie:
An MCP client only knows a server exists if something registered it: a `.mcp.json` in the directory the session started in, or a manual `claude mcp add`. A running server is otherwise invisible to it, and starting the server first does not help, because registration and ordering are two separate requirements that are easy to run together. Three of the four ways into Publisher already register themselves. The repo ships a committed `.mcp.json`, the scaffolder writes one into every workspace it creates, and a clone therefore works out of the box. `npx @malloy-publisher/server` was the only entry point that did not, which is the sole reason the getting-started docs had to teach MCP wiring before a reader could ask a question about their data.

So the server now creates a `.mcp.json` in its working directory, naming the MCP port it actually bound. The whole setup becomes `npx @malloy-publisher/server --port 4000 --host 127.0.0.1`, then open your agent in that directory. No `claude mcp add`, no `/mcp`, and nothing about ports to learn.

## What it does, precisely

It creates and never edits. If the directory already has a `.mcp.json` the server leaves it exactly as it is, does not read it, and logs the one command to add itself instead. It skips git working trees, because a checkout already has a correct config from the repo or the scaffolder and an untracked one is just a surprise in `git status`. It skips the home directory. It uses `wx` rather than checking existence and then writing, since those two disagree about what they are naming when `.mcp.json` is a dangling symlink. It reads the bound port back from `mcpServer.address()` rather than reusing the requested value, because `--mcp_port 0` is a live idiom in this repo's own e2e harness. Every failure path returns rather than throws, so a convenience file can never stop a server booting. `--no-mcp-config`, or any truthy `PUBLISHER_NO_MCP_CONFIG`, turns it off.

## Why it creates rather than merges

An earlier version of this branch merged into an existing file the way `scaffold.ts` does. Review found that nearly every hazard in the change came from that one decision, so it was removed. Replacing an existing `malloy` entry destroyed user fields, including auth headers and the `mcp-remote` stdio entry that this repo's own AGENTS.md tells stdio-only clients to write. Reading the file blocked the event loop indefinitely when that file was a FIFO or a symlink to `/dev/stdin`, taking `/health` down with it, and both of those are committable to a repository. Repointing on every boot rewrote version-controlled files. And the log line echoed the previous URL verbatim, which leaks a credential whenever an MCP endpoint carries a token in its userinfo or query string. Not reading the file at all removes all of them together.

The scaffolder can afford to merge because it runs once, deliberately, at the user's request. A server boot cannot. The tradeoff is real and worth stating: someone who already has a `.mcp.json` no longer gets this server added automatically, and is told the command instead.

## What it does not change

The file is still only discovered by a session that started in that directory, so "open your agent here" remains a genuine requirement rather than something this removes. The host still prompts on first use, which is correct for a server that can read your data.

## Verification

Typecheck clean, prettier clean, 1751 unit tests and 232 integration tests passing, and a full integration run leaves no untracked file behind. The unit tests went from 11 to 19 and encode the review findings rather than the happy path: byte-identical preservation of a credential-bearing file and of a stdio bridge entry, a dangling symlink not creating its target, a FIFO returning promptly instead of hanging, git work trees and their subdirectories, a `.git` file as worktrees store it, an unnormalised home path, and the log output itself, which previously had no coverage.

Verified live rather than only in tests. In a fresh directory the file is created with the bound port and `claude mcp list` then lists the server where it previously listed nothing. In a `packages/server` subdirectory of a repo it skips and the tracked root config is untouched. A file holding `Bearer SECRET` comes back byte-identical with no occurrence of the token or the old URL anywhere in the log. A dangling symlink's target is not created. And end to end by hand: server started in an empty directory, Claude Code opened there, `malloy_getContext` returned real data, with `claude mcp add` never typed.

## Follow-ups, not in this PR

The Docker image has `WORKDIR /publisher` and no `USER`, so a container boot creates a config as root in a place no agent will read. Setting `PUBLISHER_NO_MCP_CONFIG` in the Dockerfile is the explicit fix. A reviewer suggested gating the whole feature on `process.stdout.isTTY` instead, which was declined here as too blunt: it would also disable it under `nohup`, background starts and tmux.


jswir:
Nice change — create-never-merge is the right call, and the tests encoding the rejected alternatives rather than the happy path is the part I'd most want kept. One thing I'd pull forward out of the follow-ups section, plus some nits.

## The Docker path writes a config on every boot

The follow-up is listed as "not in this PR," but it isn't latent — nothing suppresses the feature in the container, so every boot of the published image creates `/publisher/.mcp.json`. Chain, all on this branch:

- `Dockerfile:75` — final stage is `WORKDIR /publisher`, and `CMD` runs from there, so `process.cwd()` is `/publisher`.
- `.dockerignore` excludes `.git`, so `insideGitWorkTree()` cannot fire — the guard that would otherwise cover this exact case is inert in the image.
- The final stage `COPY`s only named files, so there's no pre-existing `.mcp.json` to take the `exists` branch either.
- No `USER` directive anywhere, so the file is created as root.

In the documented setups the damage is limited: it lands in the container's writable layer and dies with `docker rm`, and `docs/deployment.md` only bind-mounts individual files at `/publisher/publisher.config.json` plus a named volume for `publisher_data/`. But anyone who mounts a project directory at `/publisher` — not documented, not unusual — gets a root-owned untracked file in their tree, which is precisely what the git guard exists to prevent. And in every container it's a write and a log line that can never do anything useful, since no agent session starts in there.

`ENV PUBLISHER_NO_MCP_CONFIG=1` next to `ENV NODE_ENV=production` (~`Dockerfile:111`) is the whole fix, and it keeps the escape hatch for the mount-at-`/publisher` case via `docker run -e PUBLISHER_NO_MCP_CONFIG=`. Shipping the feature and its one known-wrong path in the same release seems better than splitting them.

## Nits

- **`PUBLISHER_NO_MCP_CONFIG=false` disables the feature.** `server.ts:1994` is a non-empty-string check, not truthiness — `=false` and `=0` both turn it off. `docs/configuration.md` ("Set to anything") matches the code; the inline comment and PR body say "truthy," which is wrong for those two. Today it's a doc wart because nobody types `=false`. If the Dockerfile `ENV` above lands, `-e PUBLISHER_NO_MCP_CONFIG=false` becomes the natural way to opt back in and silently won't work — so I'd treat `""`/`"false"`/`"0"` as unset in the same commit.

- **`skipped-git` logs nothing, and the walk goes to the filesystem root.** For anyone whose `$HOME` is a git repo (dotfiles), the feature silently never fires in any directory under it, while the README now promises "there is nothing to configure." Silence is right for the clone case, where a `.mcp.json` is already present; it's wrong here. A debug-level line, or staying silent only when a `.mcp.json` actually exists up-tree, would close it.

- **`--mcp_port 0` leaves a permanently stale file.** First boot writes the ephemeral port; every later boot in that directory hits `EEXIST` and leaves a file naming a dead port. The module header argues "a config naming a port nothing is on is worse than no config," and create-never-edit makes exactly that state reachable and permanent. Either skip the write when the requested port was `0`, or have the `exists` log mention the existing file may name a different port.

- **The home-directory tests run against the real `os.homedir()`** (`mcp_config.spec.ts`, the two `skips home` cases). They're safe only because the home guard returns before any write — if that ordering ever regresses, the test writes `.mcp.json` into the developer's actual home directory instead of failing. Injecting the home path would make a regression show up as a red test.

- **The port read-back is the one piece of new logic with no coverage.** `bound.port` vs `MCP_PORT` is inline in the `listen` callback, so it's untestable as written; extracting `resolveBoundPort(server, requested)` makes it a three-line test. Same for `cwd`-not-`server_root`, which a future reader is fairly likely to "fix" — a comment carries it now.

- Minor: `mcpServer` is referenced inside its own `listen` callback. Not a bug — the callback is registered as a one-shot `listening` listener, so the binding is initialized by the time it runs — but `this` is the server in there and sidesteps the question entirely. And `malloyServer()` returning `Record<string, string>` lets `type: "http"` drift; a literal type or `as const` would pin it.

jswir (approved):

PR #956, opened 2026-07-31
fix: close the gaps that cost an agent a session

jswir:
An agent driving Publisher end-to-end hit a run of avoidable friction. Each fix below is traced to one of those.

## What was broken

**A boolean query param you misspell did nothing, successfully.** Four params were read as `=== "true"`, so every other spelling — `?reload=1`, `?dropTables=yes`, `?reload=TRUE`, or the parameter repeated — quietly read as `false` and the request succeeded while doing nothing it asked for.

- `reload` on a package cost the reported session ~6 tool calls: edit a model, reload, see `200`, query a model the server never recompiled.
- `dropTables` was worse. `DELETE …/materializations/{id}?dropTables=1` deleted the materialization record, left its physical tables on disk, and answered `204 No Content` — nothing in the response said half the operation was skipped.
- `bypass_filters` failed closed rather than open, but silently the same way, and sits directly beside `filter_params` and `givens`, which both already answer 400 on a malformed value.

All four now share one `parseBooleanParam`, accepting only the two spellings `api-doc.yaml` declares and refusing anything else with a 400 that quotes the value back and names a form that works. The message carries the param name and the caller's own method, so the destructive one does not suggest a `GET`. `bypass_filters` keeps absent distinguishable from an explicit `false` via `optionalBooleanParamOr400`, because `setFilterDeprecationHeaders` fires on `bypassFilters !== undefined` and collapsing absent to `false` would stamp a Deprecation header on every notebook-cell request.

**`?reload=true` on a collection route silently did nothing.** Collection routes ignored `reload` entirely and answered `200` with the list, which reads as a reload that worked. They now refuse with `400` naming the per-resource route. This covers the legacy `/projects` twins as well as the modern ones — the per-resource legacy routes and the collection legacy routes both.

**`malloy_getContext`'s `sourceName` drill-down never returned a field.** It filtered to `kind === "source"` *and then* to that source, so it returned exactly one row — the source itself, which the caller already had from the tier-3 listing. It now returns the source plus its views, dimensions, measures and named queries. There was **no `sourceName` coverage in the spec at all**, which is how this shipped; nine tests now cover it (five unit, four integration), including the unknown-source, cross-source-exclusion and `limit` cases.

The description states the empty-array invariant per tier, because it differs: the listing tier is always led by the source's own row, so `[]` means no such source, while the `sourceName` + `query` tier is ranked retrieval, where `[]` means nothing matched and the source's own row appears only if it ranks.

**Load-time render-tag validation dropped every inert tag.** `Model.validateRenderTags()` filtered to `severity === "error"`, and the renderer reports a tag that is *well-formed but inert where it sits* as `warn` — `# colspan` outside `# dashboard { columns=N }`, or a colspan wider than the grid. Those are precisely the findings an author cannot see any other way: the model compiles, the package loads, the query runs, and only the layout is silently not what was written. Both severities now ride the package's `warnings` with the renderer's own severity attached.

Publisher's own `# artifact` and `# drill` are dropped first via the existing `filterPublisherOwnedRenderLogs` — the renderer reports each as `Unknown render tag` at `warn`, which was invisible while this filtered to errors and would otherwise be a finding on every dashboard file. There is no severity filter after it: keeping only the two severities the renderer emits today would silently discard a third, which is this same bug again.

**`create-malloy-package .` gave a self-contradictory error.** It said `.` is an allowed *character* while refusing `.` as a *name*. There is no mode that scaffolds into the working directory — `createPackage` always makes `cwd/<name>/` — so the message now says that and names the two forms that do work.

**Skills taught `from()`, which no longer parses.** `from()` was removed from the language and fails with `unexpected 'from'`. Corrected in `malloy-debug`, `malloy-model`, `malloy-lookml-review` (`_concepts`, `build-derived-tables`), and `malloy-review/rubric-structure`, verified against the real compiler.

**Local data files fail in the CLI and VS Code but work in Publisher.** `duckdb.table('data/x.csv')` resolves against the DuckDB connection's `workingDirectory`. Publisher sets it to the package root; every other host reads it from `malloy-config.json`, where a *relative* value resolves against the process cwd — not the config file's directory, not the workspace root. So the same config compiles from one directory and fails from another, and the failure looks like a model bug because the first `IO Error` is followed by a "not defined" cascade for every field. `malloy-getting-started` gains a section telling agents to ask, and `malloy-gotchas-modeling` gains the mechanism.

Two tool descriptions were imprecise rather than wrong: `getContext` didn't say `retrieval` is *absent* (not defaulted) with no embedding provider; `searchDocs` didn't say it indexes the Malloy language docs only. Both are near the 2000-char truncation budget `server.protocol.spec.ts` pins, so existing prose was tightened to fit.

## The reported "Unknown error" does not reproduce

Driving a real `from()` failure through both `malloy_compile` and `malloy_reloadPackage` on `main` returns the full diagnostic. The four MCP error files are byte-identical between `v0.0.233` and `main`, so the shipped build behaves the same; the session ran an older cached build. `AGENTS.md` and `docs/ai-agents.md` therefore gain the recovery rule, the cause (stale npx cache — pin `@latest`), and the note that the server does not report its own version, which is why a stale build is invisible.

## Two things reviewers should know

**The example relayouts are checked by the server, not just by a clean compile.** `@malloydata/render-validator` has no `colspan` logic because it *delegates*: it instantiates a real `MalloyRenderer`, calls `viz.setResult(result)`, and returns `viz.getLogs()`, so the finding was always available. What discarded it was Publisher's own `severity === "error"` filter, fixed here rather than upstream. Each relayout reports zero findings, and a deliberately flex-mode `# colspan` reports `Ignored # colspan ... colspan only applies in columns mode`.

**Layout is still not visually reviewed.** Nothing in the Playwright suite asserts dashboard layout, and validation proves only that each `# colspan` is attached where it can apply, not that the result looks right. Two of the three relayouts have a spanned KPI strip over three tiles, so their last row holds one tile and a gap; `storefront` has four and fills cleanly. The extra `nest:` in `subscriptions.malloy` is cosmetic — it makes that file match the structure its two siblings already had, and `# colspan=2` binds to `kpis` identically either way.

## Compatibility

Three existing routes gain a `400`, declared in `api-doc.yaml` alongside the accepted values for each parameter. Every request that now fails was already a silent no-op, so no working behaviour changes — but a client that ignored the response body and trusted the status will now see the failure it was previously missing. Generated clients send real booleans and are unaffected; nothing in this repo sends a non-boolean value to any of the four params. `bypass_filters` already declared an enum of exactly `"true"`/`"false"`, so the server now matches what the spec always said.

`RELEASE_NOTES.md` carries a section for this. `@malloy-publisher/skills` and `@malloy-publisher/create-malloy-package` version on their own line and are published only when their version is not yet on npm, so both needed a bump — otherwise the skills and scaffold fixes here would land on `main` and be skipped with the release still green. The 0.2.0 stamp on `main` has since bumped both to the same versions this branch did (0.1.10 and 0.0.9), so those two lines are now a no-op against `main`.

## Testing

Full `bun run build`, `bun run typecheck` and `bun run lint` clean. Tests on the Bun this repo pins (1.3.13): `363 pass` sdk, `3253 pass / 3 skip / 0 fail` server unit, `406 pass / 1 fail` server integration.

The single red is `concurrent_package`, which downloads a package from `gs://publisher_test_packages/` and times out without GCS access. It is environmental, not a code failure.

Worth knowing if you run these locally: on Bun 1.4.x, 33 `config`/`theme` unit specs fail on `fs.rmdirSync(path, { recursive: true })`, which newer Node removed. That is the runtime, not this branch — the same 33 pass on 1.3.13, which is what `.tool-versions` and every workflow pin.

New coverage:
- `reload_guard.integration.spec.ts` — 11 cases over real HTTP via `startRestE2E`: both collection halves (modern and legacy), the per-resource value refusals including the legacy twins, the repeated parameter, the destructive `dropTables` sibling, and the case that catches a guard on the wrong route (per-resource must reach 404, not the collection's 400). Messages are pinned including the echoed value.
- `query_param_utils.spec.ts` — the value contract for `parseBooleanParam` on its own, plus `invalidBooleanMessage` pinned in full, in the file kept import-free for exactly this.
- `render_tags.integration.spec.ts` + a fixture package carrying one finding of each severity, a correctly-attached control, and a `# artifact`/`# drill` control asserting the Publisher-owned suppression reports nothing.
- `mcp_get_context_drilldown.integration.spec.ts` — the `sourceName`-without-`query` tier against the real compiled `storefront` package.

The three specs whose `beforeAll` timeout argument this PR replaces (`readiness_line`, `freshness_gate`, `mcp_get_context_semantic`) **were running before this change**. An earlier revision of this PR claimed Bun rejects `beforeAll(fn, timeout)` and that nine tests were silently absent; that is not true on the Bun this repo pins and CI runs (1.3.13, per `.tool-versions` and every workflow), where the argument is accepted and load-bearing — a 6s hook passes with it and times out at 5s without it. `setDefaultTimeout` is kept anyway because it covers hooks and still holds when a spec is run on its own rather than through `test:integration`'s `--timeout` flag, and the comments now say that instead.

## Companion PR

`malloy-debug`, `malloy-gotchas-modeling`, `malloy-model`, `malloy-lookml-review` and `malloy-review` are shared skills. A matching PR mirrors them byte-identically in `ms2data/agent-skills`.


mlennie:
Read the whole diff. The code is sound and better tested than the description claims. Two things in the body need correcting, and one variant of the bug this PR exists to kill is still open.

## The validator-gap caveat is misdiagnosed

Your conclusion is right: Publisher's load-time validation does not catch a `# colspan` silently ignored in flex mode. The stated cause is not, and the cause is what decides which repo the fix goes in.

It is not that the validator lacks colspan logic. `Model.validateRenderTags()` drops the finding:

```ts
// packages/server/src/service/model.ts:1716
const errors = validateRenderTags(result).filter((log) => log.severity === "error");
```

The ignored-colspan case is deliberately `log.warn`, per its own comment in `render-field-metadata.ts` ("Well-formed but inert here: 'Ignored', not 'Invalid'"), and `warn` maps to severity `'warn'` in `render-log-collector.ts`. So Publisher discards it by design. It does catch *invalid* colspan (non-numeric, out of range), because those are `log.error`.

And `@malloydata/render-validator` is not missing the check. It has no tag logic at all: `validateRenderTags` instantiates a real `MalloyRenderer`, calls `viz.setResult(result)`, and returns `viz.getLogs()`. `setResult` constructs `RenderFieldMetadata(..., this.logCollector)` (`api/malloy-viz.tsx`), and the colspan checks run during that construction and land in that collector. A source grep of the validator finds nothing precisely because it delegates. The checks landed in the same commit as columns mode (#2943, 2026-06-24, shipped around 0.0.419), and this branch pins 0.0.426.

Why it matters: as written, this sends the next reader upstream to add logic to the validator that is already there. The real fix is one line here, widening that severity filter or surfacing warn-severity findings as their own tier, and Publisher already has the `warnings` surface from #861 for exactly this. Either fix it in this PR or file it with the corrected cause.

To be clear about what I verified how: the Publisher filter is in this branch's own code and is conclusive. The delegation chain is a source trace against malloy `main` plus release dating, not an executed probe.

## The body is stale about its own test coverage

It says "No test covers the new 400. server.ts calls .listen() at module scope so its app can't be imported into a spec... The contract is pinned in api-doc.yaml instead." The branch has `packages/server/tests/integration/reload_guard/reload_guard.integration.spec.ts`: five cases over real HTTP via `startRestE2E`, including the one that catches the guard being attached to the wrong route (per-resource routes must reach 404, not the collection's 400). The "1738 pass" figure predates those additions too. As it stands, reviewers go looking for a gap that is already closed.

## Reload strictness is now inconsistent, and the lenient side keeps the same failure

Rejecting every value on the collections is defensible, and the test argues it inline. But the per-resource routes still read `req.query.reload === "true"` (`server.ts:999` and `:1479`), so `?reload=1`, `?reload=yes`, and `?reload=TRUE` on `.../packages/{pkg}` all answer 200 without reloading. That is the same session-costing failure in a different disguise: edit a model, see 200, query a model the server never recompiled. This PR closes the missing-package-name mistake and leaves the typo'd-value mistake. Cheap to close by refusing a `reload` value that is not `true` or `false`.

## The example relayouts are a visual change nobody has looked at

Three shipped examples move from flex to `columns=2` with `# colspan=2` on the KPI strip. The body concedes these rest on the renderer's logic and a clean compile rather than on the validator agreeing, and per the first point Publisher actively discards warn-severity colspan findings, so a mis-attached tag is invisible at load. Two specifics worth a screenshot each: `examples/html-data-app/subscriptions.malloy` is the only one where `kpis` is a bare view reference rather than an inline `kpis is {...}`, and the only one that gained a second `nest:`, so it is the case least like the other two; and all three end up with three tiles in a two-column grid (spanned KPI strip, then 2+1), leaving a hole in the last row. Nothing in the Playwright suite asserts dashboard layout.

## Scope

25 files spanning a REST contract change, an MCP retrieval fix, a CLI message, two tool descriptions, eight skills, three example models, and two docs. The 400 is the only piece carrying compatibility risk and it is the piece most likely to be skimmed past the skills diff. I checked the blast radius and it is genuinely small: no publisher client and nothing in the MS2 service sends `reload` to either collection route, and the parameter is not in `api-doc.yaml`, so generated clients cannot send it. Still, the skills half is already mirrored in agent-skills#77 and could land on its own.

## What holds up

The getContext fix is correct, and I checked the reasoning rather than taking it on faith: `collectEntities` gives a source entity `source: <its own name>` and a named query `source: query.sourceName`, so the single `e.source === sourceName` predicate picks up the source row plus its views, dimensions, measures, and any named query built on it, in insertion order with the source first. Collapsing two chained filters into one predicate is also just better code, and the drill-down stays uncapped by default.

The `from()` cleanup is complete, which is the kind of sweep that usually ends up half done. No `from(` remains anywhere in `skills/`, `docs/`, `examples/`, or `src/` outside the explanatory prose, and `skills_bundle.json` is regenerated to match: the new strings are in the bundle and `from(driver_stats)` is gone.

Two small ones. The unknown-`sourceName` empty array now carries a useful invariant worth stating in the tool description: a real source always returns at least its own row, so `[]` unambiguously means "no such source" rather than "a source with no fields", which is exactly what a caller cannot tell from the current wording. And three skills quote the error as `unexpected 'from'`; `FROM` is still a lexer token and appears in the parser's usable-as-identifier list (`MalloyParser.g4:596`), so the real message may differ. You said you verified it against the compiler, so this is just worth a second look given agents may pattern-match on the string.

Last thing: the branch pins render 0.0.426 while main is on 0.0.427 (#958). Rebase, not an issue.

jswir:
Thorough review — thank you. You're right on the three substantive points, and I ran executed probes at `0.0.426` (the pin this branch had) to settle the parts you flagged as source-traced rather than measured. Everything is fixed in `941e2ed`; the description's two wrong paragraphs are rewritten.

## The validator gap: your diagnosis is right, and I confirmed it by execution

You called the delegation chain "a source trace plus release dating, not an executed probe." It holds. A flex-mode `# colspan` driven through `validateRenderTags` comes straight back:

```
[warn] Ignored # colspan on 'kpis': colspan only applies in columns mode.
       Fix: add # dashboard { columns=N } to 'root', or remove # colspan.
```

So the validator never lacked the check, and my stated cause was wrong in the way that matters: it pointed the next reader upstream. Fixed here instead, by keeping both severities. The description now records the correct cause.

**One correction to your correction.** You wrote that Publisher "does catch *invalid* colspan (non-numeric, out of range), because those are `log.error`." Only non-numeric is:

| case | severity | Publisher saw it before? |
|---|---|---|
| `# colspan=abc` | `error` | yes |
| `# colspan=99` (exceeds columns, clamped) | **`warn`** | **no** |
| `# colspan=2` in flex mode | `warn` | no |

The clamp case was being dropped too, which only strengthens the argument for widening the filter.

I did not reuse #861's `warnings` surface as a separate warn tier — that surface already carries a `severity` field typed `error | warn` whose `warn` arm had no producer, so the fix was to start producing it rather than to add a tier. `warn` and `error` are the only severities the renderer emits, so nothing is withheld now. I checked this isn't a noise change: all three bundled example packages report **zero** findings.

That also closes your fourth point better than a screenshot would have. The relayouts in this PR are now validated at load time rather than resting on a clean compile, and there's a fixture package plus `render_tags.integration.spec.ts` pinning it: the `warn` surfaces, the `error` still does, a correctly-attached `# colspan` reports nothing, and the package still loads with both tags broken.

## Reload strictness: fixed, and the failure reproduced first

Confirmed against the real server before changing anything — same package, five spellings:

```
?reload=1      → 200 in  2.68ms
?reload=yes    → 200 in  1.15ms
?reload=TRUE   → 200 in  0.81ms
?reload=false  → 200 in  0.50ms
?reload=true   → 200 in 51.43ms   ← the recompile
```

That 20–60× gap is the whole bug. Fixed at all four sites through one `parseReloadParam` — including the legacy `/projects/...` twins, which had the same `=== "true"` reading and would otherwise have kept it. Only the two spellings the OpenAPI boolean declares are accepted; anything else, including a repeated `?reload=true&reload=1`, gets a 400 quoting the value and naming a form that works. I checked `@malloydata/db-publisher` never sends `reload` at all, so no legacy client depended on the lenient reading. Verified after the change that `reload=true` still genuinely recompiles (new source visible) and `reload=1` now 400s without touching the package.

## The description was stale, and its excuse was also false

You're right that `reload_guard.integration.spec.ts` was already there. Worse than stale, though: the stated reason — "`server.ts` calls `.listen()` at module scope so its `app` can't be imported into a spec" — was never true. `tests/harness/rest_e2e.ts` predates this branch, does exactly `const { app } = await import("../../src/server")`, and ~10 integration specs already use it. I've dropped the claim rather than just refreshing it, and the counts are current: `1794 pass, 0 fail` unit, `236 pass` integration.

While in there I also split the value contract into `query_param_utils.spec.ts` — that file is deliberately import-free so it can be unit-tested without booting the server, which is the testability the old note claimed didn't exist.

## Two specifics in the relayout point I'd push back on

Both are minor, but they'd mislead the next reader:

- **"subscriptions… gained a second `nest:`, so it is the case least like the other two."** It's the reverse. `storefront` and `governed-analytics` already had two `nest:` blocks before this PR — the diff only adds `columns=2` and `# colspan=2` there. `subscriptions` was the odd one out with a single flat list, and the split moves it *toward* its siblings. It's also not load-bearing: I probed split-nest against single-nest-with-bare-refs and `# colspan=2` binds to `kpis` alone in both, clean logs in both.
- **"all three end up with three tiles… leaving a hole in the last row."** Two of three. `storefront` has four tiles under the spanned strip and fills cleanly (2+2); `governed-analytics` and `subscriptions` have three and leave the gap. The description now says which is which.

## `unexpected 'from'` — settled in the branch's favour

Fair thing to double-check given agents pattern-match the string. Both forms through the real compiler:

```
join_one: ts is from(team_stats)          → [error] unexpected 'from'
source: x is from(team_stats) extend {…}  → [error] unexpected 'from'
```

`FROM` being in the parser's usable-as-identifier list doesn't change the emitted message, so the skills' quoted text is correct and I've left it.

## The rest

- **`[]` invariant** — good catch, folded into the description: "Any match includes the source's own row, so an empty array means no such source." The 2000-char budget had 20 characters spare, so this is phrased tightly; the protocol spec still passes with 3 to spare.
- **Rebased** onto `main`, so the pin is `0.0.427`.
- **Splitting the skills half** — happy to if you'd rather; agent-skills#77 already mirrors it. Now that the code half has grown a second behavioural change I lean the other way, since the render-tag fix is what validates the example relayouts and the two read better together. Your call.

One thing I found and deliberately did **not** bundle: three integration specs (`readiness_line`, `freshness_gate`, `mcp_get_context_semantic`) each pass a timeout as `beforeAll`'s second argument, which this bun version rejects, so all three error out instead of running. Pre-existing on `main` and unrelated to this PR — the description names them, and it wants its own one-line PR.

jswir:
Follow-up: I said the three erroring integration specs wanted their own PR — fixed here instead (`9282ad5f`), since leaving nine tests silently not running while asking for a review of the suite made little sense.

`readiness_line`, `freshness_gate` and `mcp_get_context_semantic` each passed a timeout as `beforeAll`'s second argument. Bun rejects it, and the throw is at file scope, so each file errored out before executing a single test — the errors then read as unrelated harness noise rather than "this file did not run".

Each keeps its intended budget via `setDefaultTimeout()`, which applies to hooks as well as tests. I checked both directions rather than assuming: a 7s `beforeAll` fails under Bun's 5s default and passes with it set. I preferred that over deleting the argument and relying on the suite's `--timeout 200000`, because that flag only comes from the `test:integration` script and these specs need the budget when run individually too.

All nine pass, and their test counts (1, 4, 4) match the `it` counts, so nothing was hidden behind the error. Integration is `245 pass`, up from 236.

The only red left is `concurrent_package`, which pulls a package from `gs://publisher_test_packages/` and fails without GCS access locally; its teardown leaves a dead server, which is the single trailing unhandled error. Both reproduce unchanged on clean `main`, and that job sets no credentials, so I left it alone rather than guess at the intended gating.

mlennie (approved):

PR #959, opened 2026-08-04
fix(server): declare engines and refuse to boot on unsupported Node

mlennie:
## What this fixes

A user on Node 18 gets no signal at any point where they would expect one. `npm install` succeeds, the server boots healthy with `load_errors=0` and `PUBLISHER_READY`, MCP discovery works, `malloy_compile` works, and only the first real query fails, with `500 crypto is not defined`. The error names neither Node nor a version, and its own suggestions ("try the request again later", "call malloy_reloadPackage, then retry") cannot help. Reproduced on published 0.0.234 with one variable changed: Node 18.20.8 gives the 500, Node 24.15.0 returns correct results.

`packages/server/package.json` is the published package and declared no `engines` field at all, so npm had nothing to enforce, even though the repo root and every `@malloydata/*` dependency already declare `node >=20`. The requirement was known; it just never travelled with the published artifact.

## What changed

Three parts, because `engines` on its own only labels the problem: npm treats a mismatch as `npm WARN EBADENGINE` and installs anyway unless the user has set `engine-strict=true`.

1. `"engines": { "node": ">=20" }` on the published package, with a spec that pins it to the version the process enforces, so it cannot be removed or drift without failing the unit suite.
2. A boot-time check that writes an actionable message and exits non-zero before anything binds. It names the required version and the version actually running, which is the entire point: the failure it replaces named neither.
3. A `## Requirements` section in the README, where a first-time user meets it. `docs/development.md` already stated the floor, but only in the contributor Prerequisites table.

## Decisions worth reviewing

**Hard exit rather than a warning.** The failure being replaced was a silent success: a server that booted healthy, reported no load errors, served MCP discovery, and only died on the first query. A warning would scroll past in that same healthy-looking boot log, and a half-working server is exactly what made the original defect expensive to diagnose.

**Bun is exempt, and this is load-bearing.** The Docker image runs the built bundle under `bun run`, as does `start:dev`. Bun reports a Node-compatible `process.version`, so it cannot be told apart from Node by version string alone, and it provides the global Web Crypto API regardless of the version it reports. A naive `process.version` check would refuse to boot the image.

**The scaffolder warns rather than refusing**, and records the floor in the manifest it generates. The scaffolder works correctly on an older Node and the workspace it writes is valid and complete; only running Publisher fails, and the server's own check catches that. Its warning deliberately does not promise that `npm start` will refuse to boot, because the pinned `SERVER_VERSION` is 0.0.234 and predates this check, so on that version the failure is still a 500 on the first query.

**`engine-strict=true` was considered and rejected** for the generated workspace. It only covers the `npm create` path, and it would change the semantics of every future `npm install` the user runs in their own project directory, so one sloppy transitive `engines` range would hard-fail work unrelated to Publisher.

**The version check fails open on an unparseable version string.** Refusing to boot a working runtime because its version had an unexpected shape would be a worse failure than the one this prevents.

## The second commit: the call site, traced

`service/query_metadata.ts` called `crypto.randomUUID()` in `mintCorrelationId()` without importing `crypto`, so it resolved against `globalThis.crypto`, which does not exist before Node 19. A repo-wide sweep confirms it was the only such reference under `packages/*/src`; every other user imports the builtin, which is present on every supported Node. Correlation ids are minted per query execution and never on the compile path, which is precisely why `malloy_compile` succeeded while the first query failed.

This is not an attempt to support Node 18. The dependencies require `>=20` and the boot check exits before this line is reachable. The pattern is wrong regardless, since relying on a global that only appears in newer runtimes is easy to copy, and `node:crypto`'s `randomUUID()` is the same RFC 4122 v4 generator, so behaviour is unchanged.

## How it was verified

The original failure was reproduced first, on the published artifact rather than a local build: `npx @malloy-publisher/server@0.0.234` under Node 18.20.8, booting to `serving` with `load_errors=0` and then returning `500 {"code":500,"message":"crypto is not defined"}` on a real REST query.

Then on the built bundle from this branch, all three runtimes:

| runtime | result |
| --- | --- |
| Node 18.20.8 | exits 1, no listener bound on the port, full message intact through a piped stderr |
| Node 24.15.0 | reaches `serving`, real query returns correct rows |
| Bun 1.3.13 (the Docker path) | reaches `serving` |

The scaffolder was run under both Node 18 and Node 24 from its built `dist`: the warning prints last on screen under 18, is absent under 24, and the generated manifest carries `engines` in both.

Full root gate green after `generate-api-types` and `build:sdk`: typecheck clean, lint clean, `prettier:check` failing only on the six git-ignored generated files that CI never sees (each confirmed with `git check-ignore`), server unit 1797 pass / 0 fail and integration 232 pass / 0 fail run as separate legs, create-malloy-package 370 / 0, skills 7 / 0, hammer 20 / 0.

Three defects were found and fixed during verification, all introduced by this change: the Bun exemption initially swallowed its own tests (the seam defaulted an explicitly-passed `undefined` back to Bun's real version, and the tests run under Bun); `process.stderr.write` followed by `process.exit` can drop the message on a piped stderr, which would recreate the silent failure this exists to remove, so the write is now `fs.writeSync` with the stream write as fallback; and the message templated the crypto explanation onto the detected version, which would have been a false claim for Node 19, since that release shipped the global unflagged but is still below the floor.

## Sequencing

This is intended for 0.0.235, landing before or with #947. #947 removes the `aws-sdk` v2 dependency whose deprecation warning is currently the only clue to the Node version anywhere in the boot log, so if that lands first the last accidental signal disappears before the deliberate one arrives. This change is independent of #947 and needs nothing from it.

## Follow-up, not in this PR

The scaffolder's pinned `SERVER_VERSION` is 0.0.234, which predates this check, so a scaffolded workspace still fails at the first query until 0.0.235 ships and the pin moves. Note also that `--help` on an unsupported Node now prints the version error rather than the help text, since the check runs before argument parsing; that seemed right for a binary that cannot run, but it is easy to reorder if reviewers prefer otherwise.


jswir:
Reviewed against the branch; new suites pass (14 in `node_version_check.spec.ts`, 7 in `node_version.spec.ts`).

**The boot check is justified, and worth saying why explicitly**, because "just declare `engines`" is the obvious objection. Confirmed `engine-strict` defaults to `false` and there is no root `.npmrc` overriding it, so npm warns and installs anyway. But the decisive point is that `engines` is an *install-time* check for a *run-time* problem: an nvm/mise user installs on 20 and later switches their shell to 18, and `engines` cannot fire again. Docker (`bun run`) and `bun run start` from a clone never consult it at all. So `engines` covers one path, with a warning. Something has to check at boot.

Three things below, then a note on weight.

### 1. Two definitions of the floor, not pinned to each other

`MIN_NODE_MAJOR = 20` is declared twice — `packages/server/src/node_version_check.ts:16` and `packages/create-malloy-package/src/node_version.ts:14` — plus four `"node": ">=20"` literals. Each spec pins its constant to *its own* `package.json`; nothing pins the server's to the scaffolder's, so the two can silently diverge. That is the drift this PR exists to prevent, reintroduced one level up.

`create-malloy-package` already depends on `@malloy-publisher/skills` (`workspace:*`), so there is an existing shared home for the constant with everything else derived from it.

### 2. The import-order comment claims more than it delivers

`server.ts:85` reads "The first statement this module runs… before anything else happens," and the import comment says it is "evaluated ahead of the rest of the graph." ESM evaluates the entire import graph before the importing module's body, and `build.ts` bundles with Bun, which inlines the entry last — so the call runs *after* every dependency's top-level code.

I checked whether that matters: with `globalThis.crypto` deleted, `@malloydata/malloy`, `@modelcontextprotocol/sdk`, `express`, `@malloydata/db-duckdb` and `winston` all import cleanly. So the guard does win the race today — by luck rather than construction. Either soften the comment to what is true ("before any argument parsing, storage init, or listener"), or make it true with a small entry stub that checks and then `await import("./server_main.js")`.

### 3. Two fixes for one defect; which is load-bearing?

`query_metadata.ts` now imports `node:crypto`, which *removes* the crash — `crypto.randomUUID()` via module import works back to Node 14.17. The boot check refuses to start below 20. If the missing global was the whole bug, Publisher plausibly runs on 18 after this PR and the guard now refuses a working configuration; if 20 is genuinely required for other reasons, the import is hygiene and the guard is the fix. The comments assert both framings and settle neither — `node_version_check.ts` still cites the crypto global as the mechanism it guards against, after the same PR removed that dependency.

The policy framing looks right to me (every `@malloydata/*` declares `>=20`, so supporting 18 means supporting a configuration the dependencies disclaim) — but then it should be stated as policy, and the import described as hygiene.

### On weight

The boot check is ~25 lines of logic inside 513 lines of diff. Most of the rest is the duplication in (1), 174 lines of comment around `node_version_check.ts`, and 14 tests for a major-version integer comparison. Per the repo's own comment-altitude convention, the V8/JSC, `writeSync`/EAGAIN and mise-PATH derivations read as commit-message material. The scaffolder's separate warning fires seconds before the server's own refusal in the same terminal; if it stays, it can be a few lines reading the shared constant rather than a module with its own constant and spec.

Also: `packages/create-malloy-package` already declared `engines: >=20` on main, so the new pin in `node_version.spec.ts` passes with no `package.json` change — worth a line in the description, since a reader will look for that edit and not find it. And `packages/cli` (`@malloydata/publisher-cli`, published) still has no `engines`; it has no `@malloydata/*` or crypto usage so it likely does work on 18, which makes it worth a deliberate call rather than an omission.

mlennie:
Thanks, this was a genuinely useful read, and the framing point in (3) was the one that needed settling rather than patching. All four are addressed in `ae2702bf`, plus `946a52aa` which fixes a miscount in my own new comments. Taking them in your order.

Also worth flagging on the commit list: the branch now carries a second signed merge of `main`, which is where #947's conflict in `packages/server/package.json` got resolved. Both keys are there, `files` then `engines`.

## 1. One floor, and it is now pinned rather than repeated

You are right that the drift this PR exists to prevent was reintroduced one level up. Fixed, but not by moving the constant into `@malloy-publisher/skills`, and the reason is your own point (2): `packages/server` does not depend on `skills`, so importing the constant from there would give the boot guard a third-party module graph. The guard's one structural property is that it pulls in nothing, and I would rather not trade that away for an integer.

Instead the declarations are now a closed chain of assertions. The "engines contract" block in `node_version_check.spec.ts` reads **four** manifests from disk (server, repo root, scaffolder, and now the CLI) and asserts each `engines.node` equals `REQUIRED_NODE_RANGE`, which is derived from `MIN_NODE_MAJOR`. The scaffolder's own spec still pins its manifest to its constant. So every declaration is asserted against the same value, and any of them drifting fails the unit suite. Reading from disk rather than importing is deliberate: it keeps the packages independent at build and run time, which a shared import would not.

One correction on the count, in the direction of your point rather than against it: there were **three** `"node": ">=20"` manifest literals, not four (server, root, scaffolder). The other two occurrences are prose in `README.md:63` and `AGENTS.md:40` quoting the error token, and `REQUIRED_NODE_RANGE` is derived, not declared. With the CLI added it is four, all four now pinned.

If you would still rather have the single shared constant, say so and I will make the swap; it is a small change and I do not feel strongly beyond the import-graph argument above.

## 2. The comment claimed more than it delivered, and now says what is true

Softened, and I verified your correction rather than taking it on trust. Two measurements:

- In the built bundle, the guard is invoked at line **292,446 of 293,520**. It is the last thing the bundler inlines, exactly as you said.
- Running the real bundle on Node 18.20.8, OpenTelemetry's `SDK initialized` line prints **before** the refusal. So dependency top-level code demonstrably runs first.

The comment now says the check imports nothing but `node:fs` and so adds no application code to the graph, that it does **not** run before that graph, and that what it does precede is everything this process chooses to do. The second comment at the call site already used that framing, so the two now agree.

I chose (a) over the entry stub, for two reasons. The weight objection at the bottom of your review cuts against adding a module and a bundler entrypoint. And once the floor is stated as policy (see 3), "before any argument parsing, storage init, or listener" is the property that matters, not "before dependency evaluation."

While checking this I also confirmed the claim in `README.md:63` and `AGENTS.md:40` that the server exits "before it binds anything" still holds: the Prometheus exporter is constructed with `preventServerStart: true`, so nothing in the instrumentation preload opens a listener before the guard runs.

## 3. It is policy, and the import is hygiene

You are right that the comments asserted both framings and settled neither. Settled as policy, in three places:

- `node_version_check.ts` now opens by saying the floor is a support policy and not a proven technical minimum: every `@malloydata/*` dependency declares `>=20`, as does the repo root, so running below 20 means running on a configuration Publisher's own dependencies disclaim and nothing here tests.
- The operator-facing message no longer names a mechanism. It said older runtimes lack the global Web Crypto API, which after this same PR describes something that no longer happens. It now states the dependency floor.
- `query_metadata.ts` describes the `node:crypto` import as hygiene: `randomUUID` from the module works on every Node the project supports, so reaching for the global raised that line's floor above the module's for no benefit. It is still recorded as *how* the problem surfaced, since the query-path-but-not-compile-path detail is the reason the failure looked like it did.

Your read is the correct one and it is now the only framing in the file. A new test pins it: the message must not contain "crypto", so the mechanism cannot creep back into the text.

That comment also had a factual error your point exposes. It said the global "does not exist before Node 20"; it arrived unflagged in Node 19. Corrected.

## 4. Weight

- `node_version_check.ts`: 174 to 154 lines, comments 80 to 61. The mise-PATH derivation is down to a clause, the `writeSync`/EAGAIN rationale to two lines, and the longer versions move to the commit message as you suggested.
- Its spec: **14 tests to 8**, while covering strictly more (four manifests instead of one). The version-comparison cases are one table-driven test rather than six, and the Node 19 test is gone because the message no longer makes a claim that needs it.
- The scaffolder: module 56 to 46 lines, spec **7 tests to 4**, now a few lines around the shared floor rather than its own explanation of why `engines` is insufficient.

One thing I could not find: there is no V8/JSC derivation in this PR. `git diff origin/main..HEAD | grep -iE 'v8|jsc|javascriptcore'` comes back empty. If you meant the Bun exemption's "the Node version Bun reports says nothing about what Bun provides", that one is staying: `docker_smoke_test` and `bun-direct test` both run the bundle under Bun, so it is the sentence that stops someone deleting the exemption and breaking the image.

I kept the scaffolder warning rather than deleting it. It fires before a workspace exists, where the server's refusal cannot: someone scaffolds on Node 18, gets a correct workspace, and would otherwise discover the problem only at `npm start`. Its own `engines` covers install time only, and with `engine-strict` false that is a warning too.

## The CLI, and a correction worth knowing

`packages/cli` now declares `engines: { node: ">=20" }`, as a deliberate call rather than for consistency. Stating the reasoning since you asked for it: its dependencies bottom out at Node `>=18` (commander 12 is the binding one) and it has no `@malloydata/*` or crypto usage, so it very likely does work on 18. The floor is a support statement, not a technical minimum, and one supported runtime for the product beats a per-package runtime matrix. It gets the manifest pin, not a boot check: there is no failure mode there that a hard exit would improve.

The correction: I could not find anything in the repo that publishes it. `release.yml` publishes `packages/sdk`, `packages/app` and `packages/server`, plus skills and create-malloy-package on their own dispatches; `packages/cli` appears in no workflow. It has no `private: true` either, so a manual publish is possible, but if `@malloydata/publisher-cli` is on npm it is not getting there from CI. Worth chasing separately, since an unpublished-but-publishable package is its own small hazard.

## The description note

Agreed, and it goes in the description: `packages/create-malloy-package` already declared `engines: >=20` on `main`, so its spec's pin passes with no manifest change, and a reader does go looking for that edit.

## Verified

Full gate green on the merged tree: typecheck, lint, unit 1856, integration 237, and 7 + 367 + 20 across skills, scaffolder and hammer, all 0 fail, `bun run test` exit 0. Prettier's usual six warnings are the gitignored generated clients.

Manual, on the built bundle rather than in tests:

| check | result |
|---|---|
| Node 18.20.8, direct | `PUBLISHER_UNSUPPORTED_NODE required=>=20 detected=v18.20.8`, exit 1, no listener |
| Node 18.20.8, stderr redirected to a file | message survives, so `writeSync` still does its job |
| Node 24.15.0 | boots to `PUBLISHER_READY`, and a real query through the worker pool returns 25,356 |
| Bun on the same bundle | boots, so the Docker path is unaffected |
| Scaffolder on Node 18 | warns, and still writes a correct workspace |

mlennie:
One correction to my note about the description: I am leaving the description as it stood when you reviewed it, so this lives in the thread instead.

For anyone who goes looking for the manifest edit and does not find it: `packages/create-malloy-package` already declared `engines: ">=20"` on `main`, so the new pin in `node_version.spec.ts` passes with no change to that manifest. The CLI is the only manifest this PR adds the field to.

jswir (approved):

PR #960, opened 2026-08-04
feat(server): add malloy_searchDatabaseSchema for warehouse discovery

mlennie:
Publisher already introspects ten dialects and serves that over REST, but an agent pointed at a database it has not modelled yet had no way to ask what is in it in plain English. This adds the ranking layer between introspection and that question, plus an MCP tool over it, so a first run can start from a warehouse rather than from a local file.

## What it does

`malloy_searchDatabaseSchema` drills down one level at a time, the same shape as `malloy_getContext`. No arguments lists the environments and their connections, a connection lists its schemas, a schema lists its tables, and a schema plus a `searchQuery` ranks those tables by relevance. Every table comes back with its columns and the `source:` line to start a model from, which is the seam between finding a table and modelling it: the response carries the exact syntax rather than leaving the agent to assemble a connection name, a dotted path, and per-dialect quoting.

## Ranking, and the egress decision

Ranking is lexical (lunr/BM25) by default. No API key, no network, works offline.

Semantic ranking needs `EMBEDDING_API_KEY` and, deliberately, a second switch: `EMBEDDING_INDEX_CONNECTION_SCHEMA`. That is the part I would most like a second opinion on. The existing key authorises embedding your own model text, which is already on your disk. A customer's table and column names are a different disclosure, and turning on semantic `malloy_getContext` should not quietly start sending them to a third party. It is the same principle `getEmbeddingConfig` already applies when it refuses to infer the key from an ambient `OPENAI_API_KEY`.

In every mode the tool reads names and types only. It never reads a row, so no data can leave, only identifiers. Sample-value indexing is an explicit non-goal, the analogue of `get_context`'s `dimensional_value` one; agents that need real values use `execute_query` with `select: distinct`.

The index holds one entity per table, carrying its column names, rather than one entity per column, which is what keeps a large warehouse inside the entity cap while still letting a column term find its table. Vectors are in memory only, LRU-capped, keyed per environment, connection and schema, and invalidated by a fingerprint over the tables and their columns.

## The separable fix in the first commit

While building this I found that Publisher's schema introspection passed no row limit, so each query inherited whatever its driver defaults to. Those defaults are not uniform and are not documented together anywhere: DuckDB, BigQuery and Trino cap at ten rows, Postgres has a 1000-row default it does not apply, and Snowflake and MySQL have none.

On the first three that returned a wrong schema rather than an error, because one row here is one column. Measured against DuckDB, a database with eight tables and thirty-seven columns reported three tables, the third missing most of its columns, which is exactly ten rows of `information_schema.columns`. This affects the REST endpoints that list a connection's schemas and tables, not only the new tool. On Snowflake and MySQL the cap it adds is a new upper bound where there was none, which is the safer direction.

The cloud path needed the same fix and is easy to miss: `listTablesForDuckDB` delegates an S3 or GCS schema to `gcs_s3_utils`, whose `DESCRIBE` returns one row per column and was capped the same way, so a twenty-five column Parquet reported ten columns.

It is a separate commit so it can be split into its own PR if you would rather land it independently.

## Evaluation, and what this does not claim

`schema_search_eval.ts` measures retrieval rather than assuming it, because the last retrieval feature did not. It builds its own warehouse-shaped fixture, since the bundled example packages address their data as files and register no tables to search.

On that fixture the lexical baseline scores 6/10 at recall@3, and 0/4 on the gap cases: queries that share no words with their target table, like "website visits" for `web_session_events` or "newsletter blasts" for `marketing_email_campaigns`. That gap is the entire argument for the semantic mode.

The semantic arm has not been run, because it needs a key and sending schema text to a provider is a decision for whoever deploys it. So nothing here claims semantic ranking is better yet, and it ships off by default partly for that reason.

## Two devex fixes that came out of running it

The per-package `duckdb` sandbox is synthesized on demand by `ConnectionController` and appears in no config listing, so the drill-down originally dead-ended at "no connections" on exactly the setup the bundled examples ship with. It is now advertised with `scope: "package"` and the packages it applies to.

An empty schema also explains itself now. A DuckDB connection over CSV or Parquet addresses files by path and registers no tables, so `information_schema` is legitimately empty while the data is right there; a bare empty list reads as "no data" and sends an agent down the wrong path.

## Verification

Typecheck, lint and `prettier:check` are clean, and 1828 unit tests pass. New coverage is 45 unit tests plus an MCP integration spec, including a test asserting connection credentials never appear in any tier of the response.

Beyond the gate I ran the tool live over MCP against a real server and a real schema, which is how the truncation bug and both devex fixes were found. None of the three would have failed a test.

The change also went through three independent review passes (adversarial correctness, security, and claim-versus-code). Those raised 21 findings and 13 were real, including a worked example in the tool description that returned nothing against the environment it named, and the cloud-path row cap above. Both are fixed here.

One known-flaky note for reviewers: the `concurrent_package` integration suite downloads `gs://publisher_test_packages/gcs_faa.zip` on every test and can exceed the 120s package-load worker timeout on a loaded or bandwidth-constrained machine. It passes on a quiet one, and the package-load path does not import either file changed in the first commit.

## Docs

`docs/configuration.md` gets the new variable and a section on exactly what the semantic mode sends where. `README.md`, `AGENTS.md` and `docs/ai-agents.md` cover the new tool. `skills/README.md` moves `search_database_schema` out of the "platform tools shared skills must not reference" line, since Publisher now has one, and adds it to the bare-name to `malloy_*` mapping.


kylenesbit (approved):
Approving. This is well-built, and the design decisions I'd have pushed back on are already made correctly: identifier positions are rejected rather than escaped, `canPasteSource` is conservative by construction, vectors are memory-only and LRU-bounded, and there's a test asserting connection credentials never appear in any response tier. I checked the dialect classification against the source of truth — `BACKSLASH_ESCAPE_DIALECTS ∪ DOUBLED_ESCAPE_DIALECTS` exactly covers the ten values in the `Connection.type` enum in `api-doc.yaml`, so the "unclassified dialect" throw really is unreachable as the comment claims.

**On the egress question you asked for a second opinion on:** the second switch is right, keep it. `EMBEDDING_API_KEY` authorizes embedding the operator's own model text; a customer's table and column names belong to a different party, and someone who enabled semantic `getContext` last month did not consent to this. The part that makes it defensible rather than merely polite is that the vectors are deliberately not persisted to the DuckDB store the package index uses.

Four things I'd like fixed. None of them block the feature, but the first is the only one that reaches beyond this tool.

---

### 1. `BadRequestError` reclassification is broader than the PR describes

`packages/server/src/mcp/handler_utils.ts` (the new `if (error instanceof BadRequestError)` branch)

`classifyToolError` is shared by `executeQuery`, `compile`, `reloadPackage`, `getContext` and the new tool, and `BadRequestError` is this codebase's general wrapper for query-time failures, not just for argument errors. `model.ts` throws it for `Invalid query: …`, `Query execution failed: …`, filter-validation errors, `Invalid query request. (Query AND !sourceName)…`, and — the one that actually matters — `Model compilation failed: …` (the fallback for compilation errors that are not `ModelCompilationError`). That last one is a genuine Malloy problem, and it now gets "change your arguments rather than retrying" instead of the syntax guidance and doc links.

For most of the others the new advice is honestly better than telling someone to check their `source:` keywords, so I am not asking for a revert. But it should be a deliberate, pinned decision rather than a side effect of a schema-introspection fix. The cleanest version is a narrower error class for the new introspection argument errors, leaving `BadRequestError` on the Malloy branch.

Either way, this test needs attention:

```
packages/server/src/mcp/handler_utils.spec.ts
it("keeps Malloy advice for an authorize denial and a malformed request", ...)
```

Its title now asserts the opposite of the behaviour, and its assertion (`not.toContain("unexpected internal error")`) passes either way, which is why nothing caught the change. Whichever direction you take it, rename it and make it assert the advice it actually expects.

### 2. `bigquery` is classified wrong in `BACKSLASH_ESCAPE_DIALECTS`

`packages/server/src/service/introspection_sql.ts`

`sqlLiteral` applies quote-doubling unconditionally and layers backslash escaping on top for the backslash dialects, and the doc comment says doubling "is valid on the backslash dialects as well." GoogleSQL does not accept `''` as an escaped quote — it is backslash-only — so `'it''s'` would not parse.

No bug ships today, because both `getSchemasForBigQuery` and `listTablesForBigQuery` go through the `@google-cloud/bigquery` client and never build SQL. I am raising it because the stated premise of taking the connection type is that a call site "cannot express the wrong answer," and the loud-failure guard specifically will not fire for BigQuery precisely because it *is* classified. Dropping it from the set would make a future BigQuery SQL path throw rather than silently mis-escape, which is the failure direction the rest of this file is built around.

### 3. Two remaining raw catalog interpolations

`packages/server/src/service/db_utils.ts` — `getSchemasForTrino` (both the configured-catalog and the `SHOW CATALOGS` loop) and `getSchemasForDatabricks` (same two spots) still interpolate `${catalog}` raw into `FROM ${catalog}.information_schema.schemata`.

Config- and warehouse-derived, so not a vulnerability, but it is the same shape you fixed for the MotherDuck and Snowflake schema listings in `perf(server): cache the lexical index…`, and it is the last instance of it.

Related, and worth a follow-up rather than this PR: hyphenated Trino catalog names are common (`hive-prod`), and `assertSafeSqlIdentifier` now hard-rejects them on the `listTables` path with a message that offers no repair — the same "no repair, invites a retry loop" problem the charset-message commit fixed. The comment's reasoning (a bare hyphenated name fails anyway) is correct, so this makes a pre-existing limitation explicit rather than regressing anything. Quoting identifiers per dialect is the real fix whenever someone gets to it.

### 4. `packageName` is reported twice in the "Ignored" warning

`packages/server/src/mcp/tools/search_database_schema_tool.ts`

Call with `connectionName` and `packageName` but no `environmentName`: the tier-1 branch adds `packageName` to `ignored`, then the unconditional `if (connectionName && connectionName !== SANDBOX_CONNECTION)` check below adds it again, so the warning reads "Ignored connectionName, packageName, …, packageName".

Cosmetic on its own. I mention it because nothing in `search_database_schema_tool.spec.ts` exercises the ignored-argument warning at all, and given how much of this tool's value is in what it tells an agent when something is off, that path is worth a test or two.

---

Two smaller notes, take or leave:

- Tier 1 fans `listPackages()` across every environment with no load shedding, and `listPackages` calls `getPackage` per package. Tiers 3–5 call `assertCanAdmit` first; tiers 1 and 2 do not. It matches what the existing REST packages endpoint does, so the risk is low, but tier 1 is the call every agent makes first and it fans out widest.
- Tier 5's match is case-sensitive (`bareTableName(t.resource) === tableName`), so asking for `orders` against `Orders.parquet` returns not-found on exactly the dialects that ignore the table filter — the branch that check exists to protect.

And I would take you up on your own offer to split the first commit. The introspection row cap is an independent bug affecting existing REST endpoints, and the history shows the dialect-escape classification was revised in three consecutive rounds. Landing it alone with `introspection_sql.ts` and `db_utils.spec.ts` would give it scrutiny proportional to its blast radius instead of arriving as commit 1 of 16.

PR #961, opened 2026-08-04
fix(server): return 413 instead of a bare 500 when a result cannot be serialized

mlennie:
## The problem

The byte cap measures a query response by stringifying it, so the measurement is the thing that fails first. Just under the threshold you get a genuinely useful 413 that names the cap, the measured size and three remedies. One step larger, `JSON.stringify` throws inside the guard before it can compare anything to the cap, and the error that escapes names neither.

Against `examples/storefront`, with `run: order_items -> { select: big is repeat(id::string, 20000); limit: 100000 }`:

| | before | after |
|---|---|---|
| `repeat(..., 2000)` | `413 Query response exceeded 50000000 bytes (was 306326576). Project fewer columns, add a LIMIT, or raise PUBLISHER_MAX_RESPONSE_BYTES.` | unchanged, byte for byte |
| `repeat(..., 20000)` | `500 {"code":500,"message":"Out of memory"}` | `413 Query response exceeded 50000000 bytes: the 25356-row result is too large to serialize. Project fewer columns, add a LIMIT, or filter wide values.` |
| the same as a notebook cell | `400 Cell execution failed: Out of memory` | the same 413 |

The notebook-cell path had the same defect with a worse status: a 400 says the request was malformed, so it blamed the caller's cell for a server-side allocation failure. Both sites now go through one helper.

## Why it matches on the error class and not the message

Both engines Publisher runs on report this as a `RangeError`, with different messages. V8 under node says "Invalid string length" at around 512 MB; JSC under bun, which runs the Docker image, says "Out of memory" and only at several GB (a direct probe serialized a 1.2 GB string without complaint). Matching the message would have fixed one and left the other broken.

A stack overflow is also a `RangeError` and is not a size problem, so it still propagates as the 500 it is. A non-`RangeError`, such as an unserializable BigInt, keeps its own error rather than being reported as too large. Both are covered by tests.

## Why it does not offer to raise the cap

There is no cap at which a response that will not serialize starts serializing, so telling the caller to raise `PUBLISHER_MAX_RESPONSE_BYTES` is advice to change a setting and hit the identical failure. That is also why this throws a `ResponseUnserializableError` subclass: REST still answers 413 by inheritance, pinned by a test, while `classifyToolError` drops its generic "raise the configured cap" suggestion for this case, so an agent over MCP is not sent the wrong way either.

## On estimating the size instead

Estimating rows times width before building the payload was the obvious alternative and I decided against it. Any estimate accurate enough not to fire where the exact measurement succeeds has to sit within a hair of the engine's own string limit, and that limit is engine-dependent, so the guard would be guessing. Firing early would also replace the good message's exact byte count with a lower bound, which is a regression in the one message here that already worked. Measured on one build, the floor costs nothing: 3.07s to reach the old 500, 3.15s to reach the new 413.

One boundary is deliberate and worth knowing: with the byte cap disabled (`PUBLISHER_MAX_RESPONSE_BYTES=0`) the query path never stringifies inside the guard, so that configuration keeps its 500. The notebook path is covered either way, because there the string is the payload rather than the measurement.

## Verification

Reproduced live before the change and confirmed after, against `examples/storefront`: the just-below case returns its exact byte count and remedies unchanged, both larger cases return 413 over REST and over MCP, and the notebook cell returns 413. Unit tests cover both engines' messages, the stack-overflow and BigInt cases that must not convert, the disabled-cap wording, and the 413 mapping of the new subclass. Reverting the conversion fails four of them.

Gate: typecheck, lint, prettier, 1794 unit tests and 232 integration tests all pass. No `api-doc.yaml` change: 413 was already documented on both endpoints, so this makes the implementation match the published contract.


jswir:
Reviewed against the branch and ran the new suites (45 pass across `model_limits` / `errors` / `handler_utils`). The mechanism is right. One gap worth fixing before merge, two accuracy items, three minors.

**Verified the portability claim on the real engines**, since the spec necessarily simulates it by throwing a hand-built `RangeError` from `toJSON`:

| Engine | `JSON.stringify` overflow | Class |
| --- | --- | --- |
| V8 (node v20.19.6) | ~512 MB → `Invalid string length` | `RangeError` ✓ |
| JSC (bun 1.2.23) | OK at 2.10 GB, threw at 2.50 GB → `Out of memory` | `RangeError` ✓ |

Matching on the class rather than the message holds, and JSC throws catchably rather than being OS-killed.

### 1. The guard is skipped when the byte cap is disabled

`model.ts:2584` only calls `stringifyQueryResponse` when `maxBytes > 0`, but `0` is a documented supported value — `config.ts`: *"A value of `0` is accepted and disables the byte cap entirely; use it only when you intend to rely on the row cap alone (e.g. for benchmarking)."*

With `PUBLISHER_MAX_RESPONSE_BYTES=0` the model-query path skips the guard entirely, and the payload's real serialization then happens unguarded at `query.controller.ts:134` — so the bare 500 this PR removes is back, on all three surfaces (REST, notebook, and MCP all route through `getQueryResults`). The notebook path has no such hole; it calls unconditionally.

A suggestion that also removes work: return the serialized string from `getQueryResults` and have the controller send it, the way the notebook path now does. Today the string is built for measurement, discarded, and then built again at `query.controller.ts:134` — twice per large response. That gives one stringify, one guard, and no `maxBytes > 0` special case to keep in sync.

### 2. `exceeded ${maxBytes} bytes` states something unmeasured

`model_limits.ts:182`. V8 throws at ~512 MB regardless of the cap, so if an operator raises the cap past that — which is what `PayloadTooLargeError`'s own MCP advice tells them to do ("raise the configured cap") — the message claims the response exceeded e.g. 1 GB when it actually died at 512 MB. `model_limits.spec.ts` pins that wording. Something like `Query response could not be serialized (byte cap: N)` is true in both cases.

### 3. The JSC threshold is ~4× the documented pod budget

`model_limits.ts` notes JSC "says 'Out of memory' and only much later". Measured, "much later" is 2.1–2.5 GB, and `constants.ts` gives the sizing target as *"a typical 1-2 GB memory budget"* — so a container at that limit is OOM-killed by the kernel before JSC ever throws. On node the fix is load-bearing: at the default 50 MB cap a 600 MB result does hit `Invalid string length` inside the guard. Worth naming the real threshold so the Docker runtime doesn't read as covered.

### Minor

- `PayloadTooLargeError` never sets `this.name`, so a `ResponseUnserializableError` logs as `Error`. The subclass exists to be told apart from its parent, and in logs it isn't. `errors.ts:140`/`160`/`176` already do this for three other classes — one line in the otherwise no-op constructor.
- `recordQueryCapExceeded("bytes", source)` now counts "could not serialize" as a bytes-cap breach, which per (2) may not have exceeded any cap. Anything alerting on that counter now mixes two distinct failures.
- The measurement serializes `wrappedResult` with plain `JSON.stringify`, while a `compactJson` request sends `compactResult` with `bigIntReplacer` (`query.controller.ts:133`). Pre-existing, but the new *"leaves a non-RangeError alone"* test now pins a 500 for a BigInt on a request whose actual payload would have serialized. Returning the real string (1) collapses this too.

Three things that look deletable and aren't, for whoever tidies this later: the `/call stack/i` exclusion (deep nesting is a `RangeError` and is not a size problem), the subclass staying a subclass so REST keeps 413 by inheritance, and the `instanceof PayloadTooLargeError` rethrow in `model.ts`'s catch — that last one is all that stands between this and the 400 the notebook path used to return.

mlennie:
Thanks, this was a good catch list. All five taken; the sixth declined with a reason. Pushed as `fc24c1cc`, plus a signed merge of main (it had moved 4 commits: #937, #947, #950, #965, none overlapping this diff).

**1. The cap-disabled hole. You are right, and I had written it off as a documented boundary, which was the wrong call.** Documenting it was not enough precisely because the payload gets serialized again downstream, so the failure moved rather than stopping. Measured both ways with `PUBLISHER_MAX_RESPONSE_BYTES=0` and the same query:

```
before: 500 {"code":500,"message":"Out of memory"}
after:  413 Query response could not be serialized: the 25356-row result is too large to turn into JSON. Project fewer columns, add a LIMIT, or filter wide values.
```

Serializing is now unconditional and only the measurement is conditional, since with no cap there is nothing to compare against. Note the message drops the cap entirely in that configuration rather than reporting a cap of zero bytes.

**Your suggested refactor is in, in the end.** I first decided against it as too broad for a review fix, then re-reading my own change found the decline had a cost in exactly the configuration the fix was for: with the cap disabled the guard built the full JSON and discarded it, so a `PUBLISHER_MAX_RESPONSE_BYTES=0` setup, which the config docs suggest for benchmarking, paid a whole extra serialization per query. `getQueryResults` now returns the string it already built and the controller sends that, which also removes the double pass on the default path, where every large response was serialized once to measure and once to send. It is additive: one field on the return type, one line in the controller. The compact form keeps its own pass since it is a different object and needs the replacer, so the mismatch you noted is narrowed rather than gone.

**2. The message stating something unmeasured. Agreed and fixed.** It now reads `could not be serialized: the N-row result is too large to turn into JSON (byte cap: N)`. The cap is context rather than the thing exceeded, which is true whether the engine limit or the cap is what fired.

**3. The JSC threshold. Agreed, and thank you for measuring it.** My comment said "only much later", which read as though bun were covered. It now names your 2.1 to 2.5 GB range and says where the guard is and is not load-bearing: on node it fires well inside a normal pod, and under bun the threshold sits above the 1 to 2 GB budget `constants.ts` sizes for, so the kernel is liable to kill the container before JSC throws anything catchable. The row and byte caps are what protect that configuration.

**Minors.** `name` is now set on both classes, so a `ResponseUnserializableError` stops logging as a bare `Error`. I set it as an explicit literal rather than deriving it from `new.target`, since the bundler could mangle class names. The counter gets a distinct `unserializable` label instead of counting as `bytes`, for exactly the reason you give: it may not have exceeded any cap.

On your closing note about the three things that look deletable: agreed, and the `/call stack/i` exclusion, the subclass staying a subclass, and the `instanceof PayloadTooLargeError` rethrow are each commented as load-bearing now, so the next person has the reason rather than having to rediscover it.

jswir (approved):

PR #962, opened 2026-08-04
fix(server): stop answering a wrong URL with the app shell and HTTP 200

mlennie:
## The problem

The SPA catch-all returned `index.html` for anything unmatched, so a request for a file got the app shell with HTTP 200, and a 200 carrying HTML is indistinguishable from a working page.

A package's data app is served at `/environments/<env>/packages/<pkg>/<file>`, while the web UI opens a model at the shorter `/<env>/<pkg>/<file>`, so the short form is an easy guess for a page too. Guessing it returned success and then displayed "Unrecognized file type: index.html", blaming the file for what was a wrong path.

| URL | before | after |
|---|---|---|
| `/examples/html-data-app/index.html` | 200, title `Malloy Publisher`, 801 bytes (the shell) | 302 to the canonical URL, which serves the same 13157 bytes and title as the correct URL |
| `/api/v0/bogus-endpoint` | 200 plus the shell | 404 as JSON |
| `/examples/html-data-app/nope.html` | 200 plus the shell | 404 naming the URL form to use |
| every app route, including `.malloy` and `.malloynb` | 200 plus the shell | unchanged, asserted |

That confusion has now caught three separate readers, including one who had read both earlier write-ups, so this decides by the shape of the request rather than case by case: a path that names a file is answered as a file request, and everything else still gets the shell.

## Deciding by extension, not by "does it have a dot"

The classifier uses a closed list of extensions the web actually serves. A dot test would have 404'd an environment or package whose name contains one, so `/examples/my.pkg` stays an app route, `.malloy` and `.malloynb` stay the model and notebook viewers, and `pages/<file>.html` stays the in-app page viewer that the package view links to. Over-reach here means a page of the product stops loading, so the spec asserts those cases first and they outnumber the ones this change exists to fix. Writing that spec is what caught a draft where an app-owned `pages/` segment fell through to a 404 instead of the shell.

The redirect is a 302 rather than a permanent one, so a path the app may later claim as a route of its own is not cached against it in every browser that guessed once. It carries the request's query string: an embedded page arrives with `?embed_token=...`, and sending it to the right path without its query would swap one broken result for another. The query rebuild that the canonicalizing package-root redirect already did is now a shared helper and both use it, with that redirect's existing integration test pinning it unchanged.

## Two smaller changes that come with it

`ModelPage`'s fallback now names the path it was asked for and links to the URL that serves package files, instead of naming the file's type. It is still reachable for an extension the server does not divert, and in development, where Vite serves the shell rather than this catch-all.

The static route's 404 was an empty body, which left a reader who guessed wrong on a blank page at the end of the redirect. It now carries a short page describing the URL form. It reflects no request input, as that route's 404s deliberately do not.

## Verification

Verified live on a production build, since the catch-all only exists outside development mode: both URLs compared on status, title and byte count, the missing-file and `/public/` guesses now 404 with an explanation, and every app route above confirmed still serving the shell. Adversarial probes confirm a protocol-relative path is refused a redirect, and that percent-encoded traversal is redirected and then rejected by the static route with `400 Resolved path is outside of root`, which is the route that owns that boundary. The reflected path stays percent-encoded and is escaped on top of that.

New Playwright spec covers the redirect, the query string, the JSON 404, the routes that must not divert, and the app's message. Gate: typecheck, lint, prettier, 1803 unit tests and 232 integration tests pass, and the full Playwright suite passes once CI's stub bigquery connection is injected the way `app-playwright.yml` does. `docs/html-data-apps.md` gains two sentences where the URL form is introduced.


jswir:
Built and ran this rather than reviewing it on paper: SDK → app → server bundle, started with `--init`, all three example packages loaded with zero load errors. **All 8 of the spec's own Playwright tests pass**, including the two browser-driven ones. The core fix works — `/examples/html-data-app/index.html` → 302 → 200 serving `<title>SaaS Subscriptions</title>`, byte-identical to the canonical URL. No regressions: `/`, `/examples`, `/examples/storefront`, `.malloy`, `.malloynb`, `pages/index.html` and nested `pages/sub/x.html` all still 200 the app shell.

Two findings, one design note, and some minors.

### 1. The `/public/` test comment is false

`spa_fallback.spec.ts:86` says *"The static route strips it, so the redirect target is what serves it."* It does not strip it. Measured against the running server:

```
/examples/html-data-app/index.html         → 302 → 200   ✓
/examples/html-data-app/public/index.html  → 302 → 404
```

`serveFromPackage` sets `publicRoot = <pkg>/public` and then joins the request path onto it, so `public/index.html` resolves to `<pkg>/public/public/index.html` → ENOENT → 404. There is no `public/` stripping anywhere in the server.

The *behavior* is still an improvement — the 404 body explains that `<file>` is relative to `public/` and excludes it. But the test name ("which was the near-miss") and the comment both claim the redirect reaches the file, and the test only asserts the classification, so it passes while the comment misleads. Either strip a leading `public/` when building the target, or correct the comment and rename the test to what it actually pins.

### 2. The redirect over-reaches, and lands in a reflecting error path

The redirect treats *any* 3+ segment path ending in a listed extension as `<env>/<pkg>/<file>`. Live:

```
/assets/foo/bar.js  → 302 /environments/assets/packages/foo/bar.js
                    → 404 {"code":404,"message":"Environment \"assets\" could not be resolved to a path."}
/assets/bar.js      → 404 (the explanatory HTML)
```

Two things fall out. The 3-segment case gets a worse answer than the 2-segment one — JSON naming an internal resolution failure instead of the HTML that explains the URL forms. And it reflects the user-controlled segment (`"assets"`) into the body, which is exactly what `serveFromPackage`'s own 404s are commented as avoiding ("no reflected request input"); the redirect routes around that policy. Not a security issue — it is JSON-escaped — but the two adjacent policies now disagree.

Not live today, since Vite emits two-segment asset paths. Worth bounding the guess (e.g. skip the redirect unless the environment resolves) or at least noting the tradeoff.

### 3. Design note: a third copy of the package-asset URL

`ModelPage.tsx` now hand-builds `${serverBaseUrl(server)}/environments/${enc(env)}/packages/${enc(pkg)}/${path}`. `PageViewer.tsx:51-53` already builds that string character-for-character, in the same package. The PR exports `serverBaseUrl` from the SDK on the grounds that "this is the one derivation of it" — while adding a second copy of the larger thing it is part of, and `spa_fallback.ts` builds the path form a third time. The export that actually removes duplication is the URL builder, not the origin helper.

### Minor

- **`${req.method}` is dead interpolation.** `server.ts:1939` can only ever render `GET`, since the branch lives in `app.get("*")`. Confirmed live: GET → `application/json`, POST/PUT/PATCH/DELETE → Express's default `text/html` 404. To be clear this is *not* a regression — the old defect was the 200-plus-shell, which only ever affected GET; non-GET already 404'd. But the interpolation implies method-generality the handler does not have, and it will mislead the next reader.
- **While in here:** the de facto error shape is `{code, message}` (`errors.ts:50`, and the SDK types it as exactly that in `useQueryWithApiError.ts`), and this PR emits it correctly. But `api-doc.yaml:4156`'s `Error` schema declares `{message, details}` — no `code`, and `details` is set nowhere in the server. Only 13 of 208 documented 4xx/5xx responses reference the schema at all. Separate cleanup, not this PR's, but this PR is the first thing to hand-write an error body and it is worth knowing which contract it is conforming to.
- `ASSET_EXTENSIONS` misses `yaml`/`yml`, `xml`, `md`, `wasm`, `webmanifest`; confirmed live that `report.yaml`, `notes.md`, `data.xml` and `mod.wasm` all still get the shell. Graceful, and the new app-side page covers it. But the Playwright test at `report.yaml` *depends* on `.yaml` being absent from the list — worth a comment there so nobody "completes" it and breaks that test.
- `PACKAGE_FILE_NOT_FOUND_HTML` is documented as "deliberately static… reflects no request input" while the new `assetNotFound` body reflects the path. Escaping is correct (verified: a raw `<script>` in the path comes back as `&lt;script&gt;`), but the two comments now tell opposite stories — pick one.
- The whole server-side change is inside `if (!isDevelopment)`, so in dev unknown API and asset paths still get the Vite shell. A test comment notes it; worth stating in the description too.

### Verified, so nobody needs to re-check

No open redirect (`location` is built from segments and always starts with `/environments/`, so it cannot go protocol-relative or off-origin); `.`/`..` excluded from the redirect branch; query string preserved (`?embed_token=abc&x=1` survives); no redirect loop (`/environments/x/notpackages/pkg/file.html` terminates in one hop → 404); `SPA_OWNED_SEGMENTS` covers every in-app `.html` link, `Package.tsx:256` being the only place one is constructed, nested paths included.

mlennie:
Thanks for building and running it rather than reading it. Both of your findings were real defects rather than comment nits, and one of them I had measured myself and still shipped the wrong explanation for. Pushed as `0c416770` and `723e62fe`, plus a signed merge of main. All 14 checks are green.

**1. The `/public/` comment was false, and worse than false.** You are right that the static route does not strip that segment: it joins onto `<pkg>/public`, so passing it through resolved to `public/public/index.html`. What makes this the worse kind of wrong is that the test only asserted the classification, so it passed with an incorrect reason attached, and I had seen the 404 in my own probe and recorded it as expected.

I took the first of your two options, because the segment is a reasonable guess: the redirect now drops a leading `public/`, so the guess reaches the file. Verified end to end rather than in the classifier alone, which is what let the wrong explanation survive:

```
/examples/html-data-app/public/index.html
before: 302 -> .../packages/html-data-app/public/index.html -> 404
after:  302 -> .../packages/html-data-app/index.html        -> 200 <title>SaaS Subscriptions</title>
```

A package that genuinely ships `public/public/<file>` is still reachable by asking the static route for it directly, and there is a test pinning that only a leading `public/` with something after it is dropped.

**2. The redirect over-reaching into a reflecting error path. Agreed on both halves.** The classifier can only see the shape of a path, and `/assets/foo/bar.js` has the same shape as `/<env>/<pkg>/<file>`. It now returns a redirect as a *candidate*, and the handler follows it only if the environment is one the server has loaded; otherwise you get the page that explains the URL forms. So the three-segment case no longer gets a worse answer than the two-segment one, and nothing echoes a user-controlled segment into a body where the adjacent route deliberately does not. Live:

```
/assets/foo/bar.js         ->  404 text/html, the explanatory page, no "could not be resolved"
/nosuchenv/pkg/index.html  ->  404 the same page, where before it redirected into the JSON error
```

You are right that it was not live today given Vite's two-segment output. It was still the wrong shape.

**And that check needed a second pass, which is worth flagging since it is a fourth commit on this PR.** My first version of it called `getEnvironment`, and CodeQL failed the build: `Missing rate limiting`, a route handler doing a file system access. It was right, and about my new line rather than pre-existing code. `getEnvironment` resolves from storage, and this is the last handler before the app shell answering unauthenticated traffic for any unclaimed path, so a caller could trigger storage work by guessing URLs. `723e62fe` uses the in-memory list of loaded environments instead, which is what the check actually needs: no I/O, the handler is synchronous again, and an environment that failed to load now gets the explanatory page too, which is the right answer given its files are not servable either. CodeQL is back to no new alerts.

**3. Your design note is correct and I had it backwards.** The export that removes duplication is the URL builder, not the origin helper. `packageFileUrl` now lives next to `serverBaseUrl` and is the one place that string is built; `PageViewer` and `ModelPage` both call it, and the `serverBaseUrl` export is withdrawn. The server-side path form in `spa_fallback.ts` stays separate, since it cannot import from the SDK, but it is now the only other copy rather than the third.

**Minors, all taken.** The `${req.method}` interpolation is gone, with a comment recording your point that the original defect was GET-only because other verbs already 404 through Express. The extension list now says why `yaml`, `yml`, `xml`, `md`, `wasm` and `webmanifest` are absent and that a Playwright test depends on `yaml` staying absent, so nobody completes it and breaks that test. The two comments about reflected input are reconciled: the static route has a resolved environment and package, so echoing the path would confirm which of them exists, while the fallback has resolved nothing and the path is all it knows. And the dev-mode limitation is now in the description as well as the test comment.

On the `api-doc.yaml` `Error` schema declaring `{message, details}` while the server emits `{code, message}`: agreed that this PR conforms to the de facto shape and that the spec is the thing that is wrong. Also agreed it is not this PR's to fix. Worth its own issue given only 13 of 208 documented error responses reference the schema at all.

jswir (approved):

PR #965, opened 2026-08-05
docs: add a security policy

Sha-Bang:
Private vulnerability reporting is on for this repo, but nothing pointed at it and nothing
told a reporter what's in scope. This adds the written half.

## What it does

- New root `SECURITY.md`. GitHub picks it up from the root (matching the existing
  `CONTRIBUTING.md`) and surfaces it on the Security tab and in the new-issue flow.
  Supported versions, the private reporting form as the only channel, when public is fine,
  scope, what to expect, and where to look for hardening.
- Links it from the three places someone actually looks: the README's Community section
  (right beside "Report issues on GitHub", which is where you land with the wrong instinct),
  the docs index, and `CONTRIBUTING.md`.

The scope section is the part that's specific to Publisher. The server is unauthenticated by
design, and `givens` / `#(authorize)` / row-level access are documented as caller-asserted
conventions rather than boundaries. Without saying so out loud, the first thing a security
policy attracts is a stream of "the REST API has no auth!" reports. So there are two lists:
what's working as documented, and what breaks a boundary Publisher does claim — a direct
query succeeding against a source the package never exported, an `#(authorize)` gate granting
what its expression should deny, credentials in a response or log, escaping a package's
static root, RCE via package load or compile, a reachable dependency vuln.

Each scope item links the doc that already documents the behavior, so the policy stays a
pointer rather than a second source of truth that can drift.

Reporting is GitHub's private advisory form only — no security@ alias to maintain.

## Where the line sits

Drawing it took two passes. The first draft got both scope bullets subtly wrong in opposite
directions, so the second commit fixes them against the docs they defer to:

- **Overstated.** "Reaches or returns rows from a source the package never exported" would
  have called a documented feature a vulnerability — an unexported source "still compiles,
  imports, joins, and extends" (`docs/discovery-and-access.md`), and `governed-analytics`
  ships a model that does exactly that. Only a *direct* query 404s, so that's the boundary.
- **Understated.** "Fails open when the given *is* supplied" correctly excludes the
  caller-asserted non-boundary, but it also waived a gate that passes when no given is sent
  at all — which `docs/authorize.md` documents as always denying. Now it reads "for the
  givens actually sent, including none."

Also added authorize's § Known limitations to the caller-asserted bullet. It documents an
ungated `/compile` raw-SQL path as a tracked follow-up, which is a real gray zone that's
better named than left for someone to discover and report.

## One deviation worth flagging

`## Hardening` deliberately does **not** link `docs/deployment.md`, though that's the obvious
target. That doc has zero security or auth content — no host binding, loopback, or gateway
anywhere in its 147 lines — so pointing an operator there dead-ends the exact reader the
section exists for. It points at the README posture instead.

## Tested

Docs only, no code touched, so no build or test run applies (CI's prettier check covers
`packages/**/*.{ts,tsx}` and there's no markdownlint).

- All 9 relative links in `SECURITY.md` resolve on disk.
- Every section anchor the policy leans on exists: `README.md#point-your-agent-at-it`,
  `docs/authorize.md#security-model`, `docs/authorize.md#known-limitations`,
  `docs/html-data-apps.md#security-model`.
- Every scope claim re-read against the doc it cites — that's what caught the two above.
- `https://github.com/malloydata/publisher/security/advisories/new` returns 200, so the
  reporting link isn't a dead end.
- Tables padded by hand rather than running prettier, which would have churned unrelated
  README content.

Worth a look at the rendered links in the preview, and confirming the Security tab shows the
policy as defined after merge.

## One thing to confirm

The response times are a proposal, not something I can commit on the maintainers' behalf:
acknowledge within 5 business days, in-scope assessment plus rough timeline within 10. Happy
to change or drop them.

## Not in this PR

Fixing `docs/deployment.md` itself — see the deviation above. The trust-model prose lives in
`docs/authorize.md`, where a self-hoster won't find it. Worth its own PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,12 @@ information on using pull requests.
 This project follows
 [The Linux Foundation's Code of Conduct](https://lfprojects.org/policies/code-of-conduct/).
 
+## Security
+
+Report a security vulnerability privately rather than as a public issue.
+[SECURITY.md](SECURITY.md) has the reporting form, what's in scope, and when filing in the open is
+fine.
+
 ## Contributing to the Python SDK (`packages/python-client`)
 
 The Python SDK is **auto-generated** from `api-doc.yaml` using OpenAPI Generator plus a thin build script.
--- a/README.md
+++ b/README.md
@@ -258,3 +258,5 @@ process and sign-off are in [CONTRIBUTING.md](CONTRIBUTING.md).
 
 - Join the [Malloy Slack](https://join.slack.com/t/malloy-community/shared_invite/zt-1kgfwgi5g-CrsdaRqs81QY67QW0~t_uw)
 - Report issues on [GitHub](https://github.com/malloydata/publisher/issues)
+- Report a security vulnerability privately — see [SECURITY.md](SECURITY.md) for the reporting form
+  and what's in scope
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,84 @@
+# Security policy
+
+## Supported versions
+
+Publisher releases continuously off `main`; there are no maintenance branches. Fixes land in the next
+release, and the ask is that you upgrade rather than expect a backport.
+
+| Version             | Supported |
+| ------------------- | --------- |
+| Latest release      | ✅         |
+| Any earlier release | ❌         |
+
+## Reporting a vulnerability
+
+> Report privately: **[Report a vulnerability](https://github.com/malloydata/publisher/security/advisories/new)**
+> (repo → Security → Advisories → Report a vulnerability).
+
+What to include, as much as you have:
+
+- The affected version or commit.
+- How Publisher was configured and reached — flags, host binding, gateway or none.
+- Reproduction steps, or a proof of concept.
+- The impact you believe it has.
+
+## Private or public?
+
+- If you're unsure, report privately. We'd rather triage a non-issue in private than have a real one
+  land in public.
+- Maintainers may move a report into a public issue once it's clear it isn't sensitive.
+- Anything that exposes data or credentials, or executes code across a boundary Publisher claims,
+  goes private — always.
+- An issue that only weakens defense in depth behind the documented trusted tier is reasonable to
+  file in the open.
+
+## Scope
+
+Publisher is unauthenticated by design, and several of its governance features are documented as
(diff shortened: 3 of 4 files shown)


sagarswamirao (commented):
Checked this out and re-verified the scope claims against the docs they defer to. The
calibration is right, and both fixes in the second commit land where they should:

- The unexported-source bullet had to be narrowed to a *direct* query — an unexported source
  "still compiles, imports, joins, and extends" (`docs/discovery-and-access.md`), so the
  original phrasing would have classified a legitimate join as a vulnerability.
- "for the givens actually sent, including none" is the correct boundary. The last two entries
  in `docs/authorize.md § Known limitations` document that an unsupplied given always denies, so
  a gate passing on zero givens genuinely is a break.

Confirmed independently: `PUBLISHER_HOST` defaults to `0.0.0.0` (`packages/server/src/server.ts`,
matching `docs/configuration.md`), and the loopback-then-gateway posture is stated verbatim in
both `README.md` and `docs/ai-agents.md` — so the "supported posture" sentence cites an existing
position rather than inventing one. All relative links resolve and all four section anchors exist.

Agreed on leaving `docs/deployment.md` out of `## Hardening`. It has no host-binding, auth, or
gateway content, and its only credential mentions are "no credentials required" for the local
samples — which would actively mislead the reader that section is for.

Two things:

**`## Hardening` could use `docs/configuration.md`.** The section takes an operator who can be
reached and hands them trust models, but the two concrete knobs are `PUBLISHER_HOST=127.0.0.1`
and `PUBLISHER_FRAME_ANCESTORS`, both in `docs/configuration.md`. Both are already cited upward
in the "working as documented" list as the mitigations for those two non-vulnerabilities, so a
reader landing here from the Security tab gets the posture and then has to scroll back into the
scope list for the flag. Adding that one link closes it — same reasoning that (correctly) kept
`deployment.md` out, one notch milder.

**Anchor nit.** `README.md#point-your-agent-at-it` is in the right section, but the posture
blockquote sits well below the heading, past the `.mcp.json` discussion and the REST fallback.
The link works; the reader just lands above the three sentences that are the point. Not worth
changing now — but if the README ever grows an explicit posture subheading, that's the better
target for both references.

On the response times: 5 / 10 business days reads as reasonable for a repo this size. Since
they're the one commitment here a reporter can hold you to, they need a maintainer's yes rather
than a reviewer's.

sagarswamirao (approved):

PR #966, opened 2026-08-05
fix(server): treat sourceName/queryName as names, not Malloy text

### LINKED ISSUE #964: Query boundary: a newline in `queryName` reaches a non-exported source and returns its rows
## Summary

The query boundary (the "queryable == discoverable" rule that makes a non-exported source a 404) can be stepped around by putting a newline in the `queryName` request field. The request reaches a source the package never exported and **returns its rows**.

Found while reviewing #963, which touches the adjacent authorize gate. This is **not** an `#(authorize)` bypass — see Scope below. It is pre-existing on `main` (reproduced at `31fbc36f`), not introduced by that PR.

## Reproduction

Package with `explores: ["index.malloy"]` (so the boundary is active in its default `declared` mode), where `customers` is exported and `helper` / `secret` are not:

```malloy
// index.malloy
// hidden and ungated
source: helper is duckdb.sql("select 1 as id") extend {
  measure: c is count()
  view: hv is { aggregate: c }
}
// hidden AND gated - used by the variant table below
#(authorize) "false"
source: secret is duckdb.sql("select 42 as z") extend {
  measure: sc is count()
  view: sv is { aggregate: sc }
}
// the only exported source
source: customers is duckdb.sql("select 1 as id, 100 as amt") extend {
  measure: total is amt.sum()
  view: v is { aggregate: total }
}
export { customers }
```

Naming the hidden source directly is correctly refused:

```jsonc
{ "sourceName": "helper", "queryName": "hv" }   // -> NotQueryableError (404), as intended
```

Naming the *exported* source and smuggling a statement through `queryName` is not:

```jsonc
{ "sourceName": "customers", "queryName": "v\nrun: helper -> { aggregate: c }" }
// -> 200, [{"c":1}]
```

Reproduced with `Model.getQueryResults("customers", "v\nrun: helper -> { aggregate: c }", undefined)` against the harness in `packages/server/src/service/query_boundary.spec.ts`.

## Root cause

Two behaviours compose:

1. **The early boundary check clears on `sourceName` alone.** In `assertQueryBoundaryEarly` (`packages/server/src/service/model.ts:1640-1651`), the named-query branch admits the request when the *source* is curated, without checking that `queryName` names a view on it:

   ```ts
   if (sourceName && curatedSources.has(sourceName)) return "cleared";
   ```

2. **`queryName` is interpolated verbatim into generated Malloy**, so a newline creates a second top-level statement:

   ```ts
   queryString = `\nrun: ${sourceName ? sourceName + "->" : ""}${queryName}`;
   ```

   yielding

   ```malloy
   run: customers->v
   run: helper -> { aggregate: c }
   ```

   and Malloy executes the **last** `run:` — which the suite already documents ("multi-statement is settled by the COMPILED run target (last statement wins)").

3. **A `cleared` admission deliberately skips the compiled backstop.** `assertQueryBoundaryCompiled` only runs for `deferred`, because an exported named query may legitimately read hidden sources internally and re-deriving its target must not re-deny it. That exemption is reasonable on its own; combined with (1) and (2) it means nothing ever re-checks the statement that actually ran.

Each piece is defensible in isolation. The gap is that `sourceName` and `queryName` are treated as identifiers by the boundary logic and as free text by the string builder.

## Scope — what this does and does not reach

- **Does:** execute and return rows from a source outside the package's exported surface. The boundary is the *what* axis (identity-free, 404).
- **Does not:** bypass `#(authorize)`. **Verified:** with `#(authorize) "false"` on the hidden source, the same injection returns `AccessDeniedError` (403), not rows. The compiled authorize gate runs unconditionally after compilation and resolves the real run target. Exposure is limited to sources that are hidden **and** ungated.
- Publisher's documented deployment posture is that the query/MCP API sits behind a trusted tier and is network-isolated from untrusted callers (`docs/authorize.md`, Security model). This is a defense-in-depth boundary rather than the last line, which is why it is being reported in the open rather than held.

## Suggested directions

Not submitting a patch with this report, since the right fix depends on a decision the maintainers should make:

1. **Validate the fields as identifiers before interpolating.** Cheapest and most direct, but it needs a call on what is legal — quoted/hyphenated source names are a supported feature (there are tests for them), so a naive `\w+` rule would break real models.
2. **Do not clear on `sourceName` for a named-view request unless `queryName` resolves to a view on that source.** Keeps the exported-named-query exemption intact while closing the case where the two fields disagree.
3. **Build the generated statement from quoted/escaped identifiers** rather than raw concatenation, so a field that is not an identifier cannot become syntax. Complements either of the above.

`sourceName` is interpolated the same way but is **not** exploitable — see the variant table below — so hardening it is defence-in-depth rather than a fix.

## Which variants actually reproduce

All run against the fixture above.

| Variant | Request | Result |
| --- | --- | --- |
| Baseline (the bug) | `sourceName: "customers"`, `queryName: "v\nrun: helper -> { aggregate: c }"` | **executes**, returns `[{"c":1}]` |
| Hidden **and gated** target | same shape, targeting `secret` | `AccessDeniedError` (403) — authorize still fires |
| Injection via `sourceName` | `sourceName: "customers->v\nrun: helper -> …"` | `NotQueryableError` — blocked |
| Injection via `sourceName`, with `queryName` | `sourceName: "customers\nrun: helper -> …"`, `queryName: "v"` | `NotQueryableError` — blocked |
| `queryName` alone, no `sourceName` | `queryName: "v\nrun: helper -> …"` | `NotQueryableError` — blocked |
| Equivalent as ad-hoc `query` text | `query: "run: customers -> …\nrun: helper -> …"` | `NotQueryableError` — blocked |

Two things follow, and they narrow the fix considerably:

- **Only the `sourceName` + `queryName` combination is affected** — the branch at `model.ts:1649`. The `queryName`-alone branch above it tests membership of the *whole* string against the curated query set, so a newline makes the lookup miss and it correctly 404s. `sourceName` is safe for the same reason: the curated-source lookup is an exact-string match, so an injected `sourceName` never matches and never clears.
- **The ad-hoc `query` path is not affected.** It returns `deferred`, and `assertQueryBoundaryCompiled` settles it against the compiled run target — the existing "last statement wins" test covers exactly this. The gap is specific to the `cleared` admission that skips that backstop.

## Note for whoever picks this up

PR #963 (open, unrelated fix in the authorize gate) modifies `extractRunTargetSourceName` in `packages/server/src/service/query_text.ts` — the `run:` branch no longer requires a trailing `->`, so it now also resolves run targets like `run: name extend { … } -> { … }`. `assertQueryBoundaryEarly` uses that helper for its **ad-hoc text** branch, so the boundary's early positive-deny gets slightly wider there. It does **not** touch or fix this issue, which goes through the named-parameter branch and never calls that helper. Mentioned only so the two changes are not mistaken for each other.

## Environment

- `main` @ `31fbc36f`, and the PR branch for #963 (which does not change this logic)
- All variants above reproduced against the fixtures in `packages/server/src/service/query_boundary.spec.ts` (`writeManifest` + `Package.create`)
- `@malloydata/*` 0.0.427

Sha-Bang:
> **Series:** stacked on #963 — this targets `nathan/fix-caller-forged-authorize-override`, not `main`. Review/merge that one first; the diff here is just the commit on top.

Fixes #964.

## What it does

`sourceName` and `queryName` were treated as identifiers by the gate logic and as free text by the string builder. A newline in either opens a second top-level statement, and Malloy runs the **last** `run:` — so this returned rows from `helper`, a source the package never exported:

```json
{ "sourceName": "customers", "queryName": "v\nrun: helper -> { aggregate: c }" }
```

Two gates key off the raw caller strings by exact match, and they fail in **opposite directions** when that match misses:

- the query boundary's curated-source lookup misses and 404s — but a `cleared` admission deliberately skips the compiled backstop, so on the `sourceName` + `queryName` shape nothing ever re-checked the statement that actually ran;
- `getFilters()` misses and returns **no** filters, so the smuggled statement reads its target completely unfiltered. That lane needs no `explores` at all, which makes it the default deployment shape, and it's reachable through either field.

The fix quotes both fields as Malloy identifiers at the single place caller strings become syntax, so a name can only ever lex as one identifier. `quoteMalloyIdentifier` already existed for the render-tag validator.

The compiled `#(authorize)` gate was never bypassed — with `#(authorize) "false"` the same injection already returned 403.

## Accepted behavior changes

1. **Malloy text in these fields now 400s** as an unresolved or unlexable name instead of being compiled. That's the fix. It also changes the *gated* variant: hidden + gated goes 403 → 400, because the request no longer compiles at all.
2. **Four shapes that happen to work today stop working:** a parameterized source (`customers(p is 1)`), a dotted view path (`carriers.by_name`), a view refinement (`by_carrier + { limit: 10 }`), and an inline source extension. No producer in the repo uses any of them — the SDK's query builder sends `.toMalloy()` text in `query` and explicitly nulls both name fields. The residual risk is an MCP agent improvising, so the doc and tool-description updates are part of the fix, not a nicety.
3. **Pre-backticked names are not unwrapped.** A caller sending `` "`gated-source`" `` now gets a 404 instead of today's accidental success. Deliberate: `sourceName` is *also* a lookup key in the curated-source, filter and authorize gates, all of which expect the bare name — so that "workaround" already skipped filter injection.

New capability from the same change: a name that *requires* Malloy quoting (hyphen, space, reserved word, leading digit) is queryable on the named path for the first time.

### Tested

Wrote each test first and confirmed it failed *in the wrong way* before touching `model.ts`:

- **`query_boundary.spec.ts`** — the reported repro returned 1 row from the unexported `helper`, both under `explores` and with none declared. Adds the reporter's verified variant table (injection through each field, `queryName` alone, the ad-hoc equivalent) as stays-blocked guards, plus a test that a hyphenated name is queryable by its bare name.
- **`filter_bypass.spec.ts`** — both vectors returned **4 rows across all three tenants**, unfiltered. This is the test that *proves* the filter lane rather than assuming it.
- **`restricted_mode.spec.ts`** — widened the named-path assertion: a disallowed construct in these fields is now refused while resolving the name, so restricted mode becomes the second line of defence rather than the first.

Full `packages/server` sweep **1898 pass / 0 fail**. `bun run test:integration` 233 pass with one pre-existing failure (`concurrent_package`, needs GCS credentials — verified failing identically on the base commit). eslint clean, and `grep 'sourceName + "->"'` finds nothing.

End-to-end against a running server on a package declaring `explores`, over REST **and** again through `malloy_executeQuery`:

| request | before | now |
| --- | --- | --- |
| the reported injection | `200 [{"c":1}]` | `400` |
| injection via `sourceName` | — | `404` |
| `helper`/`hv` named honestly | `404` | `404` |
| legitimate `customers`/`v` | rows | `200 [{"total":100}]` |

### On the tool description

`malloy_executeQuery`'s description was sitting at 1998 of the 2000-char truncation budget `server.protocol.spec.ts` enforces, leaving no room for the new contract rule. Rather than raise the budget — that cap belongs to the client, not the model, so a newer model on our end isn't a reason to move it — I applied the template's own "prefer merging over adding": the row-cap rule was stated three times over, and two clauses were rationale written for a human reader. Now 1926 chars with the *full* rule in and no fact dropped.

`malloy_compile` is at 1964 with the same latent squeeze. Left alone here to keep this diff to the fix.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3295,15 +3295,28 @@ components:
             set, the queryName parameter must be empty.
         sourceName:
           type: string
-          description:
+          description: |
             Name of the source in the model to use for queryName, search, and
             topValue requests.
+
+            A NAME, not Malloy code: a single source name exactly as the model
+            response's `sources` listing returns it. The server quotes it as a
+            Malloy identifier, so a name that needs quoting (a hyphen, a space,
+            a reserved word, a leading digit) works as-is and must be sent bare
+            — do not add backticks of your own. Anything richer than one name
+            (a parameterized source, an inline extension, a second statement)
+            belongs in `query`, and here is refused as an unresolvable name.
         queryName:
           type: string
-          description:
+          description: |
             Name of a query to execute on a source in the model. Requires the
             sourceName parameter is set. If the queryName parameter is set, the
             query parameter must be empty.
+
+            A NAME, not Malloy code, on the same terms as `sourceName`: one
+            view name as the model response lists it, quoted for you, sent
+            bare. A dotted path, a refinement (`by_carrier + { limit: 10 }`),
+            or any additional statement belongs in `query`.
         compactJson:
           type: boolean
           default: false
--- a/docs/api-overview.md
+++ b/docs/api-overview.md
@@ -63,10 +63,14 @@ curl -s -X POST \
 
 The query body takes one of two shapes: `query` alone (ad-hoc Malloy, compiled in the model's
 context), or `queryName` without `query` (a named view when `sourceName` is set; a model-level
-named query when it is not). Any other combination returns a 400. The response's `result` field is
-a JSON string, so parse it; with `"compactJson": true` it holds plain row objects, without it the
-full Malloy result envelope with type metadata. `givens` rides on either shape to supply
-model-declared [runtime parameters](givens.md).
+named query when it is not). Any other combination returns a 400. On the named shape, `sourceName`
+and `queryName` are names and not Malloy code: send one name per field, bare, exactly as the model
+response's `sources` listing returns it — the server quotes it, so a name needing quotes works and
+backticks of your own do not. Anything richer than a name goes in `query`.
+
+The response's `result` field is a JSON string, so parse it; with `"compactJson": true` it holds
+plain row objects, without it the full Malloy result envelope with type metadata. `givens` rides on
+either shape to supply model-declared [runtime parameters](givens.md).
 
 ## Live API explorer
 
--- a/packages/create-malloy-package/src/names.ts
+++ b/packages/create-malloy-package/src/names.ts
@@ -246,9 +246,12 @@ function describeUnsafeCharacter(name: string): string {
  *   with "Cannot redefine '<word>'". A parse-only sweep does not see these, which
  *   is why they were missing for so long.
  *
- * A derived source name landing on either is suffixed (below) rather than quoted,
- * so the source stays queryable by a plain name through the REST/MCP query API,
- * which composes `run: <source> -> <view>` without quoting.
+ * A derived source name landing on either is suffixed (below) rather than quoted.
+ * The query API no longer requires that — it quotes `sourceName`/`queryName` as
+ * identifiers when it composes `run: <source> -> <view>`, so a name needing
+ * quotes is queryable there — but a quote-requiring name is still hostile to
+ * every place Malloy is written by hand (ad-hoc `query` text, notebooks, the
+ * model's own views), so the generator keeps producing plain names.
  *
  * This list is derived from the installed @malloydata/malloy rather than curated
  * by hand, and names.spec.ts re-derives both sets from that package on every run.
--- a/packages/server/src/mcp/tools/execute_query_tool.ts
+++ b/packages/server/src/mcp/tools/execute_query_tool.ts
@@ -49,8 +49,18 @@ const executeQueryShape = {
       ),
    modelPath: z.string().describe("Path to the .malloy model file"),
    query: z.string().optional().describe("Ad-hoc Malloy query code"),
-   sourceName: z.string().optional().describe("Source name for a view"),
-   queryName: z.string().optional().describe("Named query or view"),
+   sourceName: z
+      .string()
+      .optional()
+      .describe(
+         "Source name for a view. A NAME, not Malloy code: one name exactly as malloy_getContext returned it, sent bare (the server quotes it, so a hyphen or a reserved word is fine — do not add backticks yourself). Anything richer, such as a parameterized source or an inline extension, goes in query.",
+      ),
+   queryName: z
+      .string()
+      .optional()
+      .describe(
+         "Named query or view. A NAME, not Malloy code, on the same terms as sourceName: one view name as malloy_getContext returned it. A dotted path (carriers.by_name), a refinement (by_carrier + { limit: 10 }), or anything containing a newline goes in query instead.",
+      ),
    filterParams: z
       .record(z.union([z.string(), z.array(z.string())]))
       .optional()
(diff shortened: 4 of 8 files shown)


sagarswamirao:
Approving. I verified the vulnerability and the fix independently rather than from the description.

## The fix is load-bearing

I reverted only the quoting call back to `sourceName + "->"` and re-ran the suites. Both leaks reproduce:

```
{ sourceName: "metrics", queryName: "by_kind\nrun: products -> by_org" }
  -> 4 rows across orgs [acme, globex, initech]   (filter bypass)
{ sourceName: "customers", queryName: "v\nrun: helper -> ..." }
  -> 1 row from the unexported helper              (injection)
```

Cross-tenant on the default deployment shape, with no `explores` required. Restoring the quoting: 47 pass / 0 fail.

## The escape holds

This is what the whole fix rests on, so I checked it against the grammar rather than the comment:

- ``BQ_STRING: BQ (STR_CHAR | ['"])* BQ``, and `STR_CHAR`'s `SAFE_NON_QUOTE` excludes the C0 control range, so a raw newline is not a valid `STR_CHAR` and can never sit inside a quoted identifier.
- ``idToStr`` calls ``ParseUtil.parseString(text, '`')``, so backslash unescaping does happen at the identifier level. That makes escaping `\` before `` ` `` required rather than cosmetic, and the added test pins it.
- Escaped forms of the payloads are all correct, including the trailing backslash case (`v\` -> ``` `v\\` ```) that a naive single-replace breaks on.
- It matches upstream `escapeIdentifier` in `to_stable.ts` character for character.

## Layer and blast radius

Grepped for any other path interpolating these fields into Malloy text: nothing. The remaining `${sourceName}` uses are error messages, materialization (server-derived names), and scaffolding. `getFilters`/`getAuthorize` use `sourceName` only as map keys, which is why item 3 (not unwrapping pre-backticked names) is right: unwrapping would desync the syntax path from the lookup keys and reopen the filter-skip lane.

Escaping at the single construction site is also the more durable layer than validating at the API boundary, since a new call path can bypass a validator but not the builder.

## Behavior-change claims

SDK producers send either a bare real `queryName`, or Malloy text in `query` with both name fields explicitly `undefined`. `get_context` returns flat view names rather than dotted paths, so an agent following the advertised contract will not emit `carriers.by_name`. The doc and tool-description updates target the actual residual risk.

Also confirmed: eslint clean; `grep 'sourceName + "->"'` finds nothing; api-doc YAML parses with valid block scalars; every response-contract fact survives the tool-description compression (the one `_query_row_limit` drop is one of the acknowledged triplicate mentions). Agreed on not raising the 2000-char budget -- that test's own comment says the cap belongs to the client.

## One correction to the numbers

"1898 pass / 0 fail" does not reproduce here. The base commit already fails 47 with 5 errors (materialization, DuckDB, theme, watch-mode -- environmental, needing GCS credentials and timing-tolerant scheduler runs). The delta is the meaningful figure and it is clean:

| | pass | fail | errors |
| --- | --- | --- | --- |
| base `9df652af` | 2061 | 47 | 5 |
| head `100f44bc` | 2085 | 44 | 1 |

Plus 21 tests, no new failures, and no overlap between the failing suites and the eight changed files.

## One non-blocking suggestion

`expectNamedBlocked` in `restricted_mode.spec.ts`, and the two sibling helpers, use a bare `catch { return; }`. Dropping `looksRestricted` on the named path is justified and documented -- rejection legitimately moves to name resolution, and the strict check correctly survives for the ad-hoc path where restricted mode is still the first line. But a bare catch also passes when the request fails for an unrelated reason (a fixture typo, a closed connection), so the test could stop exercising injection while staying green. Asserting the message matches something like `/undefined object|reserved word|Reference to/` would keep it honest without re-pinning a specific error class.

sagarswamirao (approved):

PR #967, opened 2026-08-05
fix(server): read a document's own model annotations, not its imports'

mlennie:
Slice 3 of the ten-way split proposed in #935. It stands alone, fixes a live bug on `main`, and has to precede anything that reads model-level tags.

## What it does

`modelAnnotations` folds a model's `##` tags across everything it imports. File-level `##(authorize)` needs exactly that, because a gate an import could shed would be no gate at all. Every other model-level tag wants the opposite, because it describes one document.

The notebook response used the folded set. `ownModelNotes` reads only what a file declares itself, and the notebook response now uses it. `fileLevelAuthorize` is deliberately left on the folded lineage.

A document can span several compilation nodes: a `.malloynb` is compiled a cell at a time, so its `##` tags land on an `internal://loadModel` node that the `internal://extendModel` node named by `modelID` inherits from. Reading `modelID` alone would therefore find nothing on any notebook, so the walk treats a non-`file:` scheme as more of the same document and stops at `file://`, which is the line between this document and a different one.

## The bug this actually fixes

Worth stating precisely, because it is not the symptom named in #935's description. That description says an imported model's `##` tags "were setting the title and autorun for every notebook that imported it". On current `main` there is no notebook `title` and no `autorun`: both arrive in later slices, so that symptom is not reachable yet.

What is reachable today: `model.ts` passes the folded annotations out as the notebook response's `annotations`, and its only consumer on `main` is `parseNotebookFilterAnnotation` (`packages/sdk/src/components/filter/utils.ts:55`), which scans for `##(filters)`. So on `main` the live bug is that **a shared include carrying `##(filters)` configures the filter panel of every notebook that imports it.**

That panel is removed later in the split, so this fix is worth landing on both counts: it is correct on `main` today, and it is the groundwork every later per-document tag depends on.

## How it was assembled

`ownModelNotes` and its spec are Kyle's work, taken from #935 unchanged. I checked whether `main` had moved under that branch's base for each file first:

- `annotations.ts` and `annotations.spec.ts` are untouched by `main` in the 21 commits since, so they came across wholesale.
- `model.ts` has moved (#933 per-query metadata, #938 MCP tool results), so that call site is a surgical edit rather than a copy. On `main` there is one site to switch, not the two on the branch, because the second sits inside code a later slice introduces.

One change this slice needs that the full PR does not: switching that line orphaned the `Annotations` import from `@malloydata/malloy`, whose only use in `model.ts` was the line being replaced, so it is removed here. It stays in the full branch because other new code there uses it.

## Verification

Run in the order this repo needs, with `bun run generate-api-types && bun run build:sdk` first since the root `typecheck` script chains neither.

| check | result |
|---|---|
| `bun run lint` | clean |
| `bun run typecheck` | clean |
| `bun run prettier:check` | clean apart from the six git-ignored generated files CI never sees |
| `bun run test:unit` | 1788 pass, 0 fail, 99 files |
| `src/service/annotations.spec.ts` alone | 12 pass, 0 fail, 27 assertions |
| `bun run test:integration` | 232 pass, 0 fail, 31 files |

The spec covers the notebook multi-node case, the `file://` boundary, ancestral-first ordering, and that `modelAnnotations` still folds the lineage for `authorize`.

## Follow-ups

Later slices of the split will use `ownModelNotes` in `service/dashboard.ts` and for the notebook `title` and `autorun` fields, which is where #935's original framing applies.

### DIFF
--- a/packages/server/src/service/annotation_ownership.spec.ts
+++ b/packages/server/src/service/annotation_ownership.spec.ts
@@ -1,0 +1,255 @@
+/**
+ * Compile-backed cover for model-annotation ownership.
+ *
+ * `annotations.spec.ts` exercises `modelAnnotations` / `ownModelNotes` against a
+ * hand-built registry, which pins the fold but also encodes our assumption about
+ * how malloy names compilation nodes. Both functions exist only because malloy
+ * does not export `getModelAnnotations`, so that assumption is the load-bearing
+ * part: if an upgrade renames the `internal://` scheme or stops giving a
+ * notebook's cells their own nodes, a unit test over a fixture keeps passing and
+ * every notebook silently reads the wrong tags.
+ *
+ * Two halves, and they cover different things:
+ *
+ *  1. The compiler contract: a real `Runtime`, asserting the node scheme and
+ *     that an import's `##` is reachable through the folded reader but not
+ *     through the own-notes reader.
+ *  2. The wiring: a real package through the load worker, asserting the
+ *     notebook API response carries the notebook's own `##` and not its
+ *     include's. This is the half that fails if the `model.ts` call site
+ *     reverts to the folded reader.
+ */
+import { DuckDBConnection } from "@malloydata/db-duckdb";
+import {
+   FixedConnectionMap,
+   InMemoryURLReader,
+   Runtime,
+   type ModelDef,
+} from "@malloydata/malloy";
+import { afterAll, afterEach, beforeAll, describe, expect, it } from "bun:test";
+import * as fs from "fs";
+import * as os from "os";
+import * as path from "path";
+import {
+   PackageLoadPool,
+   __setPackageLoadPoolForTests,
+} from "../package_load/package_load_pool";
+import {
+   annotationTexts,
+   modelAnnotations,
+   ownModelNotes,
+} from "./annotations";
+import { Package } from "./package";
+
+/**
+ * A shared include carrying a model-level `##(filters)`. This is the tag behind
+ * the bug: its only consumer is the SDK's `parseNotebookFilterAnnotation`
+ * (`packages/sdk/src/components/filter/utils.ts`, via `Notebook.tsx`), which
+ * matches the literal `##(filters)` prefix on the notebook response's
+ * `annotations`, so folding it across imports configured the filter panel of
+ * every notebook that imported this file.
+ */
+const SHARED_MALLOY = `##(filters) ["orders.state"]
+## title="Shared include"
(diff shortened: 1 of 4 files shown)


kylenesbit (approved):
Approving. I verified the mechanism against malloy's internals rather than taking the description on faith:

- Ordering matches malloy's own `notesInOrder` (`inherits → blockNotes → notes`), so `ownModelNotes` agrees with `Annotations.texts()`.
- The `internal://` premise is real: malloy mints `internal://${method}/${uuid}` for any URL-less compile, and the notebook path in `model.ts` builds exactly that chain (`loadModel(stmt.text)` for the first Malloy cell, `extendModel(stmt.text)` for each one after), while `.malloy` files load a real `file://` URL through `getStandardModelMaterializer`. Reading `modelID` alone really would find nothing on every notebook.
- The `file:` boundary holds today: every compile root in the server is `file:` (`pathToFileURL` in the package-load worker and pool, the `file:///storage-serve-shape/` in-memory root, the `__compile_check.malloy` virtual URL).
- The stated bug is accurate — `parseNotebookFilterAnnotation` (`packages/sdk/src/components/filter/utils.ts:51`, via `Notebook.tsx:97`) is the only consumer of the notebook-level `annotations` on main.

Three non-blocking follow-ups, none worth holding this:

1. **The document boundary is a denylist where an allowlist would fail closed.** `isSameDocument` treats anything that is neither the document nor a `file:` URL as the same document. The plumbing for non-`file:` imports already exists (`package_load_worker.ts:321` proxies any non-`file:` URL to the main thread), so the day a `gs://` or `https://` import appears, its `##` folds back in silently with no test failing. `modelID` is covered by the first clause and cells are always `internal:`, so inverting to `id.startsWith("internal:")` is strictly safer at no cost.

2. **The call site is untested.** The spec exercises the pure function against a hand-built registry, which encodes the assumption about malloy's node scheme rather than verifying it. Since this whole area is a workaround for malloy not exporting `getModelAnnotations`, one compile-backed test would catch an upgrade that renames the scheme — `source_given_probe.spec.ts` and `drill_probe.spec.ts` already compile real models with `runtime.loadModel`.

3. **Worth reconciling with the later slices before they land.** #935 keeps `allAnnotations` on the folded set on purpose ("stays whole because it is reported, not interpreted") and routes `ownModelNotes` into the notebook tag instead. As written, that slice's diff will show `annotations` widening back to the folded set, which will read as a regression to whoever reviews it. Either that comment or this call site should give.

Also needs a rebase — the branch is behind main. CI is green across all 14 checks, and `annotations.spec.ts` passes locally (12 pass, 27 assertions).

PR #969, opened 2026-08-05
docs: land the dashboards design doc and the security posture

mlennie:
Slice 2 of the ten-way split of #935, which its author describes as a handoff rather than a merge candidate. This is the docs slice: it lands the design doc the later server and UI slices are written against, plus the security posture the custom-JSX cut turns on.

## What is here

**`docs/malloyyo-dashboards-design.md`** (new), the design behind native `dashboards/*.malloy` support. It arrives ahead of its implementation on purpose, so the PRs that implement it have one reference to build against.

Its status header is rewritten rather than copied. #935 opens the doc with "Status: Shipped", which is true of that branch and false here: `dashboard.ts`, `motly.ts`, the SDK `Dashboard/` component and `examples/storefront/dashboards/` do not exist on `main`. Landing it unchanged would publish a shipped claim for a feature with no code. The header now says it is designed but not landed, that every "shipped" below is true of the branch only, and that its paths and UI labels describe the end state after the rename slice, since the doc is written throughout in post-rename vocabulary that `main` does not use yet (`pageEmbed.ts`, `PageViewer`, "Pages", "Governed Reports"). Four links to files that do not exist yet are plain text rather than dead links, and a stale claim that the SDK is on `^0.0.423` "exactly the version Malloyyo requires" is corrected to the 0.0.423 floor against the current `^0.0.427`.

**`docs/security-posture.md`** (new), the trust boundary and the four known gaps in it. All four were re-verified against `main` rather than taken on trust, and two needed correcting. The raw-HTML gap named notebooks alone, but `markdown-to-jsx` is used at three call sites (`NotebookCell.tsx`, `MutableCell.tsx`, `About.tsx`), none passing `disableParsingRawHTML`, so fixing one would have left two open and looked complete. And the isolation section claimed it would "close gap 2", the embed token, which it would not; it addresses the author-code-execution surface. The other three hold exactly as written: framing is set only for `.html` under a package's `public/` with no global `X-Frame-Options`; `publisher.js` sends `embed_token` and `Authorization: Bearer` and no server code reads either; and both `PageViewer.tsx:88` and `publisher.js:178` check `event.source` but never `event.origin`.

It is cross-linked with the security policy from #965, which landed a few hours before this branch. That policy's Hardening section dead-ended without a posture doc, so it now points here, and this doc defers to the policy on reportability. Without that, gap 1 reads like an open vulnerability when the policy deliberately classifies the permissive framing default as working as documented.

**Two verified defects in `docs/givens.md`**, both predating #935.

`string[]` was documented as a supported given type in three places, including a widget table promising "multi-value autocomplete with chip removal". It does not compile (`unexpected ']'`), and no such widget exists: `GivenInput` branches only on boolean, number and the date family, and everything else is a text box. Removed, and the type list now states that array and record givens are not accepted, which `given.ts` in #935 already explains. `timestamptz` was missing from the same table while appearing in the widget table; it is real, and its default needs an explicit cast because a bare `@...` literal is a `timestamp`.

`#(description="...")` with more than one word emits a `malformed-route` warning on every compile. **The form is kept rather than replaced.** Every alternative was run through the compiler and then through the server's own conversion and the SDK's helper-text function, and each is worse: `#(doc)` and `#(description)` with free text compile clean but render as `doc) Earliest report date to include`, and a plain `# ` tag compiles clean but is filtered out server-side as a reserved route and renders nothing at all. The documented form is the only one that works, so the warning is documented as cosmetic instead of swapped for something broken.

**`api-doc.yaml`**, prose only, no schema change. The resource hierarchy was missing materializations, package-scoped connections, the events stream and the server-level routes, and it named a `malloy-publisher.json` manifest that does not exist; the file is `publisher.json`. Rewritten against the actual path list rather than copied from the branch, so it describes what is in this spec today and does not mention dashboards.

## What was deliberately left out

Three of the five docs changes #935 groups into this slice turned out to depend on later slices rather than standing alone, and are not here:

- The rest of the `docs/README.md` index. #935 indexes `console.md`, `dashboards.md`, `choosing-a-surface.md` and `react-data-apps.md`, none of which exist yet, and drops the `html-data-app` example row, which is only correct once that example is deleted. This PR adds only the two rows for the docs it introduces, and does not touch the `publisher-app.md` line, so the renames slice can move it without a conflict.
- The `docs/development.md` edit, described as fixing a stale standalone-recipe claim. It is not stale on `main`: `examples/html-data-app/README.md` still has that recipe. It goes with the examples slice that removes the example.
- The `api-doc.yaml` tag rename and the dashboard paths, which belong to the renames and dashboard-REST slices.

The "broken `givens.md` anchor" listed for this slice has no referent. Both intra-doc anchors resolve and nothing in the repo links into them; what the branch actually does there is rename a heading, which belongs with the dashboards docs.

## A question for the reviewers

#935 also adds `docs/malloyyo-vs-publisher.md`, 172 lines of strategy and competitive comparison including a "What not to copy" section about a sibling malloydata project. It is not in this slice's list and it is the only doc in `docs/` the index does not reference. **My recommendation is that it does not belong in this repo's user-facing `docs/` at all**, and I have left it out rather than carrying it silently. Happy to be overruled, but it seemed worth an explicit decision rather than a quiet inclusion.

## Follow-up

`docs/discovery-and-access.md` quotes `examples/governed-analytics/publisher.json` as a two-entry `explores` list. The examples slice makes it three by adding a dashboard, and that doc is not in #935's diff, so nobody currently fixes it. Flagging it for whoever takes that slice.

## Verification

Full gate in the required order. `generate-api-types` and `build:sdk` clean, `lint` exit 0 with no autofix mutations, `typecheck` clean, `prettier:check` red on only the six standing git-ignored generated files, unit 1848 pass / 0 fail, integration 237 pass / 0 fail. A link and anchor checker over all five touched files reports no broken targets. The Malloy claims above were each compiled against a real DuckDB-backed runtime rather than read, and the annotation results were taken through `malloyGivenToApi` and `GivenInput`'s helper-text function so the rendered output is measured rather than inferred. No tracked file other than `api-doc.yaml` contains the rewritten description, and the generated Python client is gitignored, so the spec change causes no tracked drift.


kylenesbit (approved):

mlennie:
Pushed one more commit after your approval, so this needs another look. Sorry for the churn; the delta is four words of prose in one file.

I went back and closed the verification gap on this PR properly (live pass against a running server, rather than only the gate), and it caught an incomplete fix of my own. My first commit's message says the spec "named a `malloy-publisher.json` manifest that does not exist". It fixed exactly one occurrence, in `info.description`, and left four more in the materialization schema descriptions. So the claim was true of the sentence I edited and false of the file.

`750286e7` finishes it by enumeration rather than sampling: `grep -c malloy-publisher.json api-doc.yaml` is now 0. The real filename is pinned by `PACKAGE_MANIFEST_NAME = "publisher.json"` in `packages/server/src/constants.ts` and by the literal join in `service/environment.ts`.

Still prose only, and provably so: parsing both specs and deleting every `description` key leaves byte-identical structures, so no path, schema, or component changed. Full gate re-run on the merged tip, unit 2042/0 and integration 245/0.

Two occurrences I deliberately did **not** touch, named here rather than silently skipped because they are source comments rather than the user-facing spec and would widen a docs PR into server code: `service/package.ts:657` and `service/build_plan.ts:64`. Happy to fold them into a later slice.

PR #972, opened 2026-08-05
refactor: Pages become Data Apps, and the App becomes the Console

mlennie:
Slice 1 of the ten-PR split proposed in #935, whose own description says it is a handoff rather than a merge candidate. Extracted against current `main` rather than cherry-picked: `main` has moved a long way past that branch's base, so each file was checked individually and taken from the branch wholesale only where `main` had not touched it since. The SDK files, the test fixtures and the integration spec came across whole; `docs/README.md` and `api-doc.yaml` had moved, so those are surgical edits.

## The REST break is split out, on purpose

Kyle's description offers to split the `/pages` REST rename into its own review, and I have taken that offer. **This PR changes nothing that names the `/pages` endpoint or any user-facing URL**, so it cannot alter API behaviour: the endpoint, its `operationId`, its schema and the SPA route are all untouched here, and the comments that mention `/pages` still describe it accurately. The break lives in a second PR stacked on this one, so it can be weighed on its own merits. Please merge this one first.

## What changed

Three renames, no behaviour change.

**Pages become Data Apps, internally only.** The SDK component `PageViewer` becomes `DataAppViewer` (props unchanged), and `utils/pageEmbed` becomes `utils/dataAppEmbed`. Server test fixtures `html-pages-*` become `html-data-apps-*` and the integration directory `html_pages` becomes `html_data_apps`. The `publisher.js` runtime carries a cross-reference comment pointing at the renamed module, so that moved too.

`DataAppViewer` is reachable from the package root, so this is a breaking change for anyone embedding the SDK, and `RELEASE_NOTES.md` records it with the one-line migration. `dataAppEmbed` is not: it is not re-exported from the root barrel and `packages/sdk/package.json` declares no `./utils/*` subpath, so nothing outside the package can import it. I checked that before writing the note, because my first draft wrongly told consumers to update a deep import that does not exist.

**Publisher App becomes Publisher Console.** `docs/publisher-app.md` becomes `docs/console.md`, along with its screenshot, the `capture-screenshots.mjs` entry that writes it, and all ten referrers (`README.md` twice, `docs/README.md`, `api-overview.md`, `architecture.md`, `embedded-data-apps.md`, `explorer.md` twice, `html-data-apps.md`, `examples/storefront/README.md`). "App" deliberately survives where it names the package rather than the product surface, since `packages/app` keeps its name; that is the same line #935 draws, which is why "publisher app shell" and the Playwright README are left alone.

**Governed Reports becomes Notebooks**, the section label in the package view. Label only, same files listed, no prop or route changed.

## What is deliberately not carried over

#935 rewrites the prose of this docs page as well as renaming it (+33/-23, 52% similarity). That rewrite is not here, and this is not a style preference: it links `react-data-apps.md` and `choosing-a-surface.md`, neither of which exists on `main`, and it documents a Dashboards row and a Materializations section that later slices build. Carrying it would ship broken links and describe UI that is not there. So this PR does the file rename plus the product-name substitution and leaves the prose to the docs slice (#969).

## How it was verified

Full gate in the required order (`generate-api-types && build:sdk` first, since the root `typecheck` script chains no codegen): lint exit 0 with `git diff` byte-identical before and after, so no `eslint --fix` mutations rode along; typecheck exit 0; `prettier:check` failing only on the six standing git-ignored generated files, each confirmed with `git check-ignore`; unit **1848 pass / 0 fail** across 102 files; integration **237 pass / 0 fail** across 32 files. Unit and integration were run separately so a red unit leg could not hide the integration result.

Beyond the gate: the integration log shows `html-data-apps-test-env` being created and both renamed fixture packages loading under their new names, which is what actually proves the fixture rename rather than a green tick. A link checker over all 372 relative markdown links in 185 files reports zero broken targets after the docs rename. And a repo-wide sweep finds no remaining reference to `PageViewer`, `pageEmbed`, `html-pages`, `html_pages`, `publisher-app` or "Governed Reports" outside the deliberate references in `RELEASE_NOTES.md`.

Rendered the Console against the production bundle to confirm the label change actually reaches the UI rather than only compiling: the package view shows "Notebooks" and no "Governed Reports".

Extracted from #935; the implementation is Kyle's, credited with `Co-authored-by`.


kylenesbit (dismissed):
The rename is mechanically clean and the sweep checks out, but merging this as-is breaks the SDK build, and the breakage arrives with no conflict marker. Requesting changes on that one point.

## The silent break

`main` picked up #962 after this branch's base, which added `packageFileUrl` to `utils/pageEmbed.ts` and re-exported it from the SDK root:

```ts
// packages/sdk/src/index.ts:11 on main
export { packageFileUrl } from "./utils/pageEmbed";
```

This PR renames that module to `dataAppEmbed.ts` and does not touch `packages/sdk/src/index.ts` — it couldn't have, the line did not exist at the base. Only one side changed that file, so git merges it cleanly and nothing flags the problem. I built the merge tree to confirm: `packages/sdk/src/utils/` contains only `dataAppEmbed.ts` and `dataAppEmbed.spec.ts`, while the merged `index.ts` still re-exports from `./utils/pageEmbed`. Dangling import, failing typecheck.

The 14 green checks are honest but not reassuring here — they built the branch in isolation, never merged against current `main`.

## Both conflicts can silently revert #962

Git reports conflicts in `DataAppViewer.tsx` and `ModelPage.tsx`, and in both the tempting resolution is the wrong one.

In the viewer, `main` replaced the hand-rolled URL template with the new helper while this branch still carries the old template under a renamed variable:

```
<<<<<<< origin/main (PageViewer.tsx)
      packageFileUrl({ server, environmentName, packageName, path: pagePath }),
=======
      return `${serverBaseUrl(server)}/environments/${encodeURIComponent(
>>>>>>> pr972 (DataAppViewer.tsx)
```

Taking this PR's side — the natural instinct, since it is "the renamed file" — reverts that refactor and orphans `packageFileUrl`. `ModelPage.tsx` is the same shape but larger: `main` added an entire "Nothing to open at this path" error state built on `packageFileUrl`, and this branch's side is still the one-line `<h2>Unrecognized file type:</h2>`. Both want `main`'s side kept with only the identifiers renamed.

Worth noting the rename now touches a public surface it did not at branch time. The RELEASE_NOTES claim that `dataAppEmbed` is internal is still accurate — the exports map declares only `.`, `./client`, `./client/*` and three CSS entries, no `./utils/*` subpath — but since #962 one of that module's exports is reachable from the root barrel, which is exactly why `index.ts` needs the update.

## What checks out

The repo-wide sweep is genuinely clean: `PageViewer`, `pageEmbed`, `html-pages`, `html_pages`, `publisher-app` and "Governed Reports" survive only in the deliberate RELEASE_NOTES entries, including the historical entry at line 115 describing an earlier release, correctly left alone. The 40-file rename is consistent, `docs/console.md`'s links all resolve to files that exist on `main`, and the screenshot moved with its `capture-screenshots.mjs` entry.

## One judgment call

`Package.tsx:227` still reads `<PackageSection title="Pages" count={pages.length}>`. The sibling label was renamed here ("Governed Reports" → "Notebooks"), and a section title is not REST, so it is not covered by the #973 carve-out. As it stands the Console will show a "Pages" section while the docs and the component call them data apps, and `main`'s new error copy in `ModelPage.tsx` tells users the package view "lists this package's models, notebooks, and pages." Deliberate and deferred to a later slice is a fine answer — flagging it because it is the one user-visible piece of the rename this PR leaves behind.

mlennie:
All four points addressed. Force-pushed: #972 is now `962998c2` and #973 is `26e71422`, both rebased onto `8425fb7a`, zero behind. Thank you for the merge-tree check, it was the right way to find this and I have adopted it as a standing step rather than a one-off.

**1. The silent break.** Fixed. `packages/sdk/src/index.ts:11` now reads `export { packageFileUrl } from "./utils/dataAppEmbed";`, and `packageFileUrl` came across into the renamed module unchanged. Verified with your method rather than by reading: the merged tree has `packages/sdk/src/utils/` containing `dataAppEmbed.ts` plus its spec, and the merged `index.ts` pointing at it. No dangling import.

Your closing line on that point is the part I took most seriously, so I did what the 14 checks do not: merged current `main` into each branch and typechecked the merged tree. Exit 0 for both. I also ran `generate-api-types` and `build:sdk` on both merged trees in throwaway worktrees, and confirmed the generated client each one produces matches its own call sites, which is what proves #972 is coherent standing alone rather than only as half of the pair. On #972 the client regenerates `PagesApi.listPages` and the call sites are `apiClients.pages.listPages`; on #973 it regenerates `DataAppsApi.listDataApps` and the call sites follow.

**2. Both conflicts.** Resolved towards `main` in both, exactly as you said they wanted. `DataAppViewer.tsx` builds `standaloneUrl` from `packageFileUrl({ server, environmentName, packageName, path })` with no hand-rolled template surviving. `ModelPage.tsx` keeps your whole "Nothing to open at this path" state, including `looksLikeFile` and the `staticUrl` link; `Unrecognized file type` appears nowhere in the tree. To check I had not dropped anything else from #962 I extracted every line it added to a file this stack touches, normalised away the rename, and asserted each still appears: 987 lines, zero missing, and no hunk duplicated. #962 touched nine files, not four, and the stack touches all nine.

**3. The RELEASE_NOTES claim.** You are right that it was still accurate, and I tightened it anyway so it says what actually holds: the module move is invisible to consumers because `exports` declares no `./utils/*` subpath, while the one symbol the root re-exports keeps both its name and its root export, so only the path behind it moved.

**4. The `title="Pages"` judgment call.** Deliberate, and the reason is that the label names the same concept as the `/pages` endpoint, so it moves with it in #973. The payoff is that each PR is internally consistent at its own merge point rather than each being half-renamed:

| | #972 | #973 |
| --- | --- | --- |
| `Package.tsx` section label | `Pages` | `Data Apps` |
| `docs/console.md` hierarchy | `Pages` | `Data Apps` |
| `ModelPage.tsx` error copy | "and pages" | "and data apps" |
| `docs/screenshots/console.png` | shows Pages | shows Data Apps |

That last row is why I raise it: the screenshot is recaptured once per PR, so neither ships a figure contradicting its own text. If you would rather the label came here instead, say so and I will move it; it is a one-line change plus the figure.

**Two things your review led me to that were not in it.** Both are worth knowing beyond this PR.

Your note that "the sweep checks out" sent me to re-run it, and the sweep could not have been trusted: `\b` is silently unsupported by this repo's `git grep -E`. `git grep -cE '\bPage\b' -- api-doc.yaml` exits 1 with no output while `git grep -cP '\bPage\b'` finds it. Every sweep I had run with `\b` was a false negative. Redone with `-P`, and it surfaced a real partial rename: `docs/architecture.md` still had `- **App:** Web interface for browsing...` four lines above where the same file had been changed to say "the Console is built entirely from it", and `docs/api-overview.md` still said "The App's footer Publisher API link" in a file whose header had just been changed to say Console. Both fixed, and the completeness claim in the commit message now tests true.

Separately, two docs figures were stale in a way a rename sweep does not catch because they are images. `docs/screenshots/console.png` was a rename of `publisher-app.png` with the bytes untouched, so it still showed "Governed Reports" and "Pages", contradicting the page it illustrates. And `docs/screenshots/api-explorer.png` renders the `api-doc.yaml` overview block that #973 edits, and predated #969's rewrite of it, still showing a `malloy-publisher.json` manifest. Both recaptured at the viewport and scale factor their `capture-screenshots.mjs` entries declare, on a clean `start:init` state so no local test environment leaks into a docs image.

**One thing I did not fix, deliberately.** `docs/malloy-publisher-demo.gif`, the README hero, also still shows the "Governed Reports" label. It is a hand-recorded screencast with no entry in `capture-screenshots.mjs`, so there is no scripted way to regenerate it, and re-recording a product demo is outside a rename slice. It is already stale on `main` for unrelated reasons (it shows four semantic models where `storefront` now has one). Happy to file it as its own issue.

Also fixed while rebasing: #976 and #975 both landed on `RELEASE_NOTES.md` during review, which is where GitHub's conflict came from. The resolution keeps all four sections, #975's tidy, #976's givens entry and both of mine, with #976's 13 release-note lines and all 127 of its `api-doc.yaml` lines verified intact.

Merge order is unchanged: #972 first, then GitHub retargets #973 to `main`. One practical note, since `main` moved seven times while this was in review: the check worth running immediately before merge is `git merge-tree --write-tree origin/main <branch>` and then grepping the merged `index.ts`, not the green checkmarks. Any new commit touching `packages/sdk/src/index.ts`, `api-doc.yaml`, `ModelPage.tsx`, `Package.tsx` or `spa_fallback.ts` reopens this class with no conflict marker, and the guarantee here comes from the rebase rather than from any alias.

Sha-Bang (approved):
went through this one carefully since we're the downstream that eats the rename — nothing blocking, approving.

checked the things I'd actually worry about rather than taking the description's word for it. merge-tree against current `main` is clean, and in the merged tree `packages/sdk/src/index.ts` exports `./utils/dataAppEmbed`, so the dangling-import risk is properly gone. the three commits `main` picked up since your base are all materialization work, nothing you rename. and on our side I confirmed we don't import `PageViewer` or `pageEmbed` and don't render the SDK's `<Package>`, so this is a genuine no-op for Credible.

two small things, neither worth holding it for:

- the one user-visible change here — `Package.tsx` flipping "Governed Reports" to "Notebooks" — has no test. `packages/app/tests/playwright/package-notebooks.spec.ts` already lands on that exact view, so a one-line heading assertion would pin it, and would pin #973's flip to "Data Apps" too. right now the only evidence is the recaptured png.
- `RELEASE_NOTES.md` has two `## [Unreleased]` sections now and #973 adds a third. the file tells the releaser to copy "the relevant section" into the release page, which is an easy way to ship one and silently drop two.

mlennie:
Both of your non-blocking items are done, and this branch is now `4ca92d69`.

`package-notebooks.spec.ts` asserts the `Notebooks` heading, exactly where you suggested, plus the absence of `Governed Reports`. It does pin it: reverting `title="Notebooks"` in `Package.tsx` and rebuilding fails that one test and leaves the other three green, so your point is demonstrated rather than taken on faith. The same relabel now has a matching assertion on #973 for the flip to "Data Apps".

On `RELEASE_NOTES.md`, your count was the right one and mine was wrong: `main` has one `## [Unreleased]` section, this PR makes it two, #973 makes it three, and the `[0.0.236]` DuckLake section I had been counting is versioned rather than unreleased. The copy instruction is now plural, says to stamp each section with the version that shipped it, and says why there is usually more than one. It sits here rather than on #973 because this is the PR where reading it as singular first starts dropping entries.

Also, since it changes what you approved: the branch has been rebased onto `e8a6e98f` to clear "out-of-date with the base branch". I did that locally rather than with the Update branch button, which makes an unsigned merge commit and then rejects the next push. The diff between the old and rebased tips is exactly #978's fourteen files and none of this slice's own, and the full gate was re-run on the result.

@kylenesbit when you have a moment, could you re-review? Your changes-requested is what is holding this, and it is answered: `packages/sdk/src/index.ts` exports `packageFileUrl` from `./utils/dataAppEmbed`, checked with the merge-tree method rather than by reading the diff. #973 is stacked behind this one and cannot move until it lands.

PR #973, opened 2026-08-05
feat!: rename the /pages endpoint to /data-apps

mlennie:
Second half of slice 1 of the [#935](https://github.com/malloydata/publisher/pull/935) split, and the breaking half. It is separate on purpose: everything here is observable to a caller over HTTP, and #935's own description offers to split the REST break out for its own review. The internal renames landed in #972. 21 files, +504/-158, one signed commit.

## What changes

- **`GET …/packages/{pkg}/pages` → `GET …/packages/{pkg}/data-apps`.** Same response shape, same query parameters, same status codes. **No alias: the old path answers 404.**
- **Schema `Page` → `DataApp`**, and `operationId: list-pages` → `list-data-apps` under a `data-apps` tag. Anything generated from `api-doc.yaml` changes with it: `PagesApi.listPages` becomes `DataAppsApi.listDataApps`, and `apiClients.pages` on `<ServerProvider>`'s context becomes `apiClients.dataApps`.
- **The browser URL `/{env}/{pkg}/pages/{file}` → `/{env}/{pkg}/data-apps/{file}`, and the old form keeps working for one release.** It opens the data app and rewrites itself to the new URL in the address bar, carrying any query string and fragment so the rewrite never silently drops caller state. Deprecated, and removed one release after the release that ships this.
- **The package view's "Pages" section is now labelled "Data Apps."**

## Why the REST path breaks cleanly while the browser URL does not

The asymmetry is the whole decision, so it is worth stating rather than leaving to be inferred.

An alias in the spec means two paths, two operationIds and two generated client methods for one listing, with every future change to it made twice. The only known consumer reviewed this change and chose the clean break, having already accepted a short mixed-fleet 404 window during their worker rollout. A bookmark has no owner to consult, and the person holding one is not reading release notes, so that surface degrades gracefully instead.

**The bookmark alias also removes something worse than the break it softens.** Before it, an old link did not simply fail: it fell through to the static-file route, so a package shipping a `public/pages/` directory answered the old URL **200 with a different document**. A wrong page under a 200 is the exact failure [#962](https://github.com/malloydata/publisher/pull/962) existed to remove, and documenting it would not have helped anyone who hit it. Claiming the app's own segment deletes that path entirely.

The alias has **two halves in two packages**, and either removed alone leaves it broken: `pages` in `SPA_OWNED_SEGMENTS` (`packages/server/src/spa_fallback.ts`) is what lets the request reach the app at all, and a branch in `ModelPage.tsx` is what rewrites it. Each site carries a comment naming the other and saying when to delete it. A `.malloy` or `.malloynb` under `pages/` is excluded, since that can be a real model path the old data-app URL never named.

The removal release is named relatively rather than numerically on purpose: `packages/*/package.json` on `main` says `0.0.209` while the newest tag is `v0.0.240`, because `release.yml` bumps the version on a release branch and never merges it back. The released version is not discoverable from the repo.

## Migration

- Change the request path to `/data-apps`, and regenerate if you build a client from the spec.
- `apiClients.pages.listPages(env, pkg)` becomes `apiClients.dataApps.listDataApps(env, pkg)`.
- Update any stored link of the form `/{env}/{pkg}/pages/{file}`. It works in this release and stops working in the next.
- The standalone URL `/environments/{env}/packages/{pkg}/{file}` never changed.

## All five generated targets were exercised, not assumed

`openapi-spec-validator` OK; the SDK client emits `DataAppsApi.listDataApps` with no `PagesApi` left; the Python SDK script (the only CI job that watches `api-doc.yaml`) green; and the CLI client, which is in no workflow and not in the root typecheck, regenerates to `data-apps-api.ts` and typechecks clean.

## How it was verified

Every claim below was measured on this tree rather than carried forward.

- **Gate:** typecheck 0, lint 0 with no `--fix` mutations, prettier clean for every tracked file, SDK **88/0**, unit **2274/0**, integration **250/0**, skills 7/367/20, Playwright **22/22** on the affected specs locally.
- **Live, on the production bundle:** `…/data-apps` returns 200 with the expected payload; `…/pages` returns 404 as JSON naming the path; a malformed environment name still 400s; the new browser route renders the embedded viewer with its title arriving through the regenerated client; and the old browser route opens the data app and leaves `…/data-apps/index.html` in the address bar, with `?query#fragment` intact.
- **Mutation-tested, not accepted on a pass.** Removing `data-apps` from `SPA_OWNED_SEGMENTS` fails 2 of 27 `spa_fallback` cases; removing `pages` fails 1, plus the Playwright case that walks the alias end to end. Reverting the `api-doc.yaml` path fails 3 of 3 parity cases; reverting the route in `server.ts` fails 1 of 3. Reverting the section label fails 2 of 5 `package-data-apps` cases.

## Four test cases are new, because most of this rename was pinned nowhere

Reverting pieces one at a time and re-running the whole suite found six things nothing caught: the `api-doc.yaml` path key, the section label, the route `Package.tsx` emits, the prefix `ModelPage` strips, the old URL's behaviour, and the removal of the REST path itself.

- `packages/server/src/data_apps_route_parity.spec.ts` asserts the spec path key equals the route `server.ts` registers. This closes the only silent external-contract break found: reverting the path in `api-doc.yaml` alone regenerates a client that requests `/pages` while the server serves `/data-apps`, and typecheck, unit and integration all stay green. `Package.tsx` swallows the resulting 404 by design, so there is no UI symptom either.
- `packages/app/tests/playwright/package-data-apps.spec.ts` covers the click-through, which no Playwright test touched at all, plus the alias and its exclusion.
- The `html_data_apps` integration spec asserts `GET …/pages` is 404 **and** that the body is JSON. The JSON half is #962's fallback rather than anything this rename does, which is why it belongs here: this change is what makes the promise to callers.
- `spa_fallback.spec.ts` asserts the old `pages/<file>.html` still reaches the app, in both the asset-extension and plain forms.

**Three assertions in this PR were true too early or too coarsely and pinned nothing until mutation testing exposed them**, which is the part worth carrying into other reviews:

1. `toHaveURL(<the path just requested>)` is satisfied on its first poll, before React can navigate, so the exclusion test stayed green in 273ms while the page really did end up on `data-apps/`.
2. Asserting the post-click URL cannot detect a regression in the route the listing emits, because the alias rewrites the old form on the way. Reverting `Package.tsx` to emit `pages/` left all five cases green. It now records `pushState`/`replaceState` and asserts what the app actually pushed. That assertion should be deleted along with the alias.
3. `toHaveCount(1)` on an iframe cannot see a wrong target: removing the prefix strip renders an iframe that 404s inside itself and the count still passes. Both iframe assertions now check `src`.

## Not carried over from #935

Its restructured `ContentTypeIcon` (a new accent palette and a dashboard type) and its `Package.tsx` section layout, both of which belong to later slices. Only the "page" content type is renamed to "dataApp" here.

## Docs

`docs/html-data-apps.md` and `docs/api-overview.md` are updated, so this is a real break for anyone who took the endpoint up rather than a quiet one. Two figures are recaptured, because a rename that leaves its own screenshots showing the old labels is the confusion it exists to remove: `docs/screenshots/console.png` (the section beside "Notebooks" becomes "Data Apps") and `docs/screenshots/api-explorer.png`, which renders the `api-doc.yaml` overview block this commit edits and was additionally stale from before #969 rewrote it.

**Known and not fixed here:** `docs/malloy-publisher-demo.gif`, the README hero, still shows the old "Governed Reports" label. It is a hand-recorded screencast with no entry in `capture-screenshots.mjs`, and re-recording a product demo is not this slice's job. It is already stale on `main` for unrelated reasons. Worth its own issue.

## Downstream

The service-side sync is [ms2data/service#5948](https://github.com/ms2data/service/issues/5948), and it must not land before this does, or the lockstep deploy window returns. Worth knowing for whoever picks it up: the portal builds its own `/pages` URL, which this alias does not cover, so that surface has its own redirect decision to make.


Sha-Bang (approved):
approving — the rename is clean and the parts I was most worried about hold up. word-diffed `server.ts` and it really is renames only: depth cap value unchanged, the realpath containment block isn't in the diff at all, and the route is still registered in the same post-CORS / post-body-parser / post-`drainingGuard` position. `resource` still emits the same value at `server.ts:764`, which matters to us because we parse it with a regex downstream.

two things I'd fix before it merges, neither a code bug:

**the playwright job isn't running on this PR.** `app-playwright.yml` is gated `on: pull_request: branches: [main]` and this targets `monty/pr935-slice1-renames`, so `package-asset-urls.spec.ts` — the only automated coverage of the SPA route change, and a file this PR edits — hasn't run. #972 has the job because it targets `main`. retargeting fires the `edited` activity type, which isn't in `pull_request`'s default set, so it may not run after #972 merges either. worth pushing a commit post-retarget to force it.

**the release note and the "how it was verified" bullet disagree, and the release note is the right one.** ran `classifySpaFallback` directly:

```
/examples/storefront/pages/index.html => {"kind":"redirect","location":"/environments/examples/packages/storefront/pages/index.html"}
/examples/storefront/pages/report     => {"kind":"spa"}
```

`.html` is in `ASSET_EXTENSIONS` and `pages` is no longer in `SPA_OWNED_SEGMENTS`, so an old bookmark takes the redirect branch → 302 → static route → 404. "Unrecognized file type" is what you get from in-app client-side nav, or from a path with a non-asset extension. worth correcting since it's offered as evidence.

one non-blocking follow-on from that: after the 302 the static route serves `public/pages/<file>`, so a package that ships a `public/pages/` directory answers an old bookmark **200 with a different document** rather than failing. it's the exact mirror of the `public/data-apps/` collision you already cover well in the last paragraph — one symmetric sentence would round it out. nothing in-repo ships that dir, so it's hypothetical today.

also worth a thought: `spa_fallback.spec.ts:40` and `:157` both had old-shape assertions that got rewritten to the new shape rather than supplemented, so nothing pins old-URL behaviour now, and nothing asserts `GET /api/v0/.../pages` is a 404. two cheap tests would lock the two behavioural promises the release note makes.

fwiw on our end — we're mirroring this exactly, dropping `/pages` with no alias, and we've accepted the mixed-fleet 404 window during the worker rollout. so none of the "no deprecation period" objections are ours to make.

mlennie:
Follow-up, because I said the alias was retired and then built half of it. #973 is now `11b350bb`.

**What ships: the browser URL gets a one-release grace period, the REST path does not.** `/{env}/{pkg}/pages/{file}` keeps opening the data app and rewrites itself to `data-apps/{file}` in the address bar, carrying the query string and fragment so the rewrite never silently drops state a caller supplied. `GET …/packages/{pkg}/pages` stays a hard 404, exactly as you asked for. The split is on who owns the caller: your deployment reviewed this and chose the clean break, and a spec alias would cost two paths, two operationIds and two generated client methods on every future change to that listing. A bookmark has no owner to consult and the person holding it is not reading release notes.

**What actually changed my mind was your `public/pages/` note, not the inconvenience.** Chasing it down, the old bookmark did not merely 404: it redirected into the static route, so a package shipping that directory answered the old link 200 with a different document. That is the same "confident and wrong" shape #962 existed to remove, and it is worse than an error, because nothing about it looks like a failure. Claiming the app's own segment deletes that path rather than documenting it. So the strongest argument for the alias was in your non-blocking aside rather than in the original objection, and I would have missed it.

**The thing I would flag if you are the one who ends up deleting this:** it is two halves in two packages, and either one removed alone leaves a broken alias. `pages` in `SPA_OWNED_SEGMENTS` is what lets the request reach the app at all; a branch in `ModelPage.tsx` is what rewrites it. Each site now carries a comment naming the other. The Playwright case was mutation-proven against each half independently, so removing one and leaving the other fails the test either way rather than only in one direction. A `.malloy` or `.malloynb` under `pages/` is excluded and has its own test, since that can be a real model path.

**One thing worth raising beyond this PR.** I was about to name the removal release as a version number and it would have been wrong by thirty releases. `packages/*/package.json` on `main` says `0.0.209` while the newest tag is `v0.0.240`, because `release.yml` bumps the version on a release branch and never merges it back, then walks forward through `version_taken()` until it finds a free number. So the released version is not discoverable from the repo, and anything that reads it (a doc, a deprecation note, a support-policy statement) is reading a stale number. Happy to file that separately if it is news rather than known.

**Gate on `11b350bb`:** typecheck 0, lint clean, prettier clean for every tracked file, unit 2274/0, integration 250/0, skills 7/367/20, and Playwright 71 of 73, the two failures being the CI-fixture cases described below (73/73 with the stub injected).

**One note on the base.** `main` moved to `e8a6e98f` (#978) while this was in review, so both branches are one behind, deliberately. There is no file overlap and rebasing #972 would dismiss your approval for no benefit. But #978 changes the SDK tsconfig, adds an SDK test project, and puts `test:sdk` first in the root `test` script, which newly brings this stack's renamed SDK files under test, so I checked the merged tree rather than assuming: merge clean, typecheck 0, full suite green including #978's 88 new SDK cases.

**Two corrections to my own work, since you would find them anyway and one of them is in the code.** An earlier draft of this justified carrying the query string with `?embed_token=`, and that was false: the token belongs on the standalone URL and is read by the embedded page itself, while `<DataAppViewer>` builds that URL through `packageFileUrl`, which takes no query string, so nothing forwards the outer one into the iframe. The behaviour is right and the reason I gave for it was not. And the test I added for the `.malloy` exclusion was inert: it asserted the URL was unchanged, which is satisfied on the first poll before React can navigate, so with the guard deleted it stayed green in 273ms while the page really did end up on `data-apps/`. It now asserts text the app writes after making that decision, and it fails under the mutation it previously survived.

**One thing about our own test harness that cost me two wrong conclusions.** `playwright.config.ts` sets `reuseExistingServer: true`, so a run can be served by a server still up from a previous run, which means a mutation that is in the build is not necessarily in the process answering the requests. That is how the inert test first looked like a pinning one. If you ever mutation-test through Playwright here, kill the server or use an ad-hoc port first.

**Security pass on the redirect, since the target interpolates three URL-controlled values.** Clean. A protocol-relative `//host/...` request never reaches the server because the browser resolves it first, `%2f%2f` stays a single segment, and `pages/../` and `pages/%2e%2e%2f` classify as `spa`, which sends a fixed file and joins nothing onto a filesystem path. No probe left the origin. The alias grants no new file access either, since it lands on the same viewer and the same static route with its realpath containment untouched. Both names are `encodeURIComponent`d now, matching the "Back to" link beside them which already was.

**On the two failing browser tests, in case you see them:** `environment-connections` needs the stub `bigquery` connection that `app-playwright.yml` injects with `jq`, so it cannot pass locally; I confirmed that by injecting the same stub, after which it is 5/5. And `package-materializations.spec.ts:126` and `package-databases.spec.ts:21` time out only on cold full-suite runs and pass in isolation and warm. I settled that by running the whole suite on #972, which has none of this change, and getting the identical shape.

**A second review pass found two more, and one of them is a general point about aliases rather than about this PR.** Auditing the numeric claims in my own commit message (a stale one gave it away: "2 of 3" was written when that spec had three cases and it now has five) turned up two tests that pinned nothing.

The first is the one worth knowing. **A deprecation alias reduces the sensitivity of every test that asserts on a post-navigation URL.** I reverted `Package.tsx` to emit the old `pages/<file>` route and all five cases stayed green, because the alias rewrites it and the test only read where the click ended up. So while the alias exists, nothing detects that the listing builds the old route, and the day the alias is deleted that becomes a broken listing with no coverage. It now records `pushState`/`replaceState` and asserts the app pushed `data-apps/` and never `pages/`. That assertion is explicitly marked to be deleted with the alias, when the final URL becomes sufficient again.

The second: `toHaveCount(1)` on the iframe could not see a wrong target. Removing the `data-apps/` prefix strip in `ModelPage` left all five green, with an iframe that 404s inside itself. Both iframe assertions now check `src`, and that mutation fails two of them.

Three times in this change I wrote an assertion that was true too early or too coarsely. Each was found only by breaking the code it guarded, which is the argument for doing that on every new test rather than on the ones that look risky.

**On the flaky browser tests, now characterised properly:** one query-executing case per full local run intermittently times out, and it has been a different test each time (`package-databases:21`, `package-materializations`, `notebook-authorize:107`), each passing in isolation. That is worker contention, not a defect, and #972 with none of this change behaves identically. The two `environment-connections` failures are the `jq`-injected `bigquery` stub that only CI provides.

**@nathan-credible, this answers your alias objection as well**, and it split two ways, one of them against your comment. Everything above is addressed to Sha-Bang; this paragraph is for you. **Bookmarks: you were right, and they are aliased.** **REST `/pages`: still a hard 404**, because Sha-Bang reviewed this as the consumer and said they are mirroring the break with no alias and have accepted the mixed-fleet 404 window during the worker rollout, so I took the only known caller's preference over carrying two paths and two operationIds indefinitely. If you would rather we alias the REST path anyway, say so: it is a three-line change (Express takes an array of paths at one registration) and I have already measured that it leaves the rest of the suite green, so it is a cheap call to reverse. On the sequencing you asked about in Slack: publisher has to land first, #972 then #973, because #973 is what removes `/pages`; the service half is already open as ms2data/service#5948.

PR #974, opened 2026-08-06
fix(server): serialize the shape the caller sends, and correct six claims

mlennie:
Follow-up to #961, which merged before these three commits were pushed. Two of them fix defects that are now on `main`.

## The regression on main

#961 measures the byte cap by serializing the **full** wrapped result, unconditionally, whenever a cap is set. A `compactJson` request sends the compact rows instead, so it pays for a payload it never sends, and worse, fails on it: a result whose wrapped JSON will not serialize but whose compact rows would returns 413 for the shape the client did not ask for. I had tried to fix this twice and only removed it for the cap-disabled case, which is not the configuration anyone runs.

`getQueryResults` now takes the shape the caller is going to send, serializes exactly that one, measures it, and returns it as a plain string. That removes the regression at every cap setting, the double serialization on the compact path, and the two closures whose lifetime became the caller's problem.

One deliberate behaviour change comes with it, and it is worth an explicit look: **the byte cap now measures the payload actually transmitted.** A `compactJson` request used to be capped against the larger wrapped shape, so it could be refused on bytes the client would never receive. Capping what is sent is the defensible reading, and the row cap is unchanged and applies to both shapes. If you would rather keep capping the wrapped shape regardless, say so and I will invert it.

## The other defect on main

The compact path was still stringified in the controller, unguarded, so with the byte cap disabled an oversized `compactJson` request returned exactly the error #961 exists to remove:

```
before: 500 {"code":500,"message":"Out of memory"}
after:  413 Query response could not be serialized: the 25356-row result is too large
        to turn into JSON. Project fewer columns, add a LIMIT, or filter wide values.
```

`stringifyQueryResponse` now carries the replacer through, so the compact shape gets the same guard as the full one.

## Claims that were wrong on main, all mine

- The controller comment said neither request pays for a payload it does not send. False at the default cap, which is exactly where it mattered.
- The module header's purity carve-out named the wrong helper: `assertWithinModelResponseLimits` records the same metric, so two of the four helpers are impure, not one.
- The header also described the serializer as building bytes for the byte check, when it now produces every transmitted payload.
- The scraped Prometheus HELP text had an unbalanced parenthesis.
- `api-doc.yaml`'s shared 413 description advertised the unserializable case on the two connection-SQL endpoints, which have no such guard and answer 500. It now names the endpoints it applies to, and stops describing the engine limit as V8's string limit when JSC fails an allocation instead.
- Two spec fixtures asserted a message the code cannot produce (the cap-naming wording removed in #961's second commit), which a third spec explicitly forbids. Both now use the real message, so the MCP classification test pins something rather than nothing.

Also stated rather than implied: the MCP tool builds its own envelope with an unguarded stringify, so its 413 arrives only via the byte measurement, and with the cap disabled an unserializable result still reaches the agent as an internal error. Guarding that belongs with the envelope's own truncation budget, not here, so it is not in this PR.

## Verification

Live against `examples/storefront`, at the default cap and with the cap disabled, for both response shapes: oversized returns 413 in every combination and ordinary queries return 200 with unchanged bodies. Byte identity of the serialization change was measured rather than argued, across four queries including annotation-heavy dashboard views: responses are identical once Malloy's per-run `reference_id` UUIDs are normalised, with a same-code control confirming the normalisation hides nothing.

Rebased onto current `main`, so it carries none of the merged content. Locally: lint, typecheck and prettier clean, unit 1914 pass. The integration suite passed on the pre-rebase tree and was still running on the rebased one when this went up, so treat CI as the authority on that leg.

Everything here came out of independent review rather than from me: two adversarial lenses and a text lens over the merged diff, after #961 was already green and approved.


packages/server/src/service/model.ts
```diff
@@ -2619,21 +2633,36 @@ export class Model {
...
-      // twice per request, once to measure and once to send.
+      // What this does NOT cover: the MCP tool leaves `responseShape` at its
+      // default, so the full shape is measured as it always was, but the tool
+      // then builds its own envelope (mcp/query_envelope.ts) with an unguarded
+      // `JSON.stringify`. An unserializable result reaches the 413 there only via
+      // this measurement, which means only when a byte cap is set; with the cap
+      // disabled that envelope build still throws a raw RangeError and the agent
+      // gets an internal error. Guarding it belongs with the envelope's own
+      // truncation budget, not here.
       const serializedResult = stringifyQueryResponse(
```
kylenesbit, line 2836: The serialize still runs before the row check, so a row-cap overflow can be reported as "unserializable".

`assertWithinModelResponseLimits` does not run until a few lines below, and its row branch needs only `totalRows`. Since `resolveModelQueryRowLimit` clamps to `maxRows + 1`, the row-overflow case is a 100,001-row result that gets fully serialized here and then thrown away. If that serialize is what hits the engine limit, the caller is told "could not be serialized… project fewer columns" and the counter ticks `unserializable`, when the true answer was "you exceeded PUBLISHER_MAX_QUERY_ROWS, raise it or add a LIMIT" — the exact operator-dashboard confusion the new label exists to prevent. It is also the largest single allocation on the path, on the path whose comment above calls the row cap the primary OOM defense.

Pre-existing from #961 rather than introduced here, but this PR is rewriting these lines and its thesis is not building payloads nobody needs, so checking rows first (or splitting the assert into row and byte halves) belongs with it. The notebook-cell path further down has the same ordering.

packages/server/src/service/model.ts
```diff
@@ -2130,11 +2131,24 @@ export class Model {
...
       queryMetadataInput?: ModelQueryMetadataInput,
+      /**
+       * Which shape the caller is going to send, so exactly that one is built and
+       * measured. `"compact"` is the `compactJson` response (the rows, with the
+       * bigint replacer); `"full"` is the wrapped result. A caller that sends
+       * neither verbatim, like the MCP tool, leaves it at the default: the byte
+       * cap still has to measure something, and the full shape is what it has
+       * always measured.
+       */
+      responseShape: "full" | "compact" = "full",
```
kylenesbit, line 2251: `responseShape` appears in no spec in the repo, so neither property this commit is about is pinned.

The two `query_givens` integration specs do send `compactJson: true` and parse `body.result` as rows, so returning the wrong *shape* would fail CI. But nothing asserts that a `compactJson` request is capped against the compact bytes, and nothing covers the compact-payload guard the second commit added. Given this plumbing has been wrong in three consecutive commits, that is the gap I would close before merging.

The harness is already there: `buildModelWithFakeRun` in `model.spec.ts` hardcodes `data: { value: [] }`, so adding a `compactRows` option plus one test next to the existing byte-cap cases pins it in roughly fifteen lines — huge `wrappedJson`, small compact rows, a 100-byte cap, then `"compact"` resolves with `serializedResult` equal to the compact JSON while `"full"` throws `PayloadTooLargeError`.

kylenesbit (approved):
Approving. Net against `main` this is a real simplification: one guarded `stringifyQueryResponse` call on the shape the caller declared, both intermediate closures gone, the controller out of the serialization business. Capping what is transmitted is the defensible reading, and I would not invert it.

I checked the claims that were checkable rather than taking them on trust. The two spec fixtures do match what the code emits — `stringifyQueryResponse` appends `(byte cap: 50000000)` when `maxBytes > 0`, so this round they pin the real message. The `api-doc.yaml` scoping is accurate too: `PayloadTooLarge` is referenced by exactly four endpoints, the two connection `sqlQuery` ones plus model-query and notebook-cell, and `stringifyQueryResponse` is reachable only from the latter two. Locally `tsc --noEmit` is clean after regenerating API types (so the YAML parses), prettier is clean on the TS files, the four touched specs pass, and the unit suite passes. Six integration tests failed in my sandbox, all in "Concurrent package operations (E2E)" with `Environment could not be resolved to a path`, which is my worktree lacking a server root rather than anything in this diff.

Two comments inline, on the row-check ordering and on test coverage. Two more here:

**MCP is now the only caller measuring a shape it never sends, and the fix is one argument.** `execute_query_tool.ts` leaves `responseShape` at the default, so every successful MCP query serializes the full wrapped result at the default cap and discards the string: the envelope is built from `compactResult` and stringified by `jsonResource`. And because `buildQueryEnvelope` truncates to `MAX_RESULT_CHARS` (90k) anyway, a result whose wrapped JSON measures over 50 MB gets a 413 even though the envelope the agent would have received is 90k characters. That is the same defect class this PR fixes for REST — refused on bytes the client never receives — and passing `"compact"` would remove the wasted pass, cap the shape the envelope is actually built from, and put the unserializable guard on the rows it serializes. The deferral is reasonable, but I would rather it were an issue than a code comment.

**The behaviour change is not in the docs.** `docs/configuration.md` still describes `PUBLISHER_MAX_RESPONSE_BYTES` as "Maximum JSON-serialized response size for ad-hoc SQL and model queries," and this PR's own `api-doc.yaml` text says "the JSON-serialized response measured over the cap." After this change, two clients differing only in `compactJson` hit that cap at different result sizes. That deserves a clause in both places, not only the comment in `model.ts`.

Minor, take or leave: the metric HELP string is now ~470 characters on one line, which most scrape UIs truncate; and the `configuration.md` table cell is ~450 characters and blows out the rendered table, which would read better as a footnote beneath it (that file is already not prettier-clean on `main`, and CI only checks `packages/**/*.{ts,tsx}`, so there is no CI impact either way).

mlennie:
Thanks, this was a sharper review than the PR deserved. All six taken, including the MCP one you offered to accept as a deferral.

**The row check running after the serialize.** You are right, and the consequence is worse than wasted work: the caller was told the wrong thing. A row overflow is a `maxRows + 1`-row result by construction, so it got fully serialized and discarded, and if that serialize hit the engine limit the answer became "could not be serialized, project fewer columns" with the counter ticking `unserializable`, when the true answer was "raise `PUBLISHER_MAX_QUERY_ROWS`". Exactly the dashboard confusion the separate label exists to prevent.

`assertWithinModelResponseLimits` is now split into `assertWithinModelRowLimit` and `assertWithinModelByteLimit`, and both the query path and the notebook-cell path check rows first. Verified live with `PUBLISHER_MAX_QUERY_ROWS=10`:

```
POST .../query   {"query":"run: order_items -> { group_by: id; limit: 100000 }"}
413 Query returned more than 10 rows. Refine the query (add a LIMIT or more selective WHERE) or raise PUBLISHER_MAX_QUERY_ROWS.
```

Same on the compactJson path. You were right that it was pre-existing rather than introduced here, and right that it belongs with this PR anyway.

**MCP measuring a shape it never sends.** Done rather than filed as an issue, since as you say it is one argument. Both call arms now pass `"compact"`, so the cap measures the shape the envelope is built from and the unserializable guard lands on the rows it serializes. Your point about `MAX_RESULT_CHARS` is what settled it for me: refusing a query at 50 MB when the agent would have received 90k characters is the same defect this PR fixes for REST, so leaving it as a comment would have been documenting a bug rather than fixing one.

**responseShape pinned by no spec.** Took your recipe as given. `buildModelWithFakeRun` gained a `compactRows` option, and there are three tests: a `compactJson` request resolves with `serializedResult` equal to the compact JSON where the wrapped shape throws under a 100-byte cap; an unserializable compact payload reports the 413 rather than escaping; and a row overflow reports rows even when serializing would have failed. That last one fails if the ordering is reverted, so it pins the fix rather than describing it.

**The behaviour change missing from the docs.** Both places now carry it. `docs/configuration.md` says the cap measures the response actually sent and that two clients differing only in `compactJson` reach it at different result sizes; `api-doc.yaml` says the same where it describes the byte cap.

**Both minors taken.** The metric HELP string is down from about 470 characters to under 200, so a scrape UI will show it. The metrics table cell is a short sentence plus a footnote beneath the table, rather than 450 characters that blow out the rendering.

**On your sandbox failures**: agreed, `Concurrent package operations (E2E)` needs a server root. I get the same shape whenever another server or test run is competing for the default ports and the DuckDB lock, and it is the single most common false failure in this repo for me. With nothing else running the suite is 237 pass, 0 fail.

**On capping what is transmitted**: noted that you would not invert it, and I agree. It is now stated in both docs rather than only in a comment, so if someone disagrees later they will at least find the decision written down.

**Two follow-up sweeps since you reviewed**, both on this branch, no behaviour change from either. The first was a claim audit: I had left prose next to code describing the state before my own edits, including a comment claiming a `compactJson` request no longer builds the wrapped shape when `wrapResult` runs unconditionally on the line above, an orphaned docblock for the wrapper we deleted, and a `responseShape` docstring asserting nothing relied on its default when ninety of the ninety-four spec call sites omit it.

The second found something worth flagging to you specifically, since it is about the tests you asked for. Two of them passed whether the code was right or wrong. The row-before-`wrapResult` ordering was not actually pinned, because `wrapResult` is stubbed in the spec, so moving the check back down left the suite green with an identical 413 and message; that is now asserted as "the stub was never called". And nothing pinned the bigint replacer at all: dropping it kept every test passing, yet a DuckDB `count()` arrives as a BigInt, so `JSON.stringify` throws a `TypeError` and returns the bare 500 this PR exists to remove, on any MCP query with a count in it. Both are now covered, and I mutation-tested each one (break the invariant, confirm that test fails and nothing else does, restore) rather than trusting that a passing test means a pinned one.

One note on history: this branch was rebased past #961's squash merge and again past newer main, so a force-push was needed and your inline comments may show as outdated. Nothing was dropped in the rebase; I checked that every file the previous remote head touched is still touched here.

PR #976, opened 2026-08-06
refactor(api): give givens one meaning across the spec

mlennie:
Slice 4 of the ten-PR split proposed in #935's own description. `givens` had come to mean four different things at once: declarations, typed values, string-encoded values, and a bare list of names. It now always means a collection of `Given` declarations, and the other three have names of their own. Renames and spec corrections only, no endpoint changes what it does.

The dashboard-specific schemas are deliberately held back to a later slice so they can be born correctly named, per @kylenesbit's constraint. `dashboard.ts` is not touched. Extracted from #935 with `Co-authored-by: Kyle Nesbit`.

## What changed

- **`Givens` is renamed `GivenValues`**, and the string-encoded form that survives a URL becomes a named `EncodedGivenValues`. No field, shape or wire format changes on any endpoint.
- **`Given` gains the control contract**: `label`, `control`, `rangeMin`, `rangeMax`, and `suggest` (a new `GivenSuggest`), all optional. How a given should be presented belongs to the given rather than to any one surface. No server code populates them yet; this lands the contract so the readers in slices 5 and 7 have one place to write to, rather than each inventing a name.
- **Package warnings name their subject `subject` rather than `target`**, which meant the opposite of a `# drill` target.
- **`RawNotebook` declares what the endpoint actually returns**, which is what lets the blanket `as ApiRawNotebook` cast go, plus a new `startingGivens`.
- **`NotImplementedError` maps to 501.** Every route declaring `versionId` has documented `501 Not Implemented` all along while the error fell through to the 500 default. Verified live: it returns `HTTP 501 {"code":501,"message":"Version IDs not implemented."}` now, and returned 500 before. Both throw sites carry a fixed literal message, so the status change discloses nothing new.

## Where `GivenSuggest` landed, since the dispatch left it open

Here, in this PR. `Given.suggest` is part of the control contract being folded on, so the schema it points at has to exist in the same change. That makes the description defect this PR's to fix: it said "either a named `query`, or a `source` + `dimension` pair", but `useSuggestOptions` honours `dimension` alongside `query` to select the column, and the shipped examples and docs write both. It now documents all three forms and says the first column is used when no `dimension` is given.

## Two things found while enumerating, rather than sampling

**The `target` rename is five producers, not one.** `Package.warnings` is fed by `renderTagWarnings`, `storageWarnings`, `droppedPersistWarnings`, `persistenceCollisionWarnings` (message-only), and `materializationConfigWarnings`. #935 renames the first three. The fifth defines its own hand-written `MaterializationConfigWarning { target?: string }` and is spread into the same `ApiPackageWarning[]`, and that assignment typechecks either way, because excess-property checking fires only on a fresh object literal, never on a variable of a named interface type. Left alone it would emit `target` and `subject` on different entries of one array with no compiler complaint. All five move here. Worth knowing for whoever renames the dashboard warning side later, so it does not come back.

**`RawNotebook` declared two fields the server has never sent.** `resource` and `path` were declared; the response carries neither and sends `modelPath` instead. Confirmed against a running server, which returns exactly `annotations, malloyVersion, modelInfo, modelPath, notebookCells, packageName, queries, sources, type`. So `path` becomes `modelPath` in the schema: a breaking rename of a field that never carried a value, and anything reading it was already getting `undefined`.

## On the cast, which is the fix @kylenesbit asked for in his review

His branch fixed the symptom (the server emitting `givens` where the spec said `startingGivens`) and flagged the cause for this PR. The cause is now closed rather than narrowed: `type`, `modelPath`, `modelInfo` and `queries` are declared, so the cast is gone entirely and the response literal is bound to the schema. **Proved load-bearing rather than assumed**: deleting one declared field makes `model.ts` fail with `TS2353: 'modelInfo' does not exist in type …`, then restoring it goes green.

The symptom half of his fix could not be carried, and I would rather say so than imply otherwise. On `main` the notebook response emits neither `givens` nor `startingGivens`, because the producer that reads the file-level tag arrives in a later slice. This PR lands the field name so that slice has one to write to.

For the same reason, #935's `RELEASE_NOTES.md` line 124 (starting values "arrive as one field with one name: `givens` on `RawNotebook`") is **not** fixed here: it sits in a release note describing that unbuilt reader, and this PR does not land that section. The field is `startingGivens`, as line 235 of the same document says. Whoever lands the notebook starting-values slice should write the corrected spelling.

## Generators

`api-doc.yaml` feeds five targets and all were regenerated and checked: server types, SDK client, CLI client (regenerated and typechecks clean), and the Python SDK (builds, 4 tests pass). `openapi-spec-validator` reports `api-doc.yaml: OK`. All four generated outputs are git-ignored in this repo, so nothing tracked drifts.

**The symbol rename is breaking for a generated client that names it, and which clients those are depends on the generator.** `openapi-typescript`, which generates this repo's own server types, emits a named `Givens` on main and now emits `GivenValues` and `EncodedGivenValues` instead. The Python generator behaves the same way: the build now emits `given_values.py`, `encoded_given_values.py`, `given_suggest.py` and `given_control.py`, and `givens.py` is gone. The axios generator behind `@malloy-publisher/sdk` inlines the map and names nothing, so an SDK consumer is unaffected. All checked output, regenerated from each spec rather than assumed.

An earlier version of this description had that backwards, claiming the TypeScript generator named nothing. @kylenesbit caught it, and it mattered: it is the sentence a downstream consumer would use to decide the rename is a no-op for them. It was a sentence I carried over from #935's own release note without regenerating to check.

## Merged against current main, not just built on the branch

Following the lesson from #972, where 14 green checks coexisted with a broken merge because only one side had touched the file. `git merge-tree --write-tree origin/main HEAD` exits 0, and the merged spec contains no `schemas/Givens` reference, no `Givens:` schema and no warning `target:`, with all three new schemas present.

Also checked against the two open slices that contend for this file. Against #969: clean. Against #972 and #973: `api-doc.yaml` **auto-merges with zero conflict markers** and carries both slices' changes together, but `RELEASE_NOTES.md` conflicts, because slice 1 and this one each insert a new `## [Unreleased]` section at the same anchor. It is an adjacent-insert conflict only; the resolution is to keep both sections, and whoever lands second resolves it.

**Main then moved under this branch three times while the review was in flight**, and each time it touched files in this diff. #967 and #969 (this split's own slices 3 and 2) merged, then #960, then #974, which rewrote 103 lines of `model.ts` and touched both `errors.ts` and `errors.spec.ts`, the two files the 501 change edits. Every one of those merges reported no conflict, so each was read by eye rather than trusted: #967's `ownModelNotes` line sits directly above the literal whose cast this PR removes and both survive; #960's `InvalidArgumentError extends BadRequestError` sits behind the chain's first branch, so the 501 arm at the end cannot shadow it; #974's `ResponseUnserializableError` comment and test coexist with the 501 arm and its test. There are zero `as ApiRawNotebook` casts left in `model.ts`.

## Review round

Everything @kylenesbit raised is addressed. The generator correction is above. Also:

- **The 501 mapping now has a test** in `errors.spec.ts`, alongside the other per-error rows, proved by removing the branch and watching it fail `Expected: 501 Received: 500`.
- **`hammer/lib/scenario_md.ts`** still declared the warnings it reads as `{ model, target, message }`. Nothing failed, for the reasons he gives, but it is the same stale-shape miss on the reading side, so it moves too.
- **The render findings are spelled out rather than spread**, at both call sites. He flagged this as "nothing to change", but the PR's whole argument is that a rename needs a compiler backstop, and he had shown there was none: reverting `subject` end to end still typechecked, because a spread is exempt from excess-property checking exactly like the named interface this PR calls out. Leaving that hole in the change that sells closing it was not defensible. The same revert now fails at both sites.
- **Three spec descriptions.** The `givens` query parameter names `GivenValues` (checked against `server.ts`, which does `JSON.parse(req.query.givens)` into `Record<string, GivenValue>`); `EncodedGivenValues` no longer describes in the present tense a role nothing fills; and `rangeMin`/`rangeMax` state the rule that signals a slider. On that last one, note the rule is phrased as the contract rather than as live behaviour, because `GivenInput` on `main` has no slider at all: the condition it describes (`number` or `filter<number>` with both bounds defined) is read off the renderer on the source branch.

**A gap the review round itself introduced, found by re-reviewing that commit.** Spelling the fields out changed two call sites, and only the load path was covered end to end: the reload test asserted `model` and never `subject`, so a wrong value at the site just edited would have passed. Fixed by strengthening that test, proved by putting `w.message` into `subject` at the reload site and watching it fail, then proved live by registering a clean package, breaking its tag on disk and reloading.

## Verification

Full gate in the order this repo needs (`generate-api-types && build:sdk` first, since the root `typecheck` chains neither), re-run after each main merge: lint exit 0 with zero `--fix` mutations, typecheck exit 0, `prettier:check` red only on git-ignored generated files, unit **2055 pass / 0 fail** across 108 files, integration **245 pass / 0 fail** across 33 files. `openapi-spec-validator` OK, the Python SDK builds and its tests pass, and the CLI client regenerates and typechecks.

Four claims proved live on the built production bundle rather than by inspection: the `versionId` 501, the notebook response's exact key set, a package warning carrying `subject` over HTTP on the load path, and the same on the reload path.

## Follow-up, filed separately

The sibling `Notebook` (listing) schema has the same shape of gap: the listing response sends `environmentName`, which the schema does not declare, and declares `resource`, which the response did not carry. Filed as its own issue rather than folded in here, per @kylenesbit's request, since a later slice adds `title` and `description` to that same schema.


kylenesbit (approved):
Approving. I reviewed this against a clean worktree of the branch and re-ran the checks it claims. The slice is sound, and every verification claim I could test holds — with one exception worth fixing before merge, since it is the sentence a downstream consumer would rely on to decide the rename is a no-op for them.

## What I confirmed independently

Typecheck clean across all five packages; the touched unit tests pass (94 tests); `redocly lint` on the spec reports findings identical to `main` (52 pre-existing `security-defined` errors, same rules, nothing new); `git merge-tree` against current `main` (`7bc977cc`) exits 0.

The cast removal is genuinely load-bearing. Renaming `modelPath` back to `path` in `api-doc.yaml` and regenerating produces exactly the promised failure:

```
src/service/model.ts(2987,10): error TS2353: Object literal may only specify known
properties, and 'modelPath' does not exist in type '{ type?: string ... }'
```

The 501 claim checks out. I walked every operation in the spec that accepts `versionId`, in the query string or in the request body: all 8 declare `501`, none are left declaring only `500`, and there is a single centralized throw site going through `internalErrorToHttpError`. Dropping `RawNotebook.resource` and `path` is safe: `getNotebookModel` is the only producer and no reader in the app, SDK, or CLI touches either field.

## One thing to fix: the release note is backwards about the TypeScript generator

`openapi-typescript` does emit a named symbol for these maps. On `main`, the generated `packages/server/src/api.ts` contains:

```ts
/** @description Per-query given values that override model defaults... */
Givens: {
  [key: string]: unknown;
};
```

On this branch that symbol is gone, replaced by `GivenValues` and `EncodedGivenValues`. So "The TypeScript generator emits no symbol for any of the three … but a generator for another language may name them" is inverted for the generator this repo uses for its own server types.

It is harmless in-repo — nothing referenced `components["schemas"]["Givens"]` by name, and the file is generated and git-ignored — but a downstream TypeScript client generated the same way takes a breaking symbol rename. The claim is true only of the axios SDK generator, which does inline the map as `{ [key: string]: string; }`. Worth correcting in `RELEASE_NOTES.md` and the PR body, since the migration advice hinges on it.

## Two small follow-ups

**One consumer of the old key survives.** `hammer/lib/scenario_md.ts` (~line 2089) declares the warnings it reads as `{ model?, target?, message? }`. Nothing fails: the `## Warns` step stringifies the whole array and substring-matches, and `hammer` is not in the root `typecheck` script, so no compiler sees it. Same hand-written-type miss this PR describes catching for `MaterializationConfigWarning`, on the reading side.

**The new 501 mapping ships without a test.** `errors.spec.ts` is a per-error mapping table (422, 400, 403, 404, 424, …), and this is the PR's only user-visible behavior change. Five lines.

## Worth knowing, nothing to change

The warnings rename has no compiler backstop anywhere. I reverted `RenderTagWarning.subject` to `target` end to end, interface and producer both, and `bun run typecheck` still exits 0: `package.ts` merges the finding with a spread, `renderTagWarnings.push({ model: sm.modelPath, ...w })`, and spreads are exempt from excess-property checking exactly like the named interface this PR calls out. The rename is held in place only by the unit tests asserting `w.subject` — which exist and pass, so this is fine as it stands. Spelling the fields out instead of spreading would put the compiler on it.

## Three spec-description points, since this PR is the vocabulary

The thesis is that `givens` always names a collection of `Given` declarations, and one endpoint parameter still contradicts it: `GET …/notebooks/{path}/cells/{index}?givens=` is declared `type: string`, "JSON-encoded given values keyed by given name". The server does `JSON.parse(req.query.givens)` into `Record<string, GivenValue>` and the SDK builds it with `JSON.stringify(params)`, so it carries serialized `GivenValues` — a third transport form, under the contested name. Renaming the parameter would break callers, but pointing its description at the named schema ("a JSON-encoded `GivenValues` object") costs nothing and makes the claim true of the spec as written.

Relatedly, `EncodedGivenValues` says in the present tense "This is what a surface's URL parameters carry and what a document declares as its starting values," and nothing emits it yet: the only URL-parameter carrier is the JSON blob above, and `startingGivens` is unpopulated. The control fields got an explicit "the server does not populate them yet"; this deserves the same. On the source branch it read "a dashboard's URL parameters," concrete because that surface existed there; generalized to "a surface's," it now reads as though it describes the REST API.

Last, `control` is `select | multiselect` and `rangeMin`/`rangeMax` are described as "Slider" bounds, but nothing says how a slider is signalled. The enum is right — `GivenInput` renders a slider when both bounds are present rather than switching on `control` — but a client reading only the spec cannot derive that rule, and this PR is the contract the later readers write against, so it should state it.

On the follow-up you flagged: yes, please file the sibling `Notebook` listing-schema gap as an issue rather than folding it in here.

mlennie:
Thanks, that is a genuinely useful review, and re-running the claims rather than reading them is what caught the one that was wrong. All of it is addressed. Head is now `4f5e6986`.

**The generator claim.** You are right, and I checked it in both directions before fixing rather than taking it on your word: regenerating from main's spec produces a named `Givens`, and this branch replaces it with `GivenValues` and `EncodedGivenValues`. The axios SDK generator inlines and names nothing, so an SDK consumer is unaffected while a client generated the openapi-typescript way takes a breaking symbol rename. Corrected in `RELEASE_NOTES.md`, in the migration line, and in the PR description.

Worth naming what went wrong, because it is not a typo: I carried that sentence over from #935's own release note without regenerating to check it. Restating a claim I had not measured, on the one sentence a consumer would act on.

**Both follow-ups taken.** `errors.spec.ts` gains a 501 row, proved by removing the branch and watching it fail `Expected: 501 Received: 500`. `hammer/lib/scenario_md.ts` moves to `subject` too.

**Your "nothing to change" item, taken anyway.** You showed the rename had no compiler backstop, since a spread is exempt from excess-property checking exactly like the named interface this PR calls out. Leaving that hole in the change whose argument is that renames need a backstop was not defensible, so both call sites now spell the fields out, and the revert you ran fails at both.

That change then produced its own defect, which a re-review of it caught: it touched two call sites and only the load path was covered end to end. The reload test asserted `model` and never `subject`, so a wrong value at the site I had just edited would have passed. Strengthened, proved by putting `w.message` into `subject` at the reload site and watching it fail, then proved live by breaking a package's tag on disk and reloading.

**All three spec descriptions done.** One note on the slider rule: it is phrased as the contract rather than as current behaviour, because `GivenInput` on `main` has no slider at all. Your rule is exactly right for the source branch, and I read the condition off it (`number` or `filter<number>` with both bounds defined) rather than restating it, having just been caught doing that once.

**Main moved three times while this was in review**, each time touching files in the diff: #967 and #969, then #960, then #974, which rewrote 103 lines of `model.ts` and touched both `errors.ts` and `errors.spec.ts`. Every merge reported no conflict, so each was read by eye on the same principle as your #972 finding. Gate re-run after each: unit 2055/0, integration 245/0, spec validator OK, Python SDK green, and the four live claims re-proved on the rebuilt bundle including the reload path.

Filing the `Notebook` listing-schema gap separately as you asked. Two things in it worth flagging here: the listing also sends `environmentName`, which the schema does not declare, so it is a gap in both directions rather than just the `resource` half; and I have left `error` in it as an open question rather than a finding, since I did not check whether anything produces it.

PR #977, opened 2026-08-07
feat(server): incremental materialization — advance a persisted source by a bounded delta

kylenesbit:
## What

A persisted source can now declare that a refresh applies a **bounded delta** instead of rebuilding the whole table:

```malloy
#@ persist name="daily_orders" refresh="incremental" watermark="order_date"
source: daily_orders is orders -> {
  group_by: order_date
  aggregate: revenue is amount.sum()
}
```

Publisher records how far the table is materialized (a `covered_through` boundary, keyed by the source's content address) and each refresh recomputes only `[covered_through, frontier)` — half-open, so frontier rows still arriving are left for the next run. `merge_key="col,…"` opts into `MERGE` on identity columns for rows that can be restated with a new watermark value; without it the range is replaced wholesale (`DELETE` + `INSERT`, in one transaction). Postgres, BigQuery, and Snowflake only, gated at publish.

## How it decides

Every decision falls back to a full rebuild (SEED), which is always correct and merely expensive:

- **SEED** when no boundary is recorded, the boundary describes a different table/watermark/strategy, the target's columns no longer match what the source computes, an API build passes `forceRefresh`, or `merge_key=` lands on Postgres 14 (MERGE needs 15).
- **DELTA** when a recorded boundary matches the target and the frontier has advanced.
- **SKIP** when the frontier has not advanced.

Two auto-run integration points make deltas reachable at all:

- An incremental source is **exempt from skip-if-unchanged** — its content address does not move when data does, so the ledger, not the address, decides whether there is work.
- A scheduler-triggered `forceRefresh` defeats skip-if-unchanged but does **not** force a re-seed; an explicit API `forceRefresh` still does.

## The Postgres row_to_json bug, fixed by construction

The first delta implementation compiled a range-filtered Malloy query and inlined its SQL into the INSERT. On Postgres that fails with `column "…" does not exist`: a query's SQL is **finalized** (Malloy appends the dialect's `sqlFinalStage`, and Postgres — the only `hasFinalStage` dialect — collapses the result into one json column named `row`). This is the same root cause as #867, which upstream malloydata/malloy#2964 fixed for `PersistSource.getSQL()` — and only for it; there is no way to ask a query for unfinalized SQL.

So the delta is now composed the way the seed is built: the source's own manifest-resolved `getSQL()`, wrapped in a SQL range predicate. The delta writes the seed's shape by construction, and the publish-time trial compile, the Malloy query-text generator, and the ModelMaterializer plumbing all became dead code and are removed (net −170 lines in that commit).

Since the invariant has now bitten twice — and the previous write-up was deleted along with its patch — it is pinned in two places: a compiler-contract tripwire asserting Postgres is the only `hasFinalStage` dialect and its final stage is the `row_to_json` wrapper, and a "Materializing on Postgres" section in `docs/materialization.md`.

## Snowflake

Snowflake is on the allowlist (including `merge_key=`), and three of its behaviors need dialect-specific handling, each pinned by a test:

- **One statement per driver call, each on a possibly different pooled session** — so the range replace cannot travel as a `BEGIN;…;COMMIT;` script. It is carried as a single Snowflake Scripting anonymous block (`EXECUTE IMMEDIATE $$…$$`) whose `EXCEPTION` handler rolls back *inside* the statement; a follow-up `ROLLBACK` call would reach a different session than the one holding the open transaction. `MERGE` stays a bare single statement.
- **Bare identifiers fold to UPPERCASE** (Postgres folds to lowercase) — so table-path decoding folds per dialect, and every probe that reads a result key quotes its output alias (an unquoted `AS watermark_max` comes back as `WATERMARK_MAX`). `probeTargetColumns` also learns Snowflake's per-database `information_schema`.
- **Backslash is an escape character in its string literals** (as on BigQuery, unlike Postgres) — string bounds and probe literals now escape per dialect.

DuckDB cannot stand in for the scripting block, so its semantics are provable only against a real Snowflake account; the statement text and the dialect facts the decoder assumes are pinned in unit and contract tests.

## Safety

- The DML runs on the source connection inside one transaction, with `ROLLBACK` on failure; a failed delta leaves the table as it was and does not advance the boundary, so the next run recomputes the same range.
- Deltas are idempotent for a fixed range: a crash between the DML and the ledger advance converges on retry.
- The content address is proven stable across seed → delta → delta (and proven to MOVE when the computation changes), so a model edit re-seeds rather than mixing two definitions' rows in one table.
- Declaring `refresh="incremental"` on an unsupported dialect, on a `storage=` build, or with a watermark that is not a real, orderable, non-aggregate output column is a publish rejection — never a silent full refresh that looks like it is advancing.

## Testing

- **Hammer, on real Postgres**: new `incremental-delta-advances` scenario proves seed → delta (a restated batch is rewritten while the frontier batch stays absent — only a delta leaves those rows) → skip → forced full rebuild. Full hammer suite green apart from the three pre-existing `known-red` scenarios.
- **Unit**: 2233 pass — planner decisions, DML generation for both strategies on all three dialects, ledger lineage checks, bound rendering/canonicalization, address stability against the real compiler, and the compiler-contract facts.
- **Integration**: load-time advisories, the publish gate, and the unsupported-dialect fallback over REST.

## Docs

`docs/materialization.md` gains an "Incremental refresh" section (declaration, frontier semantics, merge vs. range replace, what falls back and why) plus the Postgres invariant above.


girishjeswani:
@kylenesbit 

Ran this end to end locally before reviewing: full unit + integration suites, the hammer suite (green apart from the three pre-existing known-reds; incremental-delta-advances passes), and a synthetic environment against a real Postgres 16 with packages exercising both strategies side by side. Everything the PR claims held up:

merge_key= replaced a restated row in place; the keyless variant duplicated it - the advisory's documented hazard, observed live
Half-open range semantics are exact: the frontier batch stays absent after a delta, and a row landing 0.5s past the truncated range end was excluded, then picked up by the next run's range
Sabotage recovery: external TRUNCATE → table_emptied re-seed; dropping a target column → shape_mismatch naming the column; API forceRefresh re-seeds
Failed delta (a CHECK constraint added to the target mid-stream): run FAILED surfacing the real Postgres error, table intact, boundary unadvanced, retry converged with no duplicates, and the pooled connection wasn't left in an aborted transaction

Publish gate: five different malformed declarations rejected in one 400 with precise, fix-naming messages
The ledger rows and the new incremental_step_total metrics matched my session's history exactly

**Findings**, in order of substance:

1. Two incremental sources sharing a content address fight over one ledger row.
Two persist sources with byte-identical SQL but different name= collapse to one sourceEntityId → one instruction, one table (pre-existing, pinned behavior). New with this PR: when their incremental declarations differ (one declares merge_key=, one doesn't), the shared ledger row's lineage flip-flops on every run - each build re-seeds both as lineage_changed ("the strategy changed from merge to range_replace", then back), forever, and the second name's table never exists. It's logged and counted, but the author gets a permanently non-advancing incremental source with no publish-time signal. Suggest the gate (or at least an advisory) flag two incremental-declaring sources resolving to one address with differing declarations - cheap to detect where collectIncrementalDeclarations runs, since the addresses are already computed there.
Repro: duplicate a persisted query source under a second name, add merge_key= to one, build twice, watch lineage_changed alternate.

2. Ledger rows orphan on every model edit.
An SQL change moves the address; the new key correctly seeds via no_boundary (and a revert-to-old-SQL is caught by the shape probe), but the old row lives until env/package delete - after one edit I had 5 rows for 4 sources. Housekeeping, not correctness: a sweep of rows whose address isn't in the current build plan would keep the table bounded.

3. Nit: step="delta" counts planned deltas, not applied ones.
recordIncrementalStep fires before applyDeltaScript, so a failed delta still increments the delta counter while the help text reads as "applied". Either count on completion or say "planned" in the help string.

4. Scope question: orchestrated builds.
Under a host that assigns generational physical names per run, ledgerLineageMismatch on physicalTableName fires every run, so incremental never engages on orchestrated workers - safely, but silently. Is standalone/auto-run the intended scope for this phase?

The every-failure-is-a-seed discipline is visible in practice - every wrong thing I did to the serving tables produced a correct rebuild and a reason code naming why

girishjeswani (approved):

kylenesbit:
@girishjeswani — thank you, this was an unusually useful review; the live Postgres sabotage runs in particular found things no unit test was going to. All four findings are addressed, but **one change contradicts a behavior you verified**, so please re-read that part before re-approving.

## `forceRefresh` no longer re-seeds — anywhere

Your verification list includes "API `forceRefresh` re-seeds". That's gone, deliberately. The equivalent is now `reseed`:

```bash
# was: -d '{"forceRefresh": true}'
curl -X POST .../materializations -d '{"reseed": true}'
malloy-pub materialize --reseed --environment default --package iv
```

`forceRefresh` had accumulated two meanings, and the second never fit: an incremental source is exempt from skip-if-unchanged anyway, so the flag has nothing to say about whether one builds — only ever about *how*, which is a different question. Overloading it also forced a carve-out for the scheduler (which sets the flag on every fire), and that carve-out keyed on a `trigger` an API caller can't set — which is what made finding 4 much worse than a naming problem. `forcesFullSeed` and its trigger coupling are deleted; build semantics no longer depend on who is believed to have started the run.

`reseed` exists at both levels, OR-ed: run-level in the request body, and per source on a `BuildInstruction` so a host can rebuild one source while the rest advance by delta.

## Finding 4 — orchestrated builds, and it was worse than generational naming

Not a scope question in the end. The blocker wasn't `physicalTableName`: it was that a control plane had **no way to express a routine incremental refresh at all**. `forceRefresh` was its only lever, that lever meant "re-seed" for any non-scheduler caller, and skip-if-unchanged (the thing it exists to defeat) never runs for an orchestrated build. So a host following the flag's own API description got a full rebuild every run, forever. Fixed as above, plus:

- Manifest entries now report `coveredThrough` / `coveredThroughType`, so a host reads incremental progress instead of inferring it. A delta reports where it advanced to, a skip reports the boundary still in force, a seed reports what it probed from the table it just wrote.
- Your generational-naming case keeps its correct fallback (the new table is empty, so a delta into it would drop history) but is no longer silent: it has its own reason code, `table_renamed`, distinct from `lineage_changed`.

## Finding 1 — shared content address

Feasible, and easier to hit than your repro suggests: I confirmed against the real compiler that two sources differing **only** in name and declarations produce byte-identical SQL and the same address, because the source name never reaches the generated SQL. Publish now advises, naming every source in the group from each of their perspectives, and distinguishes declarations that disagree (nothing ever advances) from ones that agree (merely redundant — one table built, the other `name=` never exists). The comparison covers exactly the facts the ledger records, so a difference that conflicts with nothing doesn't read as a conflict.

Two scoping notes: I left the same collapse between two **non-incremental** sources alone as pre-existing pinned behavior, and in auto-run the dedupe in `deriveSelfInstructions` means the flip-flop is really an orchestrated-only symptom — standalone just silently never builds the second source.

## Finding 3 — delta counter

`step="delta"` is now counted on commit, next to the duration, rather than at plan time. A planned delta that then failed was being counted as work the table received, so the delta:seed ratio read highest exactly when the feature was most broken.

## Finding 2 — deferred

Orphaned ledger rows are being left for now, by product call: environment/package delete reclaims them, and it's bounded housekeeping rather than correctness.

## Also worth knowing, since you ran the hammer suite

`incremental-delta-advances` needed updating: its last act forces a rebuild and asserts the batch the delta left out reappears, and it asked with `forceRefresh`. Under the new semantics that build would have skipped and the assertion would have failed — a red that would have looked like the delta path breaking rather than the scenario asking the wrong way. The publish step gained a `reseed` attribute and the scenario now uses it. **A re-run of that scenario against real Postgres would be the most valuable thing to re-verify**, along with the `reseed` recovery path replacing your `forceRefresh` one.

Snowflake support also landed since your review (single-statement Scripting blocks for the range-replace transaction, uppercase identifier folding, backslash escaping). All 15 PR checks are green.

girishjeswani:
@kylenesbit 

Re-verified everything on 234c71b3, reusing my round-1 environment so the new code read ledger state the old code wrote and the continuity held. Confirmed live: forceRefresh: true now produces deltas (4/4, zero seeds), reseed: true is the escape hatch (4 forced seeds, the excluded frontier batch appears), coveredThrough/coveredThroughType show up on delta, skip, and seed entries, a name= rename lands on table_renamed with the two-cause message, and the shared-address advisory fires from both sources' perspectives with the DISAGREE branch. 

The updated hammer scenario passes 5/5 against real Postgres with a rebuilt server, the full suite is 66 pass + the 3 known-reds, and the metric tally is exact - with the delta counter now provably unable to count a failed apply. The finding-4 root-cause analysis is convincing: splitting whether a source builds from how it builds is the right cut, and deleting the trigger coupling means an API caller and the scheduler finally mean the same thing by the same words.

girishjeswani (approved):

PR #978, opened 2026-08-07
test(sdk): add a DOM test harness, and fix two defects it surfaced

mlennie:
## Why

Every spec in `packages/sdk` was a pure function, because the package had no DOM: no `@testing-library/*`, no jsdom, no happy-dom anywhere in the repo. React components and hooks were not under-tested, they were **untestable**.

The clearest evidence is #935 itself. It is the largest PR in the split, it rewrites the given controls, and it contains **zero** component tests and no DOM environment, because it could not write one. This lands the harness on its own so the slice that reworks the shared given and drill primitives can be tested rather than only Playwright-checked. It takes nothing from `pr-935` and is independent of the other slices.

## What is here

`packages/sdk/bunfig.toml` preloads `test/setup.ts`, which registers happy-dom onto `globalThis` and installs Testing Library's `cleanup`.

**happy-dom rather than jsdom** because `@happy-dom/global-registrator` is built to install into an existing runtime, which is exactly what a Bun preload can do. jsdom expects to own the test environment and has no equivalent hook, so using it would have meant running a second test runner alongside `bun test`. Every MUI widget in `GivenInput` renders under it, including `@mui/x-date-pickers` DatePicker, which was the one I expected to break.

Two details in the setup file are load-bearing and commented there. Testing Library is imported dynamically **after** `GlobalRegistrator.register()`, because it binds to `document` at module scope and a static import hoists above the line that creates one. And the setup file lives outside `src/` because `vite-plugin-dts` emits declarations for everything under `src/`.

`packages/sdk/test/README.md` documents the pattern for both a component and a hook, since `renderHook` is the mechanism the next slice needs.

Test files are excluded from `tsconfig.json`, because that config drives the declaration build and anything left in it ships as a stray `.d.ts`. Excluded turned out to mean unchecked: `bun test` strips types without checking them and eslint here is not type-aware, so a type error in a spec was invisible to every gate. `tsconfig.spec.json` now checks exactly the excluded set and `typecheck:sdk` runs it, so it needs no new CI step. That was not hypothetical: the first spec written against this harness had a real type error, and the new config was proved to catch it by injecting one and watching the gate fail.

## SDK tests did not run anywhere, so this wires them in

Worth calling out separately because it was not obvious. The root `test` script chained server, skills, create-malloy-package and hammer, with **no `test:sdk`**, and `cross-platform-tests.yml` ran the same four. So the nine existing SDK specs never executed in CI, and a harness added without wiring would not have either. This adds `test:sdk` to the root chain and a step to the workflow, both placed first because at roughly 0.3s it is the fastest suite and gives fast-fail ahead of the much longer server leg.

## Two live defects, found by the first tests, both on published exports

**`useGivensForm` re-renders forever when `givens` is not memoized.** The sync effect keys on that array's identity and always returned a fresh `Map`, so state identity changed on every run, which re-rendered and re-fired the effect. Unfixed it emitted 645,277 "Maximum update depth exceeded" warnings and crashed the test process with exit 133. In-repo callers are shielded only because `useModelGivens` happens to memoize; the hook is a published SDK export, so a consumer passing an array literal is not. Fixed by returning the previous `Map` when nothing was dropped, so React can bail. `next` is `prev` filtered by the declared set and can never be larger, which makes size equality an exact test for "nothing was dropped".

**`GivenInput` hid the revert affordance on a number given whose value is not a number**, which is reachable through the component API since `GivenValue` is a union permitting a string. The clear button now tracks whether a value is overridden, matching what the string branch already did.

**The blank field itself is not fixable, and that is worth knowing.** `input[type=number]` discards a non-numeric value under the HTML value-sanitization algorithm, confirmed against real Chromium, where it yields `""` while `type=text` keeps the string. **happy-dom does not implement that sanitization and reports the raw string**, so a test asserting on the displayed value would pass here and be false in a browser. That trap is documented in `test/README.md` and in the spec's own comment, and it is why the fix stops where it does.

## On the defects this PR was expected to fix

Three were described to me as live on `main`. Checking them against the ref changed the plan, so recording it:

The `filter<number>` slider that reads "Any" while resting at the floor **does not exist on `main`**. There is no slider, no `decodeAtLeast` and no `rangeMin`; a `filter<...>` given falls through to a plain text box. That is #935 code, so it cannot be addressed here.

The blank number control is real, but its trigger is #935-only. `paramToGiven` is not on `main`, and `useGivensForm` seeds an empty map that only `GivenInput`'s own `onChange` writes, so a URL parameter cannot produce it. Confirmed in the running app: loading `?MIN_AMOUNT=12abc` leaves the field empty and unset. It is reachable through the **published component API**, which is how it is framed above and in the test.

The dead `array<...>` branch is real and slightly worse than described. It is unreachable twice over: Malloy's grammar emits only scalar parameter types, and if it emitted an array, `given.ts` renders the wire type as the bare discriminator `array`, which does not match the component's `startsWith("array<")`. A test pins both spellings. Left in place rather than deleted, since this is a test PR.

## Tests

Eleven component tests covering the three-state contract `GivenInput`'s own doc comment specifies, unset versus an explicit `""` or `false` versus a revert, which nothing verified before, plus which given types are actually reachable. Seven hook tests establishing the `renderHook` pattern.

I skipped a characterisation suite for `useGivensForm`'s wider behaviour: slice 7 deletes the hook, so those tests would die with it. The hook spec is labelled in-file as a migration note so the next slice knows it may rewrite the tests alongside the hook, and the last test in it is the one worth carrying over, since the replacement must not reintroduce the render loop.

## Verification

`packages/sdk` went from **70 pass / 0 fail across 9 files** to **88 pass / 0 fail across 11 files**. All 70 originals still pass with a global DOM registrator installed, which was the main risk in adding one.

Full root `bun run test` is green end to end: sdk 88, server unit 2064, server integration 245, skills 7, create-malloy-package 367, hammer 20. **2791 pass, 0 fail.** Gate run in the order this repo needs (`generate-api-types && build:sdk` first): lint exit 0 with zero `--fix` mutations, typecheck exit 0, `prettier:check` red on only the six standing git-ignored generated files, with the files touched here checked explicitly and clean. CI's exact invocation (`bun run --cwd packages/sdk test`, which differs from the local `cd` form and could have resolved the preload differently) was run separately and is also green.

**Verified in the running app, not only in tests.** Against the production bundle serving `examples/governed-analytics/orders.malloynb`, whose model declares `MIN_AMOUNT` as a number given, driven in real Chrome: at rest no revert affordance; at `MIN_AMOUNT=400` the affordance appears; clicking it clears the value and the cells re-run; and at `MIN_AMOUNT=0` the affordance is still present, which is the falsy-zero case behaving correctly in the real UI rather than only under happy-dom.

**Every test was mutation-tested rather than assumed meaningful.** Breaking the `!== undefined` default check, the `typeof value === "boolean"` override check, the placeholder gating, the null-deletes-the-key behaviour, and the declared-name filtering each turned exactly one test red; reverting the render-loop fix crashes the runner. A no-op control mutation left all tests green, which is what makes the other rows evidence. Both worked examples in `test/README.md` were also extracted verbatim and executed, since an unexecuted doc example is how this sort of thing goes stale.

**Rebased onto current `main`, three times.** `main` moved seven commits while this was open, including #967, #969, #976 and #980. #976 in particular renames givens vocabulary in `api-doc.yaml`, which regenerates the client these specs import `Given` from, so the whole gate above was re-run against each rebased tree rather than the original base. `Given` itself is unchanged. Rebased rather than using the Update branch button, so this stays one signed commit with its DCO trailer rather than gaining an unsigned merge commit.

**Merged-tree check** (`git merge-tree --write-tree origin/main HEAD`): exit 0, the harness devDependencies present in the merged manifest, both spec files and `tsconfig.spec.json` present, and the merged root scripts carrying both `test:sdk` and the spec typecheck.

## Three things a reviewer will notice

Adding happy-dom moved the hoisted `ws` from 5.2.4 to 8.21.2. `thrift` kept its own resolution through a nested pin, but `isomorphic-ws` declares `ws` as a peer with range `*` and has no pin, so it now resolves to 8.x where it previously got 5.x. That path is real rather than dev-only: `packages/server` depends on `@malloydata/db-databricks`, which reaches `@databricks/sql`, `thrift`, and then `isomorphic-ws`. I traced how far it actually goes: thrift's entry point does eagerly require its `ws_connection` module, so `ws@8` is loaded, but `@databricks/sql` only ever calls `createConnection` and never `createWSConnection`, so the transport that uses it is never exercised, and `ws@8` wants Node 10 or newer against this repo's floor of 20. So the practical risk looks like nil, but a test-only devDependency did change a resolution on the server's runtime graph and that is worth knowing rather than burying.

The published npm tarball gains three entries: `bunfig.toml`, `test/setup.ts` and `test/README.md`, about 17 KB against an existing 6.9 MB. I have deliberately not added a `files` field or an `.npmignore` to exclude them. This package has neither today, so it already publishes its whole source tree, including `vite.config.ts`, `tsconfig.json` and all nine pre-existing `.spec.ts` files; introducing one would change what consumers receive well beyond this PR, and `src/` currently ships. Flagging it rather than deciding it: if you would rather the harness stayed out of the tarball, that is a one-line `files` addition and I am happy to make it, but it belongs to you as a publishing-policy call.

`bun add` re-sorted two existing `@shikijs/*` entries in `packages/sdk/package.json` into its canonical alphabetical order. No content change, and left alone because the next `bun add` would just redo it.

## Not verified

The CI step runs on Windows and macOS as well as Linux, and only macOS was exercised locally. happy-dom is pure JS with no native bindings so the risk is low, but the Windows leg is genuinely unproven until CI runs.


Sha-Bang (approved):
approving. the harness looks right and closing the SDK's no-render-tests gap ahead of slice 7 is clearly worth doing — the `useGivensState` bug that only playwright caught is a good advert for it.

one thing worth calling out: this is titled `test(sdk)` but it ships two real behaviour changes to published exports. `useGivensForm.ts:58-66` returning `prev` when nothing changed kills an infinite re-render for any consumer passing an unmemoized array literal to a shipped hook — that's a genuine bug fix, and one only a render harness could have found. and `GivenInput.tsx:520-537` changes when the clear/revert button appears. both look right to me, but someone scanning `test(...)` commits for a changelog will walk straight past them.

couple of small ones:

- `packages/sdk/package.json` has no `files` allowlist, so the harness (~17KB) ships in the tarball. the package already ships its whole source tree so this changes little, but it's worth an explicit call rather than letting it ride on omission.
- adding happy-dom bumps the hoisted `ws` from 5.2.4 to 8.21.2 in `bun.lock`, reaching `packages/server` via `@malloydata/db-databricks → @databricks/sql → thrift → isomorphic-ws`. you traced it and thrift keeps its own nested 5.2.4 pin, so it looks fine — just noting it in case the databricks path ever picks up CI coverage.

mlennie:
Thanks. And you're right about the title. I'd landed on the same conclusion independently and retitled the commit to `test(sdk): add a DOM test harness, and fix two defects it surfaced`, but that force-push went in a few minutes after you approved, so you were reading the old one. I'll bring the PR title in line with it.

**Worth flagging that I force-pushed after your approval**, so what you reviewed is not quite what's there now. It's 9 files, +68/-56, and none of it is behavioural: the two source files changed only in their comments. What moved was a review round that found the harness itself had holes:

- The tsconfig exclude/include globs did not cover everything `bun test` actually discovers. I had the dotted `.spec`/`.test` forms in `.ts` and `.tsx`, then found the underscore spellings (`foo_spec.ts`) were uncovered, then found `.mts`/`.cts` were too. It is now four extension-open patterns (`*.spec.*`, `*.test.*`, `*_spec.*`, `*_test.*`), checked by planting all sixteen combinations and confirming which program each one lands in. A `foo_spec.ts` would previously have shipped a declaration into `dist/` while being typechecked by nothing.
- Test files were excluded from `tsconfig.json` and therefore checked by nothing at all, since `bun test` strips types without checking them and eslint here is not type-aware. There is now a `tsconfig.spec.json` covering exactly that set, run as the second half of `typecheck:sdk` so it needs no new CI step. This was not theoretical: the first spec I wrote had a real type error and no gate caught it.
- Three places explaining why the number field renders blank were causally wrong. I had blamed `input[type=number]` value sanitization. The real cause is the component's own `typeof value === "number" ? value : ""`, which substitutes before the DOM ever sees the value. happy-dom's failure to sanitize is real, but it never fires through this component, so the warning I had written would have misled the next person.
- `lint` now covers `./test`, which was the one directory eslint never saw and happens to hold the one file that manipulates globals.

Happy to re-request review if you'd rather look again rather than carry the approval across.

**On the `files` allowlist**, agreed it deserves an explicit call rather than riding on omission, so here is mine: not in this PR, and the one-line version is a trap. `packages/sdk/.gitignore` lists `src/client`, so the generated OpenAPI client does not ship today. Adding an `.npmignore` switches npm off `.gitignore` entirely, so that generated source would start shipping. A `files` allowlist avoids that, but it changes what consumers receive more broadly, since all of `src/` ships today and someone may be depending on it. That makes it a publishing-policy decision rather than a tidy-up, and I would rather it arrive as its own change with its own reasoning than ride in on a test harness. Say the word if you would prefer it here and I will add it.

**On `ws`**, agreed. For the record of what I traced, since it is slightly worse than "resolvable": thrift's entry point does eagerly `require('./ws_connection')`, so `ws@8` is actually loaded rather than merely reachable. But `@databricks/sql` only ever calls `createConnection`, never `createWSConnection`, so nothing exercises it, and `ws@8` wants Node >= 10 against our >= 20 floor. So it loads and sits there unused. Exactly as you say, worth remembering if that path ever picks up CI coverage.

One limitation I have not been able to close: the new CI step runs on Windows and macOS as well as Linux, and only macOS was exercised locally. happy-dom is pure JS with no native bindings so I expect it to be fine, but the Windows leg is unproven until CI runs it.

PR #980, opened 2026-08-07
fix(server): read notes as well as blockNotes when resolving a source-level #(authorize) gate

Sha-Bang:
## The bug

Malloy files a source-level annotation under `annotations.blockNotes` for `#(tag)\nsource: name is ...`, but under `annotations.notes` for the multi-definition block form. Gate resolution read `blockNotes` only, so a gate written the second way was silently dropped:

```malloy
#(authorize) "false"
source: gated_lineabove is duckdb.sql("select 'CA' as state") extend {
  view: v is { group_by: state }
}

source:
  #(authorize) "false"
  gated_blockform is duckdb.sql("select 'NY' as state") extend {
    view: v is { group_by: state }
  }
```

Against `0.0.237`, same gate, differing only in placement:

| Source | `GET .../models/{path}` | `POST .../query` |
| --- | --- | --- |
| `gated_lineabove` | `authorize: ['false']` | **403** Access denied |
| `gated_blockform` | `authorize: None` | **200 — rows served** |

`#(authorize) "false"` granted access. The author's only feedback is that the query works, which reads as "my gate allows this caller" rather than "my gate was discarded".

It is the same declaration slot, keyed differently purely by syntax — confirmed by compiling both forms and inspecting the resulting `SourceDef.annotations`, not inferred. Malloy binds the annotation to that source and propagates it through `extend` onto `annotations.inherits.notes`, exactly as a line-above gate propagates via `inherits.blockNotes`.

## Scope: five sites, both halves of the feature

- **`model.ts`** (enforcement) — `gateExprsForOwnAnnotations`; and both reads in `ancestorGateExprs` (each `inherits` link, and the `sourceRegistry`-resolved declared source).
- **`source_extraction.ts`** (reporting) — `extractSourcesFromModelDef`'s own-gate read and inherited-chain loop, which feed `getSources()` and therefore the reported `sources[].authorize`.

The two halves broke independently. That is why the source both *reported* ungated and *served* rows, and why fixing either alone would leave half the bug — notably, a consumer that trusts the reported gate set would still inherit it.

`ownLevelNoteTexts(annote)` in `annotations.ts` is now the single place that reads `[...blockNotes, ...notes]` at one `AnnotationsDef` node. Deliberately not `annotationTexts`, which flattens the whole `inherits` chain — authorize needs nearest-declaring-level-wins.

## Over-gating was checked before widening the read

Widening a read that decides access can fail just as silently in the opposite direction, so each form was compiled and its landing place recorded first:

| Form | Lands in |
| --- | --- |
| line-above standalone `source:` | `blockNotes` |
| block form, gate on the item | `notes` |
| gate after the `is` (`source: x is\n#(tag)\nbase`) | `notes` |
| block form, gate on item 1 of 2 — item 2 | `annotations` **undefined** (no leak) |
| field-level (`dimension:`) annotation | struct `annotations` **undefined** (no leak) |
| view-level annotation | struct `annotations` **undefined** (no leak) |
| block-form base, extended | `inherits.notes` |
| line-above base, extended | `inherits.blockNotes` |

A source struct's `notes` — and each link of its `inherits` chain — is populated only by a source-level annotation attached to that one item. Never a field/view-level annotation, never a sibling definition in the same block. So reading both keys cannot invent a gate nobody wrote; it only stops dropping one that was written. No ordering or location discriminator was needed. Three of the new tests pin that direction explicitly.

Incidentally, the comment at `materialization_eligibility.ts:303` (which already reads both) undersells this: `notes` is not model/file-only — it is the same per-item slot as `blockNotes`, chosen by syntax rather than by scope.

## `#(filter)` deliberately left alone

The `#(filter)` walk in the same function has the identical gap: a block-form `#(filter) ... required` is dropped, so `buildFilterClause` never raises the "required filter not provided" error the author asked for. Left unfixed and noted in a comment, because closing it would start rejecting requests that live models have been serving. Happy to include it if you'd rather take that break here.

## Tests

`bun test src` from `packages/server` → **2064 pass / 0 fail**, with 9 new cases in `authorize_integration.spec.ts`. (Worth noting for anyone reproducing: run it from `packages/server`, not the repo root. Two `package.spec.ts` cases copy a fixture via the relative path `tests/fixtures/xlsx/database.xlsx` and fail on cwd alone from the root, and a root-level run also picks up pre-existing `packages/cli` failures — a missing generated `../api/generated` plus four `consoleLogSpy` assertions. None are in files this branch touches.)

Worth flagging how weak the first pass at these was. The `source_extraction.ts` half of the fix — two of the five sites — initially had **no coverage at all**: reverting both left the suite at exactly the 2061/0 the first revision of this PR quoted. `sources[].authorize` assertions cannot catch it, because the `Model` constructor overwrites that field from `entryPointGatesBySource` (the `model.ts` path), masking whatever the extractor returned. The isolating case is `fails model load for a block-form gate naming an undeclared given`, which reaches the extractor through `ownAuthorizeSources` → `validateAuthorizeProbes` and fails if either extractor site is reverted.

Two other corrections from that pass:

- One test was labelled as covering the `sourceRegistry`-resolved read. It does not reach `resolveDeclaredSource`; the denial comes from the `query_source` derivation branch. Across this spec that function only ever resolves to an annotation-less copy, so **that fifth site is fixed but unverified** — changed for consistency, not because a test pins it.
- The over-gating control asserted only its negative half, so it would have passed against a model with no gate at all. It now asserts the positive side too. A block-form gate that references a *given* is also covered now; every earlier case used a constant, which short-circuits the probe path authors actually hit.

`eslint` and `prettier` clean. `tsc --noEmit` reports 6 errors, all pre-existing and confirmed identical on `main` via stash.

## One thing to decide

This changes behaviour for models that currently load and serve, in two ways:

1. **A well-formed block-form gate starts being enforced** — queries that succeed today will 403. That is the point (the author wrote a gate and meant it), but anyone unknowingly serving data through one loses access on upgrade.
2. **A malformed block-form gate now fails the whole model load**, not just the queries against that source. Authorize parse failures deliberately propagate rather than being swallowed, so a block-form `#(authorize)` that was previously ignored — including one naming a given that does not exist — now takes down every source in that file, ungated ones included.

The second has the wider blast radius and is the one a release note most needs to say out loud.

On disclosure: I held this back initially as a possible security report, then concluded the affected syntax is obscure enough (I found no documented `source:` block example in the Malloy docs) that a normal public PR is the more useful artifact. Say the word if you'd rather handle it privately instead.

### DIFF
--- a/packages/server/src/service/annotations.ts
+++ b/packages/server/src/service/annotations.ts
@@ -149,17 +149,27 @@ export function annotationTexts(
 
 /**
  * The annotation texts declared directly on ONE node of an `AnnotationsDef`
- * chain — not its `inherits` ancestors. Malloy files a source-level
- * annotation under `blockNotes` for `#(tag)\nsource: name is ...` but under
- * `notes` for every other source-level placement — the multi-definition block
- * form (`source:\n  #(tag)\n  name is ...`) and either side of the `is`
- * (`source: name is\n  #(tag)\n  base`, malloy's `getIsNotes`). Same
- * declaration slot, different key depending only on which syntax the author
- * used (confirmed by compiling each form and inspecting the resulting
- * `SourceDef.annotations`).
- * A caller walking `inherits` by hand to find the nearest declaring level —
- * rather than flattening the whole chain via {@link annotationTexts} — needs
- * this at each link so a block-form declaration is not silently skipped.
+ * chain — not its `inherits` ancestors.
+ *
+ * A source-level annotation lands under one of two keys, decided purely by
+ * which syntax the author used:
+ *
+ *     blockNotes    #(tag)                 statement form
+ *                   source: name is ...
+ *
+ *     notes         source:                multi-definition block form
+ *                     #(tag)
+ *                     name is ...
+ *
+ *     notes         source: name is        after `is` (malloy's getIsNotes)
+ *                     #(tag)
+ *                     base
+ *
+ * It is the same declaration slot in all three — confirmed by compiling each
+ * form and inspecting the resulting `SourceDef.annotations`. So a caller that
+ * walks `inherits` by hand for the nearest declaring level, rather than
+ * flattening the whole chain via {@link annotationTexts}, has to read both keys
+ * at each link, or it silently skips the latter two forms.
  */
 export function ownLevelNoteTexts(
    annote: AnnotationsDef | undefined,
--- a/packages/server/src/service/annotations.ts
+++ b/packages/server/src/service/annotations.ts
@@ -151,3 +151,23 @@ export function annotationTexts(
    const texts = new Annotations(annote).texts();
    return texts.length > 0 ? texts : undefined;
 }
+
+/**
+ * The annotation texts declared directly on ONE node of an `AnnotationsDef`
+ * chain — not its `inherits` ancestors. Malloy files a source-level
+ * annotation under `blockNotes` for `#(tag)\nsource: name is ...` but under
+ * `notes` for the multi-definition block form (`source:\n  #(tag)\n  name is
+ * ...`): same declaration slot, different key depending only on which of the
+ * two syntaxes the author used (confirmed by compiling both against a
+ * `duckdb.sql` source and inspecting the resulting `SourceDef.annotations`).
+ * A caller walking `inherits` by hand to find the nearest declaring level —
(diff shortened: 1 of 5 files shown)


jswir (approved):
lgtm

PR #981, opened 2026-08-08
feat(server): read a given's control contract off its own tags

mlennie:
Slice 5 of the ten-PR split of #935, which its author declared a handoff rather than a merge candidate. Depends on #976 (merged), and slice 6 is blocked on this one.

## What this does

A given's control contract is declared on the `given:` itself, so a notebook, a dashboard and an SDK host can render the same control without any of them restating it. `readGivenControlSpec` derives `label`, `description`, `control`, `rangeMin`, `rangeMax` and `suggest` from the declaration's own plain-`#` tags and folds them onto every `Given` the API returns. Those tags sit in Malloy's reserved namespace, so `malloyGivenToApi` drops them from `annotations`; deriving them server-side is what lets a client read them without shipping a MOTLY parser and the filter-literal workaround alongside it.

`motly.ts` is the shared MOTLY parsing layer that reading depends on. `readAutorun` and `readStartingGivens` land with it and are covered by `given.spec.ts`, but are deliberately wired to no production call site: their call site needs a `RawNotebook.autorun` field that is not on `main`, so it belongs to slice 6. `docCommentText` and `motlyParseErrors` are unwired for the same reason. Flagging that explicitly rather than leaving a reviewer to wonder.

`drill_probe.spec.ts` compiles a real model through `runtime.loadModel` and pins that a dimension's annotations reach the query result's output field, which is what lets drill be resolved in the browser off the clicked field. If a Malloy upgrade stopped propagating them, drill would stop firing everywhere at once and no other test would say so. It is deliberately not a hand-built fixture.

## Two defects found in review and fixed here rather than carried

**`quoteFilterLiterals` was quote-blind.** It rewrote `=f'...'` anywhere on the line, including inside an already-valid quoted value. `# label="Single" description="Set when status=f'open'"` parses cleanly in Malloy (label `Single`, description `Set when status=f'open'`, zero errors), and the rewrite turned it into an unparseable line, so the given lost its entire control contract, silently and with no warning anywhere. MOTLY accepts both `"..."` and `'...'` with backslash escapes, so a regex cannot tell a bare literal from one inside a string; it now walks the text and rewrites only outside quoted values. Idempotence and the escaping of quotes inside a literal body are unchanged and pinned by tests.

**A `suggest` naming only a `dimension` was emitted.** Every documented form needs a `query` or a `source` to run against; `dimension` only picks a column out of one. A `dimension`-only block named nothing to query, so a client following the contract got an unfillable dropdown where omitting `suggest` falls back to a plain input.

## Decisions worth recording

**`description` joins the control spec, read from `# description="..."`.** It does not replace `#(description="...")`, which still works, still rides on `annotations`, and is still the only form the notebook UI renders. Both can be carried at once: they are read independently and neither overwrites the other. The reason the tag form is only now viable is specific. `docs/givens.md` weighed exactly this spelling and rejected it, correctly at the time, because plain `#` tags are filtered out before `annotations` reaches a client. This change is what makes them readable, so that rejection narrows. The section now says what the tag form does at the API while still recommending `#(description="...")` for helper text today.

**The `malformed-route` warning is not fixed here, and does not need to be.** `main` already documents it as cosmetic and deliberately kept, having tried each alternative against the compiler. Every file a fix would touch belongs to slices 7 to 9. What this slice contributes is that the clean alternative becomes viable at all: measured against the compiler, `#(description="Earliest report date to include")` warns, `#(description="Earliest")` does not, and `# description="..."` does not. Malloy's own warning text recommends writing `# ...` for a tag, which is the form this slice enables.

## Two claims this invalidates, both corrected

`api-doc.yaml` said of the control fields that no endpoint populates them yet, so they are absent from every response. One populates them now. `docs/givens.md`'s comparison table concluded the plain tag form was invisible to clients, which this change narrows, so that section is updated rather than left to drift.

## Verification

Full gate in the required order: `generate-api-types && build:sdk`, then lint, typecheck, `prettier:check` (clean apart from the six git-ignored generated files CI never sees), then unit **2085/0** and integration **245/0** run separately so neither hides the other.

Both fixes are mutation-tested: reverting the quote-awareness turns exactly the two new quote tests red, loosening the suggest guard turns exactly the suggest test red, and restoring gives 18/0. The drill probe was checked for vacuity the same way, by mutating the tag in the model source and watching two of its three assertions fail.

Verified live over HTTP on the worker-pool path, not just in unit tests. A given whose quoted description contains a filter literal returns both its label and its description where it previously returned neither; a `dimension`-only suggest is omitted while its `control` survives; an untagged given returns no control fields at all; and a given carrying only `#(description="...")` keeps it on `annotations` without populating `description`.

The merged tree was checked rather than just the branch, since a green branch can still break `main`: `git merge-tree --write-tree origin/main HEAD` exits 0 and the merged tree is byte-identical to `HEAD^{tree}`, because `origin/main` is an ancestor and 0 commits ahead. Worth re-running immediately before merge, as `main` moved six times during slice 1's review.

## Follow-ups for slice 6

A MOTLY parse failure returns an empty tag rather than no tag, so `readGivenControlSpec` cannot detect it and silently returns `{}`. `motlyParseErrors` already computes the messages and has no caller. Slice 6 owns the package-warnings channel, so it is the right place to wire them; the same wiring covers an unrecognized `# control=` value, which is dropped just as silently. Slice 6 should also carry `source_given_probe.spec.ts` from #935: it imports `readDashboardModelFacts` from `dashboard.ts` and so cannot compile until that lands.


Sha-Bang (changes requested):
hate to do this on an otherwise nice PR, and to be clear the problem isn't really your code — it's what your code newly reaches. but I think it's a blocker.

this is the first thing to feed a *given's* reserved-namespace `#` tags to `parseAnnotation`, and that parser has a prototype-pollution bug. compiled this through `@malloydata/malloy` + duckdb:

```malloy
##! experimental.givens
# label="Region" description="Where to look"
given: REGION :: string is 'US'
# __proto__ { a=b }
given: EVIL :: string is 'x'
source: orders is duckdb.sql("""select 1 as x""")
```

on this branch:

```
REGION -> {"name":"REGION",...,"label":"Region","description":"Where to look"}
EVIL   -> THREW RangeError: Maximum call stack size exceeded
Object.prototype polluted with: [ "location", "properties" ]
```

on `origin/main`, same model, both givens come back fine and `Object.prototype` stays clean — because `malloyGivenToApi` there drops reserved-route notes without parsing them. so the reachability is new here: `motly.ts:151` ← `given.ts:117` ← `given.ts:205`, which sits on the model-read and package-load paths (`model.ts:1289`, `package_load_worker.ts:632,818`).

the sting is that the pollution is process-wide and sticky — a benign `# label="ok"` parsed *after* the hostile line also throws. we run publisher as a shared multi-tenant worker, so that's one tenant's five characters taking out tag parsing for every package on the pod until it restarts.

root cause looks upstream: `malloy-tag`'s `properties` is a plain object, so `properties["__proto__"] = …` walks into `Object.prototype`. worth an issue there (`Object.create(null)` or a `Map`). but I'd want a guard here too, since this is the change that makes it reachable — drop annotation lines whose MOTLY declares a `__proto__` property before parsing. worth knowing a `try/catch` alone isn't enough, the pollution happens before the throw. would make a good regression test in `given.spec.ts`.

separately `readStartingGivens` has the same shape of problem — `## givens { __proto__=x }` yields the polluted prototype's entries and dies on `value.text is not a function`. unwired today so it's harmless, but it'll bite the moment slice 6 wires it, which is a decent argument against landing it unwired.

smaller stuff, all comment-grade:

- `# range_min=-1e400` → `Infinity` → serializes as JSON `null`, but the spec declares `type: number` with no `nullable`. a `Number.isFinite` guard covers it. our gson client tolerates it; a TS client typed `rangeMin?: number` won't.
- the api-doc line "each of those six is present only when the declaration carries the tag it reads" isn't quite true in the direction that bites — `# control=radio` and `# range_min=abc` both carry the tag and silently produce nothing. one sentence saying unrecognized values get dropped would cover it.
- `drill_probe.spec.ts` is a good test, but it's entirely about `# drill` propagation — feels like it belongs with the drill/dashboards slice rather than here.
- `quoteFilterLiterals` reads as fixing a defect, but it doesn't exist on `main` — it's new here, carried from #935. "carried forward with the defect fixed" would save someone hunting for a regression in a shipped release.

for what it's worth I tried hard to break `quoteFilterLiterals` and couldn't: idempotent everywhere I poked, escape ordering right, can't corrupt a valid line (a stray unquoted apostrophe is already a MOTLY error), and the `f'say "hi"'` round-trip is exact. the design — contract on the declaration, derived server-side because plain `#` tags get filtered out of `annotations` — is right, and the mutation-testing claims reproduce exactly as written.

one downstream note for us, not something you need to do: `Given.description` is new, and our dataplane spec copy doesn't declare it. our generated client is strict, so we'll sync before we bump. flagging only so it's visible.

Sha-Bang:
Description could use an update, doesn't talk about the 2 security guards. (nit)
drill_probe.spec.ts comment is unaddressed (nit)

I'm digging into the monly default tags, it makes me nervous.  If we are accepting security gaps we should at least have an issue(Private Vulnerability Reporting tagged private issue) tracking it.

Sha-Bang:
the block is cleared — thanks for the thorough job on it. I tried to break the guard rather than read it, and couldn't: on top of your fixtures I threw backtick-escaped `` `__prot\o__` ``, nested `constructor { constructor { … } }`, `toString`/`hasOwnProperty` chain hops, `constructor { prototype { … } }`, an array-index write, and a `__proto__` block carrying a live sibling. raw `parseAnnotation` pollutes on 12 of those 14 lines; through `motlyTag` all 14 are contained, `Object.prototype`/`Object`/`Function` stay clean, and a benign parse still works after every one. the derivation behind `pollutionTargets` holds up too — against the parser's real property walk (below), after the first escape it creates a `.properties` bag *on* the escaped object rather than walking further, so "Object.prototype plus the values of its own properties" really is the complete set. cost is in the noise, and a 200-given package load lands around 1.4ms.

that said, chasing the root cause changed my picture of what we're guarding, and I think it's worth you knowing before slice 6 builds on this.

**the pollution doesn't need this PR.** `##! __proto__ { a=b }` in any model file reaches it at *compile* time on `main` today:

```
tenant A, innocent package   → compile ok
tenant B, one hostile line   → RangeError, Object.prototype += [location, properties]
tenant A again, unchanged    → THREW ×3        ← collateral
any query compile            → THREW
```

`#@ persist { __proto__ { … } }` does the same. plain `#`/`##` tags don't — only the two routes the compiler parses eagerly. so it's a tenant-authored, process-wide, sticky DoS on a shared worker, and `motly.ts` can't see it: the write happens inside `motly-ts-parser` during `getModel()`, before any of our `parseAnnotation` calls run. worth not describing this slice as closing it.

root cause is `motly-ts-parser/build/parser/src/interpreter.js`, `buildAccessPath` — and it's the same code `parseAnnotation` runs, since `malloy-tag`'s parser is a wrapper over `MOTLYSession`:

```js
if (props[segment] === undefined) { ... }   // 549: "__proto__" is NOT undefined
current = ensureDataNode(props, segment);   // 554: current = Object.prototype
setFirstLocation(current, ctx, span);       // 556: writes `location`
// then getOrCreateProperties(current):     // 848: writes `properties`
if (!node.properties) node.properties = {};
```

exactly the two keys we both saw. plain-`{}` bags plus a prototype-chain-aware read.

I patched the built parser to make every MOTLY bag `Object.create(null)` (`getOrCreateProperties`, `applyClearAll`, `clone.js`) and measured it: all 8 hostile placements compile clean, zero pollution, zero collateral. publisher's existing `src/service` suite — 1547 tests, none of them the guard's own — stays green under it, so the change doesn't regress what's on `main` today. it closes the `parseAnnotation` path from my first review too, and `# label="ok" __proto__=x` comes back as `{label:"ok"}` rather than the fail-closed `{}`. upstream is `malloydata/motly`, `bindings/typescript/parser`.

two things that follow for you. once it lands, `parseGuarded` is deletable exactly as your comment says, and `given.spec.ts:231` flips from asserting `{}` to `{label:"ok"}` — I watched it fail under the patch. so these want sequencing rather than landing in parallel. I also tried `Object.freeze(Object.prototype)` as a stopgap and it isn't viable: winston dies at import, via logform → `@colors/colors`.

**one claim I couldn't reproduce.** `motly.ts` says `#@ persist { name=@env.X }` "is hydrated and returned in `annotationFields`". the hydration is real — `persist.text("name")` gives back the live `process.env` value off a compiled model — but the `annotationFields` half didn't reproduce for me: `deriveAnnotationFields` (`build_plan.ts:212-226`) iterates top-level `tag.entries()` and calls `.text()`, which is `undefined` for the nested `persist` block, so I got `{}`. either the reach is overstated or `PersistSource.annotations` is scoped differently than my repro. worth pinning down, because that sentence will get cited later as known-and-accepted.

a few other things I turned up:

- `endOfDelimited`'s backtick branch uses a bare `indexOf` (426) where the quote branches use a backslash-pairing loop (463-472), and MOTLY *does* decode escapes inside backticks (`` # `a\`b`=1 `` yields the key `` a`b ``), so the tmGrammar's `[^`\n]*` is an approximation. measured: `` # R=f'US' `k\`=f'y'`=1 `` comes out as `` # R="f'US'" `k\`="f'y'"`=1 ``, which parses, so the rewrite is kept and the key is silently renamed to ``k`="f'y'"``. reachable only on a line that was already unparseable. it's the residual class your own doc comment names, and the pairing loop covers it.
- empty string passes every presence check, so `# control=select suggest { query="" }` ships `suggest: {query: ""}` — the unrunnable-but-runnable-looking block the guard at `given.ts:160` exists to prevent. same for `source="" dimension=""`, and `label=""` ships an empty label. treating empty/whitespace-only as absent covers it.
- the `RELEASE_NOTES.md` edit lands under `## [0.0.242]` (heading line 98), and v0.0.242 shipped. so it rewrites a released narrative to say "the server reads them off the declaration's own tags, so they are populated", which 0.0.242 didn't do — and per this file's own "How this file is used", only `## [Unreleased]` sections get copied to the release page, so the control-contract reader wouldn't appear in the release that ships it.
- `drill_probe.spec.ts` placement is the one item from my last pass I don't think got an answer either way — happy to be told it stays.
- the PR body doesn't mention either guard. nine of the ten commits and most of `motly.ts` are the prototype and `@env` work, and the description covers only the encoder and `suggest` fixes, so a re-reviewer reconstructs it from the commit log.

design's still right, and the tests are genuinely non-vacuous — every hostile fixture carrying a real `label` so `{}` proves something is the detail I'd have most expected to be missing.

Sha-Bang (approved):

PR #983, opened 2026-08-09
fix(server)!: key the incremental ledger on the table, not the package

kylenesbit:
A `covered_through` boundary is a fact about a **table**, but it was keyed on
`(environment_id, package_name, source_entity_id)` — and a package name is not stable across an
orchestrated host's versions of one package.

## What was wrong

Credible presents a version-qualified package name (`<package>|<versionId>`, which is how one worker
holds several versions of a package at once), and fires a shared table's refresh through whichever
version materialized most recently. A publish makes the new version the most recent one, so the first
refresh after **every publish** arrived under a `package_name` the ledger had never seen: no row, no
boundary, `mode: "seed"`. A full recompute, on a source whose entire purpose is not being rebuilt.

Correctness was never at stake — a seed is always right — but two things beyond the cost:

- **That seed was not a cutover.** It lands on the instructed name, which for an in-place refresh IS
  the live serving table, so its CTAS → `DROP` → `RENAME` left readers seeing the table missing for
  the width of those two DDL statements. The one place a refresh was not atomic.
- The host records `delta_applied` either way, since the wire does not report which of seed/delta ran.

## The change

`PRIMARY KEY (environment_id, connection_name, physical_table_name)`. Version-independent by
construction, and truer on its own terms: two packages materializing different tables keep separate
rows, one table refreshed through three versions of its package keeps one. `package_name` and
`source_entity_id` stay on the row as attributes naming whoever last advanced it.

## The part worth reviewing

Taking the content address out of the key takes a guarantee with it. The old key made *"different
build SQL is a different row"* free, and a standalone publisher's physical names carry no content
token — they are `#@ persist name=` or the source name — so an edited model keeps its table name and
the read now **hits**. A model edit that adds a `where:` moves neither the output columns nor the
`watermark=`/`merge_key=` declarations the read already compared, so nothing else would have caught
it: a delta applied under a definition the existing rows do not match.

So the address joins `ledgerLineageMismatch` alongside the watermark, merge keys and strategy —
recorded on write, compared on read, and a change re-seeds. Same protection, relocated from the key to
a check, which is where the other declarations already had to live anyway (changing `watermark=` does
not move the address at all).

## Re-keying an existing database

No migration, and the drop is **not optional**. The ledger is a cache of a boundary the publisher can
always re-derive and its miss path is a seed, so `initializeSchema` detects a table still keyed on the
package (`PRAGMA table_info`, checking whether `package_name` is part of the primary key) and drops it
so the ordinary `CREATE TABLE IF NOT EXISTS` pass recreates it. Self-limiting — it reads the key
rather than a schema version, so once no database carries the old key it never fires again.

Left in place, a legacy table would be worse than untidy: the new upsert's `ON CONFLICT` names a
constraint the old table lacks, so every advance would fail (a swallowed warning) and every
incremental source would seed forever. Cost of the drop is one full rebuild per incremental source —
the same cost the bug was already charging on every publish.

## Tests

- `IncrementalLedgerRepository.spec.ts`: one boundary survives across `orders|v1` → `orders|v2` (the
  regression); the new key's neighbour separation across environment, connection and table; and both
  halves of the re-key at boot — a package-keyed table is replaced, and a table-keyed one is left
  alone with its boundaries intact. That last one matters most, since the schema pass runs on every
  boot and getting it wrong would discard every boundary on every restart.
- `incremental_apply.spec.ts`: a boundary measured over different build SQL is refused.

Typecheck and lint clean; 182 unit tests and 48 materialization integration tests pass. The 3
`concurrent_package` integration failures on this tree are pre-existing (confirmed on an unmodified
checkout).

### DIFF
--- a/packages/server/src/service/incremental_apply.spec.ts
+++ b/packages/server/src/service/incremental_apply.spec.ts
@@ -623,6 +623,7 @@ describe("deltaStatements: merge", () => {
 const LINEAGE = {
    physicalTableName: "analytics.daily",
    connectionName: "wh",
+   sourceEntityId: "sid",
    watermarkName: "order_date",
    watermarkType: "date",
    mergeKeys: [] as string[],
@@ -665,6 +666,19 @@ describe("ledgerLineageMismatch", () => {
             "analytics.other",
          ) as unknown as string,
       });
+   });
+
+   it("rejects a boundary measured over different build SQL", () => {
+      // Load-bearing since the ledger stopped being keyed on the content address:
+      // a standalone publisher's table names carry no content token, so an edited
+      // model keeps its table name and the read HITS. Without this check the
+      // delta would be applied to a table its own definition no longer describes.
+      const mismatch = ledgerLineageMismatch(
+         { ...LEDGER, sourceEntityId: "sid-before-the-edit" },
+         LINEAGE,
+      );
+      expect(mismatch?.reasonCode).toBe("lineage_changed");
+      expect(mismatch?.reason).toContain("content address");
    });
 
    it("rejects a boundary advanced on another connection", () => {
--- a/packages/server/src/service/incremental_apply.ts
+++ b/packages/server/src/service/incremental_apply.ts
@@ -791,14 +791,28 @@ export async function applyDeltaScript(
 
 /**
  * Everything the ledger has to agree with before a delta may run: which table,
- * on which connection, advanced by which declaration. A change to any of them
- * means the recorded boundary describes a different table or a different meaning
- * of "covered", so it cannot be built on.
+ * on which connection, computed from which SQL, advanced by which declaration. A
+ * change to any of them means the recorded boundary describes a different table or
+ * a different meaning of "covered", so it cannot be built on.
  */
 export interface IncrementalLineage {
-   /** Logical (unquoted) physical table name, as the manifest records it. */
+   /**
+    * Logical (unquoted) physical table name, as the manifest records it. With
+    * `connectionName`, the ledger row's key.
+    */
    physicalTableName: string;
    connectionName: string;
+   /**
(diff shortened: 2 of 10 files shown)


sagarswamirao (approved):
Traced the parts this change puts weight on. Approving.

**The compensating check is genuinely wired.** Taking the content address out of the key does remove a free guarantee, and the replacement holds end to end: `ledgerLineageMismatch` compares the address first, ahead of the declaration checks, and the value compared is the real content address from `computeSourceEntityId` at both `incrementalLineage` call sites — worth confirming, since the two sites pass differently-named variables (`sourceEntityId` vs `contentSourceEntityId`) and look asymmetric until you follow both. It also fails closed: with `contentSourceEntityId` absent, `lineage` is `undefined` and the source drops to the non-delta path rather than skipping the check.

**The boot-time re-key is the highest-risk code here, and it's handled well.** Detection reads the actual key via `PRAGMA table_info` rather than a schema version, so it is self-limiting, and both directions are covered against real DuckDB — a package-keyed table is replaced, and a table-keyed one is left alone with its boundaries intact. That second test is the one that matters, since the schema pass runs on every boot. I confirmed the `ON CONFLICT` target matches the new PK exactly, and that `source_entity_id` is refreshed on conflict rather than left stale.

**Two consequences worth knowing about, both already documented in the diff:**

- `deleteByPackage` is now best-effort. The package tag names whoever last advanced the row, so a table shared across versions keeps its boundary if a different version wrote it last. Cost is at most one seed, and only if the table outlives its package. The docstring says exactly this (`IncrementalLedgerRepository.ts:122-128`).
- The package index is now load-bearing rather than a key prefix, since that delete became a non-key scan. The PR adds the comment noting it.

PR #984, opened 2026-08-10
docs: stop the dashboards design doc claiming what has landed

mlennie:
Follow-up to #969, which landed `docs/malloyyo-dashboards-design.md` three days ago with a status header asserting "Nothing here has landed yet". That was true when I wrote it and false by the time #967 and #976 merged, in the one document the remaining dashboard slices read as their map. A status snapshot in a shared reference expires silently, and this one expired inside 72 hours.

## What changed

The header no longer says which parts have merged. It says it does not track that, points at the pull requests as the authority, and keeps a short index of which file or symbol each part of the design lives in, so checking is quick.

It also names the two places this design lands **by halves**, because "has it merged" is not a yes or no there. The given control contract is the clearest case: the control fields it puts on `Given` are in `api-doc.yaml` already, from #976, while the server code that fills them from a declaration's tags is not. A single all-or-nothing check on that reads "not landed" and invites re-adding fields that exist, which in `api-doc.yaml` means duplicate keys the generators resolve silently. The dashboard REST surface is the other: two endpoints, a list and a get, that need not arrive together. The rename is called out as the two-part change it is, since the files and the `/pages` endpoint move in separate slices, and there is a window where the file names have changed and the endpoint has not.

Separately, this finishes the `publisher.json` rename #969 left partial. That PR's message said the spec named a `malloy-publisher.json` manifest that does not exist, and fixed one of five occurrences; a follow-up commit got the other four in the spec, and this gets the last two, both JSDoc comments in `service/package.ts` and `service/build_plan.ts`. `malloy-publisher.json` is now absent from spec, docs and source. The real name is pinned by `PACKAGE_MANIFEST_NAME` in `packages/server/src/constants.ts`.

## The first attempt was worse, and that is the useful part

I first replaced the stale sentence with a recipe of git commands for checking each part yourself. Three review rounds found fourteen defects in it, every one in the recipe rather than in the design. A plain pathspec resolves against your current directory, so the same command run from `packages/server/` looked for `packages/server/api-doc.yaml` and reported absent for something that is on main, biased in the direction that causes re-implementing merged work. A probe built on `git ls-tree` fails open, exiting 0 whether the path exists or not, so wrapping it in `&& echo present` called all three file rows present when none was. The recipe opened with `git fetch origin`, which part of the intended audience is not permitted to run. And the prose explaining the pitfalls contained a miscount of its own.

Two of those rounds independently observed that the sentence I had deleted ("there is no `dashboards/` discovery, no `Dashboard` component, and no dashboard REST surface") was unambiguous and could not be misread. So the recipe is gone and the page states less: 22 fewer lines than that attempt, and no shell for a reader to get subtly wrong. The lesson I would draw is that replacing an expiring claim with an executable procedure moves the defect surface rather than removing it, and a procedure written in prose has no compiler.

## Verification

Full gate: codegen and `build:sdk` clean, `lint` exit 0 with no autofix mutations, `typecheck` clean, `prettier:check` red only on the six standing git-ignored generated files with both edited `.ts` files verified clean individually, unit 2270 pass 0 fail, integration 249 pass 0 fail. Links and anchors clean across every touched doc. Security: no surface, and provably so, since every changed line in the two `.ts` files is a JSDoc body, so there is no executable change. Live: built, booted, reached `operationalState: serving`, served the spec byte-identical to source, ran a real query returning 25,356 rows, and the materialization endpoint the edited comments describe still answers 200.

Every remaining factual claim in the header was re-checked against current main rather than carried over: five control fields on `Given`, no `readGivenControlSpec`, two dashboard endpoints in the source branch, two slice-1 branches.

One thing I am not claiming, since it seems worth saying plainly: no adversarial review round has read the final simplified text, because simplifying was itself the last change. That residual is smaller than the alternative, because the simplification deleted the surface every prior finding lived in rather than adding to it, but a human read of the header is worth more here than another automated round.


Sha-Bang (approved):
the file/symbol table is a genuinely useful addition, and dropping "nothing here has landed yet" was a correctness fix rather than just a policy one — #976 already put the `Given` control fields on main, so that sentence had quietly become an underclaim.

one worth fixing while you're in here. line 195 says starting values "arrive as one field with one name (`givens` on `DashboardManifest` and on `RawNotebook`)". that's `startingGivens` now — #976 landed the rename, and :311-314 of this same page announces it. sharper than a stale name, because `DashboardManifest.givens` does exist and means the control-row declarations you describe at :307, so the sentence points at a real field with the wrong meaning.

smaller:

- :44-46 now states as fact that `docs/dashboards.md` links `choosing-a-surface.md`, while :228 still says it will "when written". both files are true of the 935 branch, so the new sentence is fine by the page's own convention — :228 is the one left stale.
- :30 gives a bare `dashboard.controller.ts` where every other row gives a repo-relative path. it's `packages/server/src/controller/dashboard.controller.ts`.

PR #985, opened 2026-08-10
feat(sdk): shared given and drill controls, and a notebook built on them

mlennie:
Slice 7 of the ten-PR split Kyle proposed in #935. **Stacked on #978**, whose `@testing-library/react` harness is what makes `useGivensState` testable at all; GitHub retargets this to `main` when #978's branch is deleted on merge.

This is the one slice that is not a straight extraction. A review of #935 found real defects living in exactly this code, so they are fixed here rather than carried. Seven verification passes then found a further 48 confirmed defects: in the extracted code, in my own fixes to it, and in my own prose about it. Everything below was measured rather than assumed.

## What this lands

The control layer behind `given:` becomes one implementation instead of one per surface: `useGivensState` for state and URL round-tripping, `paramCodec` and `filterValue` for encoding, `useSuggestOptions` for `suggest`-backed pickers, and `components/drill/` for `# drill`. `Notebook` is the first surface built on it, so a notebook's parameters are part of its URL and a filtered notebook is a link you can send. It replaces the `#(filter)` Filters panel and `useGivensForm`.

## The filter encoder did not work, and this contradicts one "holds up well" note

`encodeFilterList` quoted values with `"..."`. Malloy's string-filter grammar has no double-quote quoting: backslash is its only escape. So the quoting escaped nothing and several ordinary values silently meant something else. James's review filed this module under what holds up well; I am contradicting that with measurements, and it also means commit `68c48c27` on the branch is not closed.

Measured end to end, by generating the filter with the SDK's own `encodeFilterList` and running it through the real compiler and DuckDB against a dimension carrying these values:

| picked value | bare (old encoder) | escaped (this PR) | true count |
|---|---|---|---|
| `Ben & Jerry, Inc` | 0 (read as two values) | **2,535** | 2,535 |
| `-Outerwear` | 25,356 (a negation: everything) | **2,213** | 2,213 |
| `%` | 25,356 (filter bypassed) | **2,850** | 2,850 |
| `null` | 0 (the null operator) | **4,372** | 4,372 |

**The fix delegates instead of hand-rolling an escape**, which is the part worth reviewing. `@malloydata/malloy-filter` already prints a clause back out escaping exactly what its own grammar treats as special, so encoding is `StringFilterExpression.unparse` and decoding is its `parse`. Two reasons that beat writing the escape myself: its escape set includes the **space** character, which a by-hand reading of the grammar misses and which `Ben & Jerry, Inc` needs; and the keyword set has already grown, since 0.0.383 escapes `null` and `empty` while **0.0.427 also escapes `none`**. A hand-written escaper would have shipped stale on day one. Pinned by a round-trip test against the real parser over a hazard corpus including every ordered pair, and mutation-tested. An independent security review additionally fuzzed it over 12,000 generated values: zero negation injections, operator changes or value-count changes.

`@malloydata/malloy-filter` becomes a peer dependency. It is already in the tree of two existing peers, so an install satisfying today's peers already has it.

## `#(filter)` genuinely regresses, and the release notes said otherwise

Removing the Filters panel means the notebook sends no `filterParams`, and the server is not inert about that. On a throwaway package with `#(filter) ... required`, the cell endpoint answers **HTTP 400 `Required filter "CAT" (dimension: category) was not provided`** without filter params and **200** with them. So a model with a required filter now fails every cell and no UI can satisfy it; one with only optional filters still runs but is no longer filterable from the notebook.

`RELEASE_NOTES.md` said under 0.0.201 that such models "continue to work unchanged". That was true then and is false now, so this PR corrects it in place and states the new reality in its own section. I have not revived a filter UI, which would balloon this slice; what I have done is make the failure legible instead of invisible, since a failed cell previously rendered as blank space.

**James asked to confirm with real users that nobody is mid-migration before the panel goes. That is unresolved and is a merge gate I cannot close.** If someone is mid-migration, the alternative is to keep the panel here and defer its removal, and I am happy to do that instead. Worth noting that the security review established `#(filter)` was never an access-control mechanism: `buildFilterClause` skips any filter with no caller value regardless of the `implicit` flag, and `bypassFilters` is read straight off the request, so dropping it weakens no boundary.

## What the verification passes found

Seven passes, 48 confirmed defects. The full list is in the commit message. The ones a reviewer should know about:

- **A drill painted cells as links to a route that does not exist.** `# drill { to=<dashboard> }` navigated to `/{env}/{pkg}/dashboards/<slug>`, which the app has no route for, so the click landed on "Nothing to open at this path". Named destinations are now inert until the slice that adds the route also passes the handler, which withdraws the affordance too: measured 10 marked cells versus 0.
- **A drill onto a given the notebook does not declare** was also painted, and refused on click with a console warning. `useDrill` now asks the surface whether it can honour the given.
- **Cell state was not keyed to its notebook**, so an in-app navigation rendered one document's prose with another's results. Measured: the second notebook showed the FIRST one's heading and not its own.
- **Reset was broken in two different ways at once**, one cause: with autorun off it was completely inert, and with starting values it wrote the URL twice and ran every cell twice.
- **A superseded run is now aborted** rather than left in flight (measured 9 started, 5 finished, 4 aborted), and unmounting cancels too.
- **Several type-handling defects**: a drill on an empty cell encoded the match-everything filter (a widening, and a regression from this PR's own empty-value handling); a `timestamp` given was truncated to midnight by the URL round trip, and then, after that fix, read back as local time so the value shifted by the reader's offset on every cycle; an empty `number` parameter decoded as 0; a boolean parameter that was neither `true` nor `false` became `false`.
- **The picker and slider now refuse types they cannot encode** rather than corrupting them, which closes a hole before the slice that would open it. All 78 type-by-control combinations are enumerated in a test.
- **`order_by` in the generated suggest query did not compile for a joined dimension**, which is the ordinary case. Found by compiling the generated text for the first time.

Several of those were regressions introduced by my own earlier fixes in this same loop, including one that left every cell blank on a control change. The Playwright suite could not see it, because it asserted the URL and the control and never that a cell produced anything; it now does, proved by restoring the regression and watching only that test fail. The timezone handling took three attempts before it stopped depending on the reader's zone, and now builds the instant with `Date.UTC` rather than pattern-matching a string.

Worth stating plainly about the process: the count of findings never converged (10, 12, 10, 10, 6), but their composition did. By the fifth pass they were almost entirely defects in my own previous fixes and in my own prose rather than in the feature, which is the point at which more machine passes mostly review the edits the review provoked. The seventh pass was also unhealthy, losing 7 of 24 agents to a transport error, so it is recorded as partial. That is where I stopped, and the remaining risk is the one a human reviewer is best placed to catch: an author asserting a fix is complete when it is not.

## Where this departs from #935, deliberately

- **`annotations` is not widened back to the folded set here.** #967 narrowed it because `parseNotebookFilterAnnotation` interprets that field on `main` today. This PR removes that interpreter, so the widening becomes safe, but it is a `model.ts` change with no consumer and `model.ts` is contended by slices 5 and 6, so whichever slice needs the faithful full report should make it and cite #967. Flagging it because a reviewer expecting the widening should know it was a decision.
- **`autorun` is hardcoded true.** `RawNotebook.autorun` is not on `main`: its reader is slice 5's `given.ts`. Declaring a field nothing produces is the defect class #976 just fixed. **That wiring looks unowned:** #976's notes assign `readStartingGivens`/`readAutorun` to slice 5, but slice 5's scope is four files that exclude `model.ts` and `api-doc.yaml`. Worth settling, or `startingGivens` stays undefined forever.
- **#935's `package-notebooks.spec.ts` additions are dropped**, since they target a storefront model that only declares `given:` after slice 9. Replaced with six tests against `governed-analytics/orders.malloynb`.
- **`ResultContainer`/`RenderedResult` were split by hand**: only the drill half is here, and `chartHeight`/`RenderedResultKind` stay with slice 8's dashboard-height work.

## Known limitations, all documented at the site

A given carrying a starting value cannot be cleared (unreachable today, since nothing populates `startingGivens`); a `suggest` request carries no `givens`, so a dropdown over an `#(authorize)`-gated source reads "Options unavailable"; `markDrillableCells` matches header text, so a nested column with a colliding name is marked and does nothing; and a filter value with a trailing tab or an embedded newline does not survive the round trip. Each says what it costs and why it was not fixed.

The multiselect race a Playwright comment on the origin branch describes **does not reproduce**, with or without the host's URL round trip, so no fix was invented for it. Slice 9 inherits that spec and should re-verify before copying the comment across.

## Coverage limit worth stating

`select` and `multiselect` need `Given.control` and `Given.suggest` populated by the server (slice 5) and an example that declares them (slice 9). Neither is on `main`. So the picker path and `useSuggestOptions` are covered by unit tests only, and the live and Playwright legs exercise the text and number controls `governed-analytics` already declares. A green Playwright run here is not evidence the multiselect works end to end.

One gap I could not close from here: the naive-timestamp fix cannot be guarded by a UTC CI, because under UTC the correct and broken implementations are indistinguishable. Verified by hand across four zones. Running one suite under a non-UTC timezone in CI would close it.

## Verification

Gate in the required order: codegen and `build:sdk` clean, `lint` 0, `typecheck` 0, `prettier:check` red on only the six standing git-ignored generated files, SDK **204/0** under both UTC and America/New_York, server unit **2055/0**, integration **245/0**.

Live on the production bundle: the filter measurements above, drill exercised end to end for both `filter<string>` and plain `string` givens, the required-filter package showing the server's message where `main` showed blank space, and the notebook read by eye. Playwright: the full suite passes, bar two `environment-connections` tests that need a stub `bigquery` connection CI injects and a local config lacks; replicating that fixture makes all five pass.

**Merged-tree check, per the lesson from #972**: conflict-free, every module present, a sweep for all removed symbols clean, and the merged tree checked out and typechecked, with a control run of plain `origin/main` in the same probe worktree producing an identical error set.


packages/sdk/src/components/Notebook/Notebook.tsx
```diff
@@ -91,263 +139,235 @@ export default function Notebook({
    const [isExecuting, setIsExecuting] = useState(false);
    const [executionError, setExecutionError] = useState<Error | null>(null);
 
-   // Parse filter configuration from notebook annotations (legacy ##(filters) approach)
-   const filterConfig = useMemo(() => {
-      if (!notebook) return null;
-      return parseNotebookFilterAnnotation(notebook.annotations);
```
Sha-Bang, line 97: worth correcting the note in the body about widening `annotations`. this removes the call, but `parseNotebookFilterAnnotation` is still exported from `components/filter/index.ts`, and Credible’s app calls it on exactly this field. so “the widening becomes safe” doesn’t follow: widening back to the folded set puts #967’s problem back for a live consumer. flagging because whoever picks it up in slice 5 or 6 will read the note rather than re-check.

Sha-Bang (commented):
went at the encoder rather than reading it, and it holds up. threw a 47-value hazard corpus at it (`-Outerwear`, `%`, `null`, `empty`, `none`, `Ben & Jerry, Inc`, backslashes, backticks, `;`, `|`, a zero-width space, 500 chars), then all 2,209 ordered pairs, then the whole thing again through a real `URLSearchParams` write and read so the address bar is in the loop. zero widenings: everything comes back as a non-negated `=` carrying the right value count. `decodeFilterList` and `isPlainFilterList` never threw on hostile input either (unbalanced parens, `;;;`, a NUL, 20k chars). merge-tree clean against `b828f8a0`, typecheck 0, and the suite is green at `4a82813e` under UTC, America/New_York, Australia/Sydney and Pacific/Kiritimati.

the all-whitespace widening in the amended commit is a good catch and my corpus missed it: I had leading and trailing whitespace in there but never a value that was *only* whitespace, so I would have shipped it. re-ran against the fix, and 10 of the 11 whitespace codepoints I tried now drop. the one that does not is U+200B, because JS `trim()` does not count it as whitespace, but it is fine on its own terms: it encodes to a plain single-value `=` rather than a null clause, so it filters for a useless value instead of matching every row. worth knowing that is where the 22 stops rather than discovering it later.

one thing I would want pinned down, and it is the test rather than the code.

**the blank-cell test can pass with the regression present.** `notebook-givens-url.spec.ts:47` fills the control, asserts the URL, then asserts the count came back. but the URL is written on the applied change immediately, while the run waits out `GIVEN_SETTLE_MS` (400ms) and `executeCells` is the thing that blanks the results, so `toHaveCount(before)` resolves against the pre-change DOM on its first poll and goes green without a run having started at all. that is in tension with your note about restoring the regression and watching only that test fail, so one of us is wrong here: I reasoned it from the timing and did not run Playwright.

if it is racy, the fix that is not racy in the other direction is to pin the re-render rather than the count. `{cell.result && ...}` at `NotebookCell.tsx:629` means a run start unmounts the whole result subtree, and a detached node stays detached, so a poll cannot miss it:

```ts
const before = await results.count();
expect(before).toBeGreaterThan(0);
const stale = await results.first().elementHandle();

await page.getByLabel("MIN_AMOUNT").fill("500");
await expect(page).toHaveURL(/[?&]MIN_AMOUNT=500/);

// Nothing is dispatched for GIVEN_SETTLE_MS, so the count below means nothing
// until a run has actually started. A run rebuilds every cell from the raw
// notebook, which unmounts this node.
await expect
   .poll(() => stale.evaluate((node) => node.isConnected), { timeout: 30_000 })
   .toBe(false);

await expect(results).toHaveCount(before, { timeout: 60_000 });
```

waiting on `getByRole("progressbar")` reads better and is one line, but it is a window rather than a latch, so a fast run can close it between polls.

smaller things, whenever:

- the `encodeFilterList` doc block now says "Two kinds of value do not survive" over two bullets and closes with "All three are pinned in the spec". the newline bullet is what went missing in the amend, and the behaviour is still there: `a\nb` fails to parse, so `decodeFilterList` hands back the opaque single entry and the server sees an unparseable filter. reads like the bullet was meant to stay.
- "Trailing whitespace is dropped" is also true leading: `"\ta"` comes back as `"a"` the same way `"a\t"` does.
- `GivenValue` promises arrays the URL codec cannot carry. `givenToParam` joins on `,` unescaped, `paramToGiven` never returns an array, and `givensToRequest` sends the real one, so the URL and the request disagree. measured: `["Ben & Jerry, Inc","Nike"]` reads back as the single string `"Ben & Jerry, Inc,Nike"`. unreachable today, since `multiple` needs `isFilterType`, which makes the `else if (multiple)` arm at `GivenInput.tsx:254` dead. but the type still says arrays work and slice 8 reuses the codec.
- `resolveDrill.ts:111` and `:142` both still say the given name is "the dimension name upper-cased", which is the thing the body comment at `:118` says it fixed.
- `SDK 204/0` does not reproduce for me, I get 286/0 at `4a82813e` after `generate-api-types`. green either way, so just the number.

mlennie:
@Sha-Bang heads up before you spend more time on this: the branch has moved a long way since you started. If you are reading `4a82813e`, that tree still had eight defects in it that are fixed now. Current head is `a16e43b1`, which carries 21 fixes including all of yours.

**Your race call was right, and here are the numbers.** `toHaveCount(before)` matched at **23ms**; the run did not start until **425ms**, because the pre-change cells do not unmount until then. So the assertion was green 402ms before anything ran, exactly as you reasoned. I had reported that reverting the fix failed only that test, which was wrong: that test could not have seen the regression at all. Your latch is in, with both numbers in the comment and your spinner-vs-latch distinction kept.

**Your other three were all real.** The missing newline bullet is worse than either of us said, an embedded newline makes the filter unparseable outright rather than collapsing it to one value. A **leading** tab is dropped just as readily as a trailing one, which your wording implied and mine got wrong. And U+200B does survive `trim()`, so it is now documented as the boundary of the rule with a spec pinning all three codepoints.

**Chasing your race turned up the worst defect on the branch.** Teaching the encoder to drop whitespace-only values changed *what encodes to empty*, while the drill guard still tested for the exact empty string. So clicking a blank-looking cell (a space, a tab, or an NBSP from pasted data) encoded to the empty filter and **returned every row in the table, presented as a drilled view**. No error, nothing visible. The guard's own comment described that hazard three lines above the test that had stopped matching it.

The durable fix was not to patch the second check: `drillValueToFilter` now asks the encoder what a value encodes to instead of restating its rule, so that class cannot come back. Six of the defects on this branch had that same shape, a rule changed in one place while a second place still restated the old one from memory.

**Two more where you were right and I was not.**

**The `annotations` note in the description is wrong**, and I am correcting it here rather than editing the body so the history stays readable. "The widening becomes safe" does not follow: `parseNotebookFilterAnnotation` is still exported from `components/filter/index.ts`, and Credible's app calls it on exactly this field, so widening back to the folded set would reintroduce #967 for a live consumer in another repo. **`annotations` stays narrow permanently.** Treat the narrow shape as the decision, not a temporary state, and disregard the paragraph in the description that says otherwise. Worth adding why your flag mattered beyond the note: I had passed that same reasoning to the slices working on #981 and #987, so both had been told the narrowing was reversible. Both are corrected now. Your reason for raising it, that whoever picks it up reads the note rather than re-checking, is exactly what was about to happen.

**On `GivenValue` promising arrays, I closed half your finding and credited it as the whole thing.** Removing the dead `else if (multiple)` arm makes it unreachable today, which is not what you were pointing at. The type still promises arrays, and `givenToParam` still joins on an unescaped `,` against a `paramToGiven` that never returns one, so `["Ben & Jerry, Inc","Nike"]` still reads back as the single string `Ben & Jerry, Inc,Nike`. Since `paramCodec.ts` is added by this PR rather than pre-existing, that is a defect this branch introduces, so it is being fixed here rather than handed to slice 8. The fix is to narrow the type to what the codec can actually carry, not to build multi-value support: if slice 8 needs real arrays it should design the escaping deliberately instead of inheriting an unescaped join. Your round-trip case is going in as the test.

Still outstanding: one region of automated review, plus a re-review of the two regions that produced the fixes, since seven of the defects here were created by earlier fixes on this same branch. Not claiming this is finished until those run.

Sha-Bang (approved):
re-checked against `0ac19d27` rather than re-reading: 311 pass / 0 fail, tsc clean, and I re-ran the encoder corpus from the first pass — 47 hazard values and all 2,209 ordered pairs, zero widenings, nothing thrown on hostile input, no whitespace-only null clauses. same numbers as before the re-review commits, so none of that churn moved the encoder.

all six things I raised are closed. the one worth singling out is the blank-cell test: you took the detachment latch, then found the same shape in `settledCount` itself — two samples 250ms apart prove the count *paused*, not that it settled, so a baseline taken there pinned a partial render. that's the finding generalized rather than patched, and it's a better fix than the one I proposed.

also confirmed: `GivenValue` no longer admits `string[]` (`paramCodec.ts:42`, refused rather than joined), and the two "upper-cased" claims in `resolveDrill` are gone, leaving only the accurate note at :122.

approving.

PR #987, opened 2026-08-11
feat(server): discover, lint and serve package dashboards

### LINKED ISSUE #979: Notebook listing schema declares a field the response never sends, and omits one it always does
The `Notebook` schema in `api-doc.yaml` does not describe what the notebooks listing endpoint returns. It declares a field the response never carries, and omits one it always does.

`GET /api/v0/environments/{env}/packages/{pkg}/notebooks` returns entries shaped like this, confirmed against a running server on the bundled `examples` environment:

```json
{ "environmentName": "examples", "packageName": "storefront", "path": "storefront.malloynb" }
```

The schema declares `resource`, `packageName`, `path` and `error`. So:

- **`environmentName` is sent on every entry and is not declared.** A generated client has no typed access to it.
- **`resource` is declared and was not present** on any entry in that response. A caller following the spec reads `undefined`.

`path` and `packageName` are correct and unaffected.

This is the same defect class as the `RawNotebook` gap closed in #976, found by the same audit: an undeclared response field is unreviewed surface, and a declared-but-unsent field is a promise the server does not keep. It was deliberately left out of that PR to keep the diff to the givens vocabulary, and because a later slice of the #935 split adds `title` and `description` to this same schema and would collide with it.

### Suggested fix

Declare `environmentName`, and settle `resource` by checking whether any producer ever sets it. If nothing does, remove it; if some path does, say which. `#976` closed the equivalent `RawNotebook` gap by making the schema match the response and then removing the blanket `as ApiRawNotebook` cast that had been hiding the mismatch, so that the compiler catches the next drift. The same treatment would work here if the listing has a comparable cast.

### Worth checking at the same time

Whether `error` is reachable on this endpoint. It is declared, and a notebook that fails to compile is the obvious producer, but that was not verified while writing this up, so treat it as an open question rather than a finding.

### How it was found

Comparing the declared schema against a real response while closing the `RawNotebook` gap in #976. Reported at @kylenesbit's request rather than folded into that PR.

mlennie:
Slice 6 of the ten-part split of #935: the server half of dashboards. Discovery, the manifest, the load-time lint, and two read-only REST endpoints. No UI, so this surface is testable on its own.

**This PR targets `monty/pr935-slice5-given-control-spec` (#981), not `main`,** because it needs `readGivenControlSpec` and `motly.ts`, which are only on that branch. GitHub will retarget it to `main` automatically once #981 merges and its branch is deleted. Verified before branching: neither symbol is on `origin/main`, and that branch is zero commits behind it.

## What this adds

A `.malloy` file in a package's `dashboards/` directory carrying an `# artifact` tag becomes a dashboard, addressed by its filename slug. `GET .../packages/{packageName}/dashboards` lists them; `GET .../packages/{packageName}/dashboards/{dashboardName}` returns one manifest: the artifact tag's declarations plus the control contract derived from the givens its query references.

There is deliberately no run endpoint. A dashboard's query, each composite tile, and each control's suggest query all run through the ordinary `POST .../models/{path}/query` with `givens`, using the manifest's `path` as the model, so row caps, byte caps, authorize gates and render-tag validation all apply unchanged rather than being reimplemented on a new surface. The manifest's `path` field now says this explicitly, since it is the load-bearing fact behind having no run endpoint.

Also here: a notebook listing carries `title` and `description`, resolved from its own `## title="…"`, then its `#"` doc comment, then its first markdown heading; `RawNotebook` gains `autorun`; and `Notebook` declares `environmentName` and drops `resource`, which closes #979.

## The defect this fixes rather than carries

`discoverDashboards` was the only listing path in `Package` that never consulted `exploreSet()`. On a package that curates its query surface (`explores` declared, `queryableSources: "declared"`), a dashboard whose entry file is not listed was still served with a full manifest **advertising operations that cannot run**: the query name, the given names and the suggest-query names are all refused by the query boundary with a 404. Notebooks are uncurated in both directions; dashboards had taken "always listed" without "always queryable", so a curated package got the worst of both.

Such a dashboard is now held back and the omission reported in the package's `warnings`, so the fix does not simply hide things. Three integration tests cover it against a new fixture that lists one dashboard in `explores` and withholds another. One of them asserts that the held-back dashboard's query really is refused, so this stops being correct by coincidence if that policy ever changes.

I originally wrote this up as a disclosure fix and that overstated it, so to be accurate about what it is and is not: the manifest is a genuinely new surface, and for a file that failed to compile it also carried the raw compile error, which names tables, columns and connections. But the package `warnings` array on `main` already names the path of every model and its field, view and source names regardless of `explores`, so this is not the only place a curated package emits identifiers for files it hides, and closing it does not make the package response secret. The unarguable half is coherence: the endpoint was advertising work the server would refuse.

## Eight more, found while extracting

- **Five `Tag.text()` / `Tag.textArray()` reads went through the raw parser.** Those throw on a bad date literal rather than returning undefined. One sits in the message that reports a bad `dashboard_columns`, reached only when the value is already strange, and since the lint's `try`/`catch` wraps the whole package loop, `dashboard_columns=@2024-13-01` would have cost every dashboard finding in the package. All five now read through `tagText` / `tagNumeric`, with a `tagTextArray` helper for the list form.
- **`filterPublisherOwnedRenderLogs` had no call sites.** I took the function and not its three callers, so every dashboard query answered with "Unknown render tag 'artifact'" about its own tag. Found by the integration suite. The load-time path is deliberately left unfiltered: it already filters to error severity and these are warnings.
- **`# dashboard { columns=N }` was never validated.** Both spellings reach the manifest but only the composite `dashboard_columns` was checked, and the unvalidated one is the form the shipped example dashboards use.
- **An `## artifact` that parses but declares no `tiles=` vanished silently,** producing no dashboard and no finding, because the composite branch needs at least one tile and the lint only fired on tags that fail to parse.
- **A slug that cannot round-trip through its own URL is no longer served.** `dashboards/.malloy` published a `resource` ending in `/dashboards/`, which Express's non-strict routing folds onto the list route, so following the manifest's own link returned an array where a manifest belongs. No traversal risk; the lookup is a `Map`.
- **`motlyParseErrors` had no caller,** so a given whose own annotation failed to parse lost its entire control contract silently. A failed parse yields an empty tag rather than `undefined`, so the reader cannot detect one. Now reported per given.
- **A `suggest` block that cannot fetch options gets a finding.** Dropping an unrunnable block is right, but dropping it silently left an author with neither a control nor an explanation.
- **`DashboardManifest` described none of its seven shared fields** where `Dashboard` described each.

## What review found after this was opened, and what it changed

A security pass and an adversarial code review were run against the diff after it went up, because the first version of this branch had had neither. Between them they found thirteen defects, eleven of them mine. Ten are fixed; three were refuted. That is a poor enough result that opening on a single self-directed pass was the wrong call, and it is worth saying so rather than presenting the fixed state as if it arrived that way.

The code review (27 agents, none errored) produced 21 verified findings collapsing to ten distinct defects, every one confirmed by an independent verifier. Beyond the curation ordering below, they were: an invalid grid width that still reached the wire because `readArtifactTag` used raw `Tag.numeric()` while the lint used the strict reader, so the two disagreed about the same tag; an untagged shared include warned about as a withheld dashboard, because the gate ran before discovery decided the file was a dashboard at all; a correct `# drill` reported as naming something that is not a dashboard, because the drill lint resolved against the post-curation served set; a multi-line doc comment used verbatim as a notebook title, newline and all; an unimported-given lint that covered composite tiles only while `buildGivenSpecs` claimed otherwise; and a `list-dashboards` description promising a broken-dashboard state that is unreachable. All fixed, with the behavioural ones mutation-tested.

**Real, and it defeated the headline fix on two paths.** The new gate reads `packageMetadata.explores`, and both routes that can change curation on a live package got it wrong. In `reloadAllModels`, `discoverDashboards()` ran five lines *before* the freshly-read `explores` was installed, so the reload that first curates a package computed its dashboard set against the previous policy and went on serving the withheld manifests until some later reload. And `setPackageMetadata`, which the metadata `PATCH` goes through, re-applied the query boundary without re-running discovery at all, so that route never took effect. Discovery now runs after the refreshed metadata is installed, and `listDashboards`/`getDashboard` re-check the gate at serve time so the two cannot drift again. The serve-time check is one-directional on purpose: it can withhold a dashboard the cached map still holds, but it cannot surface one discovery already dropped, so relaxing curation by `PATCH` still needs a reload. That is the safe direction. A new integration test covers the `PATCH` route and fails if either half of the serve-time gate is removed; the reload-ordering half is correct by inspection but not pinned, because `reloadAllModels` is reached only from materialization refresh and manifest rebind, not from `?reload=true`.

**Refuted, with the reasoning kept because it corrected me.** A claim that the package-wide lints leak withheld dashboards does not hold: those findings carry no path attribution and are deduplicated package-wide, and `renderTagWarnings` on `main` already emits far more through the same array for every uncurated model. And a claim that the gate should mirror the *within-file* `export {}` half of the query boundary is not a security issue, since both boundary levels fail closed.

That second one did surface a real correctness gap, which is documented rather than fixed here. A composite dashboard in a file that only imports and re-exports nothing passes the file-level gate but has an empty export closure, so every one of its tiles 404s. Mirroring the second boundary level would mean resolving each tile against the closure before the manifest exists, which is more than this slice should change. What is fixed is the misleading part: the warning used to say "add it to `explores`", which does not make those tiles run, and now also names the re-export the author actually needs. `isQueryableEntryPoint`'s doc comment says plainly that it covers only the file-level half.

## Two things this deliberately does not claim

**The branch that lists an uncompilable dashboard with its error is not reachable from either ordinary load path.** `Package.create` aborts the package on the first model error, and `?reload=true` answers 424 and keeps the previously-loaded package serving. It is reachable only from `Package.reloadAllModels`, which keeps per-model placeholders. So the guard I added there, which stops an untagged shared include from being listed as a phantom dashboard, is defensive. The integration test beside it pins the surrounding contract rather than the guard, and its comment says so. I mention it because I first wrote a test that passed whether or not the guard was present, and only caught that by mutating the fix to check the test noticed.

**`motlyParseErrors` reports syntax only.** It is silent on an unresolved reference, silent on a line that only parses after the filter-literal rescue, and carries no position, so the new warning cannot yet say which annotation is at fault. An empty result is not proof a tag is well formed, and the warning is worded not to imply otherwise.

## Where it departs from #935

`docCommentText`, `motlyTag` and the other MOTLY primitives now live in `motly.ts`; this branch imports them, so the quote-blind `quoteFilterLiterals` fix and the prototype-cycle guard apply for free. `getNotebookListing` reads its title through `tagText` for the throwing reason above. The notebook response's comment no longer justifies an `as ApiRawNotebook` cast that no longer exists. The doc-comment spec expects newline joining, matching the reader it calls.

`drill_probe.spec.ts` is not here: it ships with #981, where it pins a compiler assumption this slice depends on, and a probe is worth most before the code relying on it lands. `source_given_probe.spec.ts` is here, because it imports `readDashboardModelFacts` from this file and cannot compile without it. It is also the compile-backed counterweight to `dashboard.spec.ts`, which is hand-built fixtures that never touch a real `ModelDef`.

The example storefront dashboards are deliberately not included. This slice has no UI and the integration suite uses its own fixtures, so they belong with the slice that adds the route rendering them.

## Verification

Gate in the required order: `generate-api-types && build:sdk`, then `lint` (exit 0, zero `--fix` mutations), `typecheck` (0), `prettier:check` (6 failures, all confirmed git-ignored generated files CI never sees), then both test legs separately.

- Unit **2393 pass / 0 fail** across 121 files, from a baseline of 2331/0 across 119 recorded on the untouched branch before any edit.
- Integration **286 pass / 0 fail** across 35 files, from 250/0 across 34.
- `openapi-spec-validator`: `api-doc.yaml: OK`. Merged tree against `origin/main` checked and clean.

Every behavioural fix was mutation-tested: disabling the curation gate fails three tests, removing the second grid-width spelling from the checked set fails two, and restoring the throwing `.text()` fails the hostile-literal test. Verified live against a built bundle serving a curated package: the list shows only the served dashboard, the held-back one 404s without leaking its manifest, its query is genuinely refused, the served one runs through the ordinary query endpoint with a given and returns rows with no render warnings, and the warning explains the omission.

Two tests pin that dashboard discovery inherits the annotation guards, since reading `# artifact` makes it a new reader of reserved-namespace tags: one that `@env` is not hydrated into a manifest, checked to be load-bearing by confirming the unguarded parser does return the value, and one that both `__proto__` spellings leave later parses working.


Sha-Bang (commented):
read this against the merge base rather than the base ref -- `monty/pr935-slice5-given-control-spec` has moved on, so a two-dot diff shows its `motly.ts` hardening as reverted by you, which it isn't. built the dashboards out and drove the real exports. two I'd want closed before it lands, both inline: the `@env.` drop and the `startingGivens` unwrap.

smaller, whenever:

- `api-doc.yaml:3728` still says of `EncodedGivenValues` that "no endpoint returns it yet" and "nothing populates that field so far", which `model.ts:3367` and `package.ts:2424` now do. you caught the same sentence in RELEASE_NOTES; this is the copy it mirrors. outside the diff, so it's here rather than inline.
- a notebook `## title=@env.X autorun=false` comes out `autorun=true`, since `readAutorun` defaults true and the tag is gone. `model.ts:3366` is new here so it's this PR's to inherit, though the wider thing is that nothing calls `motlyParseErrors` on notebook model-level tags at all -- an ordinary malformed `##` line flips it the same way, with a real parse error no caller reads. fine as a follow-up.
- the `@env.` drop is per line, not per file: `## artifact { tiles=[...] }` plus a separate `## note=@env.X` still produces the dashboard, only same-line kills it. worth a word in the docblock so the next reader doesn't over-read it.

packages/server/src/service/dashboard.ts
```diff
@@ -0,0 +1,1130 @@
...
+/**
+ * Explain a `dashboards/*.malloy` that produced no dashboard.
+ *
+ * A file with no artifact tag is a shared include and silent, as Malloyyo
+ * treats it. But a file whose tag *failed to parse* is a dashboard that simply
+ * vanished: MOTLY discards the whole tag on a syntax error, so the file falls
+ * through to the include path and never appears in a listing or at a URL. That
+ * is the least debuggable outcome in the whole feature, so it gets a finding.
+ */
+export function lintUndiscoveredDashboard(
```
Sha-Bang, line 1068: a dashboard whose artifact tag carries `@env.` lands here with nothing to report. `hasEnvReference` drops the annotation inside `motlyAnnotations`, which is upstream of `parseMotly`, so this gets `[]` from `motlyParseErrors` *and* `undefined` from `motlyTag`, and returns no findings. measured, same file both ways:

```
# artifact { title=@env.PROBE_SECRET }   -> manifest undefined, findings: 0
# artifact { title= }        (control)   -> manifest undefined, findings: 1
```

I'm not asking you to relax the guard -- the bluntness note at `motly.ts:76` is right that the leak is the worse failure. it's that the note's stated cost, "a missing control is a visible inconvenience", is what changes in this PR: the cost is now the whole dashboard, which is the outcome the docblock right above this calls the least debuggable thing in the feature.

cheapest fix I can see is to make the drop reportable rather than the guard smarter -- split the filter in `parseMotly` so an env drop is counted, and return a fixed `ENV_REFERENCE_DROPPED` the way `UNSAFE_TO_PARSE` already does. ~8 lines, filter behaviour byte-identical, and it covers the given control-contract drop too (`# label="Region" control=@env.X` gives `{}` from `readGivenControlSpec` with `lintGivenTags` silent, same shape).

I did look at neutralizing `@env.` in place and think it's the wrong trade: that needs the quote/heredoc/backtick-aware walk your own comment at `motly.ts:428` says was corrected three times, and a missed delimiter form there re-opens the leak rather than losing a tag.

packages/server/src/service/package.ts
```diff
@@ -1734,18 +1772,698 @@ export class Package {
...
+               // exact signal is available and the heuristic would be the wrong
+               // tool. This is also the drop site that pays most for a false
+               // positive.
+               if (factsCarryArtifactTag(facts)) {
+                  droppedByError.push({ modelPath, name });
+                  dashboardSlugs.add(name);
+               }
+               continue;
+            }
+            if (!manifest) {
```
Sha-Bang, line 2285: knock-on from the `@env.` drop: a file that lands here never reaches `dashboardSlugs.add(name)`, so a `# drill { to="overview" }` anywhere else in the package reports *"which is not a dashboard in this package"* -- false, and it sends the author to fix a drill tag that's correct. the path at :2011 registers the slug for exactly this reason, with a comment saying so. this is the one drop path that doesn't.

packages/server/src/service/dashboard.spec.ts
```diff
@@ -0,0 +1,1428 @@
...
+                     name: "overview",
+                     annotations: [
+                        "# artifact { title=@env.SLICE6_PROBE_SECRET }\n",
+                     ],
+                     givens: [],
+                  },
+               ],
+            }),
+         );
+         expect(JSON.stringify(manifest ?? {})).not.toContain(
```
Sha-Bang, line 1540: this can't fail on the collateral. `manifest` is undefined for an `@env.` tag, so it asserts against the string `"{}"` and passes whether the guard worked or the whole dashboard vanished. `given.spec.ts:356` pins its collateral properly with `toEqual({})` plus the sibling-drop case -- worth the same here, where the cost is higher.

packages/server/src/service/motly.ts
```diff
@@ -667,8 +793,39 @@ export function readStartingGivens(
...
+      //   @2024-03-01T23:30       America/Los_Angeles  2024-03-02T07:30Z  2024-03-02  WRONG DAY
+      //   @2024-03-01T00:30       Asia/Tokyo           2024-02-29T15:30Z  2024-02-29  WRONG DAY
+      //
+      // Nothing reports it, because the tag parses: a dashboard just opens on
+      // the wrong day depending on the worker's TZ. So the day is rebuilt from
+      // calendar fields rather than sliced off the UTC rendering.
+      collected[name] =
+         scalarTypeOf(tag) === "date"
+            ? authoredDay(text)
+            : unwrapFilterLiteral(text);
```
Sha-Bang: unwrapped unconditionally, but `givenSpec` keys the same unwrap on `declaration.type.startsWith("filter<")` and explains why: a plain string default that happens to read `f'x'` has to survive untouched. api-doc says this field follows the same rule ("`startingGivens` uses the same shape for the same reason"), and both go on the wire for the first time in this PR, so `# artifact { givens { NOTE=f'x' } }` on a `string` given publishes `x` and posts back as a different filter. needs the declarations in scope to key on.

packages/server/src/service/dashboard.ts
```diff
@@ -0,0 +1,1256 @@
...
+): DashboardLintFinding[] {
+   const findings: DashboardLintFinding[] = [];
+   const slug = manifest.name;
+   const add = (message: string, severity: "error" | "warn" = "warn") =>
+      findings.push({ subject: slug, message, severity });
+
+   // One selection, used for both the tag and the parse-failure report below,
+   // so the two cannot disagree about which annotations belong to this
+   // dashboard.
+   const ownAnnotations = manifest.tiles
```
Sha-Bang, line 810: this closes the query-level half of the "a line was dropped and nothing said" hole and leaves the file-level half open, for the single-query form only.

`ownAnnotations` is the query's annotations when `manifest.tiles` is absent, and `package.ts:2539` picks exactly one of the two lints, so a model-level `##` line that fails to parse or gets dropped in a file whose dashboard is single-query is reported by neither. measured at this head:

- `## note=@env.SECRET` + `# artifact { title="Overview" }` on a query → manifest builds, `lintDashboard` returns `[]`. same silence the `@env.` fix was about.
- worse shape: `## artifact { tiles=[orders -> by_x }` (unbalanced) beside a query-level `# artifact` → the composite tag is discarded, the file falls through to the single-query form and serves one tile, and findings are `[]`. `lintUndiscoveredDashboard` has the right message for it and never runs, because a manifest exists.

the scoping reason in the comment above is about a *sibling query's* broken tag being attributed to a dashboard that doesn't contain it, and that reason doesn't reach model-level annotations — those belong to the file either way. so I think this is just

```ts
for (const message of new Set(
   motlyParseErrors([...facts.modelAnnotations, ...ownAnnotations]),
)) {
```

which is a no-op on the composite path (same array, and the `Set` dedupes) and picks up the file-level line on the single-query one. the existing wording, "so that line had no effect on this dashboard", is already true for it.

Sha-Bang (approved):
re-reviewed at cd98e0d2, now that it's off the 981 stack and based on main, so the merge-base caveat from last round is gone. both things I raised are genuinely closed, and closed with tests that can fail:

- the `@env.` drop is reported now — `ENV_REFERENCE_DROPPED` is split out of the filter in `parseMotly` and carried through the rescue path, and `dashboard.spec.ts:1476` is no longer the vacuous `not.toContain` over `"{}"`; it asserts absence, the finding, and that a sibling tag still builds. the per-line-not-per-file scope is pinned separately at `:682`.
- `readStartingGivens` keys on the declared type, and the timestamp case that keying on `scalarType()` would have silently truncated is covered.

what I ran at this head: server unit 2975 pass / 0 fail; the dashboards integration spec 45/45; the full integration run 309/310, the one failure being `concurrent_package` fetching `gs://publisher_test_packages/gcs_faa.zip`, which is GCS creds I don't have locally and nothing to do with this. `tsc --noEmit` clean once `generate-api-types` has run.

things I checked rather than took from the comments, all of which held:

- `normalizeTileExpression` really is behaviour-identical to the expression it replaced. exhaustively fuzzed 7380 strings over `{o,-,>,space,tab,newline,_,->,double-space}` up to length 4 against your oracle: zero mismatches. (the one wrong note is in the comment, not the code — the old expression *did* produce the double space for `a->->b`, its own trailing `\s+` collapse just ate it.)
- the `authoredDay` residual reproduces exactly as documented, `@2024-03-01T16:00` under `America/Los_Angeles` → `2024-03-02`, and only there. UTC and Asia/Tokyo are correct for all four literals. the comment names the right literal and the right zone.
- `filterPublisherOwnedRenderLogs` is only needed on the query-time surface: `Model.validateRenderTags` filters to `severity === "error"` and the renderer emits `Unknown render tag` at `warn`, so the load-time package-warnings surface never carried them. the asymmetry is correct, not a miss.
- `Notebook.resource` → `environmentName`: confirmed against main that the server has never sent `resource` here, so this is the spec catching up rather than a break, and the release note says so.

one gap left, inline.

separately, and **not this PR** — it landed in 981, so there's nowhere to hang an inline: `endOfDelimited` doesn't represent the grammar's `repository.comments` rule, `#.*$`. checked against the parser rather than the tmGrammar, `# label="A" # this is a comment label="B"` really does parse to just `{label: "A"}`, so everything after a bare `#` is dead text the walk still treats as live. benign in both directions as far as I could push it — rewriting inside a comment can't change a parse that drops the comment, and the reverse case (`# artifact { title="x" # <<< } givens { R=f'US' } }`, where the walk reads the `<<<` in the comment as a heredoc and runs to the end) only *fails* to rescue, which is the direction you say to err in. I couldn't build a shape where it rewrites live text. worth taking anyway because of what the docblock claims: the delimiter set being grammar-derived rather than review-derived is the whole argument for why the fourth form won't be found by someone else, and this is a rule sitting in the grammar today. a `#` branch returning the index of the next `\n` closes it. also confirmed the walk leaves alone anything that already parses — 360 parsing annotations, zero meaning changes.

approving — the inline gap is a lint blind spot rather than anything wrong in what gets served, so it doesn't need to hold the release.

packages/server/src/service/dashboard.ts
mlennie: Taken, in `c53f2a2d`. Both shapes reproduced at my head before I changed anything, and both are exactly as you measured: `## note=@env.SECRET` beside a query-level `# artifact` builds a manifest and returns `[]`, and the unbalanced `## artifact { tiles=["orders -> by_x" }` discards the composite tag, falls through to the single-query form, serves ONE TILE, and returns `[]`. Your reading of the second as worse than a missing warning is right, and it is the part that decided this: the package does not merely stay quiet, it serves one tile as though that were the whole dashboard.

You are also right about my scoping comment, and I have said so in the code. That comment was defending against a sibling query's broken tag being blamed on a dashboard that does not contain it. A model-level line is not that case, because it belongs to the file however the dashboard's own tag was written. Worth naming the provenance too: the hole is in my own fix from two commits earlier rather than in the original code, which makes it the second time on this branch that the repair introduced the defect the next round found.

One deliberate deviation, and it is the only thing I did not take as written. I took the annotations as a SET rather than concatenating, because your no-op claim does not hold in the form you gave it. Your version passes all 87 tests, but concatenating is not a no-op: measured, a doubled syntax error comes back twice from `motlyParseErrors`, and only the message-level `Set` absorbs it. Across eight shapes, two differ that way. It works today and stops working the day a message distinguishes two occurrences by carrying a position. Deduplicating the input also keeps `parseMotly`'s per-line rescue from running over every model-level line twice, and it makes the no-op on the composite path structural rather than a property to re-measure whenever the parser changes.

So: your patch was right and your argument for it was not. Worth saying plainly rather than just reporting it fixed, because a correct patch with a wrong justification still fails the next time someone reasons from the justification, and on this branch that has now happened often enough to be the pattern rather than the exception.

Three tests, and the middle one asserts the premise before the finding (`expect(build(f)?.tiles).toBeUndefined()`) so it cannot pass for the wrong reason. Mutation-verified in both directions: reverting the scope to `ownAnnotations` fails exactly the two new single-query tests while the composite control keeps passing, which is what shows the control discriminates rather than failing along with everything else.

Gate on the merged head: typecheck 0, lint 0, unit 2978 pass / 3 skip / 0 fail across 140 files, integration 313 pass / 0 fail across 37. On the integration number specifically, an unchanged total is not evidence here, because this change alters the package `warnings` array and a fixture could have gained a finding no test asserts on. Grepped both runs for the new message text: 0 occurrences in each. So no fixture reaches the new path and the 313 is a real no-regression rather than a coincidence.

On `endOfDelimited`: agreed, and filing rather than fixing here, since it landed in #981 and is on `main` already. Your reason for wanting it is the right one and goes in the issue verbatim: the docblock's argument is that the delimiter set is grammar-derived rather than review-derived, and `repository.comments` (`#.*$`) is a rule sitting in the grammar today, so a claim to completeness is falsified by one omission even when the omission is harmless. Fix is a `#` branch returning the index of the next `\n`.

Sha-Bang: checked `c53f2a2d` on the merged head rather than taking it — your set-of-annotations version is better than what I wrote and your objection to my justification is correct. I'd retested my own patch after posting and reached the same place: concatenating duplicates the error and only the message-level `Set` absorbs it, so the equivalence was incidental. one thing I checked that neither of us mentioned: the combined parse can't cross-talk. 64 pairs that are clean apart, model-level first then the query's, including scalar-vs-block and array-vs-string conflicts on the same key — zero errors introduced by combining. so widening the input can't invent a finding.

reproduced your gate: unit 2978 pass / 3 skip / 0 fail across 140 files. mutation confirmed in both directions too — reverting the scope to `ownAnnotations` fails exactly the two new single-query tests and leaves "reports a model-level drop once on the composite form" passing, so the control does discriminate.

one thing left over, and it's cosmetic, so purely for whenever you're next in here rather than worth its own PR. the `ENV_REFERENCE_DROPPED` branch of `describeParseFailure` is the only one that ends with its own consequence clause, and all three callers append theirs, so it doubles up. rendered on the merged head:

```
lintDashboard    Annotation was dropped rather than parsed (…), so nothing on that line
                 took effect, so that line had no effect on this dashboard.
lintUndiscovered Tag was dropped rather than parsed (…), so nothing on that line took
                 effect, so the whole tag is discarded and this file is treated as a
                 shared include rather than a dashboard.
lintGivenTags    given "R" has an annotation that was dropped rather than parsed (…), so
                 nothing on that line took effect, so the whole line is discarded and the
                 given loses any label, control, range or suggest it declared. …
```

the middle one also has the two halves disagreeing about the unit — "that line" then "the whole tag". the other three branches read fine because they stop before the consequence, so dropping `, so nothing on that line took effect` from the env branch fixes all three at once and leaves each caller's own clause to say what was lost. the existing assertions are `stringContaining("@env.")`, which is why nothing caught it.

nothing here blocks anything — 987 is merged and this is prose only.

PR #989, opened 2026-08-12
fix(server): list schemas and load packages for database-less Snowflake connections

sagarswamirao:
Snowflake lets a connection omit `database`, resolving names against the session's current database instead. Publisher did not support that: a connection configured without one could not list its schemas, and could not back a source in a published package. Both failures are fixed here.

## Issue 1: listing schemas returned a 500

`getSchemasForSnowflake` built its introspection query as:

```sql
SELECT CATALOG_NAME, SCHEMA_NAME, SCHEMA_OWNER
FROM INFORMATION_SCHEMA.SCHEMATA
```

With no database configured, the `<db>.` prefix was empty and the reference was unqualified. Snowflake resolves an unqualified `INFORMATION_SCHEMA` against the session's current database, and a connection without one has no current database, so the request failed:

```
Cannot perform SELECT. This session does not have a current database.
Call 'USE DATABASE', or use a qualified name.
```

Qualifying the name is not a fix either. `INFORMATION_SCHEMA.SCHEMATA` is per-database in Snowflake, so `MYDB.INFORMATION_SCHEMA.SCHEMATA` returns only `MYDB`'s schemas. No form of that query can span databases.

**Fix.** The two cases now use different instruments. A connection with a configured database keeps the existing per-database query, byte for byte. A connection without one uses `SHOW SCHEMAS IN ACCOUNT`, a metadata command that requires no current database and reports every schema the role can see. Both branches emit the same `DATABASE.SCHEMA` name, which `listTablesForSnowflake` already parses back apart, so tables resolve as `DATABASE.SCHEMA.TABLE` either way.

Three details the branch has to get right:

- `SHOW` returns lower-case columns (`database_name`, `name`, `owner`) where `INFORMATION_SCHEMA` returns upper-case. Read with the upper-case keys alone, every owner is empty, which the `isHidden` rule treats as a system schema — hiding every row.
- `isDefault` is scoped to the configured database. Applied account-wide, a bare `schemaName === schema` marks a same-named schema in every database as the default.
- A configured `schema` is not applied as a filter when no database is set. It is a session default, not a selector; filtering on it account-wide would return that schema from every database and drop the rest.

## Issue 2: packages naming the connection failed to load

A model whose source named a database-less connection failed to publish:

```
line 8: import reference failure
  | source: overlap_report is snowflake_without_database.table(...)
```

The same source text compiled cleanly through `/compile`, which is what makes this one misleading: the error points at a model line, so it reads as a table-path problem, and the table path was already fully qualified.

The asymmetry is the cause. `/compile` runs on the main thread. Package load runs in the package-load worker pool, and the worker cannot hold native connection handles across the isolate boundary, so it builds a `ProxyConnection` from metadata fetched over RPC. That RPC calls `getDigest()`, and Malloy's `makeDigest` reads `.length` off every part it is given:

```js
parts.map(p => (p === undefined ? '{undefined}' : `${p.length}:${p}`))
```

It special-cases `undefined` only. A field omitted from config arrives as `undefined` and is fine; the `null` has a producer -- an explicit `"database": null` in `publisher.config.json`, or a client that serializes unset optionals as null over `POST /connections`. Once stored it is durable, because `ConnectionRepository` round-trips through `JSON.stringify`/`parse`, which drops `undefined` but preserves `null`. Such a value reached `getDigest()`, which threw `TypeError: null is not an object (evaluating 'p.length')` before any table was touched. The guard that should have removed it filtered on `undefined` alone, so `null` passed through:

```ts
([, fieldValue]) => fieldValue !== undefined,
```

**Fix.** `null` is treated as absent alongside `undefined`, applied to every nullable field (`database`, `schema`, `role`, `password`, `privateKeyPass`) rather than to `database` alone, since each one reaches `makeDigest`.

The load-bearing change is `removeUndefined` in `connection.ts`. That is the **key-pair path**, which bypasses Malloy's connection lookup entirely: `buildSnowflakePrivateKeyConnection` reads through `cloneApiConnection`, a shallow spread that leaves an explicit `null` intact. The `nullToUndefined` helper in `connection_config.ts` covers the config-pojo path, which Malloy's own lookup already guards on the pinned version (`if (value !== undefined && value !== null)` at both lookup sites); it is kept as defense in depth, since the dependency is a caret range and core's guard is not a contract.

The fix is identity-preserving: an explicitly-null field and an omitted one produce the same digest, so no content-addressed id splits across it.

## Verification

Against a live Snowflake account, on a connection with `database` unset:

- Schema listing returns 62 schemas across 10 databases, each named `DATABASE.SCHEMA`.
- Drilling into `ECOMMERCE_SNOWFLAKE.PUBLIC` — a database named nowhere in the connection config — lists its tables as `DATABASE.SCHEMA.TABLE`.
- A package with two sources over that connection publishes, and its queries return 10000 rows, 168063 rows, and a join across both databases in a single query.
- The same schema reached through a database-configured connection and a database-less one returns identical tables, columns, and `tablePath`.

Unit tests cover both fixes and were checked by reverting each fix in turn: dropping the database qualifier from the emitted schema name fails 5 tests, reading `SHOW` output with upper-case keys fails 4, restoring the `undefined`-only filter fails the key-pair digest test, and unscoping the blank-owner rule fails the imported-share test.

### Also addressed in review

- **Blank owner no longer hides shared databases.** `SHOW SCHEMAS` reports a blank `owner` for any schema not local to the account, so every schema of an `IMPORTED DATABASE` (a Marketplace or partner share) looked system-owned and was dropped from the Connection Explorer picker by default. The blank-owner test is now scoped to a configured database. On the verification account this moves 47 hidden to 46, surfacing exactly the share; every other blank-owner row is either `INFORMATION_SCHEMA` (caught by name) or inside a Snowflake-managed database (caught by the system set), so no system noise leaks in.
- **`SHOW SCHEMAS IN ACCOUNT` now passes an explicit `LIMIT 10000`.** Omitting it makes Snowflake error above ten thousand rows, failing the whole listing; passing it caps instead. The driver-side `rowLimit` cannot substitute, since the Snowflake connector applies it by slicing an already-fetched array. Landing on the cap is warned, as is dropping any row that yields no usable `DATABASE.SCHEMA` name.
- **A rejected database name stays a 400.** `assertSafeSqlIdentifier` throws `BadRequestError`, which the surrounding catch was flattening into a generic 500 ("retry later") — the retry-forever shape the sibling Trino and Databricks listings already guard against.
- **`storage=` federation with no database is pinned.** The generated secret carries no `DATABASE` line, so a passthrough session has no current database and its SQL must name tables in full. Asserted in `connection_federation.spec.ts`; the live build path itself is untested.
- **Spec matches the code.** The five optional Snowflake fields the code null-handles are declared nullable in `api-doc.yaml` (3.1 `type: ["string", "null"]`). `account`, `username`, `warehouse` and `privateKey` deliberately stay non-nullable, since validation rejects them absent.

## Known gap

Raw-SQL paths still require a session database or schema. `fetchSelectSchema` and `manifestTemporaryTable` in `@malloydata/db-snowflake` wrap the statement in an unqualified temp object:

```ts
const tempTableName = this.getTempViewName(sqlRef.selectStr);   // bare "tt<hash>"
await this.runSQL(`CREATE OR REPLACE TEMP VIEW ${tempTableName} AS (...)`);
```

Snowflake resolves that bare name against the session, so `/sqlSource`, `/sqlTemporaryTable`, and Malloy `sql()` sources fail with `Cannot perform CREATE TEMPVIEW. This session does not have a current database` (or `...current schema`, when `database` is set and `schema` is empty). Qualifying the statement does not help: the qualified name is inside the `SELECT`, while the unqualified object is the temp view wrapping it.

This is upstream in the driver, not in Publisher, and it is unaffected by this PR. `table()`-based models and `/sqlQuery` do not use that path and work today. The driver already handles the same problem correctly in `probeTableSize`, which qualifies conditionally via `parseSnowflakeTableName`; `getTempViewName` predates that and should follow it.

**Not carried here:** the truncation and dropped-row conditions above are logged server-side but not surfaced to the caller, so an agent asking `malloy_searchDatabaseSchema` what schemas exist sees a clean list. That tool already builds a `warnings` array, but the schema listing has no warning channel of its own (`getSchemasForConnection` returns `ApiSchema[]`), so plumbing them through means a contract change this PR otherwise does not make. Left as a follow-up.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4141,25 +4141,30 @@ components:
           type: string
           description: Snowflake username for authentication
         password:
-          type: string
+          type: ["string", "null"]
           description: Snowflake password for authentication
         privateKey:
           type: string
           description: Snowflake private key for authentication
         privateKeyPass:
-          type: string
+          type: ["string", "null"]
           description: Passphrase for the Snowflake private key
         warehouse:
           type: string
           description: Snowflake warehouse name
         database:
-          type: string
-          description: Snowflake database name
+          type: ["string", "null"]
+          description:
+            Snowflake database name. Omitted or null connects with no current
+            database, so names must be fully qualified as
+            DATABASE.SCHEMA.TABLE and schema discovery spans the account.
         schema:
-          type: string
-          description: Snowflake schema name
+          type: ["string", "null"]
+          description:
+            Snowflake schema name. Sets the session default schema; it does not
+            restrict which schemas are discoverable.
         role:
-          type: string
+          type: ["string", "null"]
           description: Snowflake role name
         responseTimeoutMilliseconds:
           type: integer
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4161,8 +4161,9 @@ components:
         schema:
           type: ["string", "null"]
           description:
-            Snowflake schema name. Sets the session default schema; it does not
-            restrict which schemas are discoverable.
+            Snowflake schema name. Sets the session default schema and forms part
+            of the connection's identity, so it determines what a materialization
+            build runs as. It does not restrict which schemas are discoverable.
         role:
           type: ["string", "null"]
           description: Snowflake role name
--- a/packages/server/src/service/connection.spec.ts
+++ b/packages/server/src/service/connection.spec.ts
@@ -1839,6 +1839,67 @@ describe("connection integration tests", () => {
             }
          });
 
+         it("should drop explicitly-null Snowflake options from private-key connOptions", async () => {
+            // This is the path that actually throws. A key-pair connection
+            // bypasses the core connection registry (which strips nulls itself)
+            // and is built directly by buildSnowflakePrivateKeyConnection, where
+            // cloneApiConnection's shallow spread leaves nulls intact. A
+            // surviving null reaches SnowflakeConnection.getDigest(), whose
+            // makeDigest reads `.length` off each part and special-cases only
+            // `undefined` -- so the digest throws, and because it is taken by the
+            // package-load worker's metadata RPC, the symptom is a whole package
+            // failing to load rather than a connection error.
+            //
+            // The fixture above omits these fields, making them `undefined`,
+            // which never reproduced the bug. They must be explicitly null here.
+            const { malloyConnections, releaseConnections } =
+               await createEnvironmentConnections(
+                  [
+                     {
+                        name: "sf_private_key_nulls",
+                        type: "snowflake",
+                        snowflakeConnection: {
+                           account: "test-account",
+                           username: "test-user",
+                           privateKey:
+                              "-----BEGIN PRIVATE KEY-----MIIB-----END PRIVATE KEY-----",
+                           warehouse: "test-warehouse",
+                           database: null,
+                           schema: null,
+                           role: null,
+                           password: null,
+                           privateKeyPass: null,
+                        },
+                     },
+                  ],
+                  testEnvironmentPath,
+               );
+
+            try {
+               const connection = malloyConnections.get(
+                  "sf_private_key_nulls",
+               ) as unknown as {
+                  connOptions: Record<string, unknown>;
+                  getDigest: () => string;
+               };
+               // Absent, not null: `null` is what breaks the digest.
+               for (const field of [
+                  "database",
+                  "schema",
+                  "role",
+                  "password",
+                  "privateKeyPass",
+               ]) {
+                  expect(connection.connOptions[field]).toBeUndefined();
+               }
+               // The real oracle: the digest is what threw, so compute it.
+               expect(() => connection.getDigest()).not.toThrow();
+               expect(typeof connection.getDigest()).toBe("string");
+            } finally {
+               await releaseConnections();
+            }
+         });
+
          it("should translate Trino Peaka credentials to core extraCredential", () => {
             const assembled = assembleEnvironmentConnections(
                [
--- a/packages/server/src/service/connection.ts
+++ b/packages/server/src/service/connection.ts
@@ -1476,10 +1476,25 @@ function entryToDuckDBOptions(
    return { ...removeUndefined(rest), name };
 }
 
+/**
+ * Drop keys whose value is absent, treating `null` as absent alongside
+ * `undefined`.
+ *
+ * `null` matters as much as `undefined` here: an optional field omitted from
+ * JSON config arrives as `null`, and the values this builds land in a Malloy
+ * connection whose `getDigest()` feeds them to `makeDigest`, which reads
+ * `.length` off each part and special-cases `undefined` alone. A surviving
+ * `null` therefore throws "null is not an object (evaluating 'p.length')" on
+ * the first digest. That digest is taken by the package-load worker's
+ * connection-metadata RPC, so the symptom is not a connection error but the
+ * whole package failing to load with "import reference failure" on the source
+ * line -- while the same model compiles cleanly through /compile, which runs on
+ * the main thread and never takes a digest.
+ */
 function removeUndefined<T extends object>(value: T): Partial<T> {
    return Object.fromEntries(
       Object.entries(value).filter(
-         ([, fieldValue]) => fieldValue !== undefined,
+         ([, fieldValue]) => fieldValue !== undefined && fieldValue !== null,
       ),
    ) as Partial<T>;
 }
--- a/packages/server/src/service/connection.ts
+++ b/packages/server/src/service/connection.ts
@@ -1536,12 +1536,14 @@ function entryToDuckDBOptions(
  * line -- while the same model compiles cleanly through /compile, which runs on
  * the main thread and never takes a digest.
  */
-function removeUndefined<T extends object>(value: T): Partial<T> {
+function removeUndefined<T extends object>(
+   value: T,
+): { [K in keyof T]?: Exclude<T[K], null> } {
    return Object.fromEntries(
       Object.entries(value).filter(
          ([, fieldValue]) => fieldValue !== undefined && fieldValue !== null,
       ),
-   ) as Partial<T>;
+   ) as { [K in keyof T]?: Exclude<T[K], null> };
 }
 
 function buildSnowflakePrivateKeyConnection(
--- a/packages/server/src/service/connection.ts
+++ b/packages/server/src/service/connection.ts
@@ -1525,16 +1525,31 @@ function entryToDuckDBOptions(
  * Drop keys whose value is absent, treating `null` as absent alongside
  * `undefined`.
  *
- * `null` matters as much as `undefined` here: an optional field omitted from
- * JSON config arrives as `null`, and the values this builds land in a Malloy
+ * This is the null guard for the connectors Publisher builds ITSELF, outside
+ * Malloy's connection lookup. Core strips nulls on its own registry path, so
+ * anything assembled into the Malloy config pojo is already covered; the two
+ * callers here are not. `buildSnowflakePrivateKeyConnection` bypasses the
+ * registry entirely, and reads through `cloneApiConnection`, whose shallow
+ * spread leaves an explicit `null` intact.
+ *
+ * `null` matters as much as `undefined` because the values this builds land in a
  * connection whose `getDigest()` feeds them to `makeDigest`, which reads
  * `.length` off each part and special-cases `undefined` alone. A surviving
- * `null` therefore throws "null is not an object (evaluating 'p.length')" on
- * the first digest. That digest is taken by the package-load worker's
- * connection-metadata RPC, so the symptom is not a connection error but the
- * whole package failing to load with "import reference failure" on the source
- * line -- while the same model compiles cleanly through /compile, which runs on
- * the main thread and never takes a digest.
+ * `null` therefore throws "null is not an object (evaluating 'p.length')" on the
+ * first digest. That digest is taken by the package-load worker's
+ * connection-metadata RPC, so the symptom is not a connection error but the whole
+ * package failing to load with "import reference failure" on the source line --
+ * while the same model compiles cleanly through /compile, which runs on the main
+ * thread and never takes a digest.
+ *
+ * A field omitted from config arrives as `undefined` and was always fine. The
+ * `null` has a producer: an explicit `"database": null` in publisher.config.json,
+ * or a client that serializes unset optionals as null over POST /connections.
+ * Once stored it is durable, because ConnectionRepository round-trips through
+ * JSON.stringify/parse, which drops `undefined` but preserves `null`.
+ *
+ * Stripping is identity-preserving: an explicit null and an omitted field
+ * produce the same digest, so no content-addressed id splits across the fix.
  */
 function removeUndefined<T extends object>(
    value: T,
--- a/packages/server/src/service/connection_config.spec.ts
+++ b/packages/server/src/service/connection_config.spec.ts
@@ -128,6 +128,67 @@ describe("assembleEnvironmentConnections — databricks", () => {
    });
 });
 
+describe("assembleEnvironmentConnections — snowflake", () => {
+   // An optional field omitted from JSON config arrives as `null`, not
+   // `undefined`. Malloy's makeDigest reads `.length` off every part and
+   // special-cases `undefined` alone, so a surviving `null` throws when a
+   // digest is taken -- which happens on the package-load worker's
+   // connection-metadata RPC, surfacing as the whole package failing to load
+   // with "import reference failure" rather than as a connection error.
+   const nullableFields = ["database", "schema", "role"] as const;
+
+   function snowflakeConnection(
+      overrides: Record<string, unknown>,
+   ): ApiConnection {
+      return {
+         name: "sf",
+         type: "snowflake",
+         snowflakeConnection: {
+            account: "acct",
+            username: "user",
+            password: "pw",
+            warehouse: "wh",
+            ...overrides,
+         },
+      } as ApiConnection;
+   }
+
+   for (const field of nullableFields) {
+      it(`omits a null ${field} from the core entry`, () => {
+         const { pojo } = assembleEnvironmentConnections([
+            snowflakeConnection({ [field]: null }),
+         ]);
+
+         const entry = pojo.connections["sf"] as Record<string, unknown>;
+         // Explicitly not null: `undefined` is what the digest tolerates.
+         expect(entry[field]).toBeUndefined();
+         expect(entry[field]).not.toBeNull();
+      });
+   }
+
+   it("preserves a configured database", () => {
+      const { pojo } = assembleEnvironmentConnections([
+         snowflakeConnection({ database: "MYDB", schema: "MYSCHEMA" }),
+      ]);
+
+      const entry = pojo.connections["sf"] as Record<string, unknown>;
+      expect(entry.database).toBe("MYDB");
+      expect(entry.schema).toBe("MYSCHEMA");
+   });
+
+   it("carries no null through to any core-entry value", () => {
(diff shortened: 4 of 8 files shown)


housejester (commented):
Good change — the diagnosis on both bugs is right, and the two-instrument design (per-database `INFORMATION_SCHEMA` vs `SHOW SCHEMAS IN ACCOUNT`) is the correct shape given `INFORMATION_SCHEMA.SCHEMATA` is per-database with no cross-database form. Holding `DATABASE.SCHEMA` as the invariant both branches emit, so `listTablesForSnowflake` keeps parsing it, is the right thing to have made load-bearing.

I checked the `SNOWFLAKE_SHOW_ROW_LIMIT` reasoning against a live account and it is exact — `SHOW SCHEMAS IN ACCOUNT LIMIT 10001` comes back `The LIMIT value for SHOW commands cannot exceed 10,000.`, so the "cannot be raised to page further" note is correct as written.

Four things below. None of them mean anything ships broken, so this is a comment rather than a change request. The first has live evidence and a one-line fix.

---

## 1. `owner === ""` hides real schemas from shared databases (`db_utils.ts:468`)

`isHidden` uses a blank `owner` as a proxy for "system schema". That was survivable inside one configured database; account-wide it now sweeps across databases the role does not own, and Snowflake returns a blank `owner` in `SHOW SCHEMAS` for schemas that are not local to the account — i.e. imported/shared databases.

Verified against an account with the same shape as the one in the Verification section (`SHOW SCHEMAS IN ACCOUNT LIMIT 10000` → 62 rows across 10 databases). Applying the rule in this PR verbatim hides 47 of 62. Exactly one is hidden *solely* because the owner is blank, in a non-system database:

```
SEC_FILINGS.CYBERSYN   owner=""
```

and:

```
SHOW DATABASES LIKE 'SEC_FILINGS'
  kind=IMPORTED DATABASE
  origin=...CYBERSYN_SEC_COMPANY_FILINGS_SHARE
  owner="ACCOUNTADMIN"
```

A Marketplace share. The *database* owner is `ACCOUNTADMIN`, but the schema's `owner` is blank because the schema is not local — so a deliberately-subscribed dataset gets flagged as a system schema. That is not cosmetic: `ConnectionExplorer.tsx:141` drops hidden schemas from the picker unless the user toggles them on, and `search_database_schema_tool.ts:331` tells the agent "Schemas marked isHidden are system schemas; your tables are in the others."

The fix is cleaner than it might look, because on real data the `""` clause contributes nothing else in the account-wide branch. Every other blank-owner row is either `INFORMATION_SCHEMA` (already caught by name — that is the one-blank-per-database pattern across all ten) or inside `SNOWFLAKE` / `SNOWFLAKE_SAMPLE_DATA` (already caught by `SNOWFLAKE_SYSTEM_DATABASES`; note the sample-data share is itself blank-owner, so the set is what actually carries it). Dropping `""` from the account-wide branch moves that account 47 → 46 hidden and surfaces exactly `SEC_FILINGS.CYBERSYN`, with no system noise leaking in:

```js
isHidden:
   owner === "SNOWFLAKE" ||                 // was ["SNOWFLAKE", ""].includes(owner)
   schemaName === "INFORMATION_SCHEMA" ||
   (!database && SNOWFLAKE_SYSTEM_DATABASES.has(catalogName)),
```

Keep `""` in the configured-database branch, where a database bounds it and the existing tests pin it.

Worth noting why the verification did not catch this: it counted schemas (62 across 10 databases) rather than checking the flag. Any account with a Marketplace or partner share hits it, so it is not an exotic edge.

## 2. The load-bearing fix is `removeUndefined`, and the tests pin the other one

The mechanism in the description is right — `makeDigest` special-cases `undefined` only, and a `null` throws. But I think the fix is attributed to the wrong guard, which matters because it is also what the tests cover.

Against the pinned Malloy (0.0.427), both connection-lookup sites already strip nulls before any connector factory runs:

```js
// api/foundation/config_lookup.js:70 — and connection/registry.js:143, identical
if (value !== undefined && value !== null) { connConfig[key] = value; }
```

So the `nullToUndefined(...)` calls in `connection_config.ts` protect the pojo path, which core already guards. The path that actually throws bypasses core: at `connection.ts:1909`, a connection with `hasSnowflakePrivateKey` skips the registry and goes to `buildSnowflakePrivateKeyConnection`, which reads `metadata.apiConnection.snowflakeConnection` — and `cloneApiConnection` (`connection_config.ts:170`) is a shallow spread, so nulls are intact there — then spreads through `removeUndefined` into `connOptions`.

Driving `makeDigest` with exactly the argument list `SnowflakeConnection.getDigest()` builds, for a key-pair connection whose optional fields are explicitly null:

```
BEFORE (undefined-only filter)  : THREW TypeError: Cannot read properties of null (reading 'length')
AFTER  (this PR's null filter)  : getDigest() -> 90230b35e9407a94...
control (fields omitted entirely): getDigest() -> 90230b35e9407a94...

AFTER digest === omitted-field digest: true
```

Two things follow.

**The one-line change at `connection.ts:1544` has no test.** All five new `connection_config.spec.ts` tests assert on `pojo.connections["sf"]`. The nearest existing test, `connection.spec.ts:1808` "should preserve Snowflake private-key auth options", omits `database` entirely — so it is `undefined`, not `null`, and would pass with the old filter. Revert `connection.ts:1544` and CI stays green while the reported bug returns. Adding `database: null, schema: null, role: null` to that existing fixture and asserting `connOptions.database` is `undefined` is a three-line change to a test that already exists, and it is the only thing that would pin this.

Worth keeping `nullToUndefined` regardless — `^0.0.427` is a caret range and core's guard is not a contract. It is defense in depth rather than the fix, and saying so in the comment would help. Relatedly, `removeUndefined`'s doc comment would be more useful describing what actually makes it load-bearing: it is the null guard for connectors Publisher builds *outside* core's lookup (this one and `entryToDuckDBOptions`).

**Also worth adding to the description:** the fix is identity-preserving — `"database": null` and an omitted `database` hash to the same digest — so no content-addressed source ids split across it. That matters given #993 just established `role`/`schema` as identity-bearing.

**And the provenance sentence is backwards.** "An optional field omitted from JSON config arrives as `null`, never as `undefined`" appears in the body and in three comments (`connection.ts:1528`, `connection_config.ts:1030`, `connection_config.spec.ts:89`). An omitted key arrives as `undefined` and has always worked. `config.ts:712` returns non-string scalars verbatim and `convertConnectionsToApiConnections` (`config.ts:1045`) is a plain spread, so the null needs a producer: an explicit `"database": null` in `publisher.config.json`, or a client that serializes unset optionals as null over `POST /connections`, which nothing validates. `ConnectionRepository.ts:55/138` then round-trips through `JSON.stringify`/`parse`, which drops `undefined` but keeps `null` — so once stored it is durable. Accurate version: "an explicitly-null field, or one a client serialized as null, survives to the connector."

## 3. The truncation guard only reaches the server log

`db_utils.ts:384` and `:411` both `logger.warn`, and the comments correctly argue that silence would present a partial list as complete. But neither consumer sees it: the SDK renders the array as-is, and an MCP agent that just asked "what schemas exist?" gets a clean list and concludes the table is not there — which is the exact failure the comment describes. `search_database_schema_tool.ts` already carries a `warnings` array in its Tier-3 response. Plumbing both conditions through `listSchemas` into it is small, and it is the difference between the guard working and being documented. Non-blocking.

## 4. Does a database-less connection still build through `storage=`?

Not a defect, a question — I have not run it. #993 made key-pair connections federatable, and `federateSnowflake` emits `DATABASE` only when supplied (`connection.ts:975`), so a database-less connection produces a secret with no `DATABASE` and `snowflake_query` runs against a session with no current database. That is the same family as the Known gap section's temp-view issue, but that section lists `/sqlSource`, `/sqlTemporaryTable` and `sql()` sources rather than `storage=` builds.

Nothing covers it either way: the three Snowflake federation tests in `connection_federation.spec.ts` all omit `database` and assert only on credential and role lines. Worth either confirming it works and pinning it with a line there, or adding it to Known gaps alongside the temp-view one.

Small related note: the api-doc text for `schema` — "Sets the session default schema; it does not restrict which schemas are discoverable" — is accurate for discovery, but #993 established that `schema` is also folded into the connection digest and determines what a `storage=` build runs as. A reader of the api-doc alone would take it as inert.

---

**Top three before merge:** (1) drop `""` from the owner test in the account-wide branch, (2) add `database: null` to the key-pair fixture so the actual fix is pinned, (3) correct the null-provenance sentence in the body and the three comments.

<sub>Reviewed with Claude Code (session "Zesty Lemur"). Live checks were read-only `SHOW` commands; the digest reproduction ran offline against the pinned Malloy.</sub>

sagarswamirao:
Thanks — all four were real, and the first two were the ones I would have shipped wrong. Three are fixed and pushed; finding 3 I have left out deliberately and would rather do separately. Details below.

## 1. Blank owner hiding shared databases — fixed

Reproduced exactly on the account from the Verification section, including the count:

```
total rows        : 62
hidden (current)  : 47
hidden (proposed) : 46
newly VISIBLE     : 1
   SEC_FILINGS.CYBERSYN  owner=(blank)
```

And it is the share you identified:

```
SHOW DATABASES LIKE 'SEC_FILINGS'
  kind:   'IMPORTED DATABASE'
  origin: '...CYBERSYN_SEC_COMPANY_FILINGS_SHARE'
```

Your safety argument also holds on real data. Categorizing all 38 blank-owner rows: 10 are `INFORMATION_SCHEMA` (caught by name, one per database), 27 are inside `SNOWFLAKE` / `SNOWFLAKE_SAMPLE_DATA` (caught by the system set), and exactly 1 is the share. So dropping the `""` test account-wide leaks no system noise.

Applied as `(Boolean(database) && owner === "")`, keeping the old behavior inside a configured database where the existing tests pin it. Live check after the change: 15 → 16 visible schemas, `SEC_FILINGS.CYBERSYN` surfaced, nothing from `SNOWFLAKE` leaking in. Pinned with a test using an imported-share row alongside a system-database row and an `INFORMATION_SCHEMA` row, so it also asserts the other two stay hidden on their own rules; reverting the scoping fails it.

You were right about why the original verification missed it: it counted schemas rather than asserting the flag.

## 2. Wrong guard credited, and the real one untested — fixed

This was the one worth catching. I confirmed your reading: both core lookup sites in 0.0.427 strip nulls before any connector factory runs, so the `nullToUndefined` calls guard a path core already covers, and the key-pair path bypasses it via `buildSnowflakePrivateKeyConnection` with `cloneApiConnection`'s shallow spread leaving nulls intact.

I verified the test gap the way you described — reverting `connection.ts` to the `undefined`-only filter left the suite green. It no longer does:

```
(fail) should drop explicitly-null Snowflake options from private-key connOptions
```

The new test sets `database`, `schema`, `role`, `password` and `privateKeyPass` explicitly null on a key-pair connection, asserts each is absent from `connOptions`, and then calls `getDigest()` — the digest being the real oracle here, since that is what threw.

`nullToUndefined` kept as defense in depth, and its comment now says so rather than claiming to be the fix. `removeUndefined`'s docblock now describes what makes it load-bearing: it is the null guard for connectors Publisher builds outside core's lookup.

The provenance sentence is corrected in the body and all three comments. You are right that an omitted key arrives as `undefined` and always worked; the null needs a producer, and once stored it is durable because `ConnectionRepository` round-trips through `JSON.stringify`/`parse`. I also added the identity-preserving note to the description — an explicit null and an omitted field produce the same digest, which matters given #993.

## 4. `storage=` with no database — pinned as a known shape

Confirmed the secret simply carries no `DATABASE` line, so the passthrough session has no current database and SQL run through it must name tables in full. Added a test to `connection_federation.spec.ts` asserting the omission, with a note that an empty `DATABASE ''` would be worse than absence since the extension would read it as a real database name. I have not exercised a live `storage=` build against a database-less connection, so I have documented the shape rather than claiming the build path works end to end.

Also took the api-doc point: `schema` now says it forms part of the connection's identity and determines what a materialization build runs as, rather than reading as inert.

## 3. Surfacing the truncation warnings — not in this PR

Agreed on the substance: a server-side `logger.warn` does not reach the caller, so an agent asking `malloy_searchDatabaseSchema` "what schemas exist?" gets a clean list and concludes the table is absent, which is the failure the comment describes.

You are right that the mechanism is already there, and it is closer to hand than I first credited: `warnings: string[]` is built at `search_database_schema_tool.ts:685` and is already used for precisely this shape of problem — the empty-schema case pushes a warning explaining that a bare `[]` reads as "no data" and saying which of the two it is. So this is appending to an existing array, not inventing a channel.

Leaving it out of this PR anyway, on scope rather than difficulty: the two conditions still have to travel from `db_utils` up through `listSchemas` into the tool, and the schema listing has no warning channel of its own today — `getSchemasForConnection` returns `ApiSchema[]`, so carrying them means either widening that return or a REST response shape, which is a contract change this PR otherwise does not make. Happy to do it as a follow-up, or to fold it in here if you would rather it not ship split.

---

Verification after the changes: 210 unit tests pass across the four spec files, `tsc` clean, and the database-configured path re-checked end to end against a live account — `listSchemas` still returns `isHidden: false` / `isDefault: true`, and the same table read through a database-configured and a database-less connection returns identical tables, columns and `tablePath`.

sagarswamirao:
image
image

housejester (approved):
All three fixes verified independently against a live account rather than read — I applied both the pre-fix and the shipped `isHidden` predicate to real `SHOW SCHEMAS IN ACCOUNT` output in both connection modes:

```
SHOW SCHEMAS IN ACCOUNT -> 62 rows

database-less                    hidden before=47  after=46  visible after=16
    now VISIBLE: SEC_FILINGS.CYBERSYN owner=""
database="SNOWFLAKE_LEARNING_DB" hidden before=47  after=47  visible after=15
    (no change -- configured-database behavior preserved)

blank-owner rows: 38 = 10 INFORMATION_SCHEMA + 27 in system dbs + 1 other
  other: SEC_FILINGS.CYBERSYN
```

**1 — the scoping is exactly right.** The share is recovered and **nothing is newly hidden**, and the configured-database mode is byte-identical: no row moves in either direction, so the existing behavior is genuinely preserved rather than approximately. My own categorization of all 38 blank-owner rows reproduces your 10/27/1 split, so the "leaks no system noise" argument holds on real data and not just on the fixture. The test is a better shape than I asked for too — asserting that the system-database row and the `INFORMATION_SCHEMA` row *stay* hidden on their own rules means it fails if someone later drops either the scoping or the system set, not only if they drop the scoping.

**2 — the new test is a better oracle than I suggested.** I asked for `connOptions.database === undefined`; calling `getDigest()` as well pins the thing that actually threw, so the test survives a refactor that changes how options are assembled. Good that you re-derived the gap rather than taking my word for it — that reverting to the `undefined`-only filter left the suite green was the part worth confirming firsthand. The comment rewrites read accurately now: `removeUndefined` explains what makes it load-bearing, `nullToUndefined` is labelled defense-in-depth with the caret-range reason, and the provenance correction landed in the body and all three comments.

**4 — pinned honestly.** Asserting the emitted shape while explicitly not claiming the live `storage=` path works end to end is the right call given neither of us ran it. The `DATABASE ''` observation is a good catch of your own — an empty value really would be worse than the omission.

**3 — agreed, and the deferral reasoning is sound.** `getSchemasForConnection` returns `ApiSchema[]`, so carrying the warnings up means widening that return or the REST response, which is a contract change this PR otherwise doesn't make. Worth tracking as owed follow-up rather than closed, though: a truncated or unparseable schema list still reads as complete to both the picker and an agent. Bounded to accounts above 10k schemas or a `SHOW` column drift, so not urgent — and you've already found the landing spot at `search_database_schema_tool.ts:685`, where the empty-schema case does exactly this. Happy for it to be a separate PR.

Nothing new introduced, CI green on `f975ee78`. Approving.

<sub>Reviewed with Claude Code (session "Zesty Lemur"). Live checks were read-only `SHOW` commands.</sub>

PR #990, opened 2026-08-12
feat(authorize): allow a trusted request to run with #(authorize) gates skipped

Sha-Bang:
## What this does

Adds a per-request way to run a query with the author's `#(authorize)` gates **not evaluated**, for trusted data-management callers.

The problem it solves: indexing runs as a machine identity with no givens, so any source carrying an `#(authorize)` gate returns 403 and is never scanned. An author who tags a dimension `#(index)` on an ACL'd source gets an empty search index — the tag opted in, and the gate silently revoked it. There is no way to express "scan this for indexing" today.

Only expressions collected from `#(authorize)` / `##(authorize)` are skipped. The author's own `where:` clauses, `#(filter)` handling, and every other semantic are untouched. This is **not** `bypassFilters`, a separate deprecated `#(filter)`-only control in the request body — the two are orthogonal and both directions are pinned by tests.

```mermaid
flowchart LR
  R["caller<br/>(trusted data-management path)"] -->|"x-publisher-bypass-authorize: true"| S[server.ts]
  S -->|"readBypassAuthorize(req)"| G["getQueryResults"]
  G --> A1["assertAuthorized<br/>(named source)"]
  G --> A2["assertAuthorizedForAllSources<br/>(compiled entry-point walk)"]
  A1 --> X{"bypass?"}
  A2 --> X
  X -->|no| E["evaluate #(authorize) exprs → 403 on deny"]
  X -->|yes| C["count + audit, skip evaluation"]
  B["QueryRequest body"] -.->|"never consulted"| X
```

The bypass is applied inside the two methods that evaluate a gate rather than at their six call sites, so there is one place to audit and no future call site can forget it. The private `assertAuthorizedExprs` they share has no other caller, which is what makes those two a complete cut.

## Decisions and tradeoffs

**A header, not a `QueryRequest` field.** This is the central decision. A deployment fronting this API generally reuses the same generated request schema for its own inbound body — in Credible's case the router's outbound `PublisherQueryRequest` and its own *public* `QueryRequest` are literally the same schema. So a body field able to reach here would also be settable by that deployment's external callers: the `bypassFilters` mistake, repeated on `#(authorize)`. A header is set on the client at construction and never passes through body conversion, so there is no strip to forget and no caller body value can reach it. `bypassAuthorize` in the body is inert, and there is a test that says so.

**Not in `api-doc.yaml` at all — not as a parameter, and (after review) not as prose either.** Declaring a header parameter would insert an optional positional argument into `executeQueryModel` for every generated SDK and CLI consumer, for a control almost none should set. The first draft documented it in the operation `description` instead; that came back out, because the API reference is read by API *consumers* who should never set this, while the party who has to act on it is the *operator* who must strip it at the edge. It now lives in [`docs/authorize-bypass-deployment.md`](docs/authorize-bypass-deployment.md) plus a section in `docs/authorize.md`. The honest cost is unchanged: documentation, not a machine-checked contract.

**Publisher bounds nothing here, deliberately.** It performs no inbound authentication, so nothing in this repo constrains who may send the header. The bounds are the deployment's: publisher reachable only from its router, and the router asserting a machine identity on the private endpoint that sets it. Worth stating plainly rather than implying a guarantee this layer cannot make — a deployment that exposes workers directly already lets a caller forge givens and pass essentially any gate, with or without this feature.

**The notebook path does not accept the bypass.** No caller needs it, so its gates stay enforced.

**Observability.** `publisher_authorize_bypass_total`, labelled only by `entry_point` (`source` | `runnable`). Org / package / model / source are unbounded cardinality, so they ride the paired `authorize bypass` audit line instead — the counter is the rate signal you alert on, the log line is what an investigation reads once it fires. Any nonzero value on a path that should not use the bypass is a finding.

## Review round (jswir)

Approved with four fixes, all applied:

- **The audit lost the source exactly when it mattered.** `assertAuthorizedForAllSources` returned before resolving the run target, logging `"(query)"` — on the one request where nothing else records the target (an ad-hoc query where the `source` entry point never fires either). Now resolved, still skipping the expensive walk. Also corrected the "emits one of each" claim: `runnable` always fires, `source` only when a target resolved pre-compilation.
- **The wiring test proved less than it looked like.** A `toContain` on source text passes from a comment, and the body-read check was blocklist-shaped. Replaced by a new integration spec sending real HTTP requests through the same middleware and call shape, plus a *positive* assertion on the extracted argument list with comments stripped.
- **Removed the dead `bypassAuthorize` parameter** from `assertAuthorizedForText` / `assertAuthorizedForRunnable` — no caller passed it, and `/compile` returns schema and, with `includeSql`, SQL.
- **Docs that this feature made untrue:** `authorize.md`'s "runs on every entry point" and its Known-limitations list, `security-posture.md`'s "modeling concern, not a request concern", and `SECURITY.md` — which listed "a gate granting access it should deny" as in-scope, i.e. invited a report against our own feature. Now scoped out for the documented header and scoped **in** for honouring one we should not.

Two of my own claims were wrong and are fixed: Node comma-joins a duplicated custom header rather than producing an array (the case that actually occurs now has a test, sent via `node:http` because supertest's `.set()` twice replaces rather than appends), and a trailing-space value is accepted by design.

Held back deliberately: the options-object refactor of `getQueryResults` (ten positional params is real, but it touches every call site) and the ingress forward-allowlist (a gateway change with blast radius beyond this feature).

## How it was tested

`packages/server`: 161 tests across `authorize_bypass_header.spec.ts` (new), `authorize_bypass_wiring.integration.spec.ts` (new), and `authorize_integration.spec.ts`; `tsc --noEmit` clean; eslint/prettier clean.

The integration tests compose the real header reader with the real gate rather than stubbing the flag, so they cover the actual path a request takes. Specifically pinned:

- bypass **honoured** with the header — a gated source returns rows;
- **body field is inert** — `bypassAuthorize: true` in the body with no header leaves gates enforced. This one pins the whole safety argument; without it a future refactor silently reopens the hole;
- gates still enforced with no header, an absent header, a **genuinely duplicated** header (comma-joined by Node, sent through `node:http` since supertest cannot express it), an array-valued header, and any value other than a lone `true`;
- the author's own `where:` still narrows a bypassed scan, and `bypassFilters` is neither read nor written by this path;
- the same two claims over real HTTP — header set → the route passes the bypass; body field only → it does not — so the safety argument is behavioural, not textual;
- a *positive* source assertion that the argument the route hands the controller is exactly `readBypassAuthorize(req)`, comments stripped, which no aliasing of `req.body` can satisfy;
- the `runnable` audit line naming a resolved source on an ad-hoc query, which is the case that used to log `"(query)"`.

Each new assertion was verified to fail against pre-fix code, one branch at a time: the audit regression, the route reading the body, the test app reading the body, and a reader that accepts a comma-joined duplicate.

## Not in scope

Any inbound authentication in publisher; enabling materialization of gated sources; and the deployment-side half (the router's private endpoint, its M2M gate, and the ingress strip of this header), which lives in the consuming repo.

---

## One CodeQL alert, left for maintainers to call

CodeQL flags `js/missing-rate-limiting` (high) at `packages/server/src/server.ts:1756` — *"This route handler performs authorization, but is not rate-limited."*

Context before anyone reaches for a fix:

- **The route is pre-existing and otherwise unchanged.** The only edit inside that handler is one added argument, `readBypassAuthorize(req)`, plus a comment. What changed is CodeQL's *classification*: reading a header that feeds an access-control decision makes it see the handler as authorization-performing. The authorization itself (`Model.assertAuthorized`) ran on this same route before this PR.
- **`main` already carries 2 open alerts for this same rule** (`server.ts:627` and `:1976`, the "file system access" variant), and there is no rate limiting anywhere in `packages/server` — no `express-rate-limit`, no equivalent. So this is an accepted finding class in the repo, not a new one.
- **I did not add a rate limiter.** Putting one on the query route is a real behavioural change with config and defaults to decide, and it belongs to whoever owns this server's operational posture — not to a PR about letting a trusted caller skip gates. `queryConcurrency()` already bounds in-flight work on this route, which is a concurrency limit rather than a rate limit and is not what CodeQL is looking for.

So: dismiss it as accepted alongside the existing two, or tell me you'd like rate limiting and I'll raise it separately for the whole server rather than one route. Happy either way — I just didn't want to quietly suppress an alert in your repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,11 @@ packages/app/playwright-report/
 
 # SPA bundle copied next to the from-source server so it can serve the web UI
 # (the server looks for app/ beside the running server file).
+# Both spellings: the trailing-slash form matches only a directory, so a
+# symlink of the same name (a common local dev convenience) slipped past it
+# and got committed once via a directory-scoped `git add`.
 packages/server/src/app/
+packages/server/src/app
 
 # Publisher runtime artifacts (DuckDB metadata + per-env package clones).
 # Created wherever the server is launched from; already covered by
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -46,6 +46,14 @@ caller-asserted conventions rather than boundaries — that shapes what counts a
   [docs/authorize.md § Security model](docs/authorize.md#security-model) and
   [§ Known limitations](docs/authorize.md#known-limitations), and in
   [docs/row-level-access.md](docs/row-level-access.md).
+- A gate skipped for a request carrying `x-publisher-bypass-authorize: true` — the documented
+  data-management bypass ([docs/authorize.md § Authorize
+  bypass](docs/authorize.md#authorize-bypass-for-trusted-data-management-callers)). Publisher
+  bounds nobody who may send it; stripping it at the edge is the deployment's job
+  ([docs/authorize-bypass-deployment.md](docs/authorize-bypass-deployment.md)). A finding that
+  Publisher *honours* the header is working as documented. A finding that it honours one it should
+  not — on a notebook or `/compile` request, from a body field, or without counting and logging
+  it — is in scope below.
 - Broad reach for whoever can publish a package or `PATCH` a connection on a bare Publisher
   ([docs/query-metadata.md](docs/query-metadata.md), [docs/packages.md](docs/packages.md)).
 - The default `Content-Security-Policy: frame-ancestors *`, which `PUBLISHER_FRAME_ANCESTORS` exists
@@ -61,7 +69,10 @@ Anything that breaks a boundary Publisher does claim, including:
   **queryable == discoverable** boundary in
   [docs/discovery-and-access.md](docs/discovery-and-access.md).
 - An `#(authorize)` gate granting access its expression should deny for the givens actually sent,
-  including none.
+  including none — absent the documented bypass header above.
+- The bypass applying where it should not: honoured on a notebook cell or `/compile` request,
+  reachable from the request body rather than the header, surviving onto a subsequent request, or
+  skipping a gate without incrementing `publisher_authorize_bypass_total` and logging it.
 - Connection credentials or secrets leaking through an API response, log line, or error.
 - Escaping a package's static root — path traversal or symlink past the rejections in
   [docs/html-data-apps.md § Security model](docs/html-data-apps.md#security-model).
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1602,26 +1602,6 @@ paths:
         as a raw Malloy query string or by referencing a named query within the model. This endpoint
         supports both ad-hoc queries and predefined model queries, making it flexible for various
         use cases including data exploration, reporting, and application integration.
-
-        **`x-publisher-bypass-authorize` (internal).** When this request header is exactly
-        `true`, the author's `#(authorize)` gates are not evaluated for this query. It exists for
-        trusted data-management callers: indexing runs as a machine identity with no givens, so a
-        gated source otherwise returns 403 and is never scanned even when the author tagged a
-        dimension `#(index)`.
-
-        Only expressions collected from `#(authorize)` / `##(authorize)` are disabled. The
-        author's own `where:` clauses, `#(filter)` handling, and every other semantic are
-        untouched, and this is NOT `bypassFilters` — a separate, deprecated, `#(filter)`-only
-        control in the request body.
-
-        Deliberately a header, not a `QueryRequest` field: a deployment fronting this API reuses
-        the same generated schema for its own inbound body, so a field able to reach here would
-        also be settable by that deployment's external callers. It is deliberately not declared
-        as a parameter either, so it does not become an argument on every generated SDK and CLI
-        client for a control almost no caller should set. Publisher performs no inbound
-        authentication, so nothing here bounds who may send it — the bound is the deployment's
-        network boundary. Every use is counted (`publisher_authorize_bypass_total`) and
-        audit-logged.
       parameters:
         - name: environmentName
           in: path
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1602,6 +1602,26 @@ paths:
         as a raw Malloy query string or by referencing a named query within the model. This endpoint
         supports both ad-hoc queries and predefined model queries, making it flexible for various
         use cases including data exploration, reporting, and application integration.
+
+        **`x-publisher-bypass-authorize` (internal).** When this request header is exactly
+        `true`, the author's `#(authorize)` gates are not evaluated for this query. It exists for
+        trusted data-management callers: indexing runs as a machine identity with no givens, so a
+        gated source otherwise returns 403 and is never scanned even when the author tagged a
+        dimension `#(index)`.
+
+        Only expressions collected from `#(authorize)` / `##(authorize)` are disabled. The
+        author's own `where:` clauses, `#(filter)` handling, and every other semantic are
+        untouched, and this is NOT `bypassFilters` — a separate, deprecated, `#(filter)`-only
+        control in the request body.
+
+        Deliberately a header, not a `QueryRequest` field: a deployment fronting this API reuses
+        the same generated schema for its own inbound body, so a field able to reach here would
+        also be settable by that deployment's external callers. It is deliberately not declared
+        as a parameter either, so it does not become an argument on every generated SDK and CLI
+        client for a control almost no caller should set. Publisher performs no inbound
+        authentication, so nothing here bounds who may send it — the bound is the deployment's
+        network boundary. Every use is counted (`publisher_authorize_bypass_total`) and
+        audit-logged.
       parameters:
         - name: environmentName
           in: path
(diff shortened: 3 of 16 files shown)


jswir:
Read through this carefully. The design is right — header instead of body field, cut at the two methods that actually evaluate a gate, `where:` left alone. I verified those rather than trusting the comments: `assertNoCallerAuthorizeAnnotation` still runs unconditionally, restricted mode still bans raw SQL, row/byte caps are unchanged, and MCP, notebooks, and `/compile` all call with the flag defaulted. The `bypassFilters` characterization checks out, and the CodeQL note is accurate — that rule already has two open alerts on `main`.

The gaps are around the mechanism, not in it.

## Worth fixing before merge

**1. The header has to be stripped at ingress, and nothing makes that discoverable.**

This is the one that matters. Many proxies forward unrecognized headers by default. Where that's true, an end user of the fronting app can set `x-publisher-bypass-authorize: true` on their own request, it rides through untouched, and the gate is off — no guessing required, since the header name and value are published in our own API docs.

FGA covers the tenant boundary, so this isn't a cross-tenant issue. The residual case is `#(authorize)` doing role- or row-level gating *within* a tenant the caller is legitimately entitled to reach: FGA correctly allows the request, and the header still disables the in-model gate.

I don't think the fix is declaring it in `api-doc.yaml` — your reasoning about polluting every generated client is sound, and `x-internal` support across generators is inconsistent. The durable fix is at the gateway: **allowlist the headers we forward rather than blocklisting this one.** A blocklist is one forgotten header away from a hole. If ingress already forwards only known headers this is close to a non-issue — but then let's say so in the deployment docs, because right now nothing tells an operator this header exists.

**2. The audit log loses the source exactly when you need it.**

`noteAuthorizeBypass` says "a bypass that does not appear here did not happen." Not quite. The `source` emission is guarded by `if (earlySource)`, and the `runnable` path passes `sourceName: undefined` because it short-circuits before `resolveAuthorizeSourceFromRunnable`. So for an ad-hoc query whose target can't be pinned up front, the line records package and model but logs the source as `"(query)"`.

That's the case where an investigator can least reconstruct what was read, and the audit trail is the main compensating control here. Either resolve the source name before returning, or narrow the claim in the docstring and in `authorize_metrics.ts` ("a single bypassed query emits one of each" is also false in that case — the telemetry test only covers a declared named source, where both do fire).

**3. The route-wiring test proves less than it looks like it does.**

`expect(serverSource).toContain("readBypassAuthorize(req)")` passes if that string sits in a comment or in dead code, and it doesn't check the value reaches the right argument position. The body-read check is blocklist-shaped — `const b = req.body; … b.bypassAuthorize` slips past it.

`oom_guards.integration.spec.ts` already mounts a real Express app with supertest using the same middleware and controller wiring. Two requests through that — one with the header, one with only the body field — would make the safety argument load-bearing instead of textual. The rest of the suite is genuinely good; this is the one test guarding the most important claim.

**4. Remove the unused `bypassAuthorize` parameter on `assertAuthorizedForText` / `assertAuthorizedForRunnable`.**

No caller passes it — both `/compile` call sites in `environment.ts` still pass two arguments, and no test exercises it. Harmless today, but it's the plumbing already run to a path that can leak schema and, with `includeSql`, SQL. A future change could enable it with one word and no review friction.

## Follow-ups

- **Docs contradict the code.** `docs/authorize.md` still says the gate runs on every entry point, and its "Known limitations" list — which carefully enumerates every known way to defeat the gate — doesn't include this one. `docs/security-posture.md` says governance is "a modeling concern, not a request concern," which this makes untrue. And `SECURITY.md` lists "a gate granting access it should deny, including for no givens" as in-scope for a vulnerability report, so as written we're inviting a report for our own feature.

- **"A repeated header arrives as an array" isn't what Node does.** I checked with a raw socket: duplicates arrive comma-joined as the string `"true, true"`. Behavior is still correct (fails closed), just not for the stated reason — and the case that actually occurs has no test.

- **`getQueryResults` is now ten positional parameters**, with `bypassFilters` at #5 and `bypassAuthorize` at #10. The test helper already has to write `undefined, undefined, "full", bypassAuthorize` to reach the last slot, and a prettier commit was needed because the new argument pushed a call over print width. Options object.

- **Comment altitude.** The router / `/private/**` / M2M narrative appears twice in slightly different words and describes infrastructure that doesn't exist in this repo. Keep the two facts checkable here (body-blindness; nothing in Publisher bounds who sends it) and let the deployment argument live in the PR description, where it already is well written.

- `{@link HeaderSetter}` in `authorize_bypass_header.ts` points at a symbol in `filter_deprecation.ts` that isn't imported, so it won't resolve.

## One open question, not a blocker

`docs/authorize.md` names identity-bound givens as a planned milestone — a trusted-proxy value the caller can't override. That shape would keep the decision with the model author (`#(authorize) "$ROLE='analyst' or $SYSTEM_CALLER='indexer'"`) instead of removing it globally, and it wouldn't apply to sources the author never tagged `#(index)`.

I don't think that blocks this — the milestone isn't built and the indexing problem is real now. But if this ships as the interim, worth saying so in the code and pointing at that milestone, so it doesn't quietly become the permanent answer.

jswir (approved):

PR #992, opened 2026-08-12
feat(server): report how a query was served and what it cost, and bound the query histogram's labels

housejester:
The server measures several things and then discards them, and the query histogram carries two labels that grow without bound. This reports the first and bounds the second.

## The histogram's labels were unbounded

`malloy_model_query_duration` labelled every sample with the **query text** and the **returned row count**. Both are unbounded — ad-hoc text gives a new series per distinct query, a row count one per distinct result size — and a histogram label multiplies by the bucket count, so the metric grew for as long as a process served traffic.

On one deployment serving real traffic, the query-text label alone carried ~637 distinct values across ~14.9k series for this one histogram.

Both move to the request log, where cardinality costs nothing. `environment` and `package` take their place: previously the only identity on the metric was a bare model path, which is not unique across packages.

## The histogram now measures execution, not preparation

The timer stopped before the warehouse round trip, so `malloy_model_query_duration` excluded the one part its own description ("how long it takes to execute a Malloy model query") named. It now spans compile, authorize, routing, prepare and execution.

**Expect every existing p95 panel to step up at deploy.** That is the metric starting to measure what it always claimed, not a regression — and it is the one operationally visible change here, so it is worth knowing before the graphs move.

## Query results now say how they were produced

`QueryResult` gains three fields:

- **`servedFrom`** — `storage`, `live_fallback`, or null. A storage-served answer is byte-identical to a live one, so this is the only way a caller can tell a materialized source did anything. `live_fallback` needs its own value rather than folding into `storage`: it is a success answered by the live warehouse, and counting it as a hit would report a healthy hit rate for a broken store.
- **`executionTimeMs`** — compile through execution, the same span the histogram records, so the two are comparable. Excludes serialization and transport, so it reads lower than a caller's own stopwatch.
- **`queryCostBytes`** — bytes SCANNED, from `runStats`, which the BigQuery connector already populates and nothing ever read. Keeps Malloy's own field name so the figure can be followed back to its source.

Plus `malloy_model_query_scanned_bytes`, the same figure as a counter.

## Builds report their duration

`ManifestEntry` gains `buildDurationMs` — already measured for the build histogram, previously sent upward as null — and `queryCostBytes`, which a **colocated** build fills from the Malloy connection's own statistics.

**A `storage=` build reports a null cost, and that is a known gap rather than a property of the path.** Its read goes through DuckDB's native query-passthrough, where no Malloy connector is in the call path to report statistics. See below.

## What this deliberately does not do

An earlier revision of this PR read the cost back from the warehouse's own accounting after a `storage=` build, correlating on a time window plus an exact match on the recorded query text — because neither passthrough returns a job handle for a rows-returning call.

That has been removed rather than fixed. Review found that embedding the recorded query text as a literal inside a **Snowflake**-parsed statement breaks on any compiled SQL containing a backslash, which Malloy emits for an apostrophe in a string literal: quote-doubling alone does not survive a dialect where backslash is an escape character, so the literal terminates early and the remainder re-parses. Confirmed against a live account — the statement fails compilation, and the failure was swallowed into a caught error and a debug log.

A dialect-aware escaper would have closed that specific hole, but the correlation key itself is the problem it comes from. Identifying the job by an **id** removes the query text from the accounting statement altogether, along with the time window, the candidate cap, and the ambiguity rule. Obtaining an id restructures how the build issues its read, so the cost reporting belongs with that change rather than ahead of it.

Removing it also drops a 10-second lookup that ran inside the measured build, inflating both `publisher_materialization_source_build_duration_ms{engine="storage"}` and the reported `buildDurationMs` by however long the accounting query took.

Everything not keyed on query text stays.

## For consumers generating clients from this spec

`QueryResult` and `ManifestEntry` both gain fields. Strict generated clients reject unknown properties — openapi-generator's Java/Gson `validateJsonElement` throws on any field absent from the client's `openapiFields` — so a consumer running this server against a client generated from an older spec fails at deserialize on every affected response. **Regenerate clients in the same change as the version bump**, not after it.

## Notes for review

- **Tests:** unit suites green. `bun test` in `packages/server` reports 2539 pass / 47 fail; the 47 are identical on this branch's parent commit and are environment-dependent integration and E2E tests, unrelated to this change. The delta against that baseline is exactly the 17 tests of the removed `build_cost.spec.ts`.
- MCP's `execute_query` does not yet surface `servedFrom` / `executionTimeMs` / `queryCostBytes`. Noted, not addressed here.


packages/server/src/service/build_cost.ts
```diff
@@ -0,0 +1,357 @@
...
+                 END_TIME_RANGE_START => TO_TIMESTAMP_LTZ(${lit(since.toISOString())}),
+                 RESULT_LIMIT => ${MAX_CANDIDATES}))
+      WHERE QUERY_TEXT = ${lit(opts.sql)}
+        AND EXECUTION_STATUS = 'SUCCESS'
+        -- QUERY_HISTORY is scoped by ROLE visibility, not by user. Under a role
+        -- carrying MONITOR, an exact text match can resolve to another
+        -- principal's run of the same SQL -- one row, so the ambiguity guard
+        -- would not catch it, and the cost would be attributed here.
+        AND USER_NAME = CURRENT_USER()`;
+   const sql = `SELECT * FROM snowflake_query(${lit(inner)}, ${lit(opts.handle ?? "")})`;
```
sagarswamirao: `lit()` doubles single quotes only, but on this path the escaped string is parsed by **Snowflake**, where a backslash is an escape character inside a single-quoted literal by default (`\'`, `\\`, `\n`, ... are all documented sequences). Flagging early so it doesn't bite downstream: `buildSQL` is compiled Malloy SQL, and this repo already records that Malloy renders string literals with backslash escaping on Snowflake (`incremental_apply.ts:712`), so backslashes reaching here are ordinary rather than exotic.

Traced with `buildSQL = SELECT * FROM t WHERE x LIKE 'a\'`. The backslash passes through both nesting levels untouched (DuckDB treats `\` as a literal character, so the outer layer preserves it), and Snowflake then reads `\'` as an escaped quote:

```
WHERE QUERY_TEXT = 'SELECT * FROM t WHERE x LIKE ''a\'''
                                                     ^^ consumed as an escaped quote
```

The literal runs past its intended terminator and the trailing `AND EXECUTION_STATUS = 'SUCCESS'` is re-parsed, putting author-controlled text in statement position -- on a session that still holds the federated warehouse credentials, and with the error swallowed into `outcome: "error"` + `logger.debug`, so it fails silently.

Worth noting this differs from `wrapPassthrough`, which embeds the same `buildSQL` with quote-only escaping at **one** level, where it is the statement Snowflake is meant to execute. The second nesting level added here is what turns that SQL into *data inside a Snowflake-parsed literal*, and that is the level quote-doubling doesn't survive.

The codebase already has the dialect-aware escaper for exactly this: `sqlLiteral(value, connectionType)` in `introspection_sql.ts`, whose `BACKSLASH_ESCAPE_DIALECTS` includes `snowflake` with the note that Snowflake has no parameter to disable backslash escapes. Its docstring records the boolean form of this being set wrong twice in successive review rounds, and `db_utils.spec.ts:1237-1241` states that doubling alone on a backslash-escaping dialect "is exploitable rather than merely cosmetic." `escapeSqlString` in `incremental_apply.ts` -- the file this module already imports `SqlRunner` from -- does the same thing.

Could you switch this path to `sqlLiteral(..., "snowflake")` (or take the `LAST_QUERY_ID()` / `QUERY_HISTORY_BY_SESSION` key this file already calls the better long-term shape, which removes author SQL from the accounting statement altogether)? Also worth correcting `lit()`'s docstring -- "these strings are server-built, never caller-supplied" isn't true of `opts.sql`, and that line is what would stop a future reader from rechecking.

packages/server/src/service/build_cost.ts
```diff
@@ -0,0 +1,357 @@
...
+   // That is worse than an error, because a null here reads as "Snowflake has no
+   // such figure". Same reasoning and same fix as `probeAlias` in
+   // incremental_apply.ts, which the codebase already learned this on.
+   const inner = `
+      SELECT QUERY_ID          AS "job_id",
+             BYTES_SCANNED     AS "bytes_scanned",
+             EXECUTION_TIME    AS "execution_time_ms",
+             ROWS_PRODUCED     AS "rows_produced"
+      FROM TABLE(INFORMATION_SCHEMA.QUERY_HISTORY(
+                 END_TIME_RANGE_START => TO_TIMESTAMP_LTZ(${lit(since.toISOString())}),
```
sagarswamirao: `since.toISOString()` always emits a trailing `Z`, and Snowflake's AUTO timestamp-format detection documents `YYYY-MM-DD"T"HH24:MI:SS.FFTZH:TZM` (an explicit `+00:00`-style offset) with no bare-`Z` variant; the docs state that if no matching format is found, Snowflake returns an error.

If that's how it resolves, the failure is invisible in the worst way: the error is caught, every Snowflake build records `outcome=error` with a null cost indefinitely, and the only signal is the counter. Given the PR notes this path's first real `storage=` build is also its first execution, it seems worth not shipping the question -- `since.toISOString().replace("Z", "+00:00")` matches a documented AUTO pattern exactly and retires it in one line. An explicit format string would do as well.

I couldn't settle this from the docs alone (they neither list nor explicitly reject `Z`) and had no live account to test against, so treating it as unverified rather than confirmed-broken.

sagarswamirao (changes requested):
Reviewed at `870f1dc`. The observability reasoning here is careful in ways worth calling out: keeping `live_fallback` distinct from `storage` so a broken store can't report a healthy hit rate, the null-is-not-zero discipline (I ran the BigQuery JSON paths through DuckDB -- a missing leaf and a missing branch both yield SQL `NULL`, so `num()` never fabricates a `0`), and returning null on `ambiguous` rather than taking the newest match. I also traced the Snowflake quote double-nesting and it round-trips byte-exactly, and the `executionTime` move is correct on every path (assigned immediately after `run()` with no interleaved statement, and on the live-fallback branch, so it can never be read as NaN or stale).

Two changes I'd ask for before merge, both on the Snowflake accounting path and both one-liners -- details inline:

1. `lit()` escapes quotes only, but the string it builds is parsed by Snowflake, where backslash is an escape character. Compiled Malloy SQL routinely contains backslashes, so this breaks the literal boundary. The repo's dialect-aware `sqlLiteral()` already handles it.
2. `TO_TIMESTAMP_LTZ` is handed an ISO string ending in `Z`, which isn't in Snowflake's documented AUTO-detect format list. If it errors, every Snowflake build reports a null cost forever behind a caught exception.

A few smaller things, not blocking:

- **The timed-out lookup is abandoned into a session that is then hard-destroyed.** `withTimeout` leaves the query pending by design, and `dispose()` then calls `closeSync()` + `disconnectSync()`. Every pre-existing session operation in that file is fully awaited, so this is the first path that can reach `dispose()` with a query still in flight -- which is the quiescence premise the `disconnectSync()` comment's safety argument rests on. Threading a real `AbortSignal` through `runner` would cancel rather than orphan, and would also fix the deferred-cancellation point below.
- **The lookup runs inside the measured build.** `durationMs` is taken after `buildSourceIntoStorage` returns, so up to 10s of accounting latency lands in `source_build_duration` and in the reported `buildDurationMs`. Given the care taken elsewhere about not letting a zero average into that series, this looks unintended. Builds are serial, so the cost is additive per source, and a user cancel waits out the full timeout.
- **`maxResults := 50` is applied before `WHERE job_type = 'QUERY'`**, so non-QUERY jobs consume the candidate budget and `not_found` is likelier than the comment suggests. `stateFilter := 'DONE'` is available and unused.
- `recordCachedBuild` sits inside the try, so a throw discards an already-read cost and relabels it `error`; the postgres `recordBuildCostLookup` call in `materialization_build_session.ts` sits outside `finish()`'s guard and could fail an otherwise-successful build. Neither looks producible today, but that's the same standing as the guard already written in `finish()`.
- `model.ts:2743-2744` still says `executionTime` is "captured after prepare and before run," which this PR deliberately inverts.
- Since `runner` is fully stubbed, no generated SQL is parsed by any engine. The one escaping test is BigQuery-only, single-level, and quotes-only -- it reads as coverage of the escaping contract while exercising the one dialect where backslash isn't an escape. A backslash-bearing `opts.sql` on the Snowflake path would have caught the first item above.

On the histogram change: dropping the query-text label is a material cardinality win, not just a tidy-up. On one deployment serving real traffic that single label carried ~637 distinct values across ~14.9k series for this one histogram, with the row-count label adding more on top. The for-the-record framing in the release notes about p95 panels stepping up is the right call to surface loudly.

One ecosystem note worth a release-note line: adding fields to `QueryResult` and `ManifestEntry` is a breaking change for anyone generating strict OpenAPI clients -- openapi-generator's Java/Gson `validateJsonElement` throws on fields not present in the client's `openapiFields`, so a consumer running a newer server against an older generated client fails at deserialize on every affected response. Telling consumers to regenerate in the same change as the version bump would save some confusing breakage.

packages/server/src/service/build_cost.ts
housejester: Confirmed against a live account.

Taking the second option rather than the escaper: the lookup is removed from this PR. `sqlLiteral()` closes this hole, but the query-text key is what created it — an id-keyed lookup carries no author SQL in the accounting statement at all, and the window, candidate cap and ambiguity rule go with it.

Obtaining an id restructures how the build issues its read, and that restructure is already required to get query-tag metadata onto a `storage=` build's warehouse read at all — the build goes through DuckDB's query-passthrough, so no Malloy connector is in the call path to apply the bag. Same change, same reason. Cost reporting comes back as a follow-up tied to that work, keyed on the job id.

packages/server/src/service/build_cost.ts
housejester: Tested against a live account: bare `Z` parses, and resolves identically to an explicit `+00:00`. A no-zone string resolves an hour differently, so `Z` is honored as UTC rather than ignored. The full `INFORMATION_SCHEMA.QUERY_HISTORY(...)` call shape works.

Not a problem. Moot here since the statement is gone with the lookup, but it should save the id-keyed version from re-deriving it.

housejester:
**Backslash:** confirmed. The build-cost lookup is removed from this PR rather than reescaped. An id-keyed lookup removes author SQL from the accounting statement altogether, along with the window, the candidate cap and the ambiguity rule. Getting an id restructures how the build issues its read, so cost reporting belongs with that change.

**`TO_TIMESTAMP_LTZ`:** tested against a live account — bare `Z` parses and resolves identically to an explicit `+00:00`. Not a problem.

The removal also resolves four of the smaller points, which all lived in that module: the timed-out lookup abandoned into a hard-destroyed session; the lookup running inside the measured build (that one mattered most — `buildDurationMs` and `source_build_duration` are what a savings figure is computed against, and up to 10s of accounting latency was landing in both); `maxResults` applied before the `job_type` filter; and the two ungated counter calls.

The two that survive are fixed. The `model.ts` comment now describes the span `executionTime` actually measures. The OpenAPI note has its own section in the release notes — additive fields on `QueryResult` and `ManifestEntry` break strict Gson clients as a deserialize failure rather than an obvious version mismatch, so it says to regenerate in the same change as the version bump.

On the stubbed runner: no generated SQL was parsed by any engine, and the one escaping test was BigQuery-only, single-level and quotes-only — the dialect where backslash isn't an escape. A backslash-bearing `opts.sql` on the Snowflake path would have caught this.

Now `+511/−40` across 7 files, down from `+1347/−39` across 10. What remains is the serve path (`servedFrom`, `executionTimeMs`, `queryCostBytes`), `buildDurationMs`, the histogram label and span fixes, and the scanned-bytes counter. A `storage=` build reports a null cost, documented as a known gap rather than a property of the path.

sagarswamirao (approved):
Re-reviewed at `1caa27d`. Both blocking items are resolved, and removing the lookup rather than patching the escaper is the better call -- `sqlLiteral()` would have closed the hole, but the query-text key is what opened it, and an id-keyed lookup carries no author SQL in the accounting statement at all. Deferring it to the change that restructures how the build issues its read (which that work needs anyway for query-tag metadata) keeps the two from being solved twice.

Verified on the new head:

- `build_cost.ts` and its spec are gone, and the removal is clean -- no dangling imports, and the only surviving `buildCostBytes` is the colocated path's local reading `runStats`, which never went through the lookup. Both counters are removed from `materialization_metrics.ts` along with their recording functions.
- The build session returns immediately after the awaited `createTableAndDescribe`, so `dispose()` is quiescent again and the `disconnectSync()`-with-a-query-in-flight exposure is gone with it. The deferred-cancellation and inflated-`source_build_duration` points go the same way -- `durationMs` now brackets only the build.
- The spec matches the narrowed code: the `buildCost` object is fully removed, and `ManifestEntry.queryCostBytes` now states plainly that only a colocated build reports a figure, with the `storage=` gap described as a gap rather than a property. The release notes say the same thing in the same terms.
- `model.ts:2740-2743` now describes the post-`run()` span correctly.
- The serve-path tests are all still there (`servedFrom` on fallback and unrouted, `queryCostBytes` null-not-zero, the two histogram-label tests). CI is green at 15/15.

Thanks for chasing the `Z` question against a live account even though it became moot -- worth having settled, and the note that a no-zone string resolves an hour differently is the part that would have bitten the id-keyed version.

What lands here is the part that was solid all along: the histogram's labels are bounded, its span matches its description, and `servedFrom` makes a materialized answer distinguishable from a live one. The two items I raised as optional both made it in -- the cardinality figure and the client-regeneration note for consumers generating strict clients.

Issue #995, opened 2026-08-12
DuckDB extension provisioning: the offline fallback can mask a failed install, and the ADBC driver only reaches the Docker image


housejester:
Two follow-ups from #993, filed together because they share a cause: DuckDB extension dependencies are provisioned in the Dockerfile, and that provisioning is allowed to fail quietly.

#993 fixed the immediate symptom (the ADBC Snowflake driver was never installed, so every `snowflake_query()` failed at run time). These are the two things it deliberately did not touch.

## 1. The offline-build fallback can mask a real failure

```dockerfile
duckdb -c "INSTALL snowflake FROM community; LOAD snowflake; SELECT snowflake_version();" || \
echo "Snowflake verification skipped (offline build)" && \
```

Any failure in that chain — including a network blip against the extension repository — prints a reassuring line and the build continues, producing an image whose extensions were never installed.

This is not hypothetical. While writing #993 I put the driver fetch inside that same `RUN`, hit a transient `503`, and got a **successful build of an image with no driver** — the exact bug #993 exists to fix, reproduced by the fallback. That's why the driver step ended up in its own `RUN` with no `||`.

The same argument applies to the step it sits next to. Tolerating an offline build is a reasonable goal; the current shape can't distinguish "no network by design" from "the download failed", and reports success either way.

Worth deciding explicitly: either detect the offline case and skip, or let the step fail. Both beat a message that says "skipped" when something broke.

## 2. The driver only reaches the Docker image

#993 installs the ADBC driver in the Dockerfile, so `snowflake_query()` works in the published image. A local clone or `npx @malloy-publisher/server` still has no driver and fails identically — installing it there is a manual step (`dbc install snowflake`, or the extension's installer script).

The natural home is `packages/server/scripts/bake-duckdb-extensions.js`, which the Dockerfile itself describes as "a single bake mechanism (the server build) instead of re-running it here" — #993 makes that two. Doing it in the bake would fix Docker, local dev and CI at once, and would get co-location right for free, since the bake knows where it put each extension (the extension resolves its driver by `dladdr` on itself, so the driver has to sit beside the loaded object).

One real tension: that script is deliberately failure-tolerant — a failure is logged and skipped — which is the opposite of what the driver needs. It would want a non-tolerant branch, which is a design decision rather than a move.

Note the platform matrix is wider than the image's: the driver publishes `macos_arm64` and `windows_amd64` assets too, so local dev on either is in scope. The installer that was removed in #819/#820 did cover them.

[...]

PR #999, opened 2026-08-13
feat(server): tag a storage= build's warehouse read, and report what it cost

housejester:
A `storage=` build reads its source warehouse through DuckDB's native query-passthrough, so no Malloy connector is in the call path to render the query-metadata bag onto the statement. Every such build has therefore reached the warehouse **untagged** — the one kind of work a deployment cannot attribute, and invisible in `INFORMATION_SCHEMA.JOBS.labels` / `QUERY_HISTORY.query_tag`.

The tag machinery already exists; the build path routes around it. This closes that, and the same change makes the read's cost recoverable.

## What changed

The bag is resolved **before** the storage branch, through the same layering the colocated path uses (connection default < model < enforced < run context), so both paths attribute identically. It is then applied by the passthrough itself.

| | how it is tagged | how its cost is keyed |
|---|---|---|
| **Snowflake** | session `QUERY_TAG`; the read is unchanged | the tag, over the session's own history |
| **BigQuery** | `@@query_label`, which requires splitting the read | the job the split returns |
| **Postgres** | no per-statement tag exists | — |
| **chained `storage=`** | reads the parent's stored table; touches no warehouse | n/a — null means "did not spend" |

`ManifestEntry.queryCostBytes` is now populated for `storage=` builds, which #992 deliberately left null. The **whole** `BuildReadCost` — billed bytes, slot or execution time, the cache flag, and the warehouse's own job ids — goes onto the build's existing log line, since the manifest has room for one figure and those exist to answer questions it cannot.

## Why BigQuery's read had to be split

Not a workaround — the split *is* the labelling mechanism. `bigquery_query()` cannot carry a label by any route:

```
bigquery_query(..., labels|job_labels|query_label|query_labels|label := ...)
  → Binder Error: Invalid named parameter          (all five)
bigquery_query('SET @@query_label = "..."; SELECT ...')
  → Binder Error: Query response does not contain a result schema
```

So a labelled read runs as `bigquery_execute` over a two-statement script, and the anonymous result table that job wrote is then read with `bigquery_scan`.

**The split is not a second scan.** Reading the anonymous table goes through the Storage Read API and creates no new QUERY job — measured by diffing the project's QUERY job ids across the read, with the window capped well above the count so it could not saturate.

**Two gates, and the second one matters more.** There must be a label to carry, *and* the connection must be able to reach the job records the split depends on — asked once, with a one-row listing, **before anything is issued**. Past that point the read has been billed and there is no cheap way back: the rows live only in a table the build can no longer locate, so the alternative to failing would be paying for the read twice. A connection that cannot list gets the unlabelled read it had before, and loses attribution rather than its build.

## Cost, without matching query text

Both engines are keyed exactly, and **no compiled SQL is embedded in any statement sent to a warehouse**.

- **BigQuery** — the job record that locates the result table already carries the cost, so it comes back on that same call. A second lookup is not available anyway: a script's child job is absent from the default `bigquery_jobs()` listing and reachable only through its parent, so the child id alone resolves to nothing. The parent is carried for that reason.
- **Snowflake** — scoped to the build's own session and filtered by the tag, then the read is picked out of that set **in code** rather than in the statement, so the compiled SQL never becomes data inside a warehouse-parsed literal.

## Verified against live warehouses

Exercising the shipped functions, not a copy of the SQL they build.

| | |
|---|---|
| label lands on the BigQuery job | ✅ `{"cred_run":"run_abc123","cred_class":"ops","cred_org":"acme_corp"}` |
| value canonicalized to BQ's grammar | ✅ `Acme Corp` → `acme_corp` |
| cost returned with the read | ✅ **5,114,816 scanned / 10,485,760 billed**, 26ms slot, not cached |
| `bigquery_scan` adds a QUERY job | ❌ none |
| unlabelled BigQuery read | ✅ unchanged, no job id, nothing executed to produce it |
| BigQuery refuses a duplicate label key | ✅ `Invalid query label list 'team:a,team:b'. Duplicate…` |
| Snowflake tag stored and re-parsed | ✅ round-trips as valid JSON with `'` **and** `\` in one value |
| Snowflake read picked out of the session | ✅ from the read plus two driver probes |
| Snowflake cost, database-backed connection | ✅ cost returned |
| Snowflake cost, database-LESS connection | ✅ cost returned (qualified — see below) |

The scanned/billed pair is the 10MB-per-query floor in the wild: 2.05× on a single small read, and refreshes are mostly small reads.

## Notes for review

**Neither connector exports its metadata renderer.** `@malloydata/db-snowflake` exports `SnowflakeConnection` and `buildPoolOptions`; `@malloydata/db-bigquery` exports `BigQueryConnection`. `snowflakeQueryTag` lives in `dist/snowflake_executor` and `toBigQueryLabels` is module-private, and reaching into either package's `dist/` is not a dependency this repo takes anywhere else. Both grammars are reimplemented in `build_query_tag.ts` and pinned by tests that state the grammar. Exporting them upstream would let both copies go.

**Two Snowflake escaping faults, both found live, both fixed with the repo's dialect-aware `sqlLiteral`.** Neither was a syntax error; both failed silently.

- The tag is JSON, and a property value may contain a backslash (the contract admits printable ASCII except `"`). Quote-doubling alone delivered `\` where the JSON needed `\\`, storing a tag that **succeeds and then does not parse** in `QUERY_HISTORY` — precisely the thing tagging exists for.
- The same value inside the lookup's `QUERY_TAG =` comparison matched **zero rows**, reporting no cost rather than an error.

**These history table functions cap their output BEFORE the `WHERE` runs.** So a predicate can only filter rows the cap already admitted, and no key — however unique — reaches past it. Measured: with the read pushed beyond the default 100, the account-wide form matched zero rows while the same predicate at `RESULT_LIMIT => 10000` matched one. The lookup is `_BY_SESSION` **with an explicit limit**, and both halves are load-bearing — the session form carries the same default and was also empty in that measurement.

**`LAST_QUERY_ID()` is not usable here**, which is why the Snowflake key is the tag. The passthrough's session is not exclusively the build's: the ADBC driver issues its own `SELECT 1` connection probes around each call, and measured live, `LAST_QUERY_ID()` after a build-shaped read returned a probe. Keyed on it, every Snowflake build would have reported a probe's zero cost.

**A database-less Snowflake connection needed the name qualified.** `INFORMATION_SCHEMA` is per-database and an unqualified reference resolves against the session's *current* database, which such a connection does not have — measured, the unqualified form raises `002004 (42601)` there. It now uses the connection's own database when it has one, and the `SNOWFLAKE` shared database when it does not. The name is emitted **bare** rather than quoted, because quoting makes it case-exact while the connection parameter that put the session there resolves case-insensitively; a name that would need quotes takes the fallback instead of being interpolated raw.

**A legal bag could kill a BigQuery build.** Property names are validated case-*preserving*, so `Team` and `team` are two properties that both render to one label key, and BigQuery refuses a list containing a duplicate. Rendered through a map now, last wins — metered like the other drops this rendering makes.

**The Snowflake cache-hit flag over-reports** and is documented as a hint rather than a fact. Snowflake exposes no cache-hit column, so it is inferred from scanning nothing while producing rows — and a `GENERATOR` returning 50,000 rows is indistinguishable from a cached read. A build's read scans real tables, so the shape barely arises on this path.

**Verifying the Snowflake path locally needs the ADBC driver**, which only the Docker image carries (#995). On macOS the published `.dylib` must additionally be renamed to `libadbc_driver_snowflake.so` — the extension hardcodes that filename in its search even on Darwin, and reports the right directory with the wrong name.

**Tests:** run the way CI runs them — `test:unit` **2404 pass / 0 fail**, `test:integration --max-workers=1` **250 pass / 0 fail**, `typecheck` and `lint` clean across all packages. Note that reproducing CI's typecheck locally needs `generate-api-types` and `build:sdk` first; without them the generated `api.ts` and the SDK go stale and produce errors that are not in the tree.

### DIFF
--- a/packages/server/src/service/materialization_build_session.ts
+++ b/packages/server/src/service/materialization_build_session.ts
@@ -137,6 +137,46 @@ async function tagSnowflakeSession(
 }
 
 /**
+ * Can this connection reach the job records the split depends on?
+ *
+ * The split locates its result table by listing the executed script's child job,
+ * which is an API surface the direct `bigquery_query()` path never touches — so
+ * turning attribution on would otherwise add a permission requirement to builds
+ * that previously needed none, and discover it only AFTER a read had been billed.
+ * This asks the question with a one-row listing, before anything runs.
+ *
+ * <b>A false answer costs attribution, not the build.</b> That is the invariant
+ * this whole path is meant to keep, and it is the reason the check exists rather
+ * than a try/catch further down: once the read has run, falling back is no longer
+ * free — the rows live only in a table this cannot find, so the alternative to
+ * failing would be paying for the read a second time.
+ *
+ * Not cached: it is one listing per build against an API the build is about to
+ * use anyway, and caching it would have to key on the connection's credentials
+ * and survive their rotation.
+ */
+async function canSplitBigQueryRead(
+   session: DuckDBConnection,
+   handle: string,
+): Promise<boolean> {
+   try {
+      await session.runSQL(
+         `SELECT job_id FROM bigquery_jobs('${escapeSQL(handle)}', maxResults := 1)`,
+      );
+      return true;
+   } catch (error) {
+      logger.warn(
+         "Cannot list BigQuery jobs for this connection, so this build's " +
+            "warehouse read will not be labelled or costed",
+         {
+            error: error instanceof Error ? error.message : String(error),
+         },
+      );
+      return false;
+   }
+}
+
+/**
  * What the CTAS reads from, and the warehouse's own id for the read when the
  * shape of that read yielded one.
  *
@@ -199,7 +239,12 @@ export async function issuePassthroughRead(
       sourceType === "bigquery"
          ? bigQueryQueryLabelValue(queryMetadata)
          : undefined;
-   if (label === undefined) {
+   // Asked BEFORE anything runs, so an unusable split costs a probe rather than a
+   // build. Past this point the read has been issued and billed, and there is no
+   // longer a cheap way back: without the job record the result table cannot be
+   // located, so the rows cannot be captured at all and re-issuing them through
+   // the direct shape would bill the same read twice.
+   if (label === undefined || !(await canSplitBigQueryRead(session, handle))) {
       return {
          selectSQL: wrapPassthrough(sourceType, handle, buildSQL),
          jobId: null,
--- a/packages/server/src/service/materialization_build_session.ts
+++ b/packages/server/src/service/materialization_build_session.ts
@@ -14,6 +14,8 @@ import { logger } from "../logger";
 import { errMessage } from "../utils";
 import { quoteIdentifier, quoteManifestTablePath } from "./quoting";
 import { projectToPublicColumns } from "./build_plan";
+import { snowflakeSetQueryTagSQL } from "./build_query_tag";
+import type { QueryMetadata } from "./query_metadata";
 import {
    attachDuckLakeReadWrite,
    escapeSQL,
@@ -86,6 +88,39 @@ export function wrapPassthrough(
             `Unsupported passthrough source type: ${String(exhaustive)}`,
          );
       }
+   }
+}
+
+/**
+ * Tag every subsequent read on a Snowflake build session with this build's
+ * query metadata.
+ *
+ * Snowflake only, and by capability rather than preference: the passthrough
+ * reuses ONE session across `snowflake_query()` calls, so a session-level
+ * `QUERY_TAG` reaches the read that follows. BigQuery has no equivalent —
+ * `bigquery_query()` accepts no labels parameter and cannot run the script that
+ * would set one — so its labelling is part of how the read itself is issued.
+ * Postgres has no per-statement tag to set.
+ *
+ * <b>Best-effort.</b> A build that produces correct rows must not fail because
+ * the warehouse refused a tag; the cost is attribution for this one build, which
+ * the log line makes diagnosable.
+ */
+async function tagSnowflakeSession(
+   session: DuckDBConnection,
+   sourceType: FederatedSourceType,
+   handle: string,
+   queryMetadata: QueryMetadata | undefined,
+): Promise<void> {
+   if (sourceType !== "snowflake") return;
+   const sql = snowflakeSetQueryTagSQL(queryMetadata, handle);
+   if (sql === undefined) return;
+   try {
+      await session.runSQL(sql);
+   } catch (error) {
+      logger.warn("Could not tag the Snowflake build session", {
+         error: error instanceof Error ? error.message : String(error),
+      });
    }
 }
 
@@ -229,6 +264,17 @@ export async function buildSourceIntoStorage(params: {
    /** Logical, unquoted physical table path (may carry a container path). */
    physicalTableName: string;
    environmentPath: string;
+   /**
+    * Per-query metadata for the warehouse read, already resolved through the
+    * same layering the colocated path uses.
+    *
+    * It has to be applied HERE rather than by a Malloy connector, which is the
+    * whole reason this parameter exists: the read runs inside DuckDB's native
+    * query-passthrough, so no connector is in the call path to render the bag
+    * onto the statement. Without this, a `storage=` build is the one kind of
+    * warehouse work the deployment cannot attribute.
+    */
+   queryMetadata?: QueryMetadata;
 }): Promise<StorageBuildResult> {
    const {
       destinationName,
@@ -237,6 +283,7 @@ export async function buildSourceIntoStorage(params: {
       buildSQL,
       physicalTableName,
       environmentPath,
+      queryMetadata,
    } = params;
 
    assertSupportedDestination(destinationName, destinationConnection);
@@ -271,6 +318,13 @@ export async function buildSourceIntoStorage(params: {
          session,
          sourceType,
          sourceFederationConfig(sourceConnection),
+      );
+
+      await tagSnowflakeSession(
+         session,
+         sourceType,
+         federated.handle,
+         queryMetadata,
       );
 
       // The CTAS target MUST be qualified with the destination catalog (the
--- a/packages/server/src/service/materialization_build_session.ts
+++ b/packages/server/src/service/materialization_build_session.ts
@@ -14,7 +14,15 @@ import { logger } from "../logger";
 import { errMessage } from "../utils";
 import { quoteIdentifier, quoteManifestTablePath } from "./quoting";
 import { projectToPublicColumns } from "./build_plan";
-import { snowflakeSetQueryTagSQL } from "./build_query_tag";
+import {
+   bigQueryQueryLabelValue,
+   snowflakeSetQueryTagSQL,
+} from "./build_query_tag";
+import {
+   BIGQUERY_COST_COLUMNS,
+   bigQueryReadCost,
+   type BuildReadCost,
+} from "./build_read_cost";
 import type { QueryMetadata } from "./query_metadata";
 import {
    attachDuckLakeReadWrite,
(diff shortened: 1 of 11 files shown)


sagarswamirao:
The spec's `queryCostBytes` description now says the opposite of what this ships.

`api-doc.yaml` (lines 5185-5196) still reads:

> A plain `storage=` build reports null because its read goes through DuckDB's query-passthrough, where no Malloy connector is in the call path to report statistics — recovering the figure means reading it back from the warehouse's own accounting, which requires a job identifier the passthrough does not return.

That names as impossible the exact thing this PR implements, and names as the blocker ("a job identifier the passthrough does not return") the thing the split read was built to obtain. `materialization_service.ts:2480` now sets `queryCostBytes: result.readCost?.bytesScanned ?? null` for that path, and `api-doc.yaml` is untouched in the diff.

`ManifestEntry` is a generated public type, so this is the contract clients and codegen consumers read. Flagging early so it doesn't bite downstream: a reader following the spec will treat a real figure as impossible, or read a null as unconditional. No schema change needed — the type is already `["integer","null"]` and null stays valid — just the description. Could you update it to reflect that a plain `storage=` build now reports scanned bytes when the read carried a label, and null otherwise?

sagarswamirao:
Two things worth settling before this lands.

**The split's permission may not be the deployer's to grant.** `canSplitBigQueryRead` needs `bigquery.jobs.list`, which per BigQuery's access-control docs comes with Admin, Resource Admin, Resource Editor, Resource Viewer, or Studio Admin — notably *not* `roles/bigquery.jobUser`, the usual least-privilege role for a connection that only runs queries. That alone will silently disable the split for a lot of ordinary setups.

The sharper case: where a BigQuery connection is configured with a caller-supplied service-account key and billed to that account's own project, the identity running the read isn't the deployment's — so `jobs.list` has to be granted by whoever owns those credentials, and the labels land in their `INFORMATION_SCHEMA.JOBS` rather than anywhere the deployment can query. For that topology, attribution coverage could be near-zero rather than incidental, and the only signal is a single `logger.warn`.

The gate itself is clearly the right design — degrading attribution instead of the build is correct, and the docblock reasons it through well. The ask is smaller: document the required permission alongside the feature, and consider a counter for skipped splits, so an operator who turns tagging on and sees nothing has a way to find out why.

**Nothing pins scanned-vs-billed on the storage path.** `materialization_service.ts:2470-2474` argues at length that this field must carry scanned bytes and not billed — they diverge by up to the 10MB floor and the field gets summed across sources — but the only `queryCostBytes` test is on the query path (`model.spec.ts:1268`). Swapping `bytesScanned` for `bytesBilled` at `:2480` passes the entire suite. Since that's the actual behaviour change here, it'd be good to have one assertion holding it.

Related: the split is covered only against the hand-rolled `fakeSession` (`passthrough_read.spec.ts:6-16`), which returns exactly the shape the code expects. The live-project verification in the description is real work, but nothing re-runs it — if `bigquery_jobs()`'s columns or `bigquery_execute`'s return shape change, every test here still passes. An integration case over the new job-record surface would be worth it if feasible.

packages/server/src/service/materialization_build_session.ts
```diff
@@ -89,6 +103,274 @@ export function wrapPassthrough(
...
+   if (parentJobId === null) {
+      throw new Error(
+         "bigquery_execute returned no job_id for the build's read, so its " +
+            "result table cannot be located",
+      );
+   }
+
+   // The destination table is read from the job record rather than constructed:
+   // it is an anonymous table whose name BigQuery chooses.
+   const child = await session.runSQL(`
```
sagarswamirao: The child-job lookup isn't wrapped, so a throw here fails a build whose read has already been billed.

`:283-291` reasons carefully about exactly this wrong state for the empty-result case ("a retry is the operator's decision, not a silent double charge"). But the *throwing* sibling reaches the same place by a different door: a transient `jobs.list` 5xx, a quota trip, or a credential rotation between the probe and this call. There's no try/catch on this statement, `buildSourceIntoStorage` wraps it in `try`/`finally` (dispose only), and `materialization_service.ts:2375` rethrows — so the run fails with the read paid for, and any retry re-bills it.

That breaks the invariant `canSplitBigQueryRead`'s docblock states: "A false answer costs attribution, not the build." It holds before the read and is dropped after it. The Snowflake side (`:338-349`) is uniformly best-effort by comparison.

Could you wrap `:272-291` so a lookup failure degrades to no-cost the way Snowflake's does, keeping the deliberate throw only for the genuinely uncapturable case (no locatable table)? A distinguishable error type there would also let a retry policy tell "already billed" apart from "safe to retry".

Two smaller things in the same block, worth folding in:

- `rows.find()` at `:282` has no `ORDER BY`, so it depends on `SET @@query_label` never yielding a child with a `destinationTable`. BigQuery documents no ordering for `parentJobId` listings and doesn't say whether `SET` produces a child job — so this is currently resting on measured behaviour rather than a guarantee. An `ORDER BY creation_time DESC` or a `statement_type` filter would pin it, and the fixture at `passthrough_read.spec.ts:21` is single-element so multi-child selection is unexercised.
- `build_read_cost.ts:110`: `/^[A-Za-z0-9_$]+$/` admits a leading digit or `$`, which Snowflake rejects for an unquoted identifier (must start with a letter or underscore). Contained — the lookup throws and the cost goes null — but it lets through exactly the quote-needing name the docblock says takes the fallback. `/^[A-Za-z_][A-Za-z0-9_$]*$/` matches the documented intent.

sagarswamirao: The child-job lookup isn't wrapped, so a throw here fails a build whose read has already been billed.

`:283-291` reasons carefully about exactly this wrong state for the empty-result case ("a retry is the operator's decision, not a silent double charge"). But the *throwing* sibling reaches the same place by a different door: a transient `jobs.list` 5xx, a quota trip, or a credential rotation between the probe and this call. There's no try/catch on this statement, `buildSourceIntoStorage` wraps it in `try`/`finally` (dispose only), and `materialization_service.ts:2375` rethrows — so the run fails with the read paid for, and any retry re-bills it.

That breaks the invariant `canSplitBigQueryRead`'s docblock states: "A false answer costs attribution, not the build." It holds before the read and is dropped after it. The owflake side (`:338-349`) is uniformly best-effort by comparison.

Could you wrap `:272-291` so a lookup failure degrades to no-cost the way Snowflake's does, keeping the deliberate throw only for the genuinely uncapturable case (no locatable table)? A distinguishable error type there would also let a retry policy tell "already billed" apart from "safe to retry".

Two smaller things in the same block, worth folding in:

- `rows.find()` at `:282` has no `ORDER BY`, so it depends on `SET @@query_label` never yielding a child with a `destinationTable`. BigQuery documents no ordering for `parentJobId` listings and doesn't say whether `SET` produces a child job — so this is currently resting on measured behaviour rather than a guarantee. An `ORDER BY creation_time DESC` or a `statement_type` filter would pin it, and the fixture at `passthrough_read.spec.ts:21` is single-element so multi-child selection is unexercised.
- `build_read_cost.ts:110`: `/^[A-Za-z0-9_$]+$/` admits a leading digit or `$`, which Snowflake rejts for an unquoted identifier (must start with a letter or underscore). Contained — the lookup throws and the cost goes null — but it lets through exactly the quote-needing name the docblock says takes the fallback. `/^[A-Za-z_][A-Za-z0-9_$]*$/` matches the documented intent.

sagarswamirao:
Four documents now describe behaviour this PR changes, and three of them ship to users.

**1. The `QueryMetadata` schema still says Snowflake tags per statement.** `api-doc.yaml:3568-3569` reads "Each backend applies it through a per-query mechanism: Snowflake a per-statement `QUERY_TAG`". The storage path now sets it with `ALTER SESSION SET QUERY_TAG` (`materialization_build_session.ts:122-138`), which is session-scoped. This is observable rather than pedantic: your own docblock at `build_read_cost.ts:129-131` notes the ADBC driver issues `SELECT 1` probes around each call, and the description measured four tagged queries for one build. `pickSnowflakeReadRow` filters those out of the cost lookup, but in the customer's `QUERY_HISTORY` all four carry the build's bag, so grouping spend by `class` or `run_id` counts driver probes as materialization work.

**2. `docs/query-metadata.md` states the opposite as a design rule.** "Session-scoped mechanisms ... are deliberately not used: a pooled session serves many queries, so a session-level tag would attribute all of them to whichever one set it last." The isolated per-build DuckDB instance is a good answer to the cross-build half of that, and it would help to say so in the doc rather than leave the rule reading as violated. Two asks: carve out this path with the reasoning, and issue `ALTER SESSION UNSET QUERY_TAG` before dispose. The reset is cheap and it also closes the case where the tag statement fails best-effort (`:131-137`) and the read inherits a previous build's tag.

**3. `RELEASE_NOTES.md` `[Unreleased]` says this feature is deliberately absent.** "A plain `storage=` build also reports null, and that one is a gap rather than a property ... it is deliberately left to the change that does so rather than approximated here." This is that change, the section is untouched by the diff, and line 9 of the same file says every `[Unreleased]` section gets copied onto the GitHub release page. As it stands the release ships notes contradicting `api-doc.yaml` in the same release. Also worth a narrative line: tagged BigQuery reads now execute as a script and need `bigquery.jobs.list`, which is a new operational prerequisite.

**4. The `queryCostBytes` null list is missing two cases.** `api-doc.yaml:5190-5199` says a plain `storage=` build "reports one when its warehouse read carried query metadata, and null otherwise", then enumerates two nulls. Postgres is a supported passthrough source (`materialization_build_session.ts:56-60`) and always reports null, since the label is BigQuery-only (`:308-311`) and the Snowflake lookup returns early (`:410`) - your own test pins it (`passthrough_read.spec.ts:74`). And a BigQuery bag whose every key fails `sanitizeBigQueryKey` yields no label and so no cost (`build_query_tag.ts:94-96`). The `QueryResult.queryCostBytes` description in the same file already states the per-engine truth, so the asymmetry sits side by side.

Could you update all four? The first three are the ones that reach users.

sagarswamirao:
The already-billed read path is carefully reasoned, and three small things stop that reasoning from reaching anyone who could act on it.

**The type is erased at the only caller.** `BilledReadNotCapturedError` exists so "a caller with a retry policy should treat it as a decision for an operator rather than something to re-drive automatically" (`materialization_build_session.ts:181-194`). It is thrown at `:242`, `:334` and `:364`, and the sole production path out is `materialization_service.ts:2355-2378`, which catches everything and rethrows `new Error(...)` with no `cause`. Nothing outside the spec does `instanceof`. So no caller can distinguish "already billed, do not retry" from "attach failed, safe to retry", and `recordStorageBuildFailure` has no `reason` label, so both land in one series.

That matters more than it looks, because this repo is itself such a caller: `materialization_scheduler.ts:126-129` says "Advance regardless of outcome so a persistent failure retries on the next cron occurrence", and the fire passes `forceRefresh: true`, which defeats skip-if-unchanged. A persistent cause therefore re-bills the same read every occurrence, which is the outcome the type was created to prevent.

For anyone orchestrating publisher the situation is worse, since prose is the only discriminator available, and substring-matching a foreign error message is fragile in exactly the way this error is expensive. A structured reason code on the materialization response would let a caller classify it properly; short of that, preserving the type (or a `reason` label on the failure counter) at least keeps the distinction alive in-process.

**The retry guards the throwing shape, not the "not yet" shape.** `:349-360` wraps only `session.runSQL`. `resultRows` and `rows.find(r => str(r.dataset) && str(r.table))` run outside it, so a listing that returns 200 with the child not yet enumerated throws on the first look, with zero attempts. Your own rationale at `:210-214` names that class of transient. `bigquery_execute` returning a job id proves the script finished; it does not prove the `parentJobId` listing already includes the child. Moving the `find` inside the retried closure covers both on the same backoff.

**The terminal message drops the one key that could recover the cost.** `:242-246` carries the attempt count and `lastError`. Its sibling at `:364-368` embeds `job ${parentJobId}` - and `parentJobId` is documented as "the id an operator needs" precisely because a script's child job is unreachable from the default listing. In the one failure the code calls operator-owned, the operator is told a read was billed and given no way to find it. One line.

**Related, worth a look rather than a fix:** the split moves the billed read ahead of the CTAS bind. Previously the read was a string inlined into one `CREATE OR REPLACE TABLE ... AS (...)`; now `bigquery_execute` runs at `:329` and the CTAS at `:1071-1073`. A destination-side bind failure now follows a billed read, and `materialization_service.ts:2358-2362` cites "schema 'analytics' not found when a `name=schema.table` target's schema wasn't provisioned" as an observed failure. With the cron re-drive above, that is a recurring charge for a build that cannot succeed until the destination is fixed.

sagarswamirao:
The new counter closes the BigQuery half of "an operator turns tagging on and sees nothing". The Snowflake half is still open, and the spec now points operators at the counter for it.

**Five paths produce a null cost; one is counted.** `recordAttributionSkipped` has a single call site (`materialization_build_session.ts:176`) and one label value.

| path | signal |
|---|---|
| BigQuery job listing denied (`:176`) | counter + warn |
| Snowflake `ALTER SESSION` tag refused (`:131-137`) | warn only |
| Snowflake `INFORMATION_SCHEMA` unreadable (`:421-426`) | warn only |
| Snowflake row-pick found 0 or >1 (`:419-420`) | nothing at all |
| Postgres source (`:410`) | nothing |

`api-doc.yaml:5197-5199` tells the reader the null case "is counted by `publisher_storage_build_attribution_skipped_total`". On a Snowflake deployment it never is. The row-pick miss is the worst of the set: no log, no counter, and it is the most likely Snowflake failure, because it depends on `QUERY_TEXT` matching `buildSQL` byte for byte after a round trip through the driver. Widening the reason union (`tag_failed`, `read_row_not_found`, `read_row_ambiguous`, `cost_query_failed`) and emitting at `:135`, `:420` and `:425` would cover it.

**The probe collapses two different answers into one label, and retries neither.** `:159-179` catches every error - permission denial, 5xx, quota, socket reset - records one reason, and returns false. The asymmetry is the tell: the same `bigquery_jobs` API gets three attempts after the read, where retrying is expensive, and none before it, where it is free. Splitting the label lets an operator tell "grant `bigquery.jobs.list`" from "BigQuery was flaky at 03:00", and only the first is actionable.

While there: `bigquery.jobs.list` is a new prerequisite for BigQuery attribution and is documented only in the `canSplitBigQueryRead` docblock. Worth stating alongside the feature, together with which permission the split actually needs - `roles/bigquery.user` grants `jobs.list` but only over the caller's own jobs, so without `jobs.listAll` the probe succeeds and the listing returns nothing, which passes the gate and behaves like the failure.

**The self-exclusion filter is redundant and creates a false negative.** `build_read_cost.ts:186` excludes any row whose text contains `INFORMATION_SCHEMA.QUERY_HISTORY`, which also excludes a legitimate persist source that models Snowflake's query history - arguably the customer most interested in query-cost data - reporting null forever. It is not load-bearing: `pickSnowflakeReadRow` already requires exact equality with `buildSQL`, which the accounting query can never satisfy, and `EXECUTION_STATUS = 'SUCCESS'` excludes the in-flight statement. Dropping it removes the false negative at no cost.

**`cacheHit` over-reports on a more ordinary shape than the comment allows.** `:236-239` infers it from `scanned === 0 && produced > 0`. Snowflake answers metadata-resolvable aggregates (`COUNT(*)`, MIN/MAX on a clustered column) with `BYTES_SCANNED = 0` and rows produced, and the `projectToPublicColumns` wrapping does not defeat that resolution. An aggregate rollup is ordinary materialization, so for it the field inverts its documented meaning, "the source data had not moved and the rebuild did no useful work". The `GENERATOR` case the comment cites is the exotic one.

**Adjacent:** the four new `logger.warn` calls (`:134-137`, `:169-178`, `:228-235`, `:422-425`) log `error.message` verbatim, while the caller wrapping the same function routes it through `redactConnectionSecrets` (`materialization_service.ts:2364`) for the stated reason that a failed federation or passthrough error "can echo source- or catalog-connection detail (connstrings, account names, service-account JSON) from the DuckDB engine". These four sites sit in helpers that never receive the connection objects, so they cannot redact. `tagSnowflakeSession`'s catch fires precisely when the Snowflake session is unreachable, the error class most likely to echo account detail. Passing the connections (or a pre-bound redactor) down would line them up with the rest of the build path.

sagarswamirao:
The scanned-versus-billed assertion is here and it is a genuine discriminator: swapping `materialization_service.ts:2480` to `bytesBilled` makes `queryCostBytes` exactly `10485760` and fails `storage_build_bigquery.integration.spec.ts:202`. Two CI wiring details currently stop it from firing.

**The spec also runs under `test:unit`, where it passes having asserted nothing.** `packages/server/package.json:24` is `bun test --timeout 100000 src`, which globs all of `src/` including this file, and there is no `bunfig.toml` exclusion. `cross-platform-tests.yml:116` runs `test:unit` on three OSes with no BigQuery secret, and the credential gate here is an inline `return` from a passing test (`:114-117`, `:156-159`, `:207-210`) rather than a skip. So every PR reports three green tests that ran no assertions. `describe.skipIf(...)` would make a skip read as a skip (the repo already uses `it.skipIf` in `mcp_config.spec.ts:148`), and either excluding `*.integration.spec.ts` from `test:unit` or failing when `CI=true` and credentials are absent would close it properly.

**`materialization_service.ts` is not in the `paths:` filter.** The assertion pins a value produced at `materialization_service.ts:2480`, but that file is absent from both trigger lists (`connection-integration-tests.yml:6-13` and `:16-23`). So the exact edit the test defends against triggers no run of the job that contains it. Combined with the point above, the CI outcome for that regression is what it was before this PR.

Also worth adding, since the workflow comment says the job exists to catch it when "`bigquery_jobs()`'s columns move, or `bigquery_execute` stops returning `job_id`": those move on a dependency change, and neither `package.json` nor `bun.lock` is in `paths:`. The DuckDB `bigquery` extension is `INSTALL`ed at runtime (`connection.ts:873`), so its surface can drift with no repo change at all, and there is no `schedule:` trigger. A minimum set would be `materialization_service.ts`, `query_metadata.ts`, `introspection_sql.ts`, `package.json` and `bun.lock`.

**The Snowflake half has no end-to-end coverage at any layer.** No test drives `buildSourceIntoStorage` with a Snowflake source; everything Snowflake is tested as a pure string builder, and the CI step added here is BigQuery only. Each of these leaves the whole suite green: deleting the `tagSnowflakeSession` call, swapping its `sourceType`/`handle` arguments, passing the wrong connection field for `database` (`:417`), or inverting the `read.cost ?? await snowflakeReadCostAfterBuild(...)` precedence (`:702-711`). The live-account measurements that justify the design are one-time manual results recorded in comments with nothing re-checking them. I appreciate the driver is the blocker - even a spec that runs only when a Snowflake credential is present would pin the argument.

One hygiene item while in `build_read_cost.ts`: `num()` at `:258` is `typeof value === "bigint" ? Number(value) : Number(value)` - both branches identical. It reads as a precision guard and is not one.

sagarswamirao (approved):

PR #1002, opened 2026-08-13
fix(server): add columns to a publisher.db that predates them

### LINKED ISSUE #998: publisher.db has no column migration: a store predating the `manifest` column 500s on every materialization create
`packages/server/src/storage/duckdb/schema.ts` creates tables with `CREATE TABLE IF NOT EXISTS` and has no mechanism for adding a **column** to a table that already exists. So a `publisher.db` created before a column was introduced never gains it, and the first write that names that column fails at the binder.

There is exactly one such column today, and it breaks materialization outright.

## What happens

Against a `publisher.db` created before 2026-06-19, every `POST /materializations` returns 500:

```
Query execution failed: Binder Error: Table "materializations" does not have a
column with name "manifest"

Did you mean: "id"
Query: INSERT INTO materializations (id, environment_id, package_name, status,
       active_key, metadata, manifest, created_at, updated_at)
       VALUES (?, ?, ?, ?, ?, ?, NULL, ?, ?) RETURNING *
```

Materialization is unusable on that store. Nothing else is affected — the failure is confined to writes naming `manifest`.

## Why

`initializeSchema` is idempotent by design, and its comment states the intent:

> Always fall through to the `CREATE TABLE IF NOT EXISTS` pass below. The statements are idempotent and let an already-initialized DB pick up new **tables** added in later builds […] without forcing operators to run `--init` and lose their existing environments / packages / materializations.

That works for a new *table*. For an existing table, `CREATE TABLE IF NOT EXISTS` is a no-op regardless of how its columns differ, so a new *column* is silently skipped. `schema.ts:236` already acknowledges the general shape of this: *"this schema has no migration."*

The two existing accommodations for on-disk drift are both destructive drops rather than alters, and both are hand-written for one specific case:

- `dropLegacyProjectSchema` — drops pre-rename `projects` tables
- `dropPackageKeyedIncrementalLedger` — drops and re-keys `incremental_ledger`

There is no schema-version marker anywhere in `packages/server/src/storage/`, so nothing records which shape a store is on.

## Reproduce

1. Start a server on a checkout from before `1673e281` (2026-06-19, *"refactor(server): implement two-round materialization; remove manifest broker"* — the commit that added `manifest` to the `materializations` DDL) so `publisher.db` is created without the column.
2. Upgrade to any later build. `initializeSchema` runs and reports success.
3. `POST /api/v0/environments/{env}/packages/{pkg}/materializations` → 500, as above.

## Scope

Diffing a store created 2026-06-08 against one created by current `main`:

```
columns the current schema declares that the older store lacks: 1
   - materializations.manifest

columns only in the older store: 9
   + build_manifests.{id,build_id,environment_id,package_name,
                      source_name,connection_name,table_name,
                      created_at,updated_at}
```

The nine are the whole `build_manifests` table, dropped in the same commit that added `manifest`. It is orphaned rather than harmful — nothing queries it — but it is a second symptom of the same absence: removals are not reconciled either.

So the blast radius today is one column. The mechanism, though, applies to every future column.

## Why CI does not see it

CI starts from a clean checkout with no `publisher.db`, so the `CREATE TABLE` path always runs with the current DDL. The bug is reachable only on a store that predates a column — which is to say, only on an upgrade, which is the case CI never exercises. Local development hits it for the same reason (a long-lived working copy), and it presents as a confusing 500 rather than as anything pointing at the store.

## Possible directions

Both are design calls rather than mechanical fixes, and the second subsumes the first.

**1. One more targeted helper**, matching the existing pattern: drop and recreate `materializations` on an already-initialized store, loudly, accepting the loss of materialization history. Consistent with the two helpers already there, and small. It does nothing for the next column.

**2. Reconcile columns generally.** Read `duckdb_columns()` for each declared table and `ALTER TABLE … ADD COLUMN` for anything declared but absent. Non-destructive, needs no version marker (the disk is the state), and it retires this class of bug rather than this instance. Bounded to additive changes: a type change or a rename still needs a hand-written step, and a dropped column would want an explicit decision rather than an automatic drop.

Worth noting whichever way it goes: `initializeSchema` currently reports success on a store it has left unusable. Even without a migration, failing or warning when the disk shape does not match the declared shape would turn a binder error at first write into a diagnosable message at boot.

housejester:
Fixes #998.

`CREATE TABLE IF NOT EXISTS` is idempotent, so an existing `publisher.db` picks up a new **table** added by a later build. It never picks up a new **column**: that same statement is a no-op against a table that already exists, however its columns differ. A store created before a column was introduced therefore never gains it, `initializeSchema` reports success anyway, and the first write naming that column fails at the binder.

There is one such column today, and it breaks materialization outright: a `publisher.db` created before 2026-06-19 returns 500 on every `POST .../materializations` with `Binder Error: Table "materializations" does not have a column with name "manifest"`.

## What this does

Schema init now reconciles columns after the `CREATE TABLE` pass. The expected shape is **not written down a second time** — it is read back out of a throwaway in-memory DuckDB that the same `createDeclaredTables()` has just been run against, then diffed against `duckdb_columns()` on the real store. The `CREATE TABLE` statements stay the single declaration of the schema, so a column added to the DDL but forgotten in a parallel list cannot reproduce this bug. The disk is the other half of the comparison, so no schema-version marker is needed and none is introduced.

It runs on any already-initialized store — nothing is on disk to have drifted before this build created it. **`--init` is deliberately not exempt**, though it looks like it should be: `dropAllTables` recreates everything, but its table list is a second, hand-maintained declaration of the table set, which is exactly the kind of parallel list this reconcile exists to avoid depending on. A table added to the DDL and forgotten there would survive the drop, no-op through `CREATE TABLE IF NOT EXISTS`, and keep its drift through the one command the warning tells operators to run. On a correctly recreated store the reconcile finds nothing, so not exempting it costs one in-memory DDL run.

The `isInitialized()` gate is airtight rather than merely cheap — `environments` is the first table created, so even a store from an interrupted first init trips it and gets reconciled.

**Strictly additive, and only for columns carrying no constraint.** That is a boundary DuckDB enforces, not one this code picked: `ALTER TABLE … ADD COLUMN` rejects a column carrying any constraint — `NOT NULL`, `PRIMARY KEY`, `UNIQUE`, `CHECK`, `FOREIGN KEY` — with or without a `DEFAULT` (`Parser Error: Adding columns with constraints not yet supported`). A bare `DEFAULT` **is** accepted and backfills existing rows, so a declared default is carried across; a constraint never can be.

Constraints come from `duckdb_constraints()` on the mirror rather than from `is_nullable`, which matters more than it sounds: a `UNIQUE` or `CHECK` column reports as *nullable*. Screening on nullability alone would add it to an upgraded store as a bare column — right name, wrong rules — so two servers on the same build would enforce differently depending on how their store was created. That is the same class of bug this PR exists to kill, moved from "column absent" to "column wrong". Such a column is refused and named in the warning instead.

Everything else the diff finds is reported rather than acted on:

- A missing constrained column, or a column already present whose type, nullability, default **or constraints** have changed, warns at boot naming the column. The constraint comparison closes the other half of the divergence above: a `UNIQUE` or `CHECK` added to an existing table's DDL is as invisible to `CREATE TABLE IF NOT EXISTS` as a column is, and it fails more quietly — the older store simply keeps accepting rows a fresh store rejects. It has already happened once here (`incremental_ledger`'s primary key was re-keyed, needing the hand-written `dropPackageKeyedIncrementalLedger` to catch it), which is also where the doc comment now points anyone who has to write such a step. This is what turns a binder error on some later request into a diagnosable line at startup, and it is the part that keeps earning its keep after this particular column is behind us.
- A column or table on disk that this build no longer declares is **left in place** and logged at debug.

## Why nothing is ever dropped

A reconciler symmetrical in both directions would delete data during a routine upgrade. Two relics prove it is not hypothetical:

- **`build_manifests`**, a whole table, added in #666 (2026-04-20) and removed by `1673e281` (2026-06-19) — the same commit that added `manifest`. Every store predating that commit both lacks the column and still carries the table.
- **`materializations.build_plan`**, added by `1673e281` and removed four days later by #844 (`fa820889`, 2026-06-23). Stores created in that window carry a column current code never names.

Both are inert — nothing queries `build_manifests`, and every INSERT names its columns explicitly — and both are cleared by `--init`, which already lists `build_manifests` in `dropAllTables`. Removing them is an operator's decision, not something an upgrade should do quietly. They are the fixture in the "leaves relics in place" test, so the boundary is pinned rather than merely intended.

## Cost on a large store

`ADD COLUMN` of a column **without a default** is a catalog operation, not a data rewrite. Against pre-`manifest` `materializations` tables of realistic width:

| rows | file before | ALTER | + CHECKPOINT | file after |
|---|---|---|---|---|
| 100k | 4.8 MB | 1.1 ms | 2.7 ms | — |
| 1M | 41.5 MB | 4.5 ms | 5.2 ms | 41.8 MB |
| 5M | 204.8 MB | 17.6 ms | 21.5 ms | 205.0 MB |

A 205MB table gained the column in 17.6ms and the file grew 0.1%; row counts were preserved and every existing row read back NULL. It scales with row *count* (~3.5µs/1000 rows, per-row-group metadata), not bytes.

**A column carrying a `DEFAULT` is the exception**, because backfilling existing rows is a real write. Same 5M-row/205MB table: 80.7ms for the ALTER (vs 19.7ms without) and 43.8ms to checkpoint (vs 19.6ms), growing the file 0.8MB instead of 0.2MB. Still a tenth of a second, still before the server accepts traffic — but it is not the free catalog operation the row above describes, and it scales with rows. Only a table actually missing a declared column is touched — `entity_embeddings`, likely the largest by bytes on a real store, is not. Steady state after the one-time fix is the catalog diff alone: 3.7–4.9ms on that 205MB file, flat in row count. All of it is at boot, inside `StorageManager.initialize()`, so it delays readiness rather than blocking queries.

## Why CI never saw it

CI starts from a clean checkout with no `publisher.db`, so the create path always runs with the current DDL and the drift cannot arise. The gap was not a missing assertion — no test had ever booted against a store older than the build. `tests/unit/duckdb/schema_column_reconcile.test.ts` is the first, seeding the pre-`1673e281` shape verbatim so it keeps describing the store that shipped even as the DDL moves on. It covers: the column is added and `MaterializationRepository.create()` (the exact failing call) succeeds, with the upgraded table then compared to a freshly created one across type, nullability, default **and table constraints**, so a future column the ALTER cannot faithfully express turns the test red instead of shipping a divergent store; a declared `NOT NULL` column is refused rather than approximated, while an unrelated addable column still lands in the same boot; existing rows survive with the column NULL; the two relics are left alone; a second boot is a no-op; and a store this build creates already matches the mirror exactly, which is what makes the reconcile a no-op on a fresh store rather than a silent source of ALTERs.

## Verification

Reproduced the failure first, then confirmed the fix on a real server boot rather than only in unit tests — seeded a pre-`manifest` `publisher.db`, started the server, created an environment and package, and issued the call from the issue:

```
info: publisher.db: added 1 column(s) this build declares that the store predates: materializations.manifest JSON
debug: publisher.db holds 1 column(s)/table(s) this build no longer declares; left in place: build_manifests (whole table)
...
POST .../packages/storefront/materializations -> HTTP 201
```

119 tests pass across the duckdb and storage suites; lint, prettier and `tsc` clean.

CI is green on every check, including **Cross Platform Tests on all three platforms**. Windows was the one I considered a genuine risk: these tests open, write and reopen real store files, and `legacy_schema_migration.test.ts` carries a comment about Windows runners failing on a close-and-reopen of the same DuckDB file. Seeding and asserting on a single open connection, the pattern that file established, holds there.

## Two judgment calls worth a reviewer's attention

1. **A missing constrained column warns rather than being approximated by an unconstrained one.** Adding a bare stand-in would make writes succeed but leave the store permanently disagreeing with its own declaration, and the warning would then fire on every boot forever and get tuned out. There are zero such columns today, so this is currently theoretical.
2. **Relics log at `debug`, not `info`.** They are permanent on affected stores, so `info` would be noise on every boot indefinitely. The default `LOG_LEVEL` is `debug`, so they are visible out of the box.

The decision logic is split into a pure `planColumnReconcile()` and unit-tested against synthetic shapes, because today's declared schema has no defaulted, `UNIQUE` or `CHECK` column — so no test driving the real DDL could reach the branches a future column will land in. Those cover the carried `DEFAULT`, the refused `UNIQUE`, constraint drift on a column both sides already have, and that `NOT NULL` is reported once (by nullability) rather than twice.

## Out of scope

Dropping `build_manifests` / `build_plan`, and any non-additive migration machinery. A versioned migration runner was considered and deferred: every store already in the wild has no version marker, so such a scheme would need to infer a baseline from the disk shape first — which is this diff. The boot warning is the trigger for building one, the first time a change lands that this cannot handle.

### DIFF
--- a/packages/server/src/storage/duckdb/schema.ts
+++ b/packages/server/src/storage/duckdb/schema.ts
@@ -328,8 +328,16 @@ function columnKey(column: ColumnShape): string {
  *   stand-in would leave the store permanently disagreeing with its own
  *   declaration, so it warns and leaves the write to fail, which is at least
  *   traceable to a named column at boot.
- * - A column already present whose type, nullability or default has changed is
- *   not an ALTER we can infer the intent of.
+ * - A column already present whose type, nullability, default or constraints
+ *   have changed is not an ALTER we can infer the intent of.
+ *
+ *   Where the hand-written step goes when one is needed:
+ *   `dropPackageKeyedIncrementalLedger` below is exactly that shape — a
+ *   targeted pre-pass, keyed off what it finds on disk rather than a version
+ *   marker, running before the CREATE pass so the tables are recreated
+ *   correctly. Copy it rather than rediscovering the pattern; the warning's
+ *   other suggestion, `--init`, is destructive and takes environments,
+ *   packages and materializations with it.
  * - A column or table on disk that this build no longer declares is left alone.
  *   Dropping is a decision, not a reconciliation, and the relics are inert:
  *   `materializations.build_plan` (added and removed within four days in June
@@ -349,12 +357,14 @@ async function reconcileDeclaredColumns(db: DuckDBConnection): Promise<void> {
 
       const declared = await mirror.all<ColumnShape>(COLUMN_SHAPE_QUERY);
       const onDisk = await db.all<ColumnShape>(COLUMN_SHAPE_QUERY);
-      const constraints = await mirror.all<ConstraintRow>(CONSTRAINT_QUERY);
 
       const plan = planColumnReconcile(
          declared,
          onDisk,
-         constrainedColumnsOf(constraints),
+         constrainedColumnsOf(
+            await mirror.all<ConstraintRow>(CONSTRAINT_QUERY),
+         ),
+         constrainedColumnsOf(await db.all<ConstraintRow>(CONSTRAINT_QUERY)),
       );
 
       const added: string[] = [];
@@ -366,7 +376,12 @@ async function reconcileDeclaredColumns(db: DuckDBConnection): Promise<void> {
             await db.run(add.sql);
             added.push(add.description);
          } catch (err) {
-            const message = err instanceof Error ? err.message : String(err);
+            // First line only: DuckDBConnection.run appends the whole failing
+            // statement after a newline, which would turn a scannable boot
+            // warning into a multi-line SQL dump.
+            const message = (
+               err instanceof Error ? err.message : String(err)
+            ).split("\n")[0];
             failed.push(`${add.description} (${message})`);
          }
       }
@@ -405,6 +420,15 @@ async function reconcileDeclaredColumns(db: DuckDBConnection): Promise<void> {
    }
 }
 
+/**
+ * The constraint kinds worth comparing between two stores, sorted so the
+ * comparison is order-insensitive. NOT NULL is dropped because `is_nullable`
+ * already carries it, and reporting both would name one difference twice.
+ */
+function enforcedKinds(kinds: string[] | undefined): string[] {
+   return (kinds ?? []).filter((kind) => kind !== "NOT NULL").sort();
+}
+
 /** Every column named by any constraint, mapped to the constraint kinds naming it. */
 export function constrainedColumnsOf(
    constraints: ConstraintRow[],
@@ -438,6 +462,7 @@ export function planColumnReconcile(
    declared: ColumnShape[],
    onDisk: ColumnShape[],
    constrainedColumns: Map<string, string[]>,
+   diskConstrainedColumns: Map<string, string[]>,
 ): {
    adds: Array<{ sql: string; description: string }>;
    needsMigration: string[];
@@ -497,6 +522,20 @@ export function planColumnReconcile(
       ) {
          differences.push(
             `default on disk ${existing.column_default ?? "none"}, declared ${column.column_default ?? "none"}`,
+         );
+      }
+      // A constraint added to a table that already exists is as invisible to
+      // CREATE TABLE IF NOT EXISTS as a column is, and the consequence is worse
+      // than a missing column: the store keeps accepting rows a fresh store
+      // rejects, so two servers on one build enforce different rules with
+      // nothing failing. It has happened here once already — `incremental_ledger`
+      // was re-keyed, and needed the hand-written detector below to catch it.
+      // NOT NULL is excluded because `is_nullable` above says it better.
+      const declaredKinds = enforcedKinds(constrainedColumns.get(key));
+      const diskKinds = enforcedKinds(diskConstrainedColumns.get(key));
+      if (declaredKinds.join(",") !== diskKinds.join(",")) {
+         differences.push(
+            `constraints on disk ${diskKinds.join(" + ") || "none"}, declared ${declaredKinds.join(" + ") || "none"}`,
          );
       }
       if (differences.length > 0) {
--- a/packages/server/src/storage/duckdb/schema.ts
+++ b/packages/server/src/storage/duckdb/schema.ts
@@ -39,9 +39,17 @@ export async function initializeSchema(
    // A CREATE TABLE IF NOT EXISTS is a no-op against an existing table no matter
    // how its COLUMNS differ, so the pass above carries a new table but never a
    // new column. Reconcile those separately, and only where the drift can exist:
-   // a store created by this build (or re-created by --init just above) already
-   // matches the declaration by construction.
-   if (initialized && !force) {
+   // nothing is on disk to have drifted before this build created it.
+   //
+   // `--init` is NOT exempt. It looks like it should be, since dropAllTables
+   // recreates everything from scratch — but that function's table list is a
+   // second, hand-maintained declaration of the table set, exactly the kind of
+   // parallel list this reconcile exists to avoid depending on. A table added to
+   // the DDL and forgotten there survives the drop, no-ops through CREATE TABLE
+   // IF NOT EXISTS, and would keep its drift through the one command the warning
+   // below tells operators to run. On a correctly recreated store the reconcile
+   // finds nothing, so the cost of not exempting it is one in-memory DDL run.
+   if (initialized) {
       await reconcileDeclaredColumns(db);
    }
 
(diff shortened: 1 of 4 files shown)


packages/server/src/storage/duckdb/schema.ts
```diff
@@ -229,6 +263,302 @@ export async function initializeSchema(
...
+            needsMigration.push(
+               `${key} (declared ${kinds.join(" + ")}; DuckDB cannot add a constrained column)`,
+            );
+            continue;
+         }
+         const withDefault = column.column_default
+            ? ` DEFAULT ${column.column_default}`
+            : "";
+         adds.push({
+            sql: `ALTER TABLE "${column.table_name}" ADD COLUMN IF NOT EXISTS "${column.column_name}" ${column.data_type}${withDefault}`,
```
Sha-Bang: Thought / question only — not blocking.

We pin `@duckdb/node-api@1.5.3-r.2`. On DuckDB 1.5.0–1.5.3, `ADD COLUMN IF NOT EXISTS … DEFAULT` against a column that **already exists** can still run the default backfill and overwrite every row ([duckdb#23209](https://github.com/duckdb/duckdb/issues/23209)). Without `DEFAULT`, `IF NOT EXISTS` is fine.

Today’s DDL has no defaulted columns, and the planner only emits this when the catalog says the column is missing, so the current `manifest` path is safe. Worth considering for the next `DEFAULT` column though: drop `IF NOT EXISTS` when a default is present (the plan already checked absence), so a catalog/ALTER disagreement errors into the catch instead of silently wiping rows?

Sha-Bang (approved):
Looks good — mirror-DDL reconcile is the right shape for this class of upgrade bug, and the judgment calls (refuse constrained columns; relics at debug) both seem right to me.

One non-blocking thought on the `IF NOT EXISTS` + `DEFAULT` path left inline.

packages/server/src/storage/duckdb/schema.ts
housejester: Good catch — it reproduces on what we pin (`1.5.3-r.2`, engine `v1.5.3`), with the upstream repro shape: the column has to have been *added* via `ADD COLUMN … DEFAULT` first, then updated. `[true,true]` → `[false,false]` on the re-run.

It's the fixed-width types:

| type | re-run |
|---|---|
| BOOLEAN, TIMESTAMP | **overwritten** |
| VARCHAR, INTEGER, JSON | no-op |

`TIMESTAMP` being in there is what decided it — that's most of this schema's non-VARCHAR columns.

Went a bit further than suggested and dropped `IF NOT EXISTS` unconditionally rather than only on the `DEFAULT` path: a version-conditional would go stale the moment the engine moves, and the plan's catalog read is what establishes absence anyway, so the clause was never the guard. Without it, a plan/catalog disagreement raises into the per-column catch and gets reported like everything else this path can't do. Pinned by a test, since re-adding it looks like a harmless safety improvement.

No exposure today (no defaulted columns, and `manifest` is JSON), so this is purely for the next one. a0b6b356.

— Jaunty Shrike

Sha-Bang: Aight Jaunty

Sha-Bang (approved):
Re-reviewed a0b6b356. Dropping `IF NOT EXISTS` unconditionally (with the engine bug pinned in a test) is the right call — better than a DEFAULT-only special case that would go stale. Still LGTM.

PR #1003, opened 2026-08-13
fix(server): make the storage= build's coverage real, and its expensive failure distinguishable

housejester:
Follow-ups from review of #999, which merged on auto-merge before these landed. Everything here is a review finding; nothing new is introduced.

## The coverage #999 added was largely notional

**The integration spec passed in CI having asserted nothing.** Its credential gate was an early `return` from a *passing* test, and the file lives under `src/`, which `test:unit` (`bun test src`) globs on every PR and on all three platforms — none of which carry a BigQuery secret. So it reported three green tests that ran no assertions. That is worse than no coverage, because it reads as coverage.

`describe.skipIf` now makes a skip report as a skip:

```
before   3 pass   0 fail      (asserted nothing)
after    0 pass   3 skip      (with credentials: 3 pass / 15 expect() calls)
```

**And the file it defends was not a trigger.** The spec's main assertion pins a value produced at `materialization_service.ts:2480`, and that file was absent from the workflow's `paths:` filter — so the exact edit it guards ran no job containing it. Added, together with `query_metadata.ts`, `introspection_sql.ts`, `package.json` and `bun.lock`: the passthrough surfaces this job exists to catch are provided by DuckDB extensions installed at runtime, so they move with a dependency bump rather than with any source file here.

## The expensive failure was indistinguishable from a cheap one

`BilledReadNotCapturedError` exists so a caller can tell "the read was billed, do not re-drive" from "the attach failed, retry freely". It was erased at its only caller, which caught everything and rethrew a bare `Error` with no `cause`.

**That matters because this service is itself such a caller.** `materialization_scheduler.ts` advances "regardless of outcome so a persistent failure retries on the next cron occurrence" and fires with `forceRefresh: true`, which defeats skip-if-unchanged. A persistent cause therefore re-runs the same read every occurrence — and on this branch, pays for it every occurrence. Exactly the outcome the type was created to prevent.

The type now survives the rethrow, and `publisher_storage_build_failures_total` carries `reason` (`build_failed` | `billed_read_not_captured`) so the two are separable on a dashboard rather than only in prose.

**The retry guarded the wrong shape.** It wrapped `session.runSQL` but not the row-pick, so a listing that answered 200 without the child job enumerated yet failed on the first look with zero attempts — the same transient class the retry exists for, and one `bigquery_execute` returning a job id does not rule out. The row-pick moved inside the retried closure.

**The terminal error now names the parent job.** It is the only handle on a read that was paid for and not captured, since a script's child job is absent from the default listing. Telling an operator a read was billed without saying which one leaves them nothing to act on.

## Attribution gaps beyond the BigQuery one are now counted

`publisher_storage_build_attribution_skipped_total` had one call site and one reason, while five paths produce a null cost. The spec meanwhile told operators the null case "is counted by" it — on a Snowflake deployment it never was.

Now emitted for a refused Snowflake tag, a cost query that fails, and a row-pick that finds nothing (`read_row_not_found`) or too much (`read_row_ambiguous`). The last is the likeliest Snowflake miss — it turns on `QUERY_TEXT` matching the build's SQL byte for byte after a round trip through the driver — and it was previously silent in every channel.

## Two smaller corrections

**Dropped the accounting query's self-exclusion filter.** `QUERY_TEXT NOT LIKE '%INFORMATION_SCHEMA.QUERY_HISTORY%'` was redundant with the exact-equality `pickSnowflakeReadRow` already requires — which that statement can never satisfy — and it blanked out a persist source that models Snowflake's own query history, arguably the customer most interested in query-cost data. The test that asserted the filter now asserts the property that made it unnecessary, plus the false negative it used to create.

**The Snowflake session tag is cleared before release.** The build session is private and disposed, so this is not about leaking to a later build; it covers the case where tagging fails best-effort on a session the driver had already tagged, which would leave the read attributed to whatever set it last. Verified live: the tag reads `""` after.

**`num()`** had `typeof value === "bigint" ? Number(value) : Number(value)` — both branches identical, reading as a precision guard while being none.

## Four documents described behaviour that changed

Three of them ship to users.

- **`QueryMetadata` schema** said Snowflake applies the bag as a per-statement `QUERY_TAG`. On the `storage=` path it is a session tag, and the driver's own connection probes therefore carry the build's bag too.
- **`docs/query-metadata.md`** states as a design rule that session-scoped mechanisms "are deliberately not used". That rule is right; this path is exempt from its reasoning rather than in violation of it, because the session is private per build rather than pooled. Written up as a carve-out, including the residue: probes scan nothing, so cost grouped by `class` or `run_id` is unaffected, but a statement COUNT reads high.
- **`RELEASE_NOTES.md`** still said a plain `storage=` build reports null and that recovering the figure "is deliberately left to the change that does so". This is that change, and line 9 of the same file says every `[Unreleased]` section is copied onto the release page — so the release would have shipped notes contradicting `api-doc.yaml`. Replaced with a section describing what shipped, including the new BigQuery prerequisite.
- **The `queryCostBytes` null list** omitted two cases: a Postgres source, which reports null on every path because the label that makes a read reportable is BigQuery's, and a bag whose every key is unusable as a BigQuery label.

## Verification

Unit **2412 pass / 3 skip / 0 fail**; integration **250 / 0**; the live BigQuery spec **3 pass / 15 expect()** with credentials, `3 skip` without. `typecheck` and `lint` clean across all packages.

Re-checked live on Snowflake after dropping the filter and adding the reset: three candidates under the tag, the read still picked correctly, and the tag cleared after `UNSET`.

## Not included

Two review items are deliberately out of scope here, and one did not reproduce.

- **Redaction of the four new `logger.warn` sites.** They log `error.message` verbatim while the caller wrapping the same function routes it through `redactConnectionSecrets`. Correct as an observation — these helpers never receive the connection objects, so they *cannot* redact — but fixing it means threading connections (or a pre-bound redactor) through the build session, which is a wider change than the rest of this.
- **Snowflake end-to-end coverage.** No secrets are provisioned for it, so it would skip everywhere including CI. Worth doing when they are.
- **`cacheHit` over-reporting on metadata-resolvable aggregates** — measured against a live account and it does not reproduce. `COUNT(*)` and `MIN`/`MAX` report `BYTES_SCANNED = 0` **and `ROWS_PRODUCED = 0`** in `QUERY_HISTORY` (the client still receives its row), so `scanned === 0 && produced > 0` never fires. A `GROUP BY` rollup — the "ordinary materialization" shape — scans normally (119,808 bytes measured) and is not metadata-resolvable at all. The existing `GENERATOR` caveat stands as the accurate one.

### DIFF
--- a/.github/workflows/connection-integration-tests.yml
+++ b/.github/workflows/connection-integration-tests.yml
@@ -10,6 +10,17 @@ on:
       - "packages/server/src/service/build_read_cost.ts"
       - "packages/server/src/service/build_query_tag.ts"
       - "packages/server/src/service/storage_build_bigquery.integration.spec.ts"
+      # Produces the value the manifest assertion pins.
+      - "packages/server/src/service/materialization_service.ts"
+      # The metadata bag's own contract, and the dialect-aware escaper the
+      # Snowflake lookup builds its literal with.
+      - "packages/server/src/service/query_metadata.ts"
+      - "packages/server/src/service/introspection_sql.ts"
+      # The passthrough surfaces this job exists to catch are provided by the
+      # DuckDB extensions, which move with a dependency bump rather than with any
+      # file above.
+      - "packages/server/package.json"
+      - "bun.lock"
       - ".github/workflows/connection-integration-tests.yml"
   pull_request:
     branches: [main]
@@ -20,6 +31,17 @@ on:
       - "packages/server/src/service/build_read_cost.ts"
       - "packages/server/src/service/build_query_tag.ts"
       - "packages/server/src/service/storage_build_bigquery.integration.spec.ts"
+      # Produces the value the manifest assertion pins.
+      - "packages/server/src/service/materialization_service.ts"
+      # The metadata bag's own contract, and the dialect-aware escaper the
+      # Snowflake lookup builds its literal with.
+      - "packages/server/src/service/query_metadata.ts"
+      - "packages/server/src/service/introspection_sql.ts"
+      # The passthrough surfaces this job exists to catch are provided by the
+      # DuckDB extensions, which move with a dependency bump rather than with any
+      # file above.
+      - "packages/server/package.json"
+      - "bun.lock"
       - ".github/workflows/connection-integration-tests.yml"
   workflow_dispatch:
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -18,6 +18,18 @@ One behaviour change to know about: `skills-npm.yml` now publishes only from `ma
 
 ---
 
+## [Unreleased] — a `storage=` build's warehouse read is now attributable
+
+A `storage=` build reads its source through DuckDB's native query-passthrough, where no Malloy connector is in the call path to apply the query-metadata bag. Every such build therefore reached the warehouse untagged — the one kind of work a deployment could not attribute. It now carries the same bag the colocated path does, resolved through the same layering.
+
+**Snowflake** takes it as a session `QUERY_TAG`; its read is unchanged. **BigQuery** takes it as `@@query_label`, which it cannot do without splitting the read: `bigquery_query()` accepts no labels parameter and cannot run the script that would set one. So a labelled BigQuery read runs as `bigquery_execute` over a two-statement script, and the anonymous result table that job wrote is then read with `bigquery_scan`. Reading that table goes through the Storage Read API and creates no new query job, so the split is not a second scan. **Postgres** has no per-statement tag and is unaffected.
+
+**New operational prerequisite on BigQuery.** A labelled read locates its result table by listing the executed script's child job, an API surface the unsplit read never touched. A connection that cannot list its own jobs keeps the read it had before and loses attribution rather than its build — the fallback is decided by a probe issued _before_ anything runs, and counted by `publisher_storage_build_attribution_skipped_total`. Separately, and independent of tagging, the passthrough streams its results over the Storage Read API on every path: `bigquery.readsessions.create` (`roles/bigquery.readSessionUser`) is a standing requirement for materializing any BigQuery source into a storage destination.
+
+`ManifestEntry.queryCostBytes` is populated for a tagged `storage=` build, and the full per-engine cost — billed bytes, slot or execution time, the cache flag, the warehouse's own job ids — goes to the build's log line.
+
+---
+
 ## [Unreleased] — queries report how they were served, and what they cost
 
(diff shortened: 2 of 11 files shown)


sagarswamirao:
Thanks for turning these round so quickly, and for splitting them out rather than letting them ride on a merged PR.

Spot-checked the three that mattered most, and they hold:

- `BilledReadNotCapturedError` now survives the rethrow (`materialization_service.ts:2379,2394`) with `cause` preserved, and `recordStorageBuildFailure` carries `reason`. That was the one I flagged as a type existing but being erased at its only caller, and the scheduler's `forceRefresh: true` re-drive is exactly why it mattered - good catch running that all the way to the consumer.
- `describe.skipIf(!hasPublicDataBigQueryCredentials())` (`storage_build_bigquery.integration.spec.ts:115`) makes a skip read as a skip. The `3 pass -> 0 pass / 3 skip` line in the description is the right before/after to state, because the old shape was worse than no coverage.
- Moving the row-pick inside the retried closure is the fix I should have asked for the first time. Wrapping only `session.runSQL` guarded the throwing shape and not the "not yet enumerated" shape, which is the same transient class.

On `cacheHit`: appreciate you measuring it rather than taking the report. `ROWS_PRODUCED = 0` on metadata-resolvable aggregates kills the premise cleanly, and a `GROUP BY` rollup scanning normally settles the "ordinary materialization" half. The `GENERATOR` caveat standing as the accurate one is the right resolution.

Two notes on the deferrals, both agreed as scoped:

- The four `logger.warn` sites are genuinely un-redactable where they sit, so threading connections through is a fair follow-up rather than something to bolt on here. Worth a tracking issue though - `tagSnowflakeSession`'s catch fires precisely when the Snowflake session is unreachable, which is the error class most likely to echo account detail, so it is the one of the four I would not leave open-ended.
- Snowflake end-to-end skipping everywhere without secrets is a real blocker. The gap that concerns me most is that deleting the `tagSnowflakeSession` call or swapping its `sourceType`/`handle` arguments leaves the suite green - so if provisioning a credential is far off, even a narrow test asserting the arguments at the call boundary would pin the wiring without needing a live account.

One thing worth calling out for anyone reading this later: the `roles/bigquery.user` case is nastier than a plain denial. It grants `jobs.list` scoped to your own jobs, so the probe succeeds and the listing comes back empty - the gate passes and the path then behaves exactly like the failure it was meant to detect. Since the docs now state the prerequisite, naming `jobs.listAll` specifically would save an operator that debug.

sagarswamirao (approved):

PR #1004, opened 2026-08-13
fix(server): report per-source build failures instead of exiting on the first one

sagarswamirao:
## What the consuming contract asks for

A control plane orchestrating this publisher binds a serving anchor per source, and its design doc states the partial-failure contract directly:

> **Run READY on partial success** — once every unit is terminal, a run rolls up READY as long as at least one in-scope unit produced (or reused) a READY artifact. **Failed units carry their own `error`; healthy units serve.**

with the complementary rule that a build only fails outright when nothing survived:

> **Run FAILED on planning/orchestration failure, or an all-failed plan** — no unit ran, the plan couldn't be formed, or **every in-scope unit the run built failed** (zero produced). ... (A reuse-only / no-op run produces nothing of its own and stays READY.)

and the visibility that depends on it:

> The graph is `getRun`-only, so **partial failures are visible per node**.

## What the publisher did instead

It exited early. The first source that failed threw out of the source loop, so the build abandoned every remaining source, reclaimed the tables already written, and surfaced one message for the whole command.

The consequence is that **there is no failed source** to report. The failed source does not arrive with an empty reason — it contributes no manifest entry at all, and an absent entry carries no cause. Nor do the healthy sources arrive: a package where one source out of five had a bad grant looked exactly like a package that was entirely broken, and the four tables that had already materialized were dropped on the way out.

For anyone running the publisher in orchestrated mode this is the case they most need visibility into. A partial failure is the normal shape of a real failure — one bad grant, one renamed column, one dataset that does not exist yet — and it is precisely the shape this reported least about. The operator sees a failed build with a single message and cannot tell which source to fix; the orchestrator has per-source anchors to bind and nothing to bind them to.

## What this changes

A source that fails is recorded in the manifest with the reason it gave, and the build continues. The sources that materialized stay usable, and a consumer can tell which source failed rather than inferring it from an absence. `ManifestEntry.error` is declared in `api-doc.yaml` so the published contract carries what the build now reports.

A build that loses *every* source still fails — it produced nothing, so it must not report itself as a success with errors attached. A reuse-only run, which builds nothing of its own, is unaffected.

## Recording a failure is not enough on its own

An entry that records a failure still carries `physicalTableName`, and that field is what the rest of the service reads to decide a table exists. Left alone, the failure entry would be indistinguishable from a healthy one at three points, so each now skips an entry that carries a reason:

- **Carry-forward.** The next run reuses a prior entry that names a table, so a failed source would be adopted as already-built and never retried. That is the worst of the three: it turns a transient warehouse error — an expired grant, a dataset created a minute later — into a source that stays unbuilt indefinitely, where the old early-exit behaviour would at least have rebuilt it on the next run.
- **Serve binding.** Both the auto-load path and the manifest split bound the entry's physical name into the package models, so queries would rewrite either to a table that was never created, or — on a failed rebuild, where names are stable — to the previous generation, served as fresh.
- **Run outcome.** The built count came from the instruction list rather than from what the build returned, so a run that lost four of five sources still reported five built and a successful outcome. The count now comes from the returned entries, the run reports a partial outcome when any source failed, and the failed count is logged and recorded.
- **Downstream seed.** A seeded entry feeds a dependent source's `FROM` verbatim, so a failed one would point a downstream build at a table that was never created.

Rather than repeat the check, all five ask a single `isFailedEntry` predicate that lives beside the type it guards, so a consumer added later inherits the rule instead of having to know it. Its own documentation carries the reason: a failed entry keeps the fields it was instructed with, so a table name on it is not evidence that a table exists.

## Why the failure path stays safe

Reclaim exists because a build that records nothing leaves tables no manifest names, and manifest-driven collection only drops names a manifest records. A partial failure now returns a manifest, so the tables it records stay reachable and are not dropped underneath it. Reclaim is left for the case it was written for — a build that recorded nothing — and the all-failed guard throws from inside the same handler, so that path is unchanged.

The reason is redacted against that source's own connection before being recorded, for the same reason the run-level message is: a warehouse error can echo the credentials it was handed, and this value is persisted and rendered downstream.

## Tests

Five, each written before the change it covers and verified to fail without it:

- a partial failure keeps healthy sources usable and records the failed one's reason
- a build whose every source fails still throws
- a per-source reason does not echo the connection's secrets
- a partial failure returns both entries, so the healthy table stays named by the manifest
- **a source whose prior entry recorded a failure is rebuilt, not carried forward**
- **a failed entry never reaches the carried set the failure count is taken over** — the count includes carried entries, so it is only correct while a failed one cannot be carried; the two rules hold each other up

Both of the last two are mutation-checked: removing the carry-forward guard turns exactly those two red and nothing else.

`bun test` on `packages/server`: **135 pass, 0 fail** across the service and manifest-loader specs. `bun run lint` and `bun run format` clean across all packages.


housejester (commented):
Nice change, and the motivation is right: a partial failure is the normal shape of a real failure, and it was the shape this reported least about. The all-failed guard and the redaction of the per-source reason are both the right calls.

The thing that makes this worth careful review isn't the logic — it's that it introduces an invariant the type system can't enforce, and that now crosses the API boundary:

> **entry present** ⇒ built  →  **entry present *and* no `error`** ⇒ built

`isFailedEntry` is a good way to carry that inside this package. My findings are about the places it doesn't reach, and the places nothing would catch if it stopped reaching them.

## A sixth consumer isn't gated, and it's the serve path

`deriveServeBindings` (`materialization_serve_transform.ts:145`) reads manifest entries and doesn't ask `isFailedEntry`. It's safe today only *incidentally* — the failure entry omits both `storageDestinationName` and `schema`, and each of those is an early-`continue` in the binding loop. So the protection is two accidental field absences rather than the rule.

That's a fragile thing to rely on: adding `storageDestinationName: instruction.destination` to the failure entry is a natural future change — an orchestrator would reasonably want to know which destination the failed source was targeting — and it silently re-opens the serve path. `buildDownstreamViaParents` reaches the same function over the in-progress entries map.

Given the predicate's own docstring says the point is that "a consumer added later inherits the rule instead of having to know it", this one should ask it explicitly.

## Four of the five guards aren't pinned by a test

I mutation-tested each guard on the PR head (`bun test materialization_service.spec.ts manifest_loader.spec.ts`):

| guard | remove it → |
|---|---|
| carry-forward `materialization_service.ts:1014` | **2 fail** — exactly the two named, as claimed ✅ |
| serve split `manifest_loader.ts:139` | 135 pass, 0 fail |
| auto-load bind `materialization_service.ts:1236` | 135 pass, 0 fail |
| downstream seed `materialization_service.ts:1529` | 135 pass, 0 fail |
| failure count → hardcode `sourcesFailed = 0` | 126 pass, 0 fail |

The three uncovered guards are the ones whose failure modes the description itself calls out as most serious — *"queries would rewrite either to a table that was never created, or … to the previous generation, served as fresh."* The mutation discipline applied to the carry-forward pair is exactly right; it just didn't extend to the others. The last row also means the whole "Run outcome" bullet — the `partial` outcome, the corrected built count, `recordSourcesOutcome("failed")` — is currently unasserted.

## The overload is invisible to consumers outside this package

`physicalTableName` is in `ManifestEntry`'s `required` list, so a failed entry is *obliged* to carry a table name for a table that doesn't exist. Inside the server that's handled by `isFailedEntry`; an orchestrator reading the manifest over the API gets a doc sentence and a strict generated client that will happily deserialize an entry that looks built.

Worth considering whether failures belong in a sibling collection on `BuildManifest` instead — structurally impossible for any consumer, present or future, to mistake for a built table, and it drops the need for five internal guards. That's a bigger change than this PR and I'd understand deferring it, but the five-guards-plus-a-sixth-I-found is the cost of the overload, so it seems worth naming before the contract ships rather than after.

## Smaller things

- **Stale metric HELP text.** Three strings still enumerate the pre-change label sets, and these publish to Prometheus as the operator-facing contract: `materialization_metrics.ts:6` (file header, `success|failed|cancelled`), `:98` (`runCounter`, same), `:107` (`sourcesCounter`, `'built'|'reused'` — now also `failed`).
- **No `RELEASE_NOTES.md` entry.** This adds a response field and a new metric label value, and a consumer generating a strict client from `api-doc.yaml` will reject `error` until it regenerates. Worth a line for anyone pinning a release.
- **A note for anyone alerting on this metric:** a success-rate expression written `success / (success + failed)` now drops `partial` into neither bucket, so a partial failure reads as a *dip in volume* rather than a failure. Not a change request — just cheaper to say in the release notes than to discover.
- **Description nits:** "Tests: **Five**" lists six bullets, and six ship. And *"The count now comes from the returned entries"* — the *failed* count does, but built is `instructions.length - sourcesFailed`, still instruction-derived, so a source skipped by `if (!instruction) continue` still counts as built.

## Verified rather than assumed

- `bun test` on the two specs: **135 pass, 0 fail** — matches the description exactly.
- `bun run typecheck:server` clean.
- The carry-forward mutation claim is precisely true: removing `!isFailedEntry(prior)` turns exactly those two red and nothing else.

**Top three:** 1) gate `deriveServeBindings`; 2) pin the three uncovered guards and the outcome/count path; 3) metric HELP strings + a release-notes line.

sagarswamirao:
Thanks — the mutation table was the part that mattered most, and it caught something in my own work I'd have shipped otherwise. All addressed in `da7e4886`.

## `deriveServeBindings` — gated

You're right that it should ask explicitly rather than rely on the failure entry omitting a destination. `isFailedEntry` is now the first check in that loop, with a comment saying why: adding `storageDestinationName` to a failure entry is a reasonable future change, and the omission is not a property worth depending on. Its test gives the entry every field a binding needs, including the destination, so it can only pass because of the `error` check.

## Every guard now has a test that fails when that guard alone is removed

Six guards, six mutations, each turning exactly its own test red:

| guard | mutation |
|---|---|
| carry-forward | 2 tests red (the pair, as before) |
| serve split | own test red |
| auto-load bind | own test red |
| downstream seed | own test red |
| serve bindings | own test red |
| counts | own test red |

**One of those tests was worthless in its first form, and the mutation is what found it.** The seed test asserted over the SQL a build emits — but that guard controls an in-memory `manifest.update` call that never reaches SQL, so it passed with the guard removed. It now asserts on the manifest call itself, and checks a *healthy* seed is still bound so a guard rejecting everything can't pass either. That's a fair hit on the original claim: I applied the discipline to one pair and then didn't extend it, which left four guards as decoration.

## The counting bullet — and a real defect underneath it

You filed this under description nits, but the arithmetic was wrong, not just the sentence. Both counts now come from the returned entries via a `tallySources` helper. That also fixes the case you spotted independently: an instruction skipped by `if (!instruction) continue` counted as built. Extracting it made the logic testable — it had been inline in a 300-line method — and it's now pinned in both directions, including that a carried source isn't counted as built either.

## Metric HELP strings and release notes

All three strings corrected (`:6`, `:98`, `:107`). Release note added, and it states the alerting consequence in your terms: `success / (success + failed)` now leaves `partial` in neither bucket, so a partial failure reads as a dip in volume rather than a failure. Cheaper to say than discover, agreed. It also warns that `physicalTableName` stays required, so presence of a name is not evidence a table exists.

Description corrected too — the test list no longer carries a count that goes stale, and the built-count claim now matches what ships.

## The overload — deferring, but the cost is now measured rather than assumed

Your structural point got stronger as I worked through this. It's six internal guards, each needing its own test, precisely because `physicalTableName` is in `required` and a failed entry is obliged to name a table that doesn't exist. A sibling collection on `BuildManifest` would make that structurally impossible for any consumer and drop all six.

I'm not doing it here — it changes the response shape rather than adding to it, and this PR is already the reproduction plus its fix. But I'd rather name the cost than let six guards read as free, so it's worth deciding before the contract ships. Happy to open it as a follow-up if you'd prefer that shape from the start.

`bun test` on the three specs: 199 pass, 0 fail. `bun run lint` and `bun run format` clean.

housejester (approved):
Verified against `a728ef44` rather than taking the table — I re-ran all six mutations and it reproduces exactly:

| guard removed | tests red |
|---|---|
| carry-forward | 2 — the pair, as before |
| auto-load bind | 1 — `does not bind a failed source's table into the pack…` |
| downstream seed | 1 — `does not seed a downstream build from a faile…` |
| serve split | 1 — `keeps a failed source out of both the serve man…` |
| serve bindings | 1 — `does not bind a source that failed, even with a …` |
| counts | 1 — `tallySources > counts a failed source as failed, never as built` |

Baseline 199 pass / 0 fail on the three specs.

`deriveServeBindings` now asks first, and its test name is the tell that the coverage is real — *"even with a destination"* means the entry carries every field a binding needs, so it can only pass on the `error` check rather than on the omission. That closes the thing I was actually worried about: the guard no longer depends on a property that a reasonable future change would remove.

The reworked seed test is two-sided now — a healthy seed is still bound, so a guard that refuses everything can't pass either. And credit for finding the worthless one yourself: a test asserting over emitted SQL for a guard that controls an in-memory `Manifest.update` is exactly the failure the mutation pass exists to surface, and you disclosed it rather than quietly replacing it.

`tallySources` is the right call and better than what I asked for. I filed the counting under description nits; you were right that the arithmetic was wrong underneath, and extracting it from a 300-line method is what made it pinnable in both directions.

## One nit, non-blocking

`tallySources` changes what `publisher_materialization_sources_total{outcome="built"}` reports — the old instruction-derived formula counted a skipped instruction as built, the new one counts what came back. A correction, but deployments trending `built` will see a step change, and the release note documents the *new* label values without mentioning that an existing one shifts meaning. You already applied this exact standard to the success-rate expression, so it seems worth the same sentence.

## On deferring the sibling collection — ship this, and open it now

To answer directly: I'd merge this and open the reshape immediately, rather than either holding this PR or leaving the follow-up for later.

Holding it isn't right. The behavior fix stands on its own — a partial failure currently discards healthy tables — and `ManifestEntry.error` is load-bearing for it, so splitting the field out would gut the PR rather than descope it.

But "later" isn't right either, and this cuts against my own earlier framing. I raised the sibling collection as probably too big for this PR; working through it made the case *stronger*, not weaker. Six guards each needing their own test is the measured cost of `physicalTableName` being required on an entry that names no table — and that cost only exists because the failure rides the same object. Moving it to `BuildManifest` makes the mistake structurally unavailable to every consumer, present and future, and drops all six.

The asymmetry is timing: the reshape is a breaking change to `error` whenever it happens, and it gets more expensive with every release consumers pin against. One release old is a footnote; ten is a migration. So yes — please open it as a follow-up, and I'd treat it as the next thing rather than a someday.

Approving.

PR #1005, opened 2026-08-14
docs(skills): governed-model pauses, labeled assumptions, and discovery breadth

jswir:
## What

Skill updates distilled from two full QA sessions (the 2026-08-13 Spotify/Kaggle modeling session and the BEAVER/MySQL session) plus a diff between the launch blog post and the skills. The through-line: **these are governed semantic models that require a human SME** — the skills now require the behavior the post describes, instead of permitting a full model to ship with zero confirmed business decisions.

## Owned skills

- **malloy-modeling** — the `(user)` pauses are real stops ("The pauses are the point"): when a decision goes unanswered, take a best-supported position, **label it an assumption in the field's `#(doc)` and in `modeling-notes.md`**, and raise it again at Review. The field-completeness bar now includes threshold provenance. Step 7 is a required review checklist (confirmed definitions vs agent-chosen thresholds vs unanswered questions; a multi-definition headline metric is a blocking question). `modeling-notes.md` is an expected output from step 2, with a compact template. Step 8 Curate: always propose, user decides. Each step's skill must be read before that step. Setup pointer to the scaffolder and `--watch-env`.
- **malloy-document** — new **Mark conventions as conventions** section (a `#(doc)` must let a reader tell a measured fact from a choice; WRONG/RIGHT example; don't hedge measured facts). Counts must name the noun counted and whether it counts distinct entities or events.
- **malloy-publish** — a flat-file package needs **no** `connections` entry (automatic per-package `duckdb` sandbox; the name is reserved and an env-level `duckdb` fails the environment); `location` is local only with a `./`/`../`/`~/`/`/` prefix; local authoring means `--watch-env` (without it, boot-time copies mean edits are never read).
- **malloy-getting-started** — `--watch-env` in the server-start line; `claude mcp add … -s user` escape hatch when the session can't start in the workspace directory; don't trust a stale `.mcp.json` URL — verify with `lsof` + `malloy_getContext`.
- **malloy** (index) — reference table reorganized into five sections covering all 29 skills in this deployment. Publisher-owned and never synced; listed here so the restructure is not silent scope.

## Shared skills (upstream sync debt)

These are shared with `ms2data/agent-skills` and were edited here first by explicit decision; copy the same text upstream to keep the pair byte-identical. `skills/README.md` says the copy is still manual, so **this list is the worklist** — every shared path the diff actually touches:

Governance (the four named in the original description):

- `skills/malloy-model/SKILL.md` — Key Rule: never write a self-chosen threshold; boundaries are user-supplied, distribution-derived, or flagged in `#(doc)` (writing-time mirror of malloy-define's data-driven-proposals rule). `rename:` composes with `include {}` in one order (`extend { rename: }` first).
- `skills/malloy-gotchas-modeling/SKILL.md` — new **Mixed-Grain Joins** section (a pre-aggregated source joined into a detail source ignores query filters by construction; two legitimate resolutions), plus a threshold pointer. `rename:` + `include {}` order rule, quoting the compiler's real messages.
- `skills/malloy-discover/SKILL.md` — prior-art signals broadened beyond `.lkml`/dbt (metadata files, KPI docs, catalog exports, READMEs, screenshots); record findings into `modeling-notes.md` as you go; profiling must run through Malloy even when the raw engine is available; sibling-tables-at-the-same-grain checklist item. `days(a - b)` replaced with `days(a to b)`.
- `skills/malloy-scope/SKILL.md` — scope is "which questions", not just "which tables"; domains framed as business processes vs entity/snapshot datasets, and the proposal must state what the model kind rules out; confirmed scope recorded in `modeling-notes.md`.

Compiler-fact corrections (copy these too, or the wrong claims keep shipping to every other consumer of the shared repo):

- `skills/malloy-analysis/SKILL.md` — `search_malloy_docs` example `"autobin"` → `"histograms"` (there is no `autobin`).
- `skills/malloy-charts/SKILL.md` — histogram example no longer calls `autobin(...)` (fails with `Unknown function 'autobin'`); arithmetic binning off the observed range, with the bin-width rule stated here rather than behind a cross-skill hop.
- `skills/malloy-debug/SKILL.md` — `"Can't find field"` + `rename:` row no longer says never use `rename:`; it describes the order rule (`extend { rename: }` before `include {}`).
- `skills/malloy-define/SKILL.md` — `days(now - last_order_date)` → `days(last_order_date to now)` (subtraction does not compile).
- `skills/malloy-patterns/SKILL.md` — `"Auto-binning Histograms"` / `"autobin"` → `"Histograms (bin by arithmetic; there is no autobin)"` / `"histogram"`.
- `skills/malloy-lookml-review/reference/_concepts.md` — `label:` mapping no longer says "Never use `rename:`"; `internal:` + `dimension:` is the lighter move and keeps the raw column reachable.
- `skills/malloy-lookml-review/reference/propose-fields.md` — `count()` "is always distinct" was backwards. `type: count` → `count()` (both count rows); `type: count_distinct` → `count(field)` (`count(field)` is already distinct; `count(distinct field)` is deprecated).

## Mechanics

- MCP skills bundle regenerated (`skills_bundle.json`); `bun test src/mcp/skills` passes (17/17, including the em-dash house-style check).

## Evidence

Each edit traces to a specific observed failure: 11 unlabeled hardcoded thresholds and a merged scope question (Spotify session), the 9.6× headline-metric spread never raised, the mixed-grain `decade_track_count` misread, the missed same-grain sibling table and the enrolments-vs-students doc (BEAVER session), and ~25 tool calls lost to undiscoverable setup facts.


skills/malloy-gotchas-modeling/SKILL.md
```diff
@@ -165,22 +178,29 @@ source: orders is conn.table('orders') include {
...
+    public: order_id, user_id
+  }
+  extend { measure: revenue is raw_revenue.sum() }
 ```
 
-If you can rename the measure or split the source instead, prefer that: it preserves `include {}` and the curated surface.
+Two ways to get the order wrong, with the errors they produce:
+
+- **`include {}` before the renaming `extend {}`** fails with `Can't find field 'X' to set access modifier`, currently surfaced as an internal compiler error. `include` runs against names that no longer exist by the time the rename is applied.
+- **Naming the pre-rename column inside `include {}`** fails with `` `revenue` not found 'revenue' is not defined ``. After a rename only the new name exists; use it.
```
Sha-Bang: Should fix: the quoted error is invented. I compiled this exact scenario against the pinned compiler (rename `revenue` to `raw_revenue`, then `include { internal: revenue }`) and the only problem returned is:

```json
{"code":"field-not-found","message":"`revenue` not found","severity":"error"}
```

There is no `'revenue' is not defined` fragment anywhere in the problem list. The advice around it is right, and the fix is just to trim the quote back to the real message.

Worth correcting precisely because of what this file is for: it teaches an agent to pattern-match on the compiler's exact text, so a fabricated string here is self-undermining. The neighbouring claim on line 200 is correct verbatim, internal-compiler-error framing included, which is what made this one stand out.

skills/malloy-model/SKILL.md
```diff
@@ -176,11 +176,12 @@ source: customer_health is customers extend {
...
 - **Mark raw columns `internal` when a derived dimension replaces them**
 - **Check for duplicate rows** before building measures
 - When both a combined table (all types) and filtered/split tables exist, prefer the split tables
 - **DRY: define measures/dimensions in base source files, not inline in views**
+- **Never write a threshold, tier boundary, or bucket cutoff you chose yourself.** Every boundary in a `pick` expression or filtered measure is user-supplied, distribution-derived (query `min`/`p25`/`p50`/`p75`/`p95` first and show the evidence; see `skill:malloy-define` § Data-driven tiers), or explicitly flagged as an assumption in its `#(doc)`. A hardcoded cutoff nobody confirmed is a business decision shipped as fact.
```
Sha-Bang: Nit: `skill:malloy-define` § Data-driven tiers points at a bolded inline phrase in a table caption (`malloy-define/SKILL.md:100`), not a heading. The nearest real heading is `## Data-driven proposals` further down. The reference resolves, so nothing dangles; the `§` just promises a section that is not one.

Sha-Bang (commented):
Fetched the branch into a scratch worktree, ran `bun install` and `bun test src/mcp/skills`, regenerated the bundle from source to diff against the committed one, and compiled the changed Malloy examples against the pinned `@malloydata/malloy` via `Environment.compileSource`.

The governance mechanism itself holds up. The labeled-assumption chain is concrete rather than vacuous: it names where the label goes (the field's own `#(doc)`), what it looks like (the WRONG/RIGHT pair), a second durable home (`modeling-notes.md` "Open decisions"), and a checkpoint that re-raises it (Step 7). An agent can tell whether it complied.

The Publisher factual claims all check out against source, which is the part that would have been expensive to get wrong:

- `environment_store.ts:1825-1832` confirms the `location` local-prefix set is exactly `./`, `../`, `~/`, `/`.
- `docs/connections.md` plus `environment_store.ts` confirm a flat-file package needs no `connections` entry, the per-package `duckdb` sandbox is automatic, and the name is reserved at env level, `shared_duckdb` workaround included.
- `inPlaceEnvs` / `PUBLISHER_WATCH` / the symlink-vs-copy branch confirm that without `--watch-env` local packages are copied at boot and edits are never re-read.

And against the real compiler: the `extend{rename}` to `include{}` to `extend{measure}` ordering compiles clean; `count(distinct x)` produces exactly the deprecation the text now quotes; `autobin(...)` fails with `Unknown function 'autobin'`; `days(a - b)` fails with `Can not offset time by '<type>'` while `days(a to b)` succeeds; `months(...)` fails on timestamps with the calendar-interval message and succeeds on dates. Each of those overturns a wrong pre-PR claim, and the replacements are accurate. `skills_bundle.json` rebuilds byte-identical from this tree, and `bun test src/mcp/skills` is 17/17.

Two inline notes, plus one on the description:

Should fix: the sync worklist is understated, and `skills/README.md` says the upstream copy is manual with no automation, so the description *is* the worklist. It names four shared skills. The diff touches ten shared files: those four plus `malloy-analysis`, `malloy-charts`, `malloy-debug`, `malloy-define`, `malloy-patterns`, and `malloy-lookml-review/reference/_concepts.md` and `.../propose-fields.md`. Whoever performs the upstream copy against this list silently drops six files' worth of real fixes, and those six are where the substantive corrections live: the `autobin()` claim, the `days(now - x)` subtraction syntax, `count()` being always distinct (backwards), and `rename:` being incompatible with `include{}` (backwards). Those keep shipping to every other consumer of the shared repo with nothing flagging the gap. Separately, `skills/malloy/SKILL.md` was substantially restructured (the reference table reorganized into five sections covering all 29 skills) and is not mentioned at all. It is Publisher-owned and never synced, and every name in the new table resolves to a real directory, so it is undisclosed scope rather than a defect. Either list every shared file actually touched, or split the four-skill governance change from the incidental fixes to the other six.

Nit: the "never ship a self-chosen threshold" rule is now stated near-verbatim in four places (`malloy-model` Key Rules, `malloy-document` "Mark conventions as conventions", `malloy-gotchas-modeling` "Thresholds Are Decisions, Not Syntax", `malloy-modeling` Step 7). They agree today. This is the same multi-copy shape that produced the stale `rename:` / `count()` / `autobin` claims this PR is fixing, so it is worth knowing it is there. Not blocking.

Sha-Bang (approved):
Approving. The governance mechanism is sound and every factual claim about Publisher internals checks out against source, which was the part worth verifying carefully.

The two items from my review are both worth doing but neither blocks: the fabricated error string in `malloy-gotchas-modeling:201` is a one-line trim, and the understated sync worklist is a description fix rather than a code change. The second one does need doing before someone runs the upstream copy, though, or six files of real corrections stay behind.

skills/malloy-model/SKILL.md
jswir: Pointed at the real heading: `skill:malloy-define` § Data-driven proposals.

skills/malloy-gotchas-modeling/SKILL.md
jswir: Trimmed to the compiler's actual message: `` `revenue` not found ``. The `'revenue' is not defined` suffix is gone.

PR #1006, opened 2026-08-14
feat(server): make a failed reload visible off-box, and give MCP a health surface

jswir:
Fixes F1 and F7 from the QA field notes ("Publisher Field Notes", BEAVER session).

## The failure

Under `--watch-env`, saving a model that fails to compile is logged to stderr and nowhere else. `/api/v0/status` keeps reporting `loadErrors: null` and keeps listing the package, so:

- a good model compiled earlier → the server serves the **stale** model; queries return confident numbers from the previous version, and nothing says so;
- no good model ever compiled → the package serves zero sources; `malloy_getContext` returns `results: []`, indistinguishable from an empty package.

An agent is the worst-affected consumer — it cannot see the terminal, has no MCP health surface at all, and the field report includes a session that told its user "compiles clean" on the strength of that status.

## The fix

**Server (F1)**
- `Environment` records a failed reload per package in a new `staleCompileErrors` map — deliberately separate from `failedPackages`, because the package is still SERVING its last good compile and the serving counters must not change. Cleared by the next successful load (`clearPackageLoadFailure`). Recording sits in `_loadOrGetPackageLocked`'s keep-serving branch, so every reload caller is covered: the chokidar watcher, `malloy_reloadPackage`, and REST `?reload=true`.
- `getStatus()` reports each entry as a `loadErrors` item with `stale: true` and `failedAt`. Additive OpenAPI change; a healthy status body is byte-for-byte unchanged.

**MCP (F7)**
- New `malloy_getStatus` tool: `operationalState`, per-environment package names, and `loadErrors` including stale entries. Payload is the REST status **reduced** — no connections, locations, descriptions, or theme — parity-minus-detail with the already-unauthenticated `/status`.
- `malloy_getContext` tier 2 now lists failed packages with an `error` marker instead of silently omitting them; tier 3 adds a `note` on the empty payload only, distinguishing "loaded but exposes no sources (curation gap — check `explores`/`export {}`)" from an empty database. Populated payloads stay byte-identical.
- `MCP_INSTRUCTIONS` (still a static literal per the SECURITY comment): after every model edit call `malloy_reloadPackage` before querying; never read an empty `getContext` as an empty package.

## Tests

- `environment_store.spec.ts`: failed reload → `stale: true` entry while the package stays listed and serving; fixed file + reload → `loadErrors` absent again (byte-for-byte-healthy contract).
- `watch_mode.integration.spec.ts` ("survives a transient compile error"): now also polls `/api/v0/status` for the stale entry appearing and clearing, end-to-end through the chokidar watcher.
- `get_status_tool.spec.ts` (new): payload reduction (asserts no connection names leak), loadErrors omission, error path.
- `get_context_tool.spec.ts`: failed-package marker in tier 2; empty-surface note in tier 3; populated tier 3 carries no note.
- `server.protocol.spec.ts`: tool list includes `malloy_getStatus`; instructions pin the reload-after-edit rule.

Full unit suite green (2419 pass), watch-mode integration suite green (11 pass).


kylenesbit (approved):
Approving. All checks green; reviewed the diff and verified the change is additive.

What I checked:
- `staleCompileErrors` is kept genuinely separate from `failedPackages`, so the serving counters and `getFailedPackages()` are untouched and a stale package keeps serving its last good compile. `clearPackageLoadFailure` clears all three maps, and recording in `_loadOrGetPackageLocked`'s keep-serving branch does cover all three reload callers (watcher, `malloy_reloadPackage`, REST `?reload=true`).
- The status message is passed through `redactPgSecrets` at the point it is recorded, so the new `loadErrors` entries and the `getContext` `error` markers inherit that redaction rather than re-implementing it.
- `loadErrors` stays unset when nothing failed, so a healthy `/status` body is unchanged as claimed.
- `malloy_getStatus` is a genuine reduction of the REST payload: names, states and load errors only, no connections, locations, descriptions or theme.
- The `note`/`error` fields are spread conditionally, so a healthy package's `getContext` payload is byte-identical to before.
- The staleness lookup in tiers 3/4 is wrapped and logged rather than thrown, so a health annotation can't take discovery down.

Merged together with #1007, #1008, #1009, #1010 and #1011 this branch conflicts with none of them, typechecks, and the combined server unit suite passes (2453 pass / 0 fail).

PR #1007, opened 2026-08-14
fix(server): one-line port-conflict exit, and never leave a stale .mcp.json behind

jswir:
Fixes CR-6 and CR-13 from the Spotify QA session (SETUP-RETRO 🐞-7/🐞-8), where the two compounded live: boot A bound MCP on 4040, **wrote `.mcp.json`**, then died on REST `EADDRINUSE` with a ~40-line uncaught-exception dump. Boot B on `--mcp_port 4041` skipped the rewrite (create-never-edit), leaving a file that pointed every future agent session at a dead port with nothing saying so.

## Changes

- **EADDRINUSE is one actionable line**, not a crash dump: `Port 4000 in use; pass --port <n>` (respectively `--mcp_port`), sibling listener closed best-effort, exit 1. Any other listen error also gets a single line + exit 1 (a listen failure is always fatal here; previously it was an uncaught `'error'` event either way).
- **`.mcp.json` is written only after BOTH listeners bind.** The write moves out of the MCP listen callback into a continuation gated on a promise resolved from the REST callback, so the failure shape above can no longer produce a fresh stale file.
- **An existing `.mcp.json` is compared, not trusted silently.** When its `mcpServers.malloy.url` parses and differs from this run's endpoint, the boot logs a **WARN** naming both URLs and the `claude mcp add` fix. The module's "never reads" rule is relaxed deliberately and narrowly: lstat-gated to a regular file under 64 KiB (a FIFO can't block the boot, a symlink is never read through), only the one URL is extracted (the file may hold other servers' credentials), the file is never rewritten, and the matching/unreadable case keeps its calm every-boot `info` line. Header comment updated to state the new contract.

## Tests

- `mcp_config.spec.ts` (57 pass): new compare cases — differing URL sets `existingUrl`, matching/unparseable/stdio-bridge/symlink stay quiet — plus the WARN text naming both URLs.
- New `port_conflict.integration.spec.ts`: real subprocess boots for both busy-port cases asserting the exact message, nonzero exit, and **no `.mcp.json` created** (cwd is a non-git temp dir so the git guard can't mask a wrong write); plus the happy path asserting the file appears with the bound MCP endpoint once both ports are up — the first test of the boot-time call site.
- Full unit suite green (2421 tests).

One incidental: `fatalListenError`'s comment is line-comments by necessity — `authorize_bypass_header.spec.ts` strips block comments from `server.ts` with a regex that pairs a route string's `/*` with the next closer, so a JSDoc after the route table swallowed the `getQuery` call it asserts on. Noted in the comment itself.


kylenesbit (approved):
Approving. All checks green, and the two behaviours are well covered by the new subprocess test.

What I checked:
- The `.mcp.json` write really is gated on both listeners: `resolveRestBound()` fires from the REST listen callback and the write moves inside `restBound.then(...)`, so the MCP-wins-the-race-then-REST-dies shape can no longer produce a fresh file. `boundAddress` is captured before the continuation rather than relying on `this` inside it, which is the subtle part and is done right.
- `conflictingMalloyUrl` keeps the module's old safety properties: `lstatSync` + `isFile()` means a FIFO can't block the boot and a symlink is never read through, the 64 KiB bound caps a surprise file, only `mcpServers.malloy.url` is extracted, and the whole thing is wrapped so it never throws. The file is still never rewritten.
- The symlink early-return happens before any compare, so `existingUrl` stays undefined there, matching the test.
- Two log tones are the right call: the mismatch case is a real WARN naming both URLs and the fix, while the matching/unreadable case stays `info` because it fires on every boot of a scaffolded package.

One note, not a blocker: `fatalListenError` is attached to the servers' generic `error` event, not only to listen failures, so any later `http.Server` `error` also becomes log-one-line-and-exit-1. For `http.Server` that event is effectively listen-time (client socket errors go to `clientError`), and the previous behaviour was an uncaught exception on the same event, so this is strictly an improvement in presentation rather than a change in what is fatal — the comment already says as much.

Merges cleanly with the other five; combined typecheck and server unit suite green (2453 pass / 0 fail).

PR #1008, opened 2026-08-14
fix(server): scope the query boundary package-wide, and exempt /compile from it

jswir:
Root-causes and fixes **HANDOFF CR-5** from the Spotify QA session: with all four model files listed in `explores` and `queryableSources: "declared"`, three of four sources 404'd on query, **every** file 404'd on compile, and `exploresWarnings` said none.

## Diagnosis (reproduced, then pinned)

The source-level gate consulted only the requested model's **own** export closure. A source declared queryable by its own `explores`-listed file was denied whenever the request addressed it through a model that imports it — reproduced with the QA session's exact error (`No queryable source "tracks"`). A client that posts every query through one model path (the session's `mq.sh` helper; any agent that pins one `modelPath`) loses every source but that file's own. The naive shape — each file queried via its own path — was never broken: the same repro **passes on the v0.0.243 tag**, ruling out an exports-population or path-normalization bug.

## Changes

1. **Package-wide queryable surface.** `Package.applyQueryBoundaryToModels` computes the union of every `explores`-listed model's export closure and pushes it to each model; `curatedSourceNames()` and the named-query gate union it in. This admits nothing new — every admitted name was already queryable via its own model path — and the file-level entry-point gate plus non-exported-source denials are unchanged (all 15 pre-existing boundary tests still pass, including the injection matrix).
2. **`/compile` (REST and `malloy_compile`) is exempt from the query boundary.** Compile is the authoring loop (`compile → save → reload`); the boundary is discovery curation, not access control — `#(authorize)` still gates compile in full (its 4-test suite untouched). **Accepted trade, stated in the code comment:** `/compile` can reveal a non-exported, non-authorize-gated source's schema (and with `includeSql`, SQL). Sources whose confidentiality matters belong behind `#(authorize)`.
3. **Listed-but-blank files are no longer silent.** `emptyDiscoveryWarnings` generalizes from import-only files to any listed model whose curated surface is empty, and now rides the package's API-visible `warnings` array (status / GET package / `malloy_reloadPackage`) as well as the load log — so `exploresWarnings: none` can't coexist silently with listed files that surface nothing. Stays advisory, out of the strict publish gate.
4. **Docs corrected:** `api-doc.yaml`'s `queryableSources` description now states the union scoping and compile exemption; `skills/malloy-publish` claimed the boundary gates compile — fixed, MCP bundle regenerated.

## Tests

- `query_boundary.spec.ts`: the CR-5 shape (4 no-export listed files, one importing the other three) — every source queryable via its own path AND via the importer (named view, ad-hoc, and ad-hoc derivation), alongside all existing denial/injection tests.
- `compile_authorize.spec.ts`: new exemption suite — compile succeeds against a non-listed file and against a hidden source, and a control asserts the **query** path still denies the same targets.
- `explore_visibility.spec.ts`: structured warning shape + rides-the-metadata assertion.
- Full unit suite green (2419 pass). MCP integration suite unchanged vs main (its 2 ambient errors reproduce on a clean main checkout with this machine's other Publisher on :4000).

Note: touches `skills/malloy-publish/SKILL.md` + the skills bundle, which #1005 also touches in different hunks — whichever merges second needs a trivial bundle regen.

---

## Follow-up commit: admit by declaration, not by name

Review of the above found the union as first written was keyed on bare source and query **names**, which falsified its own "admits nothing new" claim. Malloy resolves a name in the **requested model's** namespace; before the union that was the same namespace the curated set came from, so matching a name proved you had the right source. The union broke that coupling, and a name curated anywhere cleared the gate everywhere.

Reproduced on a running server against a control build of `main`, both axes:

- **Source axis.** `a.malloy` (listed) exports its own `customers` (1 row). `b.malloy` (listed) imports a hidden file declaring a *different* `customers` (3 rows) and exports only `safe`. `POST b.malloy/query` `run: customers -> …` returned `c=3` — the hidden source — where `main` returned 404.
- **Named-query axis, worse.** The pure-`queryName` fast path returns `"cleared"` and skips the compiled backstop entirely, so `b.malloy`'s unexported private `top` ran against its hidden source (`c=7` vs 404 on `main`).

Both leaked through REST and MCP alike. `#(authorize)` still held, so exposure was bounded to hidden sources that are *not* authorize-gated — precisely the set `queryableSources: "declared"` documents as a 404.

**The fix.** The union now maps each name to the **definition identities** exported under it (defining file URL + position, read off `modelDef.contents[name].location`). A request is admitted only when the model it names resolves that name to the very declaration a listed model exported — which is what makes "admits nothing new" actually true. A legitimate re-export still works, since the exporting model's closure carries the declaration's own location wherever it lives; with no location attached the check fails closed to the model's own export closure. The union loop also gained the `.malloy` suffix filter its sibling loops already apply, so a notebook listed in `explores` (invalid but fail-safe-served) contributes nothing.

**Second fix: the compile exemption was an existence oracle.** A source both hidden *and* `#(authorize)`-gated answered `/compile` with a 403 naming it, where the query surface answers a flat 404 — enough to enumerate the gated namespace one guess at a time, and *not* covered by the accepted trade above, since authorize protects data, not existence. The authorize denial path now re-runs the boundary and converts a hidden target's 403 into the generic 404. Compile itself stays ungated: a hidden non-gated source still compiles (the authoring loop the exemption exists for), a visible gated source keeps its informative 403, and a satisfying `given:` still compiles the hidden one.

**Also.** Removed `assertQueryBoundaryForRunnable`, dead since the `/compile` gate was dropped and still documenting itself as an active compile chokepoint; corrected the surviving claims that the boundary gates `/compile` (`package.ts`, `model.ts`, and `docs/discovery-and-access.md` — the canonical doc the first commit missed while fixing the same sentence in `api-doc.yaml` and the skill); `api-doc.yaml` gained the declaration-not-name semantics and the empty-discovery finding kind now riding the `warnings` array.

**Tests.** New collision cases on both axes plus a re-export guard against over-tightening (all three fail on the unfixed build); a compile existence-oracle suite (two of its four fail on the unfixed build). Unit suite 2423 pass / 0 fail; integration 237 pass / 1 fail / 5 errors, identical to a clean `main` checkout on the same machine.


kylenesbit (approved):
Approving. This is the highest-risk change of the batch, so I verified the security-relevant claims by running the new specs against three builds rather than reading them.

**The tests genuinely characterise the fixes.** Against the branch head, `query_boundary.spec.ts` + `compile_authorize.spec.ts` + `explore_visibility.spec.ts` are 44 pass / 0 fail. Against `main` (140b4be2), 5 fail — the CR-5 shape and the re-export guard, i.e. the behaviour the union adds. Against your own intermediate commit 45e4a460 (the name-keyed union), 6 fail, and they are exactly the ones that matter:

- `denies a hidden source that merely shares a name with a sibling explore's export`
- `denies a private named query that shares a name with a sibling explore's exported query`
- all four `compile exemption is not an existence oracle` cases, including the `includeSql` and multi-statement-decoy variants

So the name-collision leak and the existence oracle were both real, and both are now pinned by tests that fail on the unfixed build. The self-review that caught the first-commit regression is the reason I'm comfortable with this.

On the design: keying the union on definition identity (`modelDef.contents[name].location`, file URL plus position) rather than bare names is the right primitive, and failing closed to the model's own export closure when no location is attached is the right default. `admittedByPackage` requiring `mine !== undefined && identities.has(mine)` means an absent identity can never accidentally admit. Hoisting `ownCuratedSourceNames()` out of the `derivesFromCurated` walk keeps the chain check cheap while still identity-checking per link.

Two things I want on the record rather than as objections:

1. **The compile exemption is a deliberate widening of what `/compile` reveals**, and merging this ratifies it: a non-exported, non-`#(authorize)`-gated source's schema (and its SQL under `includeSql`) becomes readable through `/compile` in a package that set `queryableSources: "declared"`. Your code comment states this plainly and the existence-oracle mask closes the part that authorize *does* protect, so the residual exposure is bounded to exactly the set the docs describe as curation rather than access control. It is a policy call, correctly reasoned and documented — just one worth being conscious of, since `queryableSources: "declared"` reads like a stronger guarantee than it now is on that one endpoint.
2. `denyHiddenAsNotQueryable` rethrows the original `AccessDeniedError` unless `convert()` throws, which is the intended control flow, but it also means an unexpected internal throw from `convert()` would surface instead of the 403. That fails closed, so it is cosmetic.

`hasEmptyDiscoverySurface()` broadening from import-only to any empty curated surface is advisory-only and rides the `warnings` array, so no publish gate changes.

Merges cleanly with #1006, #1007, #1009, #1010 and #1011; combined typecheck clean and the full server unit suite green on the merged tree (2453 pass / 0 fail). Note your own caveat still applies: whichever of this and #1005 lands second needs a skills bundle regen.

PR #1009, opened 2026-08-14
fix(mcp): report a restricted-mode rejection as its cause, not its fallout

jswir:
Fixes F5 from the BEAVER QA field notes.

## The failure

Ad-hoc `malloy_executeQuery` text runs restricted (no `connection.table(...)`, raw SQL, `import`, `##!` — sound policy). But one forbidden construct produced five diagnostics, three actively misleading: the refused construct leaves names undefined, so the caller read `'ACADEMIC_YEAR' is not defined` at the top of the payload plus generic "verify your Malloy syntax" suggestions, and went hunting for column typos in the warehouse. The restriction also appeared nowhere on the tool surface, so an agent only discovered it by tripping over it.

## The fix

- **`getMalloyErrorDetails` gains a first-priority restricted branch.** The compiler already marks the real cause structurally (`problems[].code: 'restricted-construct-forbidden'` — documented by `restricted_mode.spec.ts`); when present, the payload carries **only** those diagnostics plus one suggestion: these constructs ARE allowed in the package's model files — add the definition there, `malloy_compile` → save → `malloy_reloadPackage` → query by name — and any other diagnostics are fallout, not separate problems. A message sniff covers re-wrapped errors, deliberately narrowed to `cannot be used in a restricted` so a user's own `restricted_data` source name can't suppress a genuine not-found error (test pins this).
- **The restriction is documented on the tool surface**: the `query` param's `describe()` carries the full rule and the model-file alternative; the tool description gets a one-line contract pointer (the 2000-char truncation budget had ~60 chars of headroom, so the full text lives on the param).

## Tests

`error_messages.spec.ts`: cascade suppressed (fallout absent from message and suggestions), sniff path, restricted-looking identifier not hijacked, non-restricted `problems`-bearing errors untouched. `execute_query_tool.spec.ts`: end-to-end classification through `classifyToolError` (verified it forwards `MalloyError` unwrapped with `problems` intact). Full unit suite green (2420 pass).


kylenesbit (approved):
Approving. Smallest and lowest-risk change of the batch, and the narrowing is done carefully.

What I checked:
- The structured branch filters on `problems[].code` containing `restricted`, and the early return replaces the payload rather than appending, so the three misleading downstream diagnostics genuinely disappear from both `message` and `suggestions` instead of being reordered.
- The message sniff is correctly narrower than the structured check: `/cannot be used in a restricted/i` can't be tripped by a user's own `restricted_data` source name, and it only runs when no structured problem matched. Pinned by a test, which is the right place for that guarantee.
- Non-restricted errors that happen to carry `problems` fall through untouched, so this doesn't change any other error's shape.
- Hoisting `RESTRICTED_CONSTRUCTS` into one exported constant is worth it given the two surfaces (the suggestion after the fact, and the `query` param doc before it) — the drift it prevents is the failure mode where a rule missing from the list reads as permission.
- The comment explaining why `given:` is deliberately absent from the list (its `##!` flag is refused first, so the diagnostic is unreachable) while `sql_*` is present is the kind of thing that would otherwise get "helpfully" added later.

The suggestion text points at the right loop (`malloy_compile` -> save -> `malloy_reloadPackage` -> query by name) and says outright that other diagnostics are fallout, which is the part that would have saved the QA session.

All checks green. Merges cleanly with the other five; combined typecheck and server unit suite green (2453 pass / 0 fail).

PR #1010, opened 2026-08-14
fix(create-malloy-package): accept spaced paths, keep the server pin current, bridge the editor

jswir:
Four scaffolder fixes from the Spotify QA session — HANDOFF **CR-7**, **CR-11**, **F6**, and the in-repo half of **CR-2**. Bumps the package to 0.0.7.

## CR-7 — the space-in-path guard refused directories that provably work

The guard's premise was that Publisher absolutizes the model's relative `duckdb.table('data/…')` against the package directory before malloy's `DUCKDB_FILE_PATH_RE` check runs, so any space/accent in the workspace path breaks every model. Measured against the current server, the premise does not hold: the per-package DuckDB sandbox resolves the model's **relative** path via its `workingDirectory`, so workspace-path characters never reach the path parser. Verified end-to-end (install → load → query a CSV through `Environment`/`Package`) under paths containing a space, an apostrophe, and a double quote — and the QA session served a 784 MB CSV from a `YYYY-MM-DD Project Name` path, the exact convention the guard turned away with an incorrect explanation.

The guard now refuses only **control characters** (below U+0020, plus DEL), which genuinely corrupt the shell commands and briefing files the scaffold writes verbatim. Tests inverted accordingly (spaced and accented paths scaffold; a U+0001 path is refused by codepoint; nothing half-written). README's "workspace path" section rewritten to match reality.

## CR-11 — the server pin goes stale with nothing to say so

`SERVER_VERSION` is hand-bumped; the latest scaffolder pinned `0.0.237` while the server was at `0.0.244`. Bumped the pin to `0.0.244` (npm `latest`, verified), and `create-malloy-package-npm.yml` gains a publish gate: fail unless `SERVER_VERSION` exists on npm **and** equals `dist-tags.latest`, with the error naming the exact edit (and the `--host` re-check the pin's comment demands). A scaffolder published ahead of an unpublished pin — the other failure the commit history warns about — fails the same gate.

## F6 — the quickstart assumed port 4000 free

On a machine already running a Publisher, the documented final step died with EADDRINUSE and no guidance. The closing output now prints the alternate-port boot (`npm start -- --port 4100 --mcp_port 4140`, via a shared `portOverrideCommandFor` so this line and AGENTS.md's cannot drift) plus the `.mcp.json` url edit that keeps MCP reachable after the move.

## CR-2, in-repo half — scaffolded packages were broken in the editor out of the box

The VS Code/Cursor Malloy extension resolves relative table paths against the **editor workspace root**; Publisher against the **package directory**; this tool scaffolds the workspace one level above the package — exactly the layout that breaks. The package now gets a generated `<pkg>/malloy-config.json` with an absolute `workingDirectory` (absolute because the extension resolves relative values against its own per-window process cwd — that bug stays upstream), and AGENTS.md + README explain the mismatch and the committing caveat (drop the file and open the editor at the package root instead).

## Verification

All 368 scaffolder tests pass; `check-pack` green (9 files); scaffolded end-to-end into a `2026-08-13 scaffold test` directory — package created, `malloy-config.json` carries the spaced absolute path, generated scripts pin 0.0.244, the port-override line prints.


kylenesbit (commented):
Not approving yet — the Windows CI failure is caused by this PR, though the fix is one line.

## The blocker: a new test stages a filename Windows cannot create

`Cross Platform Tests (windows-latest)` is red on exactly one test, the new one:

```
(fail) scaffold: unservable workspace path > accepts an apostrophe and a double quote
 361 pass / 13 skip / 1 fail
```

`"` is one of Windows' reserved filename characters (`< > : " / \ | ? *`), so `fs.mkdirSync(path.join(tmp, 'jim\'s "data"'))` throws `EINVAL` before `run()` is ever called. You already guard the two control-character tests for exactly this reason:

```ts
const controlCharTest = process.platform === "win32" ? test.skip : test;
```

This new test needs the same treatment, or wants splitting so the apostrophe half (legal on Windows) keeps running and only the double-quote half is POSIX-only. The latter is probably better, since the apostrophe is the character most likely to show up in a real `~/Users/O'Brien` path.

Worth noting the product code is fine here: a Windows user cannot put a `"` in a directory name in the first place, so nothing is untested in a way that matters to users. It is purely test portability.

## The premise reversal is correct — I verified it independently

Because this inverts a guard rather than tightening one, I checked the claim end-to-end myself rather than taking the manual verification on trust. I laid out a workspace at `/tmp/space check/2026-08-13 Project Name/` — your exact convention — with a package whose model is `duckdb.table('data/sales.csv')`, booted the server from it, and queried:

```
PUBLISHER_READY ... environments=1 packages=1 load_errors=0
GET .../models/sales.malloy -> sources: [sales (views: overview)]
POST .../query "run: sales -> overview"
  -> [{"region":"east","total_amount":250,"row_count":1},
      {"region":"west","total_amount":175,"row_count":2}]
```

Correct data. So yes: the model's relative path resolves through the sandbox's `workingDirectory` and workspace-path characters do not reach DuckDB's path parser on the query path. The old guard's premise was wrong and refusing these paths was the worse bug.

## One factual correction: there is still one path that absolutizes

The PR's comments state the general case a bit too strongly — "workspace-path characters never reach the path parser". One code path does reach it. `Package.getDatabaseInfo` builds `source: temp is duckdb.table('<absolute path>')`, and under a spaced path that fails:

```
warn: Could not read package database
  databasePath: data/sales.csv
  error: Invalid DuckDB table path:
    "/tmp/space check/2026-08-13 Project Name/publisher_data/default/sales/data/sales.csv"
    — expected an identifier path, a quoted identifier path, ...
```

The visible consequence is on the databases endpoint. Spaced path:

```json
[{"path":"data/sales.csv","type":"embedded",
  "error":"Error(s) compiling model: ... Invalid DuckDB table path ..."}]
```

Identical workspace copied to `/tmp/nospace/proj` as a control:

```json
[{"path":"data/sales.csv",
  "info":{"name":"data/sales.csv","rowCount":3,
          "columns":[{"type":"number","name":"id"},
                     {"type":"string","name":"region"},
                     {"type":"number","name":"amount"}]},
  "type":"embedded"}]
```

So under a spaced path a scaffolded package loads, serves, and answers queries correctly, but loses row counts and column types for its embedded data files, and the package still reports `load_errors=0`.

This is pre-existing server behaviour, not something this PR introduces, and it degrades gracefully (the entry carries `error` instead of `info`, exactly as that catch block intends). It is **not** a reason to keep the old guard — losing a schema panel is far better than refusing the workspace. But it is worth recording, because every package this scaffolder generates uses `duckdb.table('data/…')`, so every scaffolded workspace under a spaced path hits it, and the current comments would tell the next reader it cannot happen. A server-side fix (quote the literal in `getDatabaseInfo`) would close it properly.

## The rest looks good

- The npm publish gate is `workflow_dispatch`-only, so it cannot redden unrelated PRs, and the pin is genuinely current — `npm view @malloy-publisher/server dist-tags.latest` returns `0.0.244`. The `tail -n 1` handling of npm's stderr warnings and the untrusted-output prefixing in the error path are both the right kind of paranoid. Do be aware it couples the scaffolder release to server-release ordering: a dispatch that lands before the server publish completes fails the gate, which is the intended loud failure but will need the release order respected.
- `portOverrideCommandFor` / `resetPortOverrideCommandFor` sharing one `withAlternatePorts` fixes the real drift risk, and keeping `--` on the npm-script form only is correct.
- `malloy-config.json` using `path.resolve(packageDir)` (rather than trusting `options.cwd` to be absolute) is the right defensive choice given `scaffold()` is exported, and gitignoring it unanchored is right for a file carrying a machine-specific absolute path.

kylenesbit (approved):
Approving. The substance of this PR is right, and I verified its central claim independently rather than trusting it (details in my earlier comment): a workspace at `/tmp/space check/2026-08-13 Project Name/` loads, serves, and returns correct query results, so the old guard's premise was genuinely wrong and refusing these paths was the worse bug.

Approving with one **must-fix before merge**, since it is currently what makes CI red.

## Must fix: the new double-quote test cannot run on Windows

`"` is a reserved filename character on Windows (`< > : " / \ | ? *`), so `fs.mkdirSync` throws `EINVAL` before the assertion is reached:

```
(fail) scaffold: unservable workspace path > accepts an apostrophe and a double quote
error: EINVAL: invalid argument, mkdir 'D:\Temp\cmp-scaffold-gDmCJI\jim's "data"'
 361 pass / 13 skip / 1 fail
```

Splitting the case is better than skipping it wholesale, because the apostrophe is the character most likely to appear in a real path (`~/Users/O'Brien`) and it is perfectly legal on Windows:

```ts
test("accepts an apostrophe", () => {
   const quoted = path.join(tmp, "jim's data");
   fs.mkdirSync(quoted);
   expect(run({ cwd: quoted }).packageCreated).toBe(true);
});

// Windows forbids `"` in a filename, so the hazard cannot be staged there.
controlCharTest("accepts a double quote", () => {
   const quoted = path.join(tmp, 'jim "data"');
   fs.mkdirSync(quoted);
   expect(run({ cwd: quoted }).packageCreated).toBe(true);
});
```

`controlCharTest` already exists in the file for exactly this purpose; it may be worth renaming it to something like `posixOnlyTest` now that it guards more than control characters.

Nothing about the product code is at stake here — a Windows user cannot create such a directory in the first place — so this is purely test portability.

## Please also correct the overstated comment

Worth doing while you are in the file, because the comments will otherwise mislead the next reader. `Package.getDatabaseInfo` does absolutize and does reach DuckDB's path parser, so "workspace-path characters never reach the path parser" is true of the query path but not in general. Under a spaced path the databases endpoint returns `error` instead of row counts and column types, while status still reports `load_errors=0`:

```
warn: Could not read package database
  error: Invalid DuckDB table path: "/tmp/space check/.../data/sales.csv"
```

Pre-existing server behaviour, degrades gracefully, and not a reason to keep the old guard — but every package this scaffolder generates uses `duckdb.table('data/…')`, so every scaffolded workspace under a spaced path hits it. Narrowing the claim to the query path (and noting the `getDatabaseInfo` exception as a known server-side gap worth a follow-up) would keep this honest.

## The rest is good

The publish gate is `workflow_dispatch`-only so it cannot redden unrelated PRs, and the pin is current — `npm view @malloy-publisher/server dist-tags.latest` returns `0.0.244`. The `tail -n 1` handling of npm's stderr warnings and the untrusted-output prefixing are both appropriately paranoid; just respect the release ordering, since a scaffolder dispatch landing before the server publish will fail that gate by design. `portOverrideCommandFor` / `resetPortOverrideCommandFor` over one `withAlternatePorts` removes the real drift risk, and `path.resolve(packageDir)` for the generated `malloy-config.json` is the right defensive choice given `scaffold()` is exported.

PR #1011, opened 2026-08-14
docs: self-contained server quickstart, honest OpenAPI descriptions, and a skills version stamp

jswir:
Documentation fixes from the Spotify QA session — **CR-1**, the residue of **CR-9**, **CR-10**, **F6** — plus the in-repo half of **F3**.

- **`packages/server/README.md`** gains a self-contained quickstart at the top: the scaffolder one-liner (`npm create @malloy-publisher/malloy-package@latest`) and the hand-rolled path (config + package dir + `npx @malloy-publisher/server --server_root . --watch-env <env>`), with a minimal config inline. This README ships in the npm tarball; the repo-root README it linked to does not, so that link was dead for every npm consumer — and this file was the QA session's first stop, where the authoring command was findable only by grepping the minified bundle. It now states the three facts a first config gets wrong: the automatic per-package `duckdb` sandbox (no `connections` entry needed; the name is reserved and an env-level `duckdb` fails the environment), the `./`-prefix rule for local `location`s, and copy-vs-watch serving. Also links the previously-unreferenced duckdb example config and corrects the stale aws-sdk note.
- **`api-doc.yaml`**: `Package.location` states the actual local-path rule (`./`, `../`, `~/`, `/`; a bare name is silently not local) and the copy-vs-watch-mount semantics; `Connection.name` documents the reserved `duckdb` sandbox name. Both facts were previously learnable only from `dist/server.mjs`.
- **`docs/configuration.md`**: one paragraph making the copy-semantics trap explicit — without `--watch-env`, a local package serves its boot-time copy and source edits are never read. (Per the QA follow-up decision: this ships as docs/skills prose, not a server-side warning.)
- **Root `README.md`**: the `--data` snippet gets a three-line sample CSV so the expected shape is copy-pasteable (F6's second half; the port half landed in #1010).
- **`packages/skills/scripts/copy-skills.ts`**: stamps each `SKILL.md`'s frontmatter with the package version at pack time, so a stale installed copy is identifiable on disk — F3's in-repo half (a QA session followed installed skills that named tools the server no longer serves, with nothing saying how old they were). Pack-time only: the repo's source skills stay unstamped so the upstream-sync copies remain byte-identical in both directions. Skills package tests pass (7/7; 30 skills stamped in a local run).

**Deferred, outside this repo** (recorded so they aren't lost): publishing a deprecated `@malloydata/publisher` npm stub pointing at `@malloy-publisher/server` (CR-14, one `npm deprecate` command), and the VS Code extension's relative-`workingDirectory` resolution bug (CR-2's upstream half; the scaffolder-side bridge landed in #1010).


kylenesbit (commented):
Not approving yet. The Windows CI failure is caused by this PR, and unlike a flaky test it breaks the skills package build on that platform outright.

## The blocker: the version stamp assumes LF line endings

`Cross Platform Tests (windows-latest)`:

```
$ bun run copy-skills
Cannot stamp 30 of 30 skills with version 0.1.4:
  malloy (no frontmatter block), malloy-analysis (no frontmatter block),
  ... (all 30) ...
error: script "copy-skills" exited with code 1
error: script "test" exited with code 1
```

All 30 skills, not a subset, which is the tell. The frontmatter detection is:

```ts
const close = text.startsWith("---\n") ? text.indexOf("\n---", 4) : -1;
```

There is no `.gitattributes` at the repo root normalising line endings (the only one is scoped to `packages/create-malloy-package`), so on a Windows checkout the skills come out CRLF and the file starts `---\r\n`. `startsWith("---\n")` is then false, `close` stays `-1`, every skill is classified "no frontmatter block", and the `unstampable.length > 0` guard exits 1.

Reproduced directly against your logic:

```
LF   -> close index: 38  ok
CRLF -> close index: -1  UNSTAMPABLE (fails)
```

Because `copy-skills` is part of that package's `test` and `build`, this is not just a red check: a Windows contributor can't build or pack the skills package at all. The fix is to make both ends CRLF-tolerant, e.g. match `/^---\r?\n/` and search for `\r?\n---`, and preserve whichever newline the file actually uses when inserting the `version:` line so a CRLF file doesn't end up with one stray LF inside its frontmatter.

Ironically the fail-closed design is what surfaced this — a silent skip would have shipped unstamped skills on Windows and nobody would have known, which is the exact staleness problem the stamp exists to expose. The guard is right; only the parsing is wrong.

## Everything else looks good

- The two example configs added to `files` in `packages/server/package.json` do exist (`publisher.config.example.duckdb.json`, `publisher.config.example.bigquery.json`), so the README's links resolve inside the tarball rather than pointing at repo paths npm consumers never receive. That was the actual CR-1 defect and this is the right fix for it.
- Refusing to append a second frontmatter block, and failing the pack rather than emitting a duplicate `version:` key, is the correct trade: a duplicate-key YAML error would drop the skill from a strict host entirely, which is worse than staleness. The reasoning about upstream owning these files is sound.
- Keeping the stamp pack-time only, so the repo's source skills stay unstamped and the upstream sync stays byte-identical in both directions, is the detail that makes this safe to add at all.
- The `api-doc.yaml` additions (the local-path `./`/`../`/`~`//` rule with the silently-not-local bare-name case, and the reserved `duckdb` connection name) document things that were previously only discoverable from `dist/server.mjs`, and they don't overlap textually with the `api-doc.yaml` edits in #1006 or #1008 — all six branches merge cleanly together.

Worth flagging for the batch: this is the second of the six with a Windows-only failure, both from new code assuming POSIX (see #1010, where a test stages a filename Windows forbids). Nothing systemic, but the cross-platform job is earning its keep here.

kylenesbit (approved):
Approving. The documentation changes are all correct and the pack-time stamp is a good idea, well-scoped. Approving with one **must-fix before merge**, since it currently breaks the skills package build on Windows outright rather than just reddening a check.

## Must fix: the stamp assumes LF line endings

There is no `.gitattributes` at the repo root (the only one is scoped to `packages/create-malloy-package`), so a Windows checkout produces CRLF and every `SKILL.md` starts `---\r\n`. `startsWith("---\n")` is then false for all 30 files:

```
Cannot stamp 30 of 30 skills with version 0.1.4:
  malloy (no frontmatter block), ... (all 30) ...
error: script "copy-skills" exited with code 1
error: script "test" exited with code 1
```

Reproduced against your logic in isolation:

```
LF   -> close index: 38  ok
CRLF -> close index: -1  UNSTAMPABLE (fails)
```

Because `copy-skills` is part of that package's `test` and `build`, a Windows contributor cannot build or pack the skills package at all.

The fix needs to be tolerant at both ends, and to preserve whichever newline the file actually uses, or a CRLF file ends up with one stray LF inside its frontmatter:

```ts
const open = /^---\r?\n/.exec(text);
const close = open ? text.search(/\r?\n---/) : -1;
if (close === -1) {
   unstampable.push(`${entry.name} (no frontmatter block)`);
   continue;
}
const newline = open![0].endsWith("\r\n") ? "\r\n" : "\n";
// ...
fs.writeFileSync(
   skillFile,
   `${text.slice(0, close)}${newline}version: ${version}${text.slice(close)}`,
);
```

Pinning `SKILL.md` to LF via a root `.gitattributes` would also work and is arguably tidier, but it is a broader change to the repo and does not protect the script from a CRLF file arriving some other way (an upstream sync, an editor). I would do the parser fix regardless. A test that feeds the stamper CRLF input would keep this from regressing, since the Windows job is the only thing that currently catches it.

Credit where due: your fail-closed guard is what surfaced this at all. A silent skip would have shipped unstamped skills on Windows and nobody would have known, which is precisely the invisible-staleness problem the stamp exists to remove. The guard is right; only the parsing is wrong.

## The rest is good

- Both example configs added to `files` in `packages/server/package.json` do exist, so the README's links now resolve inside the tarball instead of pointing at repo paths npm consumers never receive. That was the actual CR-1 defect and this is the correct fix for it.
- Refusing to append a second frontmatter block, and failing the pack rather than emitting a duplicate `version:` key, is the right trade: a duplicate-key YAML error would drop the skill from a strict host entirely, which is worse than staleness.
- Keeping the stamp pack-time only, so the repo's source skills stay unstamped and the upstream sync stays byte-identical in both directions, is the detail that makes this safe to add.
- The `api-doc.yaml` additions document things previously discoverable only from `dist/server.mjs`, and they do not overlap textually with the `api-doc.yaml` edits in #1006 or #1008 — all six branches merge cleanly together, typecheck clean, and the combined server unit suite is green (2453 pass / 0 fail).

PR #1012, opened 2026-08-14
docs(skills): finish the shared-set sync and retire the em-dash allowlist

jswir:
The reverse leg of the shared-skill reconciliation started in [#1005](https://github.com/malloydata/publisher/pull/1005) and [ms2data/agent-skills#84](https://github.com/ms2data/agent-skills/pull/84). #1005 edited four shared skills here first, against the upstream-first rule; reconciling that turned up a wider drift across the two trees. Eight shared files where this repo was ahead travel upstream in agent-skills#84. This PR is everything that travels the other way, plus the two loose ends that pulling on it exposed.

## 1. Cardinality-gated doc values (`malloy-lookml-review/reference/document.md`)

A `#(doc)` for a categorical field should list its values only while the list stays short (roughly ten or fewer). Past that, listing them crowds out the meaning and goes stale as the data changes, so describe what the field captures instead. The line it replaces said "and valid values for categorical fields" with no ceiling, which is how a doc string ends up mostly enum.

## 2. The em-dashes, and the allowlist that held them

`skills_bundle.spec.ts` banned em-dashes with a named allowlist for three shared reference files and a comment saying the fix was "an upstream-first change of its own, not something to bundle into the build-time asset that surfaced it." That change has now been made upstream, so this copies the rewritten files back and deletes both the allowlist and the test that policed it. A new em-dash anywhere in the bundle now fails outright, with no exception list to keep in sync.

23 em-dashes across `lazy-load.md`, `verification-harness.md` and `build-derived-tables.md`, each replaced by the colon, comma, semicolon or parentheses the sentence already wanted. No meaning moves.

This was worth doing rather than working around, because the punctuation was actively costing something: the two repos have different rules for it, only this one enforces its rule in CI, and a straight verbatim copy of an upstream file in this direction fails the build. That is a trap for every future sync, and the character carries nothing semantic. `skills/README.md` loses its own seven for the same reason: the document that teaches the sync rules should not model the banned punctuation.

Untouched upstream: `credible-*` and `eval-*` skills and the root `malloy` index (~700 lines with em-dashes). None of those upstream copies is shared, and none of them reaches this bundle.

## 3. The same gate on `malloy-document`

Fixing (1) in the lookml-review reference left the identical sentence standing in `malloy-document`, the skill that actually owns `#(doc)` writing: "Include units (USD, count, percentage) and valid values for categorical fields", no ceiling. Both ship in the same MCP bundle, so an agent asking how to write a doc string got one answer that caps the list at roughly ten and one that does not. #1005 extends that same bullet without adding the gate, so neither PR was going to close it on its own.

The `Include units` line here is already the hand-merge of the two edits: the gated categorical bullet stays as its own line, and #1005's counts sentence (name the noun counted; distinct entities vs events) is appended to units. A later conflict on that markdown resolves to this branch's text.

The rule already has a home upstream, in `credible-document`, which states the gate and names `#(index)` as the mechanism for a long value space. `malloy-document` is this repo's Publisher-facing counterpart and is not synced, so it is brought in line here directly; the `#(index)` half stays out, since that annotation never appears in this tree.

The duplication itself is structural and stays. `malloy-lookml-review` is a shared skill and may only `skill:`-reference other shared skills, so it cannot point at Publisher-specific `malloy-document`. Two statements of one rule is the cost of that boundary; two statements that disagree is not.

## 4. What the em-dash check actually checks

Retiring the allowlist in (2) left one assertion, `expect(offenders).toEqual([])`, running against a tree that now has no em-dashes anywhere. That passes for free: if the predicate ever stopped detecting, through a renamed field, a different codepoint, or an empty `skills`, the test would stay green and the house style would go unenforced. The deleted allowlist test ruled that out incidentally, by asserting the listed entries still held one. Two positive controls restore the property without the list, one per field the predicate reads: an em-dash in `body` with an empty `description`, and the reverse. Dropping either clause now fails that control rather than leaving the suite green.

The check also read `body` only. A `SKILL.md`'s frontmatter `description` is not part of its body, and it ships as the MCP prompt's description, so it is bundle text like any other and an em-dash there reached a host unchallenged. Both fields are checked now. Verified by injecting an em-dash into a description alone and confirming that `carries no em-dashes` fails, rather than only the sync test.

The tree is clean on both fields today, so this closes a hole rather than fixing a live violation.

## After #1005, this, and agent-skills#84 merge

Every shared file is byte-identical across both repos. Verified file by file, `reference/` files included, not just `SKILL.md`.

**#1005 is a prerequisite for that claim.** Diffing this branch's `skills/` against agent-skills#84 leaves twelve shared files still differing: eleven of them are carried by #1005, and the twelfth, `malloy/SKILL.md`, is correctly divergent, being one of the Publisher-specific host/router entry points that is never synced.

**Merge order matters.** #1005 and this PR both rewrite the single-line `skills_bundle.json`. Whichever lands second conflicts there, and the resolution is to regenerate the bundle, not to hand-merge the line. The `malloy-document` source conflict is already resolved on this branch (gated list + #1005's counts sentence); a conflict there resolves to this text.

`bun test src/mcp/skills` and `packages/skills`: 16 and 13 pass (one test removed with the allowlist, two positive-control assertions on the em-dash check). The 13 is the skill-version stamping suite this branch picked up in its merge from main, not a discrepancy. `skills_bundle.json` regenerated.


Sha-Bang (approved):

packages/server/src/mcp/skills/skills_bundle.spec.ts
```diff
@@ -137,35 +137,26 @@ describe("skills_bundle.json (generated dual-channel asset)", () => {
...
+      const hasEmDash = (s: SkillEntry) =>
+         s.body.includes("—") || s.description.includes("—");
+
+      // Positive control. With the allowlist gone this assertion runs against
+      // a clean tree, so it would pass for free if the predicate ever stopped
+      // detecting; the deleted allowlist test proved that incidentally, by
+      // asserting the listed entries still held one.
+      expect(
+         hasEmDash({ name: "control", description: "", body: "a — b" }),
+      ).toBe(true);
```
Sha-Bang, line 160: Should fix: this control only exercises the `body` clause. Its `description` is `""`, so the half of the predicate this PR adds is the half the control does not cover.

Dropping the description check back out, i.e. reverting exactly what section 4 closes:

```ts
const hasEmDash = (s: SkillEntry) => s.body.includes("\u2014");
```

leaves the suite green:

```
bun test src/mcp/skills/skills_bundle.spec.ts
 10 pass
 0 fail
```

That is the same pass-for-free the retired allowlist test used to rule out, moved to the other field. It will not be caught incidentally either, since no description in the tree currently holds an em-dash.

One more line closes it:

```ts
expect(
   hasEmDash({ name: "control2", description: "a \u2014 b", body: "" }),
).toBe(true);
```

The other two failure modes the comment names do hold, for what it is worth: a changed codepoint fails the control, and a renamed field fails the `Typecheck` CI job, since the literal is structurally typed against `SkillEntry`.

skills/malloy-document/SKILL.md
```diff
@@ -39,7 +39,8 @@ measure:
 Doc strings power natural-language search: users type plain-English questions and the system matches against your `#(doc)` strings. Write descriptions that match how analysts would search:
 
 - **Include business meaning**, not code mechanics: what it represents, not how it's implemented
-- **Include units** (USD, count, percentage) and valid values for categorical fields
+- **Include units** (USD, count, percentage): a unit is part of what a number means
+- **List a categorical field's values only while the list stays short** (roughly ten or fewer). A handful of values makes a description concrete; past that, say what the field captures instead, because the dump crowds out the meaning and goes stale the moment someone adds a value. Treat ten as a rule of thumb, not a hard cap.
```
Sha-Bang, line 43: Should fix: #1005 rewrites this same bullet, appending a sentence about naming the noun counted and whether the count is of distinct entities or of events. Two different replacements for one line, so whichever lands second conflicts here in the source markdown, not only in `skills_bundle.json`.

"Merge order matters" in the description covers the JSON and says the resolution is to regenerate rather than hand-merge the line. True for the JSON; this one does need a hand-merge, keeping the gated bullet and re-appending #1005's counts sentence to the `Include units` line, and it has to happen before the bundle is regenerated.

Sha-Bang (commented):
Reviewed the em-dash check, the sync claims, and the rewritten prose against a checkout of the branch.

The bundle is genuinely regenerated in sync (the `is in sync with skills/` test passes on the branch), and the em-dash arithmetic matches the diff line for line: 8/9/6 across the three reference files is exactly the 23, and those three plus `skills/README.md` were the only files anywhere under `skills/` carrying U+2014 before this PR. The tree really is clean afterward.

Three inline notes. On the description itself:

Should fix: the test counts are stale. `bun test src/mcp/skills` gives 16 as claimed, but `packages/skills` gives 13, not 7 (run through its own `bun run test`, so the `copy-skills` prestep happens). The extra tests came in with the skill-version stamping this branch picked up in its merge from main. Everything passes, so nothing is broken; the figure just sends a reviewer chasing a discrepancy that is not there.

Nit: `skills/README.md` carried seven em-dashes, not eight.

Nit: "Untouched upstream: `credible-*` and `eval-*` skills" holds for `eval-*` only because no such directory exists. `packages/skills/scripts/exclusions.ts` and `build_skills_bundle.ts`'s `isCredible()` special-case the `credible-` prefix alone, and `.gitignore` likewise lists only `credible-*`. There is no `eval-*` equivalent anywhere. Nothing to change in this PR, but the sentence reads as a structural guarantee that is not actually in place.

skills/malloy-html-data-apps/reference/verification-harness.md
```diff
@@ -64,12 +64,12 @@ page.on("console", (m) => m.type() === "error" && errors.push(m.text()));
...
 await page.goto("http://localhost:4173/index.html", { waitUntil: "load" });
 // publisher.js holds an SSE stream open (even with watch off), so networkidle NEVER fires.
-// Wait on CONTENT — and for ALL tiles to RESOLVE, not just the first to appear.
+// Wait on CONTENT, and for ALL tiles to RESOLVE, not just the first to appear.
 // Asserting after only the first .value renders races the others and yields a
 // false "stuck skeleton" / "empty value" FAIL. "Resolved" = every tile has a
 // value and no skeleton remains. This single wait subsumes the DELAY_MS delay
 // AND the stuck-skeleton check; if it times out, a tile really is stuck.
-// NOTE: assumes a NON-lazy page — every tile loads on open. For a lazy-loaded
+// NOTE: assumes a NON-lazy page, every tile loads on open. For a lazy-loaded
```
Sha-Bang: Nit: comma splice. `assumes a NON-lazy page, every tile loads on open` joins two independent clauses with a bare comma, where the em-dash was carrying the break. A colon or semicolon does the job.

This is the only one of the 23 rewrites where the replacement reads worse than the original; the rest are meaning-preserving and several are clearer for the explicit "so" or "because".

Sha-Bang (approved):
Approving. The sync is real, the bundle is regenerated in sync, and the em-dash arithmetic matches the diff exactly.

The one-line second positive control is worth adding before merge, since without it the description half of the predicate is unguarded, but it is a test-coverage gap rather than a behaviour bug and I would not hold the PR for it. Do watch the `malloy-document/SKILL.md` conflict with #1005: that one needs a hand-merge of the two bullet edits before the bundle is regenerated.

packages/server/src/mcp/skills/skills_bundle.spec.ts
jswir: Added. The second control puts the em-dash in `description` with an empty `body`, so dropping the description clause now fails that assertion rather than leaving the suite green.

skills/malloy-document/SKILL.md
jswir: Pre-merged on this branch: the gated categorical bullet stays, and #1005's counts sentence is appended to the `Include units` line. Bundle regenerated after that, so a later conflict in `skills_bundle.json` is still regenerate-don't-hand-merge.

skills/malloy-html-data-apps/reference/verification-harness.md
jswir: Replaced the comma with a colon.

PR #1013, opened 2026-08-14
feat(server): compile scopes — validate an edit pre-save, and dry-run the package

jswir:
Implements the compile-loop audit's recommendations: validate edits before saving, dry-run package-wide breakage, and document the authoring modes.

## The gap

`/compile` previously appended submitted text to a model. That validates new definitions, but an edit to an existing definition collides with the model's own copy (`Cannot redefine`), and diagnostics use concatenated-file coordinates. The only reliable edit check was save-then-reload.

## Compile scopes

- **`append`** (default): source appended to `modelPath`, for new definitions and queries.
- **`file`**: source compiled as `modelPath`, replacing its on-disk content for the check. This validates an edit before saving at true file coordinates.
- **`package`**: reload's worker compiler runs over every `.malloy` and `.malloynb` file selected by reload, without swapping the result into the served package. An optional source replaces `modelPath`, so importer breakage surfaces before save. Dotfiles are ignored like reload, a non-exact replacement path is warned and treated as new, and worker admission/timeout behavior bounds package compilation.

Every diagnostic carries **`model`**, the package-relative file it points at, on REST and in MCP's flattened response.

## Guard rails and disclosure trade

Caller `#(authorize)` annotations remain rejected on every source-bearing scope. Both the early text gate and compiled runnable backstop apply. A model file added since the last load gets an ephemeral disk-backed gate model, while a genuinely new path evaluates gates carried by its compiled runnable; neither path can skip authorization. Authorization denials for genuinely new paths are masked as not-queryable.

Package scope intentionally reports compiler diagnostics from all reload-selected files, including files hidden from discovery. It returns no rows or SQL, and caller-provided text remains authorize-gated. This is an accepted authoring-surface disclosure trade, documented in the API and agent guidance.

## Validation

- Full server unit suite: **2485 passed, 3 skipped, 0 failed**
- Repository typecheck: passed
- Repository lint: passed
- MCP protocol description budget and generated skill bundle parity: passed


packages/server/src/service/environment.ts
```diff
@@ -574,7 +646,7 @@ export class Environment {
          // compile below. If the model isn't loaded, there's nothing to enforce
          // and compilation surfaces its own error.
          const gateModel = pkg.getModel(modelName);
-         if (gateModel) {
+         if (gateModel && source !== undefined) {
```
Sha-Bang: Should fix: this guard makes the authorize gate fail open for any `modelPath` that is not a loaded model, and this PR turns that from a latent edge into a supported, tested, documented path.

`Package.getModel` (`package.ts:1445`) returns `undefined` for anything not in the loaded model map. Both this early text gate and the compiled backstop at `:801-806` are guarded on `gateModel`, so when it misses, no authorize enforcement runs at all.

Concrete: package `hr` has `payroll.malloy` carrying `#(authorize) "$ROLE = 'admin'"` on `source: payroll`.

```
POST /api/v0/environments/e/packages/hr/models/scratch.malloy/compile
{"scope":"file","includeSql":true,
 "source":"import \"payroll.malloy\"\nrun: payroll -> { aggregate: total_comp }"}
```

`scratch.malloy` does not exist, so `gateModel` is `undefined`. The compiled `payroll` structDef still carries the author's annotation, but nothing looks at it, and the response is a 200 with the gated source's SQL. The same request against a real model path is correctly denied.

The guard is unchanged from main, so the hole is pre-existing rather than introduced here. What changes is that a missing `modelPath` is now a blessed path: `compile_scopes.spec.ts:94` pins "a brand-new file compiles", `compile_tool.ts:209-210` tells the agent "a missing modelPath does NOT error here", and the new docs teach it. #1008's stated premise was that `#(authorize)` still gates compile in full; on this path it does not.

Direction: when `pkg.getModel(modelName)` misses, fall back to a package-level gate rather than skipping enforcement, so the miss fails closed.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+               recursive: true,
+            })) as string[];
+            const modelFiles = entries
+               .filter(
+                  (entry) =>
+                     entry.endsWith(MODEL_FILE_SUFFIX) &&
+                     !entry.endsWith(NOTEBOOK_FILE_SUFFIX),
+               )
+               .map((entry) => path.join(packagePath, entry))
+               .sort();
```
Sha-Bang: Should fix: this enumeration diverges from what a reload actually compiles, in both directions, and the comment just above it claims otherwise.

The real load path is `recursive(packagePath, [ignoreDotfiles])` (`package_load_worker.ts:464`, `package.ts:1886`) followed by `filterModelPaths`, which keeps both `.malloy` and `.malloynb` (`package_load_worker.ts:470-474`). So:

Notebooks are excluded here and compiled by reload. A `.malloynb` broken by the what-if edit passes the dry-run and fails the reload. The tool description, `docs/ai-agents.md`, `AGENTS.md` and the skill all describe package scope as having "reload's reach", and it does not. It also skips the loader's non-compile validations (`assertPersistNamesQuoted`, render-tag validation, explores checks), so a `#@ persist name=` quoting error still surfaces only at reload.

Dotfiles and dot-directories are walked here and skipped by reload. `utils.ts:16-24` documents why the loader filters them: they "have been a source of spurious ENOENTs when their contents disappear mid-scan". This `readdir` is not inside any `try`, so an ENOENT from `.git` churn under a watch-mounted package, where `publisher_data/<env>/<pkg>` symlinks into a live git working tree, rejects the whole request as a 500. That contradicts the code's own comment at `:784-785`: "A file the dry-run cannot even read is a finding, not a 500."

Reusing `recursive(packagePath, [ignoreDotfiles])` and wrapping the enumeration fixes both halves at once.

Nit, same block: `!entry.endsWith(NOTEBOOK_FILE_SUFFIX)` on line 735 is dead. `".malloynb".endsWith(".malloy")` is `false`, so the first predicate already excluded them.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+            // A what-if source for a NEW file is still part of the dry-run.
+            if (source !== undefined && !modelFiles.includes(modelPath)) {
+               modelFiles.push(modelPath);
+            }
```
Sha-Bang: Should fix: this and the substitution test at `:771` are both exact string comparisons, and when `modelPath` does not match an enumerated path the result is a clean pass rather than an error.

A typo, or a case difference on macOS or Windows (`Tracks.malloy` against `tracks.malloy`), gives you both halves of the wrong answer at once: `filePath === modelPath` is false at `:771`, so the real file compiles from disk with its old content, and the caller's what-if is appended here as a phantom extra entry, which compiles standalone and cleanly.

Concrete: an agent renames `base_source` to `base_renamed` and asks for a package dry-run with `modelPath: "Base.malloy"`. Every importer compiles against the unrenamed `base.malloy` on disk, so the response is `status: "success"`. The agent saves, reloads, and `tracks.malloy` breaks. That is precisely the failure this scope exists to catch, reported as a pass, with no signal of any kind.

Direction: when `source` is given at package scope and `modelPath` is not among the enumerated files, emit a warn-severity diagnostic saying it was treated as a new file and no existing file was replaced. A genuine new-file dry-run still works; a typo becomes visible.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+               // The runnable backstop applies only to the CALLER's text (the
+               // what-if replacement); the package's own files are author
+               // content, gated at query time like a reload's. Outside the
+               // compile try so its denials propagate as denials.
+               if (
+                  source !== undefined &&
+                  filePath === modelPath &&
+                  gateModel &&
+                  materializer
+               ) {
```
Sha-Bang: Should fix, though possibly only with a comment and a doc sentence rather than code: "Guard rails (unchanged posture)" in the description is not accurate for package scope.

With no `source`, `source !== undefined` is false both here and at the early gate on `:649`, so zero gating runs. Every `.malloy` file on disk is compiled and every diagnostic comes back tagged with its package-relative path, including files hidden by `explores` / `queryableSources: "declared"` and files whose sources are `#(authorize)`-gated.

Concrete enumeration: `POST .../models/dashboards.malloy/compile {"scope":"package","source":"<edit renaming a symbol a widely-imported file exports>"}`. Every importer errors, and the response names each one by package-relative path, including files `GET .../packages/{pkg}/models` deliberately does not list, with error text quoting identifiers from `#(authorize)`-gated files. #1008 went out of its way to make hidden indistinguishable from nonexistent on `/compile`, converting a hidden target's 403 into a flat 404; this reaches the same information by another door.

I do not think it blocks on its own. No rows and no SQL cross the boundary, and #1008 already states the boundary is discovery curation rather than access control. But it is a posture change, and it deserves what #1008 gave its own trade: an explicit accepted-trade comment at this branch and a sentence in `api-doc.yaml`'s `scope` description. As written the PR body asserts the opposite, and the `#(authorize)` claim holds only for caller text, and only when `gateModel` exists.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+                  const key = `${problem.model ?? ""}|${start?.line ?? -1}|${
+                     start?.character ?? -1
+                  }|${problem.severity}|${problem.message}`;
+                  if (seen.has(key)) continue;
+                  seen.add(key);
+                  problems.push(problem);
+               }
+            };
+
+            for (const filePath of modelFiles) {
```
Sha-Bang: Should fix: this loop is the worker dual-path concern inverted. `package_load_worker.ts:1-46` says the whole-package compile lives in a worker deliberately, being "intentionally pure-CPU" and kept off the main isolate. `scope: "package"` does N synchronous Malloy compiles in-process, inside `withPackageLock`, with no `assertCanAdmitQuery()` (never called on `/compile` at all, `server.ts:1825-1846`), no cap on N, no per-request time bound, and N retained compiled `ModelDef`s on one `Runtime` with no memory-governor consultation.

On an unauthenticated localhost endpoint that makes a 200-byte POST into a full package compile plus N warehouse schema fetches, blocking the event loop and holding the package mutex, so reloads and installs queue behind it.

Routing package scope through the existing load worker would fix this and the enumeration divergence at `:728` in one move. Failing that, `assertCanAdmitQuery()` plus a file-count bound covers the amplification.

Sha-Bang (commented):
Reviewed the single commit `42efc5df` on top of `32b7f2cd`, by reading code rather than re-running the suite, so treat the judgements below as static analysis. All 18 checks are green.

First, the banner is stale: #1008 is already merged and in this PR's history (`d7e32c0a`), and this already targets `main`, one commit behind. The "merge #1008 first, then retarget" note at the top of the description should come out before anyone acts on it.

The parts I have no complaint about, since they are the parts that would have been easy to get wrong: `assertSafeRelativeModelPath` blocks `..`, backslash, NUL, absolute and leading-dot segments, `safeJoinUnderRoot` does resolve-and-contains, Node's recursive `readdir` does not descend symlinked directories, and `bodyParser.json` caps the body at 1 MB. The `#(authorize)` forgery gate is applied on every scope, and #1008's `denyHiddenAsNotQueryable` mask is correctly reused at the new package-scope backstop. `file`-scope coordinates involve no offset arithmetic at all, since the submitted text simply is the file, so there is no off-by-one to get wrong across REST, the MCP flattening, and the text block. The `model` tag cannot leak an absolute path: `path.relative` plus the `!rel.startsWith("..") && !path.isAbsolute(rel)` check covers the POSIX and the Windows cross-drive case. `scope` is optional and defaults to `append` on both wires, `api-doc.yaml`'s enum and `CompileProblem.model` match the implementation, and `skills_bundle.json` matches the frontmatter-stripped `malloy-modeling/SKILL.md`.

Inline notes below. Three of them I would want resolved before this merges: the authorize gate skipping entirely when `modelPath` is not a loaded model, the silent clean result when `modelPath` misses an on-disk file at package scope, and the enumeration divergence from what a reload actually compiles.

A few smaller things that have no natural line to sit on:

Nit: "`append` is byte-for-byte unchanged" is not literally true, in three harmless ways. The notebook rejection message lost "for namespace context" (`environment.ts:560`); every diagnostic now carries `model`; and omitting `source` went from appending the literal string `"undefined"` to the model, since main coerced `undefined` through the regex `.test()`, to a clean 400. All three are improvements. The claim just does not need to be that strong.

Nit: `COMPILE_DESCRIPTION` measures 1973 of the 2000 chars enforced by `mcp/server.protocol.spec.ts:83`. 27 characters of headroom means the next wording tweak trips it, and the failure will look unrelated to whatever change caused it.

Nit: `api-doc.yaml` drops `required: [source]` from `CompileRequest` outright, so generated clients now type `source` as optional at every scope, including the two where it is mandatory. The description carries the real rule. An `if`/`then` would encode it, or this is fine as an accepted loosening, but it is a real change to the generated client surface.

Nit: at package scope with no `source`, `modelPath` is unused but still validated and notebook-rejected (`environment.ts:513`, `:557`). An agent holding only a `.malloynb` path gets `Cannot compile against a notebook` for a whole-package dry-run that does not involve that path.

packages/server/src/service/compile_scopes.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+            "tracks.malloy",
+            "run: tracks -> v",
+            false,
+            undefined,
+            "buffer" as never,
+         ),
+      ).rejects.toBeInstanceOf(BadRequestError);
+   });
+
+   it("rejects a caller #(authorize) annotation at file scope too", async () => {
```
Sha-Bang, line 245: Should fix: this pins `assertNoCallerAuthorizeAnnotation` with the simplest spelling, which is the direction that was already safe. The case that matters at file scope is the opposite one: the caller submitting the author's file with the `#(authorize)` line deleted.

That case is currently caught, because the early text gate resolves `run: gated` against the cached model, which still carries the annotation. It is also untested, and one refactor away from breaking.

The append-scope defence-in-depth pinned by `compile_authorize.spec.ts:102` does not transfer either. At file scope the caller can restate the gated declaration without its annotation and put a decoy `run: open_src` first: the first-`run:`-only early gate resolves the decoy, and the compiled backstop finds no gate on the restated struct. That is not a new capability, since it still needs the source's base table and append-scope aliasing already allowed that, but it collapses two gates to one, and `compile_authorize.spec.ts` gained no scope-aware case.

Three worth adding: file-scope resubmission with the annotation deleted expecting `AccessDeniedError`, the decoy-first-`run:` variant, and the `gateModel === undefined` case from the `:648` thread.

packages/server/src/service/compile_scopes.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+      await fs.writeFile(
+         path.join(rootDir, "env", "pkg", "tracks.malloy"),
+         TRACKS_MODEL.replace("n.sum()", "nope.sum()"),
+      );
+      const broken = await compile("base.malloy", undefined, "package");
+      const errors = broken.problems.filter((p) => p.severity === "error");
+      expect(errors.some((p) => p.model === "tracks.malloy")).toBe(true);
+   });
+
+   it("package: does not touch the served model", async () => {
```
Sha-Bang, line 173: Nit: both assertions here are trivially true. `compileSource` never calls `reloadAllModels`, so `getFailedPackages().size === 0` and "a query still answers" hold without the property under test holding. The comment invokes the `stale` flag, which the test does not check.

The code does make the property true, so this is about the test not being able to detect a regression rather than about the behaviour.

packages/server/src/service/compile_scopes.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+      expect(inBase.length).toBe(keys.size);
+   });
+
+   // -- validation ---------------------------------------------------------
+
+   it("rejects a missing source at append and file scope, naming the fix", async () => {
+      for (const scope of ["append", "file"] as const) {
+         await expect(
+            compile("tracks.malloy", undefined, scope),
+         ).rejects.toBeInstanceOf(BadRequestError);
```
Sha-Bang: Nit: `rejects.toBeInstanceOf(BadRequestError)` does not check what the description promises. The claims are "400 naming the valid set" and, in this test's own name, "naming the fix"; neither message is asserted, so the naming could regress silently.

Same for `:189`. Also worth noting there is no REST-level test for `scope` passthrough or for the 400 on an invalid scope, and the MCP passthrough test in `compile_tool.spec.ts` asserts positional `_args[5]`, which rots quietly if the signature gains a parameter.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+                     }
+                  } catch {
+                     // Not a resolvable file URL — leave the tag off.
+                  }
+               }
+               if (
+                  model !== undefined &&
+                  path.basename(model) === "__compile_check.malloy"
+               ) {
+                  model = modelName;
```
Sha-Bang: Nit: this relabels any diagnostic whose basename is `__compile_check.malloy` as `modelName`. At package scope, where the files are author content rather than the virtual entry, a real file with that name gets mislabelled as whatever `modelPath` was.

packages/server/src/service/environment.ts
```diff
@@ -615,6 +687,151 @@ export class Environment {
...
+                        at?: {
+                           range?: {
+                              start?: { line?: number; character?: number };
+                           };
+                        };
+                     }
+                  ).at?.range?.start;
+                  const key = `${problem.model ?? ""}|${start?.line ?? -1}|${
+                     start?.character ?? -1
+                  }|${problem.severity}|${problem.message}`;
```
Sha-Bang: Nit: the key falls back to `""` when `model` is absent, which happens for a diagnostic originating outside the package tree. Two positionless problems with the same message and severity, reached through different out-of-package imports, collapse into one. Unlikely enough to leave alone, but worth a comment saying the empty-model case is knowingly lossy.

Sha-Bang (approved):
Approving on the understanding that the findings get addressed before this merges.

The design is right and the parts that were hard to get right are right: path handling, the forgery gate on every scope, the reuse of #1008's existence-oracle mask, and the absence of any coordinate arithmetic at file scope. Nothing here is architectural.

The three I would not merge without, restated so they are easy to check off:

1. `environment.ts:648` — the authorize gate skips entirely when `modelPath` is not a loaded model, so a compile against a nonexistent path returns a gated source's SQL. Pre-existing on main, but this PR makes it a taught path.
2. `environment.ts:740`/`:771` — a `modelPath` that misses an on-disk file at package scope yields `status: "success"` while the real file compiles from its old content. Silent wrong answer in the headline use case.
3. `environment.ts:728` — the enumeration skips notebooks that reload compiles, walks dotfiles that reload skips, and is unwrapped, so `.git` churn under a watch-mounted package 500s the request.

Also worth dropping the stale "stacked on #1008, merge that first" banner from the description, since #1008 is already merged and this targets main.

PR #1017, opened 2026-08-14
feat(server): let an #(authorize) gate filter rows, not just admit a source

Sha-Bang:
An `#(authorize)` gate is a whole-source boolean, decided by a synthetic one-row DuckDB probe. That probe has no real columns, so a gate can reference neither a row field nor a joined field, and the only answers available are the whole source or a 403. A gate that reads the row it protects — `org_id in $GROUPS` — is the case operators actually have.

This makes such a gate a row filter. `#(authorize) "org_id in $GROUPS"` and `#(authorize) "childtable.name = $BOB"` are now valid and filter rows. A gate that reads no row field is untouched and still decided by the probe.

## Mechanism

The compiled `FilterCondition` is lifted through the model's own materializer, grafted onto a `structuredClone` of the `ModelDef` at the entry point's own `contents` key, loaded via `_loadModelFromModelDef`, and the caller's **unmodified** text recompiled against it. No caller text is ever rewritten, so no parser sits on the security path.

Three cheaper mechanisms were tried first and each returns rows it must not. All were measured against `@malloydata/malloy` 0.0.427, not reasoned about:

- **Appending `+ {where:}`** binds to the *last* pipeline stage, not stage 0. A query that reprojects the gate's column (`-> { group_by: org_id is 1, … }`) filters against a literal and serves every tenant.
- **Grafting onto the `SourceDef`'s `filterList`** reaches `run: X` only — every derivation snapshots its base at compile time, so `Y is X extend {…}`, `Z is X -> {…}` and joined copies come back ungated.
- **Grafting onto `pipeline[0].filterList`** emits the `WHERE` without the `JOIN`, because joins come from `activeJoins`/`expandedFieldUsage` rather than `filterList`: `Binder Error: Referenced table "child_0" not found!`

Each is written out here rather than filed away, because the next person to reach for one will reach for it in a review comment.

## Semantics and posture

Entry-point-only semantics are unchanged: a gate fires for the source the caller named, never for one reached through a join, though a boundary gate may *reference* a joined field. Inheritance still walks derivations until something declares its own gate.

**Gates authored on different sources AND; a single source's own gates OR.** That distinction has to be carried structurally, because Malloy's composite resolution copies a composite parent's annotation note object onto the resolved member, so reading a struct's notes as one list silently turns two sources' gates into a disjunction — a caller entitled to nothing reads rows through a query source over the composite. An entry point's gates are therefore grouped by declaring source and classified independently, and the parent's notes are identity-subtracted before the member's are read.

The posture is a pure row filter, so **a caller a gate admits nowhere gets 200 with zero rows, not 403**. That leaves no boolean to fall back on, which makes fail-closed the only backstop — attach-or-deny returns the runnable to execute, and a caller supplying no recompile hook is denied rather than served, so an unmigrated path fails shut. Rows are protected; the schema is not, stated in `docs/security-posture.md` rather than left to be discovered.

The gate grammar is a positive allowlist read from the compiled IR, not the annotation text, so Malloy has already decided which names are fields and which are givens. `not` is refused: with an empty array given it compiles to `COALESCE(NOT (FALSE),TRUE)` and returns every row, inverting fail-closed into fail-open. A field compared against a given carrying a **declared default** is refused for the same reason — `tenant != $EXCLUDED` at the documented empty default is `WHERE tenant != ''`, which is roughly every row.

## Breaking changes

There is no flag; this ships on. Four of these break packages that exist today, each because today's behaviour is quietly wrong:

- **A colocated `#@ persist` on a gated source is refused.** Not because the gate is lost — measured, a colocated substitution replaces only the source's relation SQL while the gate applies as the reading query's own `WHERE`, so the two compose and rows come back filtered. What it refuses is authorization decided against a **frozen copy of the gating column**: the artifact is built once, so a row that changes hands keeps being served to its former owner, and because the content address does not include the annotation, an artifact built before the gate stays addressable while every rebuild is refused. This fails a build that currently succeeds, and the check's reach is deliberately wider than the hazard — it also refuses a source that merely *joins* a gated one.
- **An `#(authorize)` in a position nothing enforces fails the model load** — a top-level `query:` statement, a field inside a source, or a `join_one:`/`join_many:` line whose declaration resolves inside the model. Today each silently protects nothing: the annotation lands on a node no enforcement path reads, and the source the author meant to gate serves every row.
- **A file-level `##(authorize)` is refused**, its gate belonging on each `source:` it was meant to protect. It was never in the spec. Refusing rather than ignoring is deliberate: dropping support silently would turn a working gate into a no-op.
- **A near-miss `authorize` spelling fails the load.** What counts as the tag is now Malloy's own annotation routing rather than a prefix regex, because the regex diverged from the compiler in both directions and one of them was a live fail-open. `#|(authorize)` — the block form — and `#[authorize]`/`#<authorize>`/`#{authorize}` all route to `authorize`, so an author writing one had a source the compiler agreed was gated, publisher registered nothing, every row served, and the load was clean; worse, such a source was also materialization-eligible, so it could be frozen into an artifact and served to everyone. In the other direction the regex blessed `#( authorize )` and `#(authorize)X`, which Malloy calls `malformed-route` — a branch test pinned the first as a working gate, so narrowing it is the part that breaks a package that works today. `# (authorize)` and `## (authorize)` are route `''`, Malloy's reserved MOTLY/render namespace (the same route as `# bar_chart`), and are currently silent no-ops. All of those near misses now fail the load naming the spelling and the fix. The near-miss rule is keyed on Malloy's route too (`routeOf`/`payloadOf`, public in the pinned version) rather than on a text superset, so another app's deliberate namespace is left alone — an earlier text test refused the whole `#(authorize-v2)` / `#(authorize.audit)` family and `#(authorized)`, failing a model load with advice aimed at someone else. A case-variant of our own name (`#(AUTHORIZE)`) is still refused, since honouring it is not an option and silence is the exact fail-open this exists to catch. Refusing rather than interpreting is deliberate: honouring `# (authorize)` would assign application meaning inside a namespace Malloy reserves, and would silently start enforcing a filter on packages that have been serving every row.
- The 403 → 200-with-zero-rows change **cannot** affect anything existing, since a row-field gate could not be written before this.

**Before merge:** the published-package store should be scanned for real `#(authorize)` usage, to size the blast radius. Grep `#|(authorize)` **first** — those packages are unguarded *and* materialization-eligible right now — then `#(\s+authorize` and `#\(authorize[^)]` (gates that work today and stop being honoured, the set that decides whether this needs a transition release), `#\s+\(authorize` (silent no-ops today, for whom failing the load is strictly an improvement), and a case-insensitive `#\(authorize\)` (case variants, also silent today).

## Decisions and trade-offs

- **Schema disclosure accepted.** A row-gated source is readable-but-empty, so a caller can name it, compile against it, and enumerate columns through compile errors. Resolving a gate out of untrusted text before compiling is the resolution-from-text this design refuses elsewhere.
- **A gate on a join line fails the load only where authorship is provable.** Malloy copies a base's notes onto joining structs by reference, so an inherited note is indistinguishable by value from an authored one. Where the join's declaration resolves inside the model, note-object identity settles it and a mismatch is fatal. Where it resolves nowhere — a selective or two-hop import — it is let through silently, because every legitimate cross-file join of a gated source reaches that branch and a warning there fires on correct packages. The residual cost is stated at the branch: a gate typed on such a join line goes unenforced with no signal.
- **The defaulted-given refusal is about the default, not the operator.** Narrowing the allowlist to `=` and `in` would have been simpler and is wrong: it kills `clearance <= $MAXLVL` and `effective_date <= $AS_OF`, which are legitimate no-read-up gates, and contradicts the operator set this PR's own docs publish. `> $FLOOR` at `0` fails open while `<= $MAXLVL` at `0` fails closed, and a given with no default cannot fail open at all.
- **A valid gate one derived entry point can't express** (a `rename:`/`except:`/`accept:` that removed the column) warns at load and denies at request time, rather than failing the whole model. One bad derived view shouldn't stop a file from serving.
- **The colocated-persist refusal has a second job, now written down.** `#@ preaggregate` synthesizes each rollup as a colocated `#@ persist` over an import of the annotated base, and no `preaggregation_*` module has any `#(authorize)` awareness — so that refusal is the only thing standing between a gated source and the pre-aggregation tier at build time. It holds because the eligibility walk follows the import → rename → `query_source` chain. `docs/materialization.md` justified it entirely on the frozen-gating-column hazard, which is right for the case it was written for but would let a later cleanup narrow it believing that is the whole story. A refused rollup also now names `#@ preaggregate` and the gated source rather than the synthesized rollup name the author never wrote, via the `origin` that `CompiledBuildPlan.preaggregatePlans` already distinguishes.
- **Pre-aggregation routing is blocked for a gated entry point as defence in depth, not as a leak fix.** Measured: a rollup over a gated base compiles and is *planned*, but both build gates refuse to materialize it (`referencesAuthorize` finds the base's gate in the rollup's compiled subtree), so no rollup table exists, the composite's rollup member recomputes from the gated base, and a covered query already returns correctly filtered rows. What the guard removes is the serve path's dependence on a *build* path's refusal — and that refusal is the blind walk whose join reach is documented as partial, so a synthesis change that emitted its base import in a shape the walk cannot see would build a rollup pre-aggregated across every tenant, with `preaggRouted` never reset and `effectiveBuildManifest` handing it the rollup manifest regardless. Both tiers now guard on one hoisted answer, short-circuited on a model-level "any authorize note anywhere" sweep so an ungated pre-aggregation deployment pays nothing per query.
- **That short-circuit has to be a superset of what the gate walks reach, and `contents` ∪ `sourceRegistry` is not.** Two of the five links the entry-point walk follows land on structs in neither collection: through an import, a `query_source`'s base arrives as an inline `SourceDef` on `query.structRef` rather than as a name into `contents`, and a composite's resolved member is synthesized into neither. The layered shape is ordinary — `import { gated }` in a middle file, `source: qs is gated -> {…}`, `import { qs }` locally — and it answered "no gates anywhere", un-guarding both tiers. Both hops already had resolvers, so the traversal is now one shared transitive, identity-guarded worklist, used by the routing pre-check and by the near-miss spelling refusal, which had the same blind spot at the cost of a missed error message rather than a missed guard.
- **The post-hoc storage-routing undo was deleted, not repaired.** It was unreachable: once routing swaps in the serve shape, that shape carries no `#(authorize)` bytes, so the walk could never find a gate. Correctness rests on the pre-check, which now fails closed on any throw and on any resolution that is not `given_only`.

## Testing

`bun test src` — 2842 pass / 3 skip / 0 fail (skips are from `main`); `tsc --noEmit` clean.

A dedicated row-level suite covers the entry-point matrix (`run: X`, `Y is X extend`, `Z is X -> {}` with and without the gate's column, `rename:`/`except:`/`accept:`, query sources, composites, joins), both P0 directions, fail-closed on every path that cannot apply the filter, cross-file and two-hop imports through the real worker pool, and the notebook paths.

Verified by execution rather than inspection where it mattered: gate predicates are absent from `drill_filters`; an empty array given yields zero rows; a row-gated query is not served from the storage tier; a missing-column gate failure returns no column name to the caller; two sources' gates AND rather than OR, asserted on rows for an admitted caller, an excluded one, and a caller the boolean denies outright.

The annotation ROUTING table is covered end to end — every spelling asserted as gate-enforced, refused-as-near-miss, or left alone, plus the caller-text rejection for each — and the block form is covered on all three consumers (gate enforced, caller text rejected, materialization refused). The `#@ preaggregate` + row-level-gate combination, which nothing covered, now has tests for the build refusal and for both covered and uncovered serve paths. The layered-import shape has its own pair — one pinning that the routing pre-check's sweep finds a gate reachable only through a derivation hop, one that the near-miss refusal reaches the same struct — each shown red without its fix.

`malloy_annotation_invariants.spec.ts` pins the Malloy IR behaviours this design is built on, by **reference identity** rather than deep equality — a deep-equality check would still pass if Malloy switched to copying by value, which is the change most likely to matter and least likely to be noticed. Each assertion names the code that depends on it and what breaks if it moves, and the resolved Malloy version is asserted alongside them so a failure names what changed. Two invariants were added: the annotation routing table itself (asserted against Malloy's own `Annotations.forRoute`, since a change in either direction silently changes what is a gate), and the exact wording behind `NO_DEFAULT_GIVEN_PATTERN`, which `authorize.ts` substring-matches as the one probe outcome it treats as safe.


housejester (commented):
Reviewed the whole change. The mechanism choice is right, and the three rejected alternatives in `docs/row-level-authorize-spike-findings.md` are the most valuable artifact here — each one looked correct until it was run, and recording that is worth more than the conclusion. The fail-closed discipline is genuinely thorough: I went looking for an execution path that obtains a runnable and runs it without going through `authorizeAndBindRunnable`, and didn't find one.

Findings below are ordered by what I'd act on. Everything hangs off the model shift in row 5 of the table, or off the breaking-change set.

## 🧠 Mental-model delta

**In one sentence:** `#(authorize)` stops being a boolean answered by a synthetic probe and becomes a *compiled filter condition grafted into the model's own IR* — so the enforcement point moves from "before the query" to "inside the query's compilation", and "denied" moves from 403 to 200-with-zero-rows.

|---|---|---|
| Where a gate is evaluated | synthetic one-row DuckDB probe | the real query's own `where:` on the entry source (row-field gates); probe retained for given-only |
| What a gate returns | admit / deny | a row set |
| Wire signal for "no access" | 403 | 200 + zero rows |
| What enforcement *depends on* | the probe's return value | **the caller reassigning `runnable`** to `authorizeAndBindRunnable`'s return value |
| Load-time validation scope | gates a source *declares* | every *entry point* in `authorizeMap`, two-pass, fatal-vs-warn decided by annotation **note-object identity** |
| Fail-closed surface | one probe | ≥6 points: no graft scope, no graft target, lift throws, classification rejected, graft build throws, landing proof fails |
| Storage tier defence | post-hoc undo (deleted — it was unreachable) | a **pre-check that is the only line**; nothing downstream can recover a miss |
| `##(authorize)` file-level | applied model-wide | fatal load error, *including folded in from an import* |
| New ownership on `Model` | — | retains the Malloy `Runtime` (`gateRuntime`), plus a shape cache and a cache of full `structuredClone`d `ModelDef`s |

**What a future dev must now hold in their head**

1. Anything that obtains a `QueryMaterializer` and runs it must pass it through `authorizeAndBindRunnable` **and use the return value** — the reassignment *is* the enforcement.
2. A gate's shape is decided by the **compiled IR** (`refSummary.fieldUsage`), never the annotation text.
3. Two identical gate strings are two different security facts. Note-object identity is the only safe discriminator for "inherited" vs "independently authored".
4. Anything that swaps the runnable for a shape carrying no `#(authorize)` bytes must be vetoed *before* the swap.
5. Load-time validation probes the **concatenated** gate list (OR); the request path enforces **separate entries** (AND). Those are not the same predicate — that is where the top finding lives.

**New coupling:** `Model` → Malloy's internal `Runtime._loadModelFromModelDef` (a graft is a model reload); `Model` → the storage-routing decision; `authorize.ts` (worker-bundleable) → `annotations.ts` note types; the materialization build path → `referencesAuthorize` on the colocated branch.

**Deliberately unchanged, so reviewers can skip these regions:** given-only gates — same probe, same 403, and that's essentially every published gate today. Entry-point-only semantics; joins still not traced. The REST/MCP `authorize` field shape.

---

## 🔴 Worth acting on

**[HIGH — availability: a synthesized predicate fails a whole model's load; this is the bug `50e1915e` fixed, reopened through the new check]**

`packages/server/src/service/authorize.ts:1391` calls `assertNoVacuousDefaultAtom` on the `row_level` success path with **no `ownNotes` discriminator** — unlike the `rejected` branch 18 lines above at `authorize.ts:1373`, which grew exactly that escape in `50e1915e`.

For a `query_source` over a composite, `effectiveAncestorGateExprs` concatenates the composite's whole-source boolean with the member's row-level gate. The concatenation classifies `row_level`, `literalAtoms` picks up the boolean's `$ROLE != 'admin'`, the vacuous-default probe returns true, and it throws — failing the **whole model**, at an entry point that carries no `#(authorize)` note of its own and whose probed text no author wrote.

Measured against `@malloydata/malloy` 0.0.427, driving `extractSourcesFromModelDef` + `validateAuthorizeProbes` exactly as `package_load_worker` does:

```malloy
##! experimental.givens
##! experimental.composite_sources

given: ROLE :: string is ''
given: GROUPS :: number[]

#(authorize) "org_id in $GROUPS"
source: member_a is duckdb.sql("SELECT 1 AS amount, 7 AS org_id") extend {
  dimension: only_a is 1
}

source: member_b is duckdb.sql("SELECT 2 AS amount, 8 AS org_id") extend {
  dimension: only_b is 2
}

#(authorize) "$ROLE != 'admin'"
source: combo is compose(member_a, member_b)

source: qs is combo -> { group_by: org_id, amount }
```

```
row-level gate alone  -> loads, no warning
boolean gate alone    -> loads, no warning
both together         -> Invalid #(authorize) annotation on source "qs":
                         the atom `$ROLE != 'admin'` evaluates to TRUE when a
                         caller supplies no givens (its given's own declared default)...

ownNotes counts: [["member_a",1],["member_b",0],["combo",1],["qs",0]]
```

That is `50e1915e`'s own "Measured: either gate alone loads, both together do not", one commit later.

Recommend routing it through the same `ownNotes.length === 0` escape (warn + `onRowLevelGateUnexpressible`). Nothing is under-enforced by relaxing, by that commit's own argument: the request path AND-s the entries separately, so the boolean's probe still runs and the member's filter still applies. The deeper fix is to probe atoms from each **declared** gate list rather than the walk's concatenation — which also closes the load-validates-OR / request-enforces-AND mismatch in row 5 above, where load-time proves resolution for a predicate that is not the one enforced.

Two incidental things the same measurement showed: `qs`'s `authorizeMap` entry carries `$ROLE != 'admin'` **twice** — `effectiveAncestorGateExprs` pushes the query-source base's own notes and then the composite-resolved member's, which carry the same note by reference. It is doubled in `filterText` and in what `sources[].authorize` reports over the API.

**[MED — the positive allowlist refuses `not in`, then admits the identical fail-open through `!=` / `>`]**

`authorize.ts:539` refuses `not org_id in $GROUPS` as "not an access rule; write the gate as the set of rows a caller MAY read". But the scalar-comparison branch at `authorize.ts:569` accepts a field comparison for every operator in `SCALAR_COMPARISON_NODES`. Measured, same fixture, with the documented empty/zero default convention:

```
org_id in $GROUPS      -> ACCEPTED
not org_id in $GROUPS  -> REFUSED (`not` is not permitted in a gate)
tenant != $EXCLUDED    -> ACCEPTED    # default ''  -> WHERE tenant != ''  -> ~every row
amount >  $FLOOR       -> ACCEPTED    # default 0   -> ~every row
amount >= $FLOOR       -> ACCEPTED
```

`assertNoVacuousDefaultAtom` structurally cannot catch these: they are field-vs-given, so they never become `literalAtoms`. The one shape that gets a default-value probe is the given-vs-literal atom, and the shapes that produce the same admits-everything-by-default outcome through a *field* comparison get nothing — while `not in` is refused for precisely that failure mode.

Options, least to most work: document `!=` / `>` / `<` / `>=` / `<=` against a field as carrying the identical default hazard `not in` was refused for; or narrow the allowlist to the positive operators (`=`, `in`) with explicitly-reasoned exceptions; or generalise the default probe to the whole condition rather than to isolated literal atoms. I'd lean toward narrowing — it reads most consistent with `authorize.ts:369`'s own "small enough to read in one screen". Thoughts?

**[MED — memory: a grafted materializer retains a full `ModelDef` clone, cached for the `Model`'s life, never evicted]**

`model.ts:433` / `model.ts:2297`. `graftedMaterializerCache` holds a `ModelMaterializer` built from `structuredClone(modelDef)`, keyed on `cacheScope` + the sorted graft set. `package_load/protocol.ts:63` describes these payloads as "multi-MB". A model with K row-level-gated entry points retains up to K clones, and a notebook multiplies that across `cell:N` / `cell-self:N` scopes.

The field doc argues the *correctness* of caching for the instance lifetime (a reload constructs a fresh `Model`) but never a bound. Not a DoS — no cache key is caller-controlled — but it is an unmeasured resident-set step function per gated package. Recommend measuring RSS for a realistic gated model plus a gated notebook, and capping it if that turns out material (an LRU miss costs one clone, not correctness).

## 🟡 Smaller correctness notes

**[MED — doc-vs-reality: `RELEASE_NOTES.md` still carries the stale claim `65ca79b2` corrected in `authorize.md`]** Under "Author-facing behavior worth knowing": "`rename:`, `except:`, and `accept:` can remove the field a gate was written against; package load fails, naming the source." `docs/authorize.md:238` has the correct two-branch behaviour — an entry point declaring its **own** gate fails the load with a 424; one that only **inherits** it loads with a warning and denies every request at that entry point. `65ca79b2`'s message flags fixing exactly this sentence elsewhere. Recommend mirroring authorize.md's two-branch wording.

**[MED — the colocated-persist refusal's justification doesn't match how a colocated persist is served; the *break* is real, the *leak* is the part I can't substantiate]**

`materialization_eligibility.ts:147`, plus `RELEASE_NOTES.md` and `docs/persist-storage-tutorial.md`. The `storage=` refusal has a clean mechanism: the serve shape carries no `#(authorize)` bytes, so the gate is genuinely gone. Colocated looks different — `#@ persist` substitutes the table via `buildManifest` against the **real, annotated** model, so the query still compiles against the source carrying the gate and `authorizeAndBindRunnable` still finds it (given-only → probe → 403; row-level → graft, or deny when the persisted projection dropped the gated column). Yet the tutorial says "a per-request gate on it is absent rather than unevaluated", and the release note says such a package "serves frozen rows TODAY, which is the leak being closed".

This matters because it is the one breaking change the PR itself gates on a store-wide scan, and it fails builds that currently succeed. Two asks:

1. Can you point at the concrete path where a colocated persist loses the gate? If it exists, it is the strongest justification in the set and the docs should name it.
2. If it does not, the refusal is still defensible as defence-in-depth over a frozen relation — but say *that* instead, because "closing a leak" and "refusing a shape we would rather not reason about" earn very different amounts of author patience for a build that breaks.

Separately: `referencesAuthorize`'s reach includes a *joined* source's gate, which entry-point-only semantics never enforced anyway — the function's own doc says as much at `materialization_eligibility.ts:44-49`. So the refusal is strictly broader than any leak it could close.

**[LOW — the vacuous-default rationale reads general but is scoped]** `authorize.ts:1133`. "An OR'd atom that is true by default makes the whole row filter admit every row" applies just as well to a whole-source gate — `#(authorize) "$ROLE != 'admin'"` on its own admits every caller at the documented empty default — and that one is not refused. Grandfathering it is the right call (refusing would break shipped packages), but one sentence saying so would stop the next reader from "fixing" the asymmetry in the wrong direction.

## 🟢 Minor / non-blocking

**[LOW — `queryEntryPointHasRowLevelGate` forces a live compile on every query in a routing deployment]** `model.ts:3862`. Ordering it *after* the routing preconditions is the right call and the comment earns it. But inside a routing deployment it runs per query, and `resolveRunTargetStruct` calls `getPreparedQuery()` on the **live** runnable — which a successfully-routed query previously never compiled at all (`liveRunnable` was only prepared in the degradation catch). So the storage tier's happy path now pays a live compile it used to skip, on the path whose whole purpose is latency. The fail-closed reasoning is right and I am not asking to weaken it; a sound cheap short-circuit exists (`entryPointGatesBySource` being empty everywhere — unlike `hasAuthorize()`, it *does* see derivation-inherited gates) if a measurement says it matters.

**[LOW — one worth a code comment]** The whole discriminator design rests on annotation note objects staying reference-shared across the package-load worker boundary. They do: transport is V8 structured clone (`protocol.ts:63`), which preserves internal aliasing, so `findSourceByOwnAnnotationIdentity` and `authorizeOwnNotes` behave identically in-process and under the worker pool. Worth one line near `protocol.ts:63` saying so, because a future switch to JSON transport "for speed" would silently turn every identity check into a fail-closed deny with no test naming the reason. (Also checked and clear: a `query:` statement over a gated source does **not** inherit the source's note — `NamedQueryDef.annotations` is `undefined` on 0.0.427 — so the new fatal `"query"`-kind misplacement cannot fire on `query: q is gated -> {...}`. The join-field case needed the note-identity escape hatch; this one genuinely does not.)

---

**Top three to resolve before merge:** 1) the vacuous-atom whole-model load failure at `authorize.ts:1391` — measured, and a repeat of the bug the previous commit fixed; 2) the `!=` / `>` vs `not in` allowlist asymmetry; 3) settle the colocated-persist justification, then run the published-package `#(authorize)` scan the PR body already commits to.

<sub>Reviewed by an assisted review pass (session codename "Spicy Bandicoot"); the measurements above were run locally against `@malloydata/malloy` 0.0.427 on this branch.</sub>

housejester (commented):
Re-reviewed at `80a28007`. Everything I raised last round is resolved. Three things left — one new, two carried over that no commit mentions.

**[MED — the new `field_given_has_default` check's only input is a private Malloy field, and it is not in the new canary suite]**

`service/given.ts:102` sources `ApiGiven.default` from `(given as { _internal?: { defaultText?: string } })._internal?.defaultText` — Malloy's private surface, as the comment there says. Everything downstream degrades silently if it moves:

- `model.ts:786` filters on `g.default != null`
- `authorize.ts:1408` is `options.declaredDefaults ?? new Map<string, string>()`
- the refusal itself is `declaredDefaults.has(given)` (`authorize.ts:677`)

So if `_internal.defaultText` is renamed or moved upstream, every `ApiGiven.default` becomes `undefined`, `declaredDefaults` is empty at every call site, `has()` is always false, and every defaulted-given row-level gate starts loading again — the fail-open `73dd7dc0` just closed, reopened with no signal at any layer. There is no "input was missing" path to distinguish an empty map from a model whose givens genuinely have no defaults.

This is the risk class `531e080f` exists for, in its own words: "Prose does not go red on a version bump, and the failure direction here is not uniformly fail-closed." The five behaviors it pins are all annotation identity; this input is not among them, and it now gates a security refusal. Recommend one more invariant in `malloy_annotation_invariants.spec.ts`: a given declared with a default yields the expected `defaultText` literal, and one declared without yields `undefined`.

Worth noting how easy the silence is to hit — driving `validateAuthorizeProbes` without `declaredDefaults` populated accepts all four hazardous shapes (`tenant != $EXCLUDED`, `amount > $FLOOR`, `tenant = $EXCLUDED`, `clearance <= $MAXLVL`) with no warning, which is exactly what a rename would produce in production.

**[MED — carried over: `graftedMaterializerCache` is unbounded]**

`model.ts:432`, unchanged this round. Each entry holds a `ModelMaterializer` built from a full `structuredClone(modelDef)` (`model.ts:2311`-ish), keyed on `cacheScope` + the sorted graft set, and nothing evicts. `package_load/protocol.ts` describes these payloads as multi-MB. A model with K row-level-gated entry points retains up to K clones, multiplied across notebook `cell:N` / `cell-self:N` scopes.

The field doc argues the *correctness* of caching for the instance lifetime (a reload constructs a fresh `Model`) but never a bound. No cache key is caller-controlled, so this is not a DoS — it is an unmeasured resident-set step function per gated package. Recommend a measurement on a realistic gated model plus a gated notebook before merge, even if the answer turns out to be "negligible"; a cap is cheap if it is not (an LRU miss costs one clone, not correctness).

**[LOW — carried over, non-blocking: `queryEntryPointHasRowLevelGate` forces a live compile per query on the storage tier]**

`model.ts:3862`-ish, unchanged. Ordering it after the routing preconditions is the right call and the comment earns it. But within a routing deployment it runs on every query, and `resolveRunTargetStruct` calls `getPreparedQuery()` on the **live** runnable — which a successfully-routed query previously never compiled at all, since `liveRunnable` was only prepared in the degradation catch. So the storage tier's happy path now pays a live compile it used to skip, on the path whose whole purpose is latency.

The fail-closed reasoning is right and I am not asking to weaken it. Flagging only so it is a known cost rather than a surprise later; a sound cheap short-circuit exists if a measurement says it matters (`entryPointGatesBySource` being empty everywhere — unlike `hasAuthorize()`, it does see derivation-inherited gates).

---

**Merge bar, for my money:** the `defaultText` invariant, and a number on the graft cache. The `queryEntryPointHasRowLevelGate` cost is fine to ship as-is.

<sub>Assisted review pass, session codename "Spicy Bandicoot"; measurements run locally against `@malloydata/malloy` 0.0.427 on this branch.</sub>

housejester (commented):
Re-reviewed at `f2a73876`. Both merge-bar items are done and check out: the graft cache is capped with a real LRU (re-insert on hit so `Map` order is genuinely least-recently-used, evict after insert so the entry just built is never the one dropped), and `_internal.defaultText` is now a canary covering four default types plus the absent case, which is what keeps the positive assertions from passing vacuously. I also read `skipOwnSourceGate` from the rebase reconcile — it skips only the name-keyed early gate, defaults false, has one call site, and the compiled entry-point walk still runs unconditionally, so it is not a fail-open.

One thing left, and it is an interaction with what main landed while this branch was open rather than a defect in the branch — so mostly a confirmation request.

**[MED — `#@ preaggregate` × `#(authorize)`: the containment holds, but by accident, with a misleading message and no test]**

Pre-aggregation is a second serve path that reassigns `runnable` (`model.ts:4328`) before the authoritative gate runs (`model.ts:4425`) — structurally the same shape as the `storage=` serve shape, whose only line of defence is `queryEntryPointHasRowLevelGate`. Unlike that one it has no eligibility test, deliberately and reasonably ("there is no eligibility test here and no metric for 'declined'" — from out there a covered query and an uncovered one are the same call). And `grep -n authorize packages/server/src/service/preaggregation_*.ts` returns nothing across all six modules.

The good news is that it is contained. Building the synthesized shape from `preaggregation_synthesis.ts`'s own header — `import { orders__preagg_base is orders }`, a `#@ persist` rollup over that alias, then `source: orders is compose(rollup, base)` — over a base carrying `#(authorize) "org_id in $GROUPS"`, both eligibility checks refuse the rollup. `referencesAuthorize` follows the import → rename → `query_source` chain and finds the gate:

```
persist sources:  ["orders__preagg__category__deadbeef"]
colocated check:  REFUSED: Source 'orders__preagg__category__deadbeef' cannot be materialized (colocated '#@ persist')…
storage= check :  REFUSED: Source 'orders__preagg__category__deadbeef' cannot be materialized into a storage destination…
```

So a gated source cannot get rollups built, and the serve path never has a frozen rollup to route to. Three things follow:

1. **The author-facing message is misleading for this origin.** It names `orders__preagg__category__deadbeef` — a synthesized source the author never wrote and cannot find in their model — and prescribes "Drop `#@ persist` from this source", when what they actually wrote is `#@ preaggregate`. `build_plan.ts:783` already carries `origin: rollup ? "preaggregate" : "persist"`, so the refusal has what it needs to name the annotation the author typed and the source it sits on.

2. **This is a second, independent reason the colocated-persist refusal earns its keep.** `materialization.md` now justifies it entirely on the frozen-gating-column hazard, which is the right justification for the case it was written for — but it is also, today, the only thing standing between a gated source and the pre-aggregation tier, which has no authorize awareness of its own. Worth a sentence there, so a later cleanup does not narrow the check believing the frozen-column hazard is the whole story.

3. **No coverage in either direction.** No `preagg` in any authorize spec; no `authorize` in any `preaggregation_*.spec.ts`. Given the containment is emergent rather than designed, and it is the load-bearing property, a test asserting a gated source's rollup is refused is worth more than the message fix.

Happy for the message and the test to be a follow-up rather than this PR's — it is main's feature that moved under you, and the security outcome is already correct. Mainly flagging it because a cross-PR interaction like this is the kind nobody owns by default. Is the refusal reaching pre-aggregation something you had already traced, or is this new?

<sub>Assisted review pass, session codename "Spicy Bandicoot"; measurements run locally against `@malloydata/malloy` 0.0.427 at `f2a73876`.</sub>

housejester (commented):
Re-reviewed at `3b8a13ea` (the rebase moved every SHA; the authorize commits are `ea4d557f` / `65c85214`). Last round's three asks are all done and check out — the `#@ preaggregate` origin threads through `preaggregatePlans[sourceID]`, the same key `build_plan.ts` derives `origin` from; `materialization.md` records why the colocated refusal is load-bearing for a second reason; and the rollup-refusal test exists in both directions. Thanks for answering the cross-PR question with a commit.

The headline of this round is a good catch nobody reviewing found: `#|(authorize)` was unenforced at query time **and** materialization-eligible. Moving classification onto Malloy's own routing is the right permanent fix, and `everyMemberOf` is the right shape for drift you'd already been bitten by twice.

Two findings, both in this round's new code. Measured against `@malloydata/malloy` 0.0.427 (what `bun.lock` pins) at this head. Suites green locally: `malloy_annotation_invariants`/`annotations`/`authorize`/`materialization_eligibility`/`authorize_gate_walk` (87), `authorize_integration`+`row_level_authorize.integration` (242), `preaggregation_seams` (17).

## 🔴 The `hasAnyAuthorizeNote()` short-circuit is not the superset its doc claims

`model.ts:1012`, gating the hoisted answer at `model.ts:4387`. The doc at `model.ts:1000` asserts:

> That is a superset of everything `collectEntryPointGates` can reach. Every link it follows (own notes, the inherits chain, a `sourceRegistry` declaration, a `query_source`'s base, a composite's resolved member) lands on a struct in one of those two collections.

Two of those five don't. `resolveQuerySourceBase` resolves `query.structRef`, which is an **inline `StructDef`** rather than a string into `contents` when the base arrived through an import; `resolveCompositeResolvedBase` returns a synthesized member struct that is in neither collection. The sweep reads `contents` plus *non-reference* `sourceRegistry` entries only, so it answers `false` and the walk never runs.

Measured, driving publisher's own `extractSourcesFromModelDef` and `effectiveAncestorGateExprs` (the sweep is a line-for-line replica, since the method is private — I diffed it against this head to be sure):

```
base.malloy:  #(authorize) "org_id in $GROUPS"   on   source: gated
mid.malloy:   import { gated };  source: qs is gated -> { group_by: org_id, amount }
m.malloy:     import { qs }                       # `gated` never enters local contents

  contents:                        ["qs"]
  authorizeMap["qs"]              = [["org_id in $GROUPS"]]   <- registered as gated
  effectiveAncestorGateExprs(qs)  = [["org_id in $GROUPS"]]   <- the walk finds it
  hasAnyAuthorizeNote(modelDef)   = false                     <- so the walk never runs
```

Same for the composite variant (gated member, only `compose(...) -> {...}` exported): found via `compositeResolvedSourceDef`, missed by the sweep.

So `routingBlockedByRowLevelGate` is `false` for a row-level-gated entry point. For the **storage** tier that is a behaviour change this round — before the short-circuit the condition was `storageRoutingPossible && !bypassAuthorize && walk`, the walk ran, and it blocked. It lands on the line the design describes as the only one: *"nothing downstream can recover a miss."*

Scoping honestly: **not a live leak.** `referencesAuthorize`'s blind deep walk does recurse `query.structRef` and finds the note, so the build refuses, no artifact exists, and there is nothing to route to. But that is precisely the dependence `ea4d557f` added the pre-aggregation guard to remove — *"What the guard removes is the serve path's dependence on a BUILD path's refusal for its own correctness."* The short-circuit hands it back for this one IR shape, and `materialization.md`'s new parenthetical ("so that containment does not rest on this refusal alone") is true only for gates the sweep can see.

Fix is the two hops the walk already takes. Verified at 0.0.427 — closes both shapes while an ungated model with a `query_source` still answers `false`, so the short-circuit keeps doing its job:

```ts
for (const struct of [...structs]) {
   const q = (struct as unknown as { query?: {
      structRef?: unknown; compositeResolvedSourceDef?: unknown } }).query;
   if (!q) continue;
   if (typeof q.structRef === "object" && isSourceDef(q.structRef))
      structs.push(q.structRef as StructDef);
   if (isSourceDef(q.compositeResolvedSourceDef))
      structs.push(q.compositeResolvedSourceDef as StructDef);
}
```

```
query_source over an imported gated base:  as-shipped=false   with-two-hops=true
composite member, imported and gated:      as-shipped=false   with-two-hops=true
genuinely no gate anywhere:                as-shipped=false   with-two-hops=false
```

Worth a test in that shape — the new `preaggregation_seams` gate tests all declare the gate in local `contents`, so none of them can see this. Smaller sibling: `source_extraction.ts`'s near-miss sweep walks the same two collections, so a near miss authored in that layered-import shape isn't refused either; that one costs an error message rather than a guard.

## 🟡 The near-miss test refuses valid, distinct routes — and `routeOf` is exported, so the exact rule is available

`authorize.ts:74`. The lookahead is `(?![\p{L}\p{N}_])`, but a Malloy route name runs to the closing bracket and admits punctuation. Measured:

```
#(authorize-v2)     -> routes to "authorize-v2"     -> refused, whole model load
#(authorize.audit)  -> routes to "authorize.audit"  -> refused
#(authorize/v2)     -> routes to "authorize/v2"     -> refused
#(authorize:v2)     -> routes to "authorize:v2"     -> refused
   (also every other punctuation char probed: + * ~ ! @ $ % ^ & = ? | \ ' " ; , #)
```

Those aren't near misses — they're another route's deliberate annotation, and the refusal tells their author to write `#(authorize) "<expression>"`, which is the wrong advice for them. `collectAuthorizeNearMisses`'s doc claims false positives are confined to a note starting with the sigil followed by `authorize`, offering `#(authorized)` as the boundary; the word-char lookahead catches that one and misses this class.

The reason worth flagging it rather than filing it as a nit: this is the one place the implementation diverged from the shape we'd agreed. The near-miss rule was specified as route-based and exact — a note whose route is `''` with content beginning `(authorize)`, or a malformed-route note whose bracketed text mentions authorize — and what shipped is a text superset, which is what admits this class. I suspect the reason is `annotations.ts`'s own TODO ("this route classification belongs in `@malloydata/malloy` core, beside `parsePrefix` — remove when core exports an equivalent"). **It already does, at the pinned 0.0.427:** `routeOf` is exported from the public barrel, `(a: Malloy.Annotation) => string | undefined`, giving exactly the three-way split — the route name, `''` for MOTLY, `undefined` for malformed. Verified:

```
#(authorize)        -> "authorize"       GATE
# (authorize)       -> ""                near miss -> refuse
#( authorize )      -> undefined         near miss -> refuse
#(authorize)X       -> undefined         near miss -> refuse
#authorize          -> undefined         near miss -> refuse
#(authorize-v2)     -> "authorize-v2"    someone else's route -> ignore
#(authorized)       -> "authorized"      someone else's route -> ignore
# bar_chart         -> ""                MOTLY, payload isn't (authorize -> ignore
```

Adapting is `routeOf({ value: note.text })`, since `routeOf` takes the stable wire shape while the IR notes are `{text, at}`. That closes the TODO and retires the superset regex from the *near-miss* half entirely. Note the caller-text rejecter still needs its regex and still needs to be a superset — that asymmetry is right and shouldn't change.

If you'd rather not take the `routeOf` dependency in this PR, the cheap equivalent is to require the route name to end at `authorize`. Verified still a superset of all seven gate spellings, still refuses all six near misses, drops the four false positives:

```ts
// was: authorize(?![\p{L}\p{N}_])
const AUTHORIZE_TAG_LIKE = String.raw`##?\|?[ \t]*[([{<]?[ \t]*authorize(?=[)\]}>]|[ \t]|$)`;
```

Same change also stops the unanchored rejecter 400-ing a submitted query that merely contains `#(authorize-v2)`.

## 🟢 Two small ones

- **`#(AUTHORIZE)`** is the motivating fail-open with a different shift key: `routeOf` gives `"AUTHORIZE"`, so it's not a gate, and `AUTHORIZE_TAG_LIKE` is case-sensitive so it's not a near miss either — author reads the source as locked, every row serves, load clean. Under the `routeOf` rule it reads as someone else's route, which is defensible; a case-insensitive *tag-like* test would refuse it without publisher honouring anything. Your call — mentioning it because the refusal machinery now exists and this is the obvious remaining hole in it.
- **The per-query walk's surface widened**, correctly. It used to be reachable only when storage routing was possible; a gated pre-aggregation deployment now pays it too. `hasAnyAuthorizeNote` is memoized per instance so an ungated deployment pays one IR sweep for the model's life and nothing per query — which is exactly why the short-circuit is worth keeping rather than dropping. Non-blocking.

---

**Merge bar:** the two hops plus a test in the layered-import shape. The near-miss rule I'd take in the same commit, `routeOf` for preference, but wouldn't block on the lookahead version alone.

Still open from earlier rounds, unaddressed by any commit: the published-package scan the PR body commits to. Worth two more patterns in that grep — `#\(authorize[^)]` for the distinct-route class above, and a case-insensitive `#\(authorize\)` for the uppercase one.

<sub>Assisted review pass, session codename "Stoic Gibbon"; measurements run locally against `@malloydata/malloy` 0.0.427 at `3b8a13ea`.</sub>

housejester (approved):
Both findings fixed at `105fe73c`, and each went further than what I asked for. Approving.

**The sweep gap.** `derivedStructsReachable` in `gate_registry_walk.ts` is the better shape than the inline version I sketched — one transitive, identity-guarded traversal, used by *both* sweeps rather than just the routing pre-check, so the near-miss refusal picks up the same reach at the same time. Verified through the real code at the pinned 0.0.427:

```
query_source over an imported gated base   contents=["qs"]    sweep=true   authorizeMap[qs]=[["org_id in $GROUPS"]]
same, re-derived locally                                      sweep=true   authorizeMap[local]=[[...]]
composite over imported gated members     contents=["cqs"]    sweep=true   authorizeMap[cqs]=[[...]]
no gate anywhere                                              sweep=false  <- still short-circuits
near miss on the hop-only source                              REFUSED      <- the sibling case
```

I checked the "goes red without its fix" claim rather than taking it: dropping the single `structs.push(...derivedStructsReachable(...))` line turns the sweep test red and nothing else, which is the right blast radius for that assertion.

**The near-miss rule.** Route-keyed on `routeOf`/`payloadOf` is exactly right, and it retiring `noteContentOnRoute` plus that stale "remove when core exports an equivalent" TODO is a nice bonus. Measured through the real exports: all seven gate spellings including `#|(authorize)` classify as gates; `# (authorize)` / `## (authorize)` / `#\t(authorize)` / `# [authorize]` / `#( authorize )` / `#(authorize )` / `#(authorize)X` / `#authorize` and the case variants all refuse; `#(authorize-v2)`, `#(authorize.audit)`, `#(authorize/v2)`, `#(authorized)`, `#(authorize_v2)`, `#(doc)`, `#(filter)` and the render tags are all left alone.

One thing worth recording, since it briefly looked like a regression on my end and isn't: `payloadOf` and the old `forRoute().content` return byte-identical payloads on every gate spelling, block form included. My first check said otherwise, but that was my fixture — I'd written the `|#` closer into the note text, and Malloy's lexer strips it when building the IR note. A real compile shows the block form parsing to its expression cleanly.

Suites green locally: 334 across the authorize/annotation files, 44 across source-extraction/compile-authorize/ownership, 19 preagg.

Two non-blocking nits, both fine to fold into a follow-up:

- `# authorize "expr"` (space, no bracket) is now silent, where the previous commit's text superset refused it — route `""`, and `MOTLY_AUTHORIZE_PAYLOAD` requires a bracket. The comment's reasoning holds: a bare MOTLY tag named `authorize` is a legitimate thing and claiming it is the overreach we're avoiding. It does leave `#authorize` refusing while `# authorize` doesn't, and the latter is the same "author reads locked, every row serves" shape. Your call whether the payload test should also admit a bare `authorize` followed by whitespace and a payload.
- `RELEASE_NOTES.md:136` enumerates the refused spellings and now under-states them: case variants (`#(AUTHORIZE)`) also fail the load, and other `authorize*` routes are deliberately left alone. Both commits touched only `.ts`, and that bullet is the breaking-change contract.

<sub>Assisted review pass, session codename "Stoic Gibbon"; measurements run locally against `@malloydata/malloy` 0.0.427 at `105fe73c`.</sub>

PR #1018, opened 2026-08-14
fix(server): carry author-declared query metadata onto served queries

housejester:
Author-declared query metadata now rides **every statement touching a source**, not only the ones a
build issues.

A package's `materialization.queryMetadata`, a model file's `## materialization.queryMetadata.*` and
a source's `#@ persist queryMetadata.*` reached materialization statements and stopped there. A served
query arrived carrying the server's own context — `class`, `package`, `model`, `query_id` — and none
of the author's vocabulary. So a deployment could attribute its builds and not the interactive
traffic that is most of a warehouse bill.

Confirmed against live warehouses on both BigQuery (`INFORMATION_SCHEMA.JOBS.labels`) and Snowflake
(`QUERY_HISTORY.query_tag`): a build statement carried 14 properties, and a served statement from the
same package on the same connection carried 9.

## What a served query resolves

Same layers as a build, same precedence, through the same function:

| layer | source of truth |
|---|---|
| package `materialization.queryMetadata` | the controller, which holds the package handle |
| model file `## materialization.queryMetadata.*` | `modelDef`, which survives the worker serialization boundary |
| source `#@ persist queryMetadata.*` | the run-target source's own annotation |

`composeDeclaredQueryMetadata` is factored out of the build-path resolver and takes tags rather than a
`PersistSource`, so precedence cannot drift between the two paths.

The source layer resolves from the run target the server already computes for the authorize gate —
the explicit `sourceName`, a `queryName`'s source, or the run target in ad-hoc text. A statement with
no single run target, such as some notebook cells, carries the package and model-file layers only.

The MCP query boundary resolves the same layers as the HTTP controllers. It builds its own metadata
input, and agent traffic is interactive traffic.

## Context names are reserved against a declaration

Context is applied above every declared layer, but it only overwrites a name it has a **value** for —
and a served query has no `source`, `trigger` or `run_id`. Without a rule, a model file declaring
`queryMetadata.source="orders_daily"` would stamp it on interactive statements, which in the
warehouse's own history read exactly like a build of that source.

A declared context name (`class`, `environment`, `package`, `model`, `source`, `trigger`, `run_id`,
`query_id`, `version`) is therefore **dropped rather than attached**, and metered as `reserved_name`.

A silent drop is not much better than a silent attach, so the publish gate reports it as a message
naming the level that declared it. That gate now walks all three declaration levels: the package
manifest, each persist source, and each model file's own `##` block — including a model file that
persists nothing, which contributes a layer to every query served from it while having no source to
be reported under.

Scoped to the model-side declaration. A connection default and a per-request bag can still set these
names, as they always could: a request setting `queryClass` is a caller describing its own call, and
narrowing either is a compatibility decision of its own. What the rule covers is the layer that
describes *other people's* statements — a declaration rides every query against that source,
including ones its author never sees.

**This changes the build path slightly.** A declared `source` was already overwritten by context there,
so reserving it is a no-op; `model` and `version` are context names a build does not set, so a
declaration of either survived before and now drops. Both are server-owned names, and a uniform rule
is easier to state and to trust than a per-path exemption list.

## Documentation

`docs/query-metadata.md` and `docs/materialization.md` stated the rule this changes, in the sections
an author reads before declaring anything. The OpenAPI descriptions did too — including
`PersistSourcePlan.queryMetadata`, which invited a caller to copy the reported bag onto its own
queries, and which can now report a name the publisher will refuse. All corrected here.

## Tests

The regression test fails without the fix — verified by reverting `model.ts`:

```
Expected: "finance"
Received: undefined
```

There was no serve-path query-metadata test before this. Added: all three declared layers composing
with source-beats-model-beats-package precedence; the source layer resolving from a named query and
from ad-hoc text separately, since those are different branches; a request property overriding a
declared one without evicting what it does not mention; a declared `source`/`run_id` refused while the
author's own property still lands; and the publish gate warning about a reserved name at the level
that declared it, including a model file with no persist source.

Suite: **2,422 pass, 0 fail, 3 skip** across 123 files. `tsc --noEmit` and prettier clean.

## Behaviour that is unchanged

Wire and API shape, precedence order, the enforced layer, and the feature's default —
`PUBLISHER_QUERY_METADATA` is `off` unless a deployment opts in, so none of this reaches a deployment
that has not asked for attribution.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -3682,12 +3682,16 @@ components:
         `environment`, `package`, `model`, `source`, `trigger`, `run_id`,
         `query_id`, `version`) is applied on top and cannot be overwritten — it
         is what the server is actually doing, and a caller must not be able to
-        relabel it. Those names are also RESERVED against a declaration: a
-        connection, package, model file or source declaring one has it DROPPED
-        rather than attached. Context only overwrites a name it has a value for,
-        and a served query has no `source`, `trigger` or `run_id` — so a declared
-        one would otherwise ride interactive traffic and read, in the backend's
-        own history, exactly like a build. Put no secrets or personal data here: a tag is visible to
+        relabel it. Those names are also RESERVED against a MODEL-SIDE
+        declaration: a package block, a model file or a `#@ persist` annotation
+        declaring one has it DROPPED rather than attached. Context only
+        overwrites a name it has a value for, and a served query has no
+        `source`, `trigger` or `run_id` — so a declared one would otherwise ride
+        interactive traffic and read, in the backend's own history, exactly like
+        a build. A connection default and a per-request bag are not covered: a
+        request setting `queryClass` is a caller describing its own call, while a
+        model-side declaration rides every query against that source, including
+        ones its author never sees. Put no secrets or personal data here: a tag is visible to
         anyone who can read the backend's query history, and on a backend with no
         native tag mechanism it is visible in the statement text.
 
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -1407,10 +1407,19 @@ paths:
         package-level properties. The package must exist and be accessible.
 
         A supplied `materialization` block REPLACES the one on disk, so re-send every
-        field of it you want to keep. Two exceptions, because neither is expressible
-        in the block a caller sends: `scope` (a top-level field) and `queryMetadata`
-        (orthogonal to the schedule and freshness policy this call usually carries) are
-        preserved when omitted. Send `queryMetadata: null` to clear it.
+        field of it you want to keep. Two exceptions are preserved when omitted:
+        `scope`, which is a top-level field and not expressible in the block at all,
+        and `queryMetadata`, which is orthogonal to the schedule and freshness policy
+        this call usually carries — a client setting a schedule has no reason to
+        re-send the package's tags, and dropping them silently untags every statement
+        the package issues.
+
+        `queryMetadata` has two homes: the top-level field, which is canonical, and
+        `materialization.queryMetadata`, which is DEPRECATED and still accepted. The
+        canonical field wins if a request sends both, and both are populated on read.
+        At EITHER home, omitting it preserves the package's tags and so does sending
+        null — a client that serializes unset fields as null must not untag a package
+        by accident. Send an empty object to clear.
       parameters:
         - name: environmentName
           in: path
@@ -2955,6 +2964,32 @@ components:
             URI (gs:// or s3://) of the externally-computed manifest for this package.
             On (re)load the publisher reads it and binds persist references
             (sourceEntityId -> physicalTableName). Null = serve live.
+        queryMetadata:
+          oneOf:
+            - $ref: "#/components/schemas/QueryMetadata"
+            - type: "null"
+          description: |
+            The package-level layer of per-query metadata. Applies to every
+            statement the package's sources issue — a served query as much as a
+            build — and is overridden per property by a model-file, source or
+            per-request declaration.
+
+            **This is the canonical home, here and in `publisher.json`.**
+            `materialization.queryMetadata` is the original one and is
+            DEPRECATED: the properties are not a build setting, so carrying them
+            inside the build-policy block described a scope the feature does not
+            have.
+
+            Both are populated on read and both are accepted on write, so a
+            client can migrate on its own schedule. On write the canonical field
+            wins if a request sends both; in the manifest, the two declaring
+            different bags warns and the root wins, rather than failing the load
+            the way a conflicting `scope` does — a tag is observability only and
+            must never be the reason a package refuses to load.
+
+            Omitting this on a PATCH preserves the package's existing tags, and
+            so does sending null — a client that serializes unset fields as null
+            must not untag a package by accident. Send an empty object to clear.
         materialization:
           oneOf:
             - $ref: "#/components/schemas/PackageMaterializationConfig"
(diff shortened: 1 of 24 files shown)


Sha-Bang (changes requested):
## Pre-landing review — request changes

Core design (shared `composeDeclaredQueryMetadata`, reservation rule, publish warnings, MCP package layer, docs) looks solid. Two correctness bugs should be fixed before merge; a few P2s are worth landing in the same pass.

### Must fix

**1. Metadata uses `earlySource`, not the compiled run target** (`model.ts` ~2826 / live-fallback ~2987 / notebook ~3500)

Authorize correctly resolves `compiledSource` from the prepared query (last `run:`). Metadata still passes `earlySource` from `extractRunTargetSourceName` (first `run:`). Multi-statement text like `run: cheap …\nrun: expensive …` can execute one source while tagging another’s team/tier. `compiledSource` is already available above the resolve call.

**Fix:** Pass `compiledSource ?? earlySource` into both serve and live-fallback resolves (and the analogous notebook path).

**2. Model-file publish warnings put the model path in `subject`** (`materialization_config_validation.ts` ~114–120)

OpenAPI `Package.warnings` uses `model` for the model path and `subject` for source/query/view. `modelDeclarations` currently pass `declaration.modelPath` into `metadataWarnings`’s subject slot, so clients filtering by `model` miss them and `subject: "marts.malloy"` looks like a source name.

**Fix:** Extend `MaterializationConfigWarning` with `model?: string` and emit `{ model: declaration.modelPath, message }` for model-file findings.

### Should fix (same PR or immediate follow-up)

**3. Inherited declarations still warned as `#@ persist`** (`materialization_config_validation.ts` ~124–129)

`modelDeclarations` was added because resolved source bags mislabel model-file properties — but the sources loop still validates the full resolved bag under `#@ persist queryMetadata`. A `##` reserved name still gets a second, wrong-labelled warning on every inheriting persist source.

**Fix:** For sources, validate only own `#@ persist queryMetadata` (or skip properties not declared on that source).

**4. `safeModelFileTag` uses import-folded `modelAnnotations`** (`model.ts` ~3250)

`annotations.ts` says non-gate `##` tags should use `ownModelNotes` (document-local). Folding lets an import’s `queryMetadata` ride every importer and attributes publish warnings to the wrong path via `getDeclaredQueryMetadata()`.

**Fix:** Own-file notes for publish; decide explicitly whether serve should inherit includes (and test it).

**5. PersistSourcePlan OpenAPI text understates request-copy risk** (`api-doc.yaml`)

Says copying the reported bag onto a request “can carry a name the publisher will not accept.” Reservation applies only to model-side declarations; request-layer `source`/`trigger`/`run_id` still attach. Reword to match the QueryMetadata schema text.

**6. Tag parse runs when `PUBLISHER_QUERY_METADATA=off`** (`model.ts` `resolveQueryMetadata`)

`composeDeclaredQueryMetadata` / `safe*Tag` run before `mergeQueryMetadata`’s `off` early-return. Default deployments pay annotation parse on every query for nothing. Guard with `getQueryMetadataMode() === "off"` first.

### Out of scope for this PR (fine to leave)

- Request/connection layers still setting context names (documented deliberate compatibility boundary)
- Build path dropping declared `model`/`version` (acknowledged in the PR body)
- Circular `model.ts` ↔ `build_plan.ts` (follow-up extract)

CI is green; suite coverage for the happy path is good. Happy to re-review after the two P1s land.

Sha-Bang (changes requested):
## Re-review — still one blocker

Thanks for `71e1a507` and `8d576463`. Cleared from the prior review:

| # | Finding | Status |
|---|---------|--------|
| 1 | `earlySource` vs compiled run target | ✅ `compiledSource ?? earlySource` on serve, live-fallback, notebook |
| 4 | Import-folded `modelAnnotations` | ✅ `ownModelAnnotations` |
| 5 | PersistSourcePlan understated request-copy risk | ✅ Corrected |
| 6 | Tag parse when metadata off | ✅ Early return in `resolveQueryMetadata` |

Tests for multi-`run:` attribution and import non-folding look right. CI green.

### Still must fix

**Model-file publish warnings put the model path in `subject`** (`materialization_config_validation.ts` ~114–120)

OpenAPI and every other warning path in `package.ts` use:

- `model` → package-relative model path  
- `subject` → source / query / view name  

`modelDeclarations` still pass `declaration.modelPath` into `metadataWarnings`’s subject slot, and the spec asserts `warnings[0].subject === "marts.malloy"`. Render-tag / dropped-persist / storage warnings all spell out `model:` + `subject:` explicitly so a rename is caught; this path does not.

**Fix:** Add `model?: string` to `MaterializationConfigWarning`, emit `{ model: declaration.modelPath, message }` for model-file findings (omit `subject`), and update the test accordingly.

### Non-blocking follow-up

Inherited package/model properties in a source’s *resolved* bag still get a second warning labelled `#@ persist queryMetadata`. Message text names the level, so this is debuggable — but it is the misattribution `modelDeclarations` was added to fix. Prefer validating only source-own `#@ persist queryMetadata` on that loop.

Happy to approve as soon as the `model`/`subject` wire shape is corrected.

Sha-Bang (approved):
## Re-review — approve

Prior blockers are cleared on `8fb03528`:

- Metadata tags the **compiled** run target (`compiledSource ?? earlySource`) on serve, live-fallback, and notebook paths
- Model-file publish warnings put the path in `model`, not `subject`
- Publish gate checks **declarations** (not resolved bags), so inherited properties are no longer mislabelled `#@ persist`
- `ownModelAnnotations`, off-mode early return, and PersistSourcePlan request-copy wording all look correct

The follow-on rename (canonical root `queryMetadata`, dual-home deprecation of `materialization.queryMetadata`, PATCH serve/build agreement, merged budget floor) is coherent and CI is green. One non-blocking nit: `resolveQueryMetadata`'s JSDoc in `build_plan.ts` still spells the deprecated layer names — fine as a follow-up.

LGTM.

PR #1019, opened 2026-08-14
test(sdk): pin a given's date to UTC, and run the suite east and west of it

mlennie:
## Why

CI runs entirely in UTC. `git grep TZ` across `.github/workflows/` and all three manifests returns nothing, and GitHub runners default to UTC, which is the one zone where a local-vs-UTC date bug is invisible. That class has already shipped in this repo: a date given published the wrong **day** depending on the worker's zone.

## The obvious fix would have caught nothing, and that is worth its own line

The natural change is a non-UTC CI run. On `main` today it would have been decoration. Measured, by reverting `GivenInput`'s `dayjs.utc(value)` to `dayjs(value)`, which is the shipped bug's own shape:

| zone | failures, before this PR |
|---|---|
| UTC | 0 |
| Asia/Kolkata (+05:30) | 0 |
| Pacific/Chatham (+12:45) | 0 |

Zero even at +12:45, because **no spec in `packages/sdk` exercised a date at all**. There was nothing for any zone to catch.

That is a finding in its own right, independent of the timezone question: the date branch was the one branch of `GivenInput` with no test, and the DatePicker rendered in no spec. It is also why the timezone question looked answerable when it was not. I added the harness these specs run on, rendered the DatePicker in a throwaway probe while building it, and shipped no assertion for it.

## What is here

**Date tests first.** Two literals, because one can only catch half the bug: a late-UTC instant crosses midnight only for zones ahead of UTC, an early-UTC one only for zones behind it. Both assert the same UTC day. 2024 being a leap year makes the west-shifted wrong answer `02/29`, which is doubly wrong. The trap being avoided is a fixture that happens to sit on the same UTC day in the runner's zone and therefore cannot fail however exotic the zone looks.

**Then two extra runs of the SDK suite**, one east of UTC and one west, in the existing step.

## The measurement, after the tests exist

Same mutation, counting which tests fail rather than only how many:

| zone | tests that fail |
|---|---|
| **UTC** | **none** |
| Australia/Sydney (east) | late-instant, plus both shared-branch types |
| Pacific/Auckland, Asia/Kolkata (east) | same three |
| America/Los_Angeles, America/New_York (west) | early-instant |

Three things follow.

UTC still catches nothing, which is the whole case for running anywhere else.

A half-hour offset is not the poor relation it looks. Kolkata ties the far-ahead zones, so whether a zone catches anything is a property of the fixture straddling UTC midnight in that zone's direction, not of the zone being large or unusual.

**And the two directions are not interchangeable.** A first version of this ran UTC plus Sydney only, both at-or-ahead of UTC, which left the early-instant test unable to fail anywhere it ran. Counting failures without looking at *which* ones is what hid that. So it runs one zone each side, and they are representatives rather than special: any ahead zone scores the same as Sydney, any behind zone the same as Los Angeles. Please do not add more for symmetry.

## Shape

Extra runs in the existing step rather than a matrix entry: the suite is about 0.3s, a matrix entry would double the whole job, and setting `TZ` on the existing run would trade UTC coverage rather than add to it.

The two types that share the date branch are separate `it.each` cases rather than a loop, because a loop stops at the first failing type and names neither.

## Verification

Clean code is **15 pass / 0 fail** under each of UTC, Sydney and Los Angeles. Every one of the new tests was proved load-bearing by mutation, and every one now fires in one of the two non-UTC runs.

Full gate in the order this repo needs (`generate-api-types && build:sdk` first): lint exit 0 with zero `--fix` mutations, typecheck exit 0, `prettier:check` red on only the six standing git-ignored generated files, server unit 2412 / 0. Merged-tree check exit 0 with both steps present.

One integration failure appeared in a full-chain run, `first-boot readiness line` timing out at 120s in `initializing`, while two other local worktrees were running servers. Re-run in isolation it passes in **1.88s**. Environmental contention, and this diff is a spec file plus a workflow, so it cannot reach server integration.

An independent adversarial review also confirmed the `03/01/2024` literal is not locale-driven (green under `de_DE`, `en_GB`, `ja_JP` and `C`), which was the main risk in asserting a formatted string, and that the `TZ=` prefix cannot alter any other step's semantics.

## Not verified

Whether `TZ=` takes effect on `windows-latest`. The steps run on all three platforms under `shell: bash` and are verified on macOS locally. The failure mode is benign either way: if Windows ignores it, those legs degrade to duplicate UTC runs costing ~0.3s each, and cannot go falsely red.

## Scope left alone

The same `dayjs.utc` hazard exists at three more live call sites in `components/filter/DimensionFilter.tsx`, and neither it nor `Notebook.tsx` has any spec. So these runs have exactly the new tests to act on today. Widening that coverage is a separate change rather than something to fold in here.


Sha-Bang (approved):
went at the tests rather than reading them — reverted `dayjs.utc(value)` back to `dayjs(value)` at `GivenInput.tsx:174` and every new case goes red somewhere:

```
Australia/Sydney       89 pass / 3 fail   (late-in-UTC-day, timestamp, timestamptz)
America/Los_Angeles    91 pass / 1 fail   (early-in-UTC-day)
UTC                    92 pass / 0 fail
```

that reproduces the asymmetry your comment claims at `cross-platform-tests.yml:86-87` digit for digit, so both zones earn their keep rather than one being decoration. baseline green in all three, so these pin real behaviour rather than an assumed one. TZ does propagate through `bun run` to the test child, checked.

approving. one for a follow-up rather than here: this pins the read direction, and the write direction has the same bug live today. when a date given starts unset `dateValue` is `null` (:174), so `AdapterDayjs` has no UTC-mode value to infer from and builds local midnight; `next.toDate()` then goes on the wire as `toISOString().slice(0, 10)`. east of UTC the user picks a day and the server is queried for the day before, box still showing the day they picked:

```
Sydney       calendar click day 5 -> 2026-08-04T14:00:00.000Z   wire 2026-08-04
Los_Angeles  same click           -> 2026-08-05T07:00:00.000Z   wire 2026-08-05
```

pre-existing on main, and the new suite is green through it in all three zones. `timezone="UTC"` on the `LocalizationProvider` at :178 would close it, plus one spec that fires a pick and asserts the `onChange` Date — that spec fails east of UTC today.

smaller:

- the two extra zone runs cover `packages/sdk` only. the server suites at :149-172 still run in the runner's UTC zone alone, which is where the date bug actually shipped — your own note at :85, "the one zone where a local-vs-UTC date bug is invisible", applies there too.
- the resolved-zone echo at :100-106 deliberately doesn't assert, because Bun's TZ handling on the Windows leg is unverified. if Bun-on-Windows ignores IANA `TZ`, that leg's two extra steps become duplicate UTC runs — green, fast, detecting nothing, which is the silhouette the comment says it wants to avoid. fatal on non-Windows only would keep it honest on two legs without risking the third.
- :249 says "one case per type rather than a loop" directly above an `it.each`. "one test case per type" reads truer.

PR #1020, opened 2026-08-14
feat(skills): a dashboards authoring skill

mlennie:
Adds `skills/malloy-dashboards`, so an agent can write a Publisher dashboard instead of a person hand-writing the tag. This is slice 9c of the #935 split. The skill is Kyle Nesbit's, cut from the handoff branch; what is new here is checking it against what actually shipped and fixing what had drifted or was never true.

**Based on `monty/pr935-slice8-dashboard-rendering`, not `main`.** The skill's further reading points at `docs/dashboards.md`, which lives on that branch, and the behaviour it documents is that branch's renderer. Note that a PR based on a feature branch silently does not run `app-playwright.yml` or `connection-integration-tests.yml`, since both filter to `branches: [main]`. `build.yml` and the cross-platform unit, integration and examples suites do run.

## What is in it

`skills/malloy-dashboards/SKILL.md`, a `.claude/skills` symlink so Claude Code sessions in the repo pick it up, a line in `skills/README.md` classifying it Publisher-specific and therefore never synced upstream, and a regenerated `skills_bundle.json`.

## What is deliberately not in it

The five `examples/storefront/dashboards/` files. They compile on no branch that exists today: all of them import `../givens.malloy`, which is on the 9b branch and the handoff branch only, and they further need model fields and `# drill` tags spread across two more branches. There are six such example files rather than five, since `examples/governed-analytics/dashboards/governed-overview.malloy` has the same blocker. They land once 9b, slice 6 and slice 8 are all on `main`. Shipping a skill with no runnable example seemed a smaller loss than shipping examples that do not compile.

## How it was verified

Every claim below was reproduced here rather than taken from the handoff. Against a live server on a package built to the skill's own templates, the four layout claims hold: four cards at `# colspan=3` render 253px each, two tiles at 6 render 521px, a table at 12 renders 1058px, and all three rows end flush at 1058, which is the criterion the skill states. `# break` does start the row. The controls render as documented, a select as a dropdown and a range as a slider. A URL parameter beats a `givens { }` starting value. Discovery, the manifest, the composite form and the shared-include pattern all behave as described, and the three lint strings the skill quotes are emitted verbatim.

The acceptance test is whether an agent can follow it, which nobody had run. Two fresh agents were given the skill and a package. Both produced working dashboards. The second was asked for drill-through between two dashboards and got it right, including writing `given=` explicitly and choosing a table tile for the drillable view, citing the corrected text for both.

## What was wrong, and the one that matters

The skill said that without `given=` a drill seeds the dimension name **upper-cased**. The code uses it verbatim, and `resolveDrill.ts` records the upper-casing as a bug it fixed because it "dropped the click in silence". Measured end to end: a drill written that way still paints the cell clickable, still navigates, and arrives as `?category=...`, which the destination drops, so the reader lands on a completely unfiltered page. The `to=self` destination quietly disappears from the menu. The package lint stays green throughout, because it upper-cases when it checks, so it confirms the wrong model rather than catching it. The same sentence was in `docs/dashboards.md`; slice 8 has corrected it there and found a related encoder defect underneath it.

Fifteen other corrections, each measured: no lint checks that a drill's destination declares the given it seeds; charts render to SVG rather than canvas, so the stated mechanism was invented even though the conclusion was right; a broken dashboard does not stay "listed with its error", and more importantly a failed reload leaves the previous version serving, so the listing answers 200 and looks healthy while the edit has silently not taken; wrapping a `##` tag only errors when the continuation is not valid Malloy, and otherwise the file quietly stops being a dashboard; the `# colspan` warning is real but arrives as a render log on the query response rather than in the package warnings; a guessed field name fails the whole package load rather than degrading; a broken `suggest` keeps its dropdown arrow rather than becoming a text box; and the composite example carried two tiles against a three-column grid, nine lines above the rule telling the reader to divide evenly.

The frontmatter description was silently truncated. It contained an unquoted ` # drill`, and a space-preceded `#` opens a YAML comment, so a YAML reader saw 160 of 309 characters and lost the entire "Use when the user asks..." clause, which is what makes an agent select a skill. The bundle generator reads the same line with a regex and saw all 309, so two consumers of one file disagreed and neither reported anything. Quoting it makes them agree. Worth knowing that `skills/malloy-document/SKILL.md` on `main` has the same bug from ` #(doc)` and is truncated to 22 characters, which is a separate one-line fix.

## Gate

Run in the required order after `generate-api-types` and `build:sdk`, three times across the review rounds. Typecheck and lint clean, prettier clean apart from the six standing git-ignored generated files, unit 2473/0, integration 297/0, sdk 304/0, skills 7/0, create-malloy-package 367/0, hammer 20/0. The bundle was regenerated with the committed generator rather than hand-edited, and `skills_bundle.spec.ts` was confirmed to fail on a stale bundle before regenerating, so it is pinning rather than passing vacuously.

## Follow-ups

One sentence is owed. Slice 8 has since made drill cells keyboard-reachable, so the affordance line should gain focus and Enter or Space alongside the pointer and hover underline. That code is not in this branch, so it is not claimed here; whoever rebases can add it. Separately, a given value of the form `f'X'` sent over the API still returns zero rows silently, though the manifest no longer advertises `f''` as the default, so the specific path that led into it is closed.


Sha-Bang (approved):
reviewed at 3df13c21. spot-checked the claims against code rather than taking them from the body, and the headline correction holds from three directions that agree: `drillGivenName` returns `field.tag?.tag("drill")?.text("given") ?? field.name` verbatim with no upper-casing (`resolveDrill.ts:132`), `Dashboard.tsx:191-202` folds case for its own given so `to=self` survives, and `paramCodec.ts` matches an arriving parameter exactly so `to=<slug>` drops it. the skill describes all three, correctly and separately. both drill lint strings are verbatim (`dashboard.ts:1191`, `:1273`), `warn` really is the `add` helper's default (`:804`), the 424 and the `stale: true` loadErrors entry both exist and the entry clears on the next clean reload, and the bundle entry matches SKILL.md byte for byte in description and body — regenerated, not hand-edited. symlink resolves, both reference docs are on main, CI green including Playwright at 7m30s.

**the body is stale in three places now that slice 8 has landed, and this repo squashes.** it leads with "Based on `monty/pr935-slice8-dashboard-rendering`, not `main`" and the note that `app-playwright.yml` silently does not run — the base is `main` and Playwright ran and passed. "What is deliberately not in it" says the example dashboards land once 9b, slice 6 and slice 8 are all on main; slice 8 is. and Follow-ups says the keyboard-affordance sentence is owed because "that code is not in this branch" — it is on main, and the skill already has the sentence, tab order, button role, focus styling, Enter and Space included. worth a pass before merge, since the body becomes the commit body.

**worth pulling in the `malloy-document` one-liner.** i swept all 31 skills: it is the only other instance, raw description 452 characters, a YAML reader sees `'Add documentation with'` — 22, exactly as you say. same bug, same one-line fix, found by this work. shipping it here closes the class instead of leaving a skill on main whose description a YAML consumer can't select on.

mlennie:
Both asks addressed. Body left as written, corrections here so the history stays intact.

**The `malloy-document` one-liner is in.** Quoted with single quotes, since the description already contains double-quoted trigger phrases. Its parsed description goes from 22 characters to 452, verified with `yaml.safe_load` rather than by eye. Swept all 31 skills afterwards and there are no remaining truncations, and I proved the sweep was live by feeding it the pre-fix line (flagged) and its quoted form (not flagged), so a zero there means zero rather than a broken check.

One correction worth recording, because I got it wrong first: I expected this to require regenerating `skills_bundle.json` and asserted in the original commit message that it had. It does not, and the bundle blob is byte-identical either side. Neither in-repo reader parses YAML: `build_skills_bundle.ts` and `packages/skills/src/index.ts` both match `/^description:\s*(.+)$/m` and pass the result through an `unquote()` helper, so the bundle always carried the full 452 characters. The truncation was only ever visible to YAML-parsing consumers outside this repo, which is exactly how it surfaced. Commit message amended to state the measured scope.

**Four stale claims in the body, your three plus one.** All are the same class: a claim whose truth lives in another branch, so nothing scoped to this diff could catch them.

The basing note. The base is `main`, and `app-playwright.yml` ran and passed rather than silently not running.

"What is deliberately not in it" said the example dashboards land once 9b, slice 6 and slice 8 are all on main. Slice 8 is.

Follow-ups said the keyboard-affordance sentence was owed because that code is not in this branch. It is on main, and the skill already carries the sentence, including tab order, button role, focus styling, Enter and Space.

The fourth was mine, found by re-reading the section rather than patching the three you named: I had written that the lint section "maps each warning to its cause", which oversells it and contradicts an earlier finding on this branch that two causes emit identical logs. What the section actually does is name the places the lint misleads.

**One thing I deliberately did not claim.** You reported Playwright green at 7m30s and I believe you, but I cannot verify CI from here, and the rule I have been applying on this branch is that a quoted measurement gets attributed or re-derived rather than absorbed into my own prose. So it is not asserted anywhere in the branch.

Branch is up to date with `main` via a local signed merge rather than the Update branch button, which would have produced an unsigned commit and failed the DCO check.

PR #1022, opened 2026-08-14
feat(server): retire the materialization.* spelling for query metadata

housejester:
Query metadata is not a build setting, and the `materialization.` spelling says it is.

Its properties ride **every statement a package's sources issue** — a served query as much as a
build. Declaring them inside the build-policy block describes a scope the feature does not have, and
sends an author hunting through build settings for a way to label their traffic.

Stacked on #1018, which is what made the mismatch real: before it, a declaration only reached
statements a build issued, so the name was narrow but accurate.

## The canonical home is the manifest root

```json
{
  "name": "orders",
  "queryMetadata": { "team": "finance", "tier": "gold" },
  "materialization": { "scope": "version" }
}
```

`materialization.queryMetadata` still works and warns, on the pattern `scope` already established in
this manifest: both homes read, both written whenever the server rewrites the file, so an older
publisher still reads the right value.

**One deliberate difference from `scope`.** A conflict between the two homes warns and prefers the
root, where a conflicting `scope` throws and fails the load. Scope decides whether an artifact is
version-owned, so guessing could reuse a table across versions that was never meant to be shared —
worth failing a load over. A tag is observability only, excluded from `sourceEntityId` and from build
identity, so the worst a wrong guess yields is a mislabelled statement. Failing a package load over a
label would break the rule the feature rests on.

## Two things the docs had wrong, not the code

**The bare model-file form already worked.** `## queryMetadata.*` resolves today, with
`## materialization.queryMetadata.*` overriding it where both appear. The docs listed only the longer
spelling.

**Any source can be tagged, not only a persisted one.** `queryMetadata` is a sibling of `persist` in
the `#@` namespace rather than a key inside it, so this has always worked:

```malloy
#@ queryMetadata.tier="gold"
source: daily is orders -> { aggregate: n is count() }
```

The docs listed layer 4 as `#@ persist queryMetadata.…`, which reads as though `persist` is required.
It never was.

## The non-persist source had no validation path

That second one was not just undocumented. The publish gate reads persist sources from the build
plan, so a source that persists nothing was invisible to it — a reserved or malformed name there
published clean and vanished with only a metric behind it, which is the failure the publish warnings
exist to prevent.

It now has its own level, filtered against the build plan so a persist source is not reported twice
under two labels. With this, every one of the five declaration levels has a validation path.

## Tests

**2,431 pass, 0 fail, 3 skip** across 123 files; `tsc --noEmit` and prettier clean.

Added: the root home read with no warning; the enveloped form honoured with a deprecation; both homes
agreeing staying quiet, because that is the state the server itself writes; the two disagreeing
warning and preferring the root; a manifest that declares tags and no build policy still producing a
config, which is the shape the canonical form makes; and the publish gate warning about a reserved
name on a source that persists nothing.

## Sequencing

Worth landing before the feature is announced. Nobody has written a `queryMetadata` block outside a
test yet, so the deprecation costs nothing today. After release it is either the `materialization.`
spelling forever or a second migration aimed at people who just learned the first one.


packages/server/src/service/package.ts
```diff
@@ -744,6 +754,104 @@ export class Package {
...
+            .map(({ sourceName, queryMetadata }) => ({
+               subject: sourceName,
+               modelPath: model.getPath(),
+               size: Object.keys({
+                  ...(packageDeclaration ?? {}),
+                  ...(modelDeclaration ?? {}),
+                  ...queryMetadata,
+               }).length,
+            }));
+      });
```
Sha-Bang: **P1 — Merge budget gate misses package+model-only overflows**

This only sizes sources returned by `getDeclaredSourceQueryMetadata()` — i.e. sources that declare their own `#@ queryMetadata`. Package + model layers still ride every query via `composeDeclaredQueryMetadata`, including sources with no source-level bag.

Author budget is 11 (`MAX_PROPERTIES − RESERVED_CONTEXT_PROPERTIES`). Package 6 + model 6 (disjoint keys) → merge size 12 → should warn. With no source-level declarations, `effectiveBagSizes` is `[]` and the package publishes clean — exactly the failure the new comments/tests describe.

The unit test only mocks `effectiveBagSizes`; nothing asserts `Package` actually produces sizes for package+model-only.

**Fix:** For each model, compute merge size for every top-level source (`isSourceDef`), with source-own possibly `{}`. Add an integration test on `Package.getPackageMetadata().warnings` for package+model only (no `#@ queryMetadata`).

packages/server/src/service/package_manifest.ts
```diff
@@ -99,6 +99,86 @@ const SCOPE_ROOT_DEPRECATION =
...
+      materializationRaw && typeof materializationRaw === "object"
+         ? (materializationRaw as { queryMetadata?: unknown }).queryMetadata
+         : undefined;
+   const rootDeclared = rootRaw !== undefined && rootRaw !== null;
+   const envelopeDeclared = envelopeRaw !== undefined && envelopeRaw !== null;
+
+   if (rootDeclared && envelopeDeclared) {
+      // Both homes agreeing is the transition state the server itself writes,
+      // so it is not worth a word to an operator who cannot act on it.
+      if (JSON.stringify(rootRaw) === JSON.stringify(envelopeRaw)) {
```
Sha-Bang: **P2 — Dual-home “agreement” is key-order sensitive**

Equal bags with different key order (`{"a":"1","b":"2"}` vs `{"b":"2","a":"1"}`) are treated as a conflict. Root still wins with the same values (no mislabel), but operators get a spurious conflict warning. Server dual-writes usually keep order; hand-edited manifests may not.

Prefer comparing canonicalized entries (sorted keys / stable serialize) rather than raw `JSON.stringify`.

packages/server/src/service/build_plan.ts
```diff
@@ -417,15 +417,29 @@ export function resolveQueryMetadata(
...
+   // is the legacy one. Query metadata migrates the other way: the bare `##
+   // queryMetadata.*` is canonical and `## materialization.queryMetadata.*` is
+   // deprecated. So the list is consumed as-is rather than reversed, putting the
+   // deprecated spelling underneath. Reversing it let the spelling we tell
+   // authors to stop writing silently override the one we tell them to write.
    const ordered: QueryMetadata[] = [
-      layers.packageMaterialization?.queryMetadata ?? {},
-      ...modelTagLayers(layers.modelTag).map(tagQueryMetadataLayer).reverse(),
+      layers.packageDeclaration ?? {},
+      ...modelTagLayers(layers.modelTag).map(tagQueryMetadataLayer),
```
Sha-Bang, line 447: **P2 — Function JSDoc above still documents the old precedence**

This comment correctly describes bare-over-envelope. The function-level JSDoc a few lines up still says the opposite: source `#@ persist queryMetadata.*` > `## materialization.queryMetadata.*` (bare underneath) > package `materialization.queryMetadata`.

OpenAPI was updated; that docstring was not. Please align it with this comment and the OpenAPI wording.

Sha-Bang (changes requested):
Request changes — dual-home migrate/PATCH/clear and model-file ordering look solid; one claimed gate is still incomplete.

### Summary
- **P1:** Merge budget sizes only sources that declare their own `#@ queryMetadata`, so package+model overflows with no source tags publish clean (the case the new comments claim to catch).
- **P2:** Stale JSDoc on `composeDeclaredQueryMetadata` still documents envelope-over-bare.
- **P2:** Dual-home agreement uses `JSON.stringify` (key-order sensitive → spurious conflict warning).

Looks good otherwise: null-preserve / `{}`-clear at both wire homes, PATCH dual-write without inventing a `materialization` block, model-file order fix, `safeSourceTag` + `isSourceDef`, OpenAPI dual-home wording, and strong schedule/surface + manifest tests.

Sha-Bang (approved):
Prior findings addressed in d30785d4 — approving.

- **P1 budget floor:** Sized per model as package ⊕ model, with source merges only when the floor is under budget; e2e test covers package+model with no `#@`.
- **P2 key-order compare:** `sameQueryMetadataBag` / `keySortedJson` with tests for agreement and real conflict.
- **P2 JSDoc:** Canonical precedence, with the freshness trap called out.

No new blocking issues.

PR #1023, opened 2026-08-14
fix(server): report build failures in their own collection, deprecating ManifestEntry.error

sagarswamirao:
Follow-up to #1004, which made a build that loses one source keep the rest. The behavior stands; this changes where the failure lives, and starts a deprecation cycle for where it used to.

## The problem

#1004 recorded a failed source as a `ManifestEntry` carrying an `error`. But `ManifestEntry` requires `physicalTableName`, so a failed entry named a table that was never created:

```yaml
ManifestEntry:
  required: [sourceEntityId, physicalTableName]
  properties:
    error: { type: [string, "null"] }   # present only on a source that failed
    physicalTableName: { type: string } # required, so a failed entry names a table anyway
```

Every consumer that resolved an entry to a real table had to ask about `error` first, and nothing in the type made it do so. #1004 routed all of them through one shared predicate (`isFailedEntry`), which was the best available fix while the two states shared an object -- but it is a convention enforced by review, not by the compiler. A new consumer that read `physicalTableName` without asking was well-typed and wrong.

That cost was measurable: six guards, each needing its own test, all of them existing only because the failure rode the same object.

## The shape

`entries` keeps the sources that built. A sibling collection carries the ones that did not:

```yaml
BuildManifest:
  properties:
    entries:   { additionalProperties: { $ref: ManifestEntry } }
    failures:  { additionalProperties: { $ref: SourceFailure } }   # keyed by sourceEntityId
SourceFailure:
  required: [sourceEntityId, reason]
  properties:
    reason:              { type: string }   # redacted, as before
    materializedTableId: { type: string }   # the caller's row identity from the instruction
    physicalTableName:   { type: string }   # the table it was HEADED for; names nothing built
```

Once the deprecation below completes, a consumer iterating `entries` cannot reach a failed source, and the guard stops being a rule and becomes a property of the shape.

`SourceFailure` keeps the `physicalTableName` the source was headed for, because correlating a failure with the request that asked for it is genuinely useful. That it is not a table this build created is stated on the field -- and, for a failed *rebuild*, the prior generation still sits under that same name, so resolving it serves stale data rather than nothing.

## `ManifestEntry.error` is deprecated, not removed

An earlier revision of this branch removed it, on the argument that it had never shipped. That was wrong, and @housejester caught it: the newest published `@malloy-publisher/server` is **0.0.246**, and both 0.0.245 (2026-08-15) and 0.0.246 post-date #1004's merge, so both declare the field.

```
$ npm view @malloy-publisher/server version   → 0.0.246
$ npm pack @malloy-publisher/server@0.0.246 && \
    grep -A2 'error:' package/dist/app/api-doc.yaml   → type: [string, "null"]
```

So `error` is marked `deprecated: true` and a failure is written to **both** collections for one cycle. A consumer reading `error` keeps working unchanged while it moves to `failures`.

Removing it now would also have broken it *silently*: the field is optional, so a client pinned at 0.0.245/0.0.246 reads `undefined` on every entry and sees a failed source as **absent** rather than **failed** -- indistinguishable from a source nobody asked about, which is the fail-dangerous reading.

There is a known consumer in this position. `ms2data/service`'s control plane reads the field directly (`MaterializationFinalizer.reportedFailure` → `entry.getError()`), documented there as *preferred over* its own synthesized "no table produced" message. Dropping the field would have regressed it to that placeholder with every test still green on both sides -- an inert-field failure, which is the defect class this work exists to remove.

### Legacy shape at rest

A manifest committed by 0.0.245/0.0.246 on a partially-failed build stores the failed source **inside** `entries`, with its required `physicalTableName`. That survives the upgrade, and auto-run physical names are stable (`selfAssignTableName`), so the name resolves to the **prior generation's real table** -- served as though fresh, not a lookup miss.

Rather than restoring the six guards, the legacy shape is dropped at the two boundaries a persisted manifest is read through:

- `splitManifestEntries` -- the single chokepoint for both the host-manifest fetch and the local-store serve rebind
- `getMostRecentManifestEntries` -- reuse (skip-if-unchanged) and reference resolution

Producers and internal consumers stay clean, and both call sites are commented for deletion with the field.

This is not hypothetical: `ms2data/service` pins **0.0.246**, so that deployment is already writing this shape.

## A failed source could also land in `entries` beside its failure

Separately from the legacy case, and also @housejester's find. `entries[sourceEntityId] = entry` ran for every seed unconditionally, and the failure path wrote to `failures` without removing it -- so a source that is both seeded and instructed-and-failed appeared in both, the entry naming the prior generation's table.

Reachable because `referenceManifestToEntries` applies no exclusion for instructed sources, and `seedFromBoundManifest` excludes on `instruction.sourceEntityId` while the failure keys on the content-addressed id -- which the surrounding contract explicitly allows a caller to derive by another scheme. Under #1004 it was masked: `entries[id] = {…error}` overwrote the seed and the shared guard skipped it.

The failure now overwrites the seeded entry rather than sitting beside it.

## What the guards become

Verified individually rather than assumed. After the deprecation completes, five call sites and the shared predicate are gone:

| site | after |
|---|---|
| `manifest_loader` serve split | legacy tolerance only, deleted with the field |
| `materialization_serve_transform` serve bindings | gone |
| `materialization_service` auto-load bind | gone |
| `materialization_service` downstream seed | gone |
| `materialization_service` carry-forward reuse | legacy tolerance only, deleted with the field |
| `tallySources` | counts two collections instead of filtering one |

## Tests

The specs #1004 added to pin the six guards are the regression net. Where a guard's premise became structurally unreachable, the test was retargeted at the reason a skip can still happen rather than left asserting a state the types now forbid, and the two-sided seed assertion stays two-sided so a change that binds nothing cannot pass either.

Three tests added for this revision: the seed/failure overlap, and the legacy tolerance at each read boundary.

Mutations run against the result, each individually:

- revert the split so a failure is written to both collections → 2 red
- drop `failures` from the persisted manifest → initially **nothing** red, because every test stubbed `commitManifest`; the one link carrying a reason to a consumer was untested. Now covered by two tests against the repository write (the reason persists; the key is omitted rather than `{}` on a clean run).
- remove either legacy tolerance → its test red

`bun test` 202 pass / 0 fail on the three affected specs; `tsc --noEmit` and `lint:server` clean.

## Metrics note

`outcome="built"` on the sources counter is counted from what the build returned, where before this release it was the length of the instruction list, so a skipped instruction no longer reports as built. The change came with #1004 rather than this diff (`git log -S` lands on `ceeacb46`); it is in the release note because the counter predates the release and a deployment trending `built` sees a step change.


housejester (commented):
The shape change is the right one, and the test work behind it is strong — the two-sided producer assertion and the mutation that caught `commitManifest` being entirely stubbed are exactly the checks that keep a refactor like this honest. Two things below, though: the "No migration" argument that the guard-deletion rests on doesn't hold, and the new invariant has one reachable hole.

## The published version is 0.0.246, not 0.0.244 — `ManifestEntry.error` shipped

The PR says the newest published `@malloy-publisher/server` is `0.0.244`, "which predates it, so no consumer can have pinned a version that declares it." The registry disagrees:

```
$ npm view @malloy-publisher/server version         → 0.0.246
$ npm view @malloy-publisher/server time --json     → 0.0.245  2026-08-15T03:31Z
                                                       0.0.246  2026-08-15T14:08Z
$ git log -1 --format=%ci ceeacb46   # #1004        → 2026-08-14 20:38 +0000
$ npm pack @malloy-publisher/server@0.0.246 && \
    grep -A3 '^    ManifestEntry:' package/dist/app/api-doc.yaml
                                                    → error: type: [string, "null"]
```

Both published versions post-date #1004 and carry the field.

The wire-compatibility consequence is mild — `error` was optional, so a client pinned there won't break structurally, it will just read `undefined` on every entry and see a failed source as *absent* rather than *failed*. That default is fail-dangerous rather than fail-safe, and the release note should say `error` was removed and name `failures` as the replacement; right now it documents only the addition.

The consequence that matters more is **data at rest**. A manifest committed by a 0.0.245/0.0.246 server on a partially-failed build stores the failed source *inside* `entries`, with its required `physicalTableName` — in `publisher.db` and in the bound manifest file. That state survives the upgrade, and every guard that used to handle it is now gone:

- `environment_store.ts:768` → `rebindServeBindingsFromLocalStore` → `splitManifestEntries` (`manifest_loader.ts:131`) binds the name on package load. Auto-run physical names are stable (`selfAssignTableName`, `materialization_service.ts:998`), so that name is the **previous generation's table** — the "served as though it were fresh" case the deleted comments named, not a benign "table not found".
- `getMostRecentManifestEntries` (`:1198`) → the reuse branch at `:1027-1033` carries the failed prior entry forward, retiring the source from every later run. That is precisely what the removed `!isFailedEntry(prior)` existed to stop.
- `seedFromBoundManifest` → `fetchManifestEntries` → the same split, seeding a downstream `FROM` from a table that was never created.

Suggested fix, which keeps the structural win rather than undoing it: don't reinstate the six guards — add legacy tolerance at the two **read** boundaries only. `splitManifestEntries` is a single chokepoint (its own doc says it is "shared by the URI-fetch path (host manifest) and the local-store rebind"), and `getMostRecentManifestEntries` covers reuse plus `resolveReferencesFromStore`. Two lines of `if ((entry as {error?: unknown}).error) continue;`, commented as legacy shape written by 0.0.245–0.0.246 and removable after a few releases. Producers and internal consumers stay clean.

Could you confirm whether any server on 0.0.245/0.0.246 has committed a manifest from a build where some sources failed and some succeeded? If we're confident none has, this reduces to the release-note correction and I'm happy to drop the rest.

## A failed source can still land in `entries`

`entries[sourceEntityId] = entry` at `materialization_service.ts:1554` runs for **every** seed entry unconditionally, and the failure path at `:1695` writes into `failures` without removing the seed. So a source that is both seeded and instructed-and-failed ends up in both collections. Confirmed with a throwaway test against this branch:

```
entries:  {"bbadbbbbbbbbbbb": {… "physicalTableName":"bad_v1_prior_generation" …},
           "bgoodaaaaaaaaaa": {…}}
failures: {"bbadbbbbbbbbbbb": {… "physicalTableName":"bad_v1", "reason":"boom"}}
```

It is reachable because `referenceManifestToEntries` (`:1064`) applies no exclusion for instructed sources at all, and `seedFromBoundManifest`'s exclusion keys on `instruction.sourceEntityId` while the failure keys on the content-addressed `computeSourceEntityId` — which the surrounding comments explicitly allow a caller to derive by a different scheme. Under #1004 this was invisible, because `entries[id] = {…error}` overwrote the seed and the shared guard then skipped it.

`delete entries[sourceEntityId];` beside the `failures[…] =` write fixes it. One line, and it makes "a consumer iterating `entries` can no longer reach a failed source" unconditionally true instead of true-given-disjoint-seeds. The auto-run path is disjoint by construction at `:1027-1041`, so this is the orchestrated path only.

## Smaller notes

- `[Unreleased]` isn't evidence of unreleased. `RELEASE_NOTES.md`'s own header says sections are copied into the GitHub release and version-stamped *manually*, after CI publishes — and several sections above this one still read `[Unreleased]` although 0.0.242 and later shipped, #1004's among them. Not a problem with this PR's code, but it's the label the no-migration argument leaned on; `npm view` is the check when an argument depends on what's published.
- The `outcome="built"` metrics note is accurate for the release, but the instruction-list → build-returned change came in #1004, not here (`git log -S "sourcesFailed: returned.filter(isFailedEntry)"` lands on `ceeacb46`), and this PR's `tallySources` edit leaves the count identical. Worth attributing so a later reader doesn't look for it in this diff.
- In the overlap case above, `tallySources` (`:357-365`) counts one source as both `reused` and `failed`. Falls out of the same one-line fix.
- `SourceFailure` drops the `materializedTableId` that #1004's failed entry carried. `physicalTableName` is the stated correlation key, but `materializedTableId` is the row identity the caller actually sent in the instruction — free to add now, an additive schema change later. Worth a check with whoever consumes the manifest.
- Reproduced the test claim locally: `bun test` on the three specs is 199 pass / 0 fail.

**Top three before merge:** 1) legacy-shape tolerance at `splitManifestEntries` + `getMostRecentManifestEntries`, or establish that no 0.0.245/0.0.246 server has written such a manifest; 2) `delete entries[sourceEntityId]` on the failure path; 3) correct the "No migration" section to 0.0.246 and document `error`'s removal in the release note.

<sub>Review pass by Claude Code (session "Deft Possum"), on behalf of @jimmeh — findings verified against the branch at c845bce8.</sub>

housejester (approved):
👍 once build failure fixed.

PR #1025, opened 2026-08-14
docs(html-data-apps): the page runtime does pass givens, and the safety advice built on that

mlennie:
## What this fixes

`docs/html-data-apps.md` tells data-app authors that the no-build page runtime **does not** pass per-query `givens`. It does, and has since #891. `packages/server/src/runtime/publisher.js:99` sends them, and the runtime's own header comment already documents them as safe parameterization rather than string interpolation. The `opts` table on that page never listed `givens` at all.

## Why it is worth a PR rather than a follow-up note

The false premise carried the security advice. The paragraph tells authors never to interpolate untrusted input into a `run:` string, and then denies that the parameterized alternative exists, which leaves hand-constraining the string as the only route the page offers. **A doc that denies a safe mechanism exists pushes readers toward the unsafe one.**

## Two files, not one, and the second is the more interesting half

The replacement wording was lifted from `skills/malloy-html-data-app-runtime/SKILL.md` rather than composed fresh, deliberately: that text was already reviewed, already shipped, and carries two limits the doc lacked (the `filter<T>` validation caveat, and that parameterization is not an authorization boundary).

**The reviewed prose was itself wrong.** It named "an operator's per-package config" as something that strips or finalizes a client-supplied given. There is no such control: the only `finalizeGivens` in the tree is a comment about a Malloy config Publisher never sets, and `docs/authorize.md:230` says identity-bound givens are "a planned milestone… It is not implemented yet", with `docs/row-level-access.md:75` agreeing.

So the skill is corrected in the same commit. Fixing only the doc would have recreated the drift on a file whose own text says it tracks that skill.

Two smaller corrections came out of the same check: binding happens server-side rather than in the page runtime, and `sourceName` alone is a 400.

## Verification

The capability is not merely present, it is exercised: the storefront data app on `monty/pr935-slice9b-storefront-data-app` drives its filters entirely through this path, measured live against a running server, 25,356 line items narrowing to 2,535.

The three false claims were each checked against the tree with a working control rather than by reading, and the "no per-package control exists" negative was established by grepping for the mechanism and finding only a comment about a config Publisher never sets.

Merges clean with `main` at `90a32fa3`.


Sha-Bang (approved):
checked the central claim against the runtime rather than the prose. `publisher.js:99` (`if (opts.givens) body.givens = opts.givens;`) sits in `rawQuery`, which both `query` (:131) and `queryFull` (:134) go through, so "does pass givens" holds on both paths and the flat statement isn't an overcorrection. the old "does not pass per-query givens" was simply wrong, and the runtime skill already said the opposite — so this settles a contradiction that was live between the two files.

the two new `opts` claims hold as well, per `model.ts:4279-4312`: `sourceName` alone falls past both branches into the `BadRequestError`, and `queryName` alone takes the `queryName && !query` branch and emits `run: <queryName>` with no source prefix. dropping the "or an operator's per-package config" reassurance is right too — there's no givens handling in `package.ts` or `environment_store.ts` to point at, so the more conservative wording is the accurate one.

one coordination thing, which is why I'd flag rather than nit it. #79 merged on the 17th and its whole job was making these two skill files byte-identical to publisher, so the copy-back could happen. after 1025, `skills/malloy-html-data-app-runtime/SKILL.md` differs from agent-skills in four places: two where publisher is now ahead (:21, :70 — this PR), and two where publisher is still behind (:46, :133 — #79's fixes that never came back). `skills/malloy-html-data-apps/SKILL.md`, which this PR doesn't touch, is behind in two more (:44, :81).

concretely: publisher main still names `html-data-app`, the example #935 deletes, in three places, and still carries "The shipped example renders rows with a plain chart library instead, so this path is not exercised there" — the dangling sentence #79's head commit fixed. worth folding the copy-back in while you're already in the file, otherwise the skills keep serving the superseded safety advice and publisher keeps naming a deleted example.

aside, and not yours: the 400 at `model.ts:4310` reads "(Query AND !sourceName) OR (queryName AND sourceName) must be defined", which misdescribes what the code does, since `queryName` alone is legal. your new doc text is more accurate than the error a caller actually sees.

Issue #1016, opened 2026-08-14
packages.spec.ts leaks its environment when the git-clone test flakes, cascading into ~62 unrelated failures


mlennie:
`packages/app/tests/playwright/packages.spec.ts:70`, "create env → add package from git → open → back → delete package → delete env", is flaky and fails in a way that poisons the rest of the run.

The test creates a disposable environment, adds a package by cloning `https://github.com/credibledata/malloy-samples/tree/main/ecommerce` over the network, exercises it, and deletes the package and the environment at the end. It carries a 180 second timeout, which is a fair signal of how long that clone can take.

Measured: two identical runs, one passed and one failed.

The problem is not the flake, it is what the flake leaves behind. When it times out it does so **after** creating the environment and **before** reaching the delete, and there is no `afterEach` or `afterAll` in the file, so nothing removes the environment. Every later test that assumes a known set of environments then fails against the leftover. One flaky test turns into roughly **62 failures**.

That is worse than it sounds, because of how it presents. You are usually running this suite against a change of your own, and what you see is a red wall across areas you did not touch. It reads exactly like a catastrophic regression in your own branch. It cost about an hour to work out that the branch was fine and one networked test had littered.

Two fixes, and the first matters more than the second:

1. **Clean up regardless of outcome.** An `afterEach` that deletes the environment if it still exists turns this from 63 failures into 1. Worth doing even if the flake is never addressed, because it makes the failure honest: one test failed, and it says which.
2. **Reduce the flake.** The test depends on cloning a GitHub repo mid-run, which is the least reliable thing in the suite. A local fixture, or a pre-seeded package, would take the network out of the hot path. Whether the git-add path deserves its own dedicated coverage is a separate question worth asking.

Related in kind: the legacy `e2e/` suite has a known problem where a killed run leaves an orphaned `git clone` behind. Both are network-dependent tests that strand state when interrupted, and both are found the hard way.

PR #1026, opened 2026-08-14
feat(sdk): Console polish — say what this server does, and give each content type its own signal

mlennie:
**Slice 10 of the ten-way split of #935**, and the last one. Kyle's PR body proposed ten smaller PRs in dependency order; this is the Console-side remainder.

**Base is `monty/pr935-slice8-dashboard-rendering`, not `main`.** Read it against the merge base. At submit time this is 13 commits and 16 files; if GitHub shows ~25, the base reverted to `main` and slice 8's work is bundled in.

## What this does

- **Six feature cards on the home page instead of three.** One was titled "Notebook dashboards", which names neither of the two surfaces it straddled, and linked the publishing setup guide. Dashboards, data apps and the governance model had no card at all. A closing paragraph names connections, materialized tables and the REST API, which have docs but do not earn a card.
- **`DOC_LINKS` gains a `REPO_DOCS` block**, six links to Publisher's in-repo reference docs, which for several features are the only write-up there is. A spec checks each target exists in the tree.
- **`docs/choosing-a-surface.md`**, the notebook / dashboard / HTML data app comparison. `docs/malloyyo-dashboards-design.md` has referenced this file at four places since it merged; it now exists.
- **One color and one glyph per content type on the package page.** Four of the six rows were being handed the same teal by four separate call sites, so color distinguished two kinds out of six. The row now derives both from one `type` prop.
- **`Add Connection` matches its siblings**, contained with a start icon, the only one of the three add-triggers still outlined.

## Two defects found by looking at the running product, not the diff

**Four of six rows shared a color.** Correcting the four would have left the seventh caller free to get it wrong again, so the pairing moved into the row. `CONTENT_TINT` and `CONTENT_GLYPH` are both `Record<ContentType, …>`, so adding a content type is a compile error at both tables rather than a row that silently paints itself like its neighbour or renders an empty `<svg>`. Both proven by adding a seventh type and reading the errors.

**The dashboard glyph and the Package Data glyph were the same icon.** Same rounded square, same horizontal rule, differing only in whether one vertical stroke ran full height at `x=9.5` or half height at `x=12`. At 18px they read as one glyph, and both are sections on the same page. Confirmed by cropping both at 4x. Replaced with a four-pane grid, which still satisfies the original comment's constraint of not colliding with the data app's browser window; that comment had simply reasoned about the wrong neighbour.

## Colors, measured rather than asserted

The three added accents each clear 3:1 against white, WCAG's minimum for a graphical object, since they sit behind a white glyph: violet 5.5, magenta 4.6, moss 3.9. The nearest chart-series hues do not (sage 1.8, pink 2.7, series violet 2.9), which is why these are their own family rather than borrowings.

**`MALLOY_BRAND.teal` is 2.5:1, below that bar.** It is inherited from the logo, it backs exactly one row, and it is a brand decision rather than a component one, so it is named where the numbers are rather than quietly fixed or quietly asserted away. The spec grandfathers it **by content type rather than by color**, so a new type handed teal is not exempt, and fixing teal later leaves the entry unused rather than turning a correct fix into a red suite.

## One test change that is not cosmetic

`packages/app/playwright.config.ts` gains `workers: 1`. `fullyParallel: false` serializes only **within** a file; separate files still run concurrently, and `notebook-givens`, `notebook-authorize` and `notebook-readme-links` all write fixture files into the same `publisher_data/examples/storefront` and all POST `?reload=true` on it. Reproduced rather than reasoned about: a full run failed `notebook-givens` on a navigation timeout while another file was mid-reload, and the same spec passes 9/9 in isolation, which is the tell that separates contention from a broken assertion. Related to #1016.

## Verification

- Full gate green: typecheck 0, lint 0 with no `--fix` mutations, prettier red only on the six git-ignored generated files, SDK 321/0, unit 2473/0, integration 297/0.
- Playwright 98 pass. Three failures, none caused by this change: two `environment-connections` cases need the `bigquery` stub CI injects and reproduce in isolation, and `package-dashboards:261` is a pre-existing layout-timing flake on the base branch, since fixed there.
- Browser-verified on the built bundle in both light and dark, with the colors read from `getComputedStyle` rather than from a screenshot: six kinds, six colors, none repeated, unchanged across the theme toggle, zero console errors.
- Every `REPO_DOCS` target verified present in the actual merged tree of `main` + slice 8 + this branch, using a probe with a known-present and a known-absent control so the check can come out both ways.
- `docs/screenshots/console.png` regenerated, because this change made the page's own illustration contradict the sentence above it.

## Review notes

Two review rounds, seven findings then eight, all fixed. The two worth knowing about:

- The API-spec link was the only root-relative `href` in the SDK. It works in stock Publisher, where the Console and the server share an origin, and 404s in any host that embeds `@malloy-publisher/sdk`. Now built from `serverBaseUrl(server)`, and re-verified in a browser.
- `choosing-a-surface.md` opened by saying nothing about governance changes with the choice. `docs/security-posture.md` says the opposite in its own words: an HTML data app runs author-written JavaScript with the viewer's authority and notebooks and dashboards execute no author code, which it calls a deliberate property. That is the single biggest thing that does change, so the guide now leads with it.

**Not claiming convergence.** Round two returned eight findings on a tree round one had already cleaned, which is the expected shape when each pass reviews the diff the previous pass wrote. The severity trend is the real result: round one had two defects that would have shipped, round two had one factual reversal and a set of precision corrections.


Sha-Bang (approved):
reviewed at 2c0e6734. checked the numbers rather than reading them — all seven contrast figures are exact, teal is the only tint under the bar and it's the one disclosed, the connections and "four of the six rows" claims are exact, every doc link and anchor resolves, and `choosing-a-surface.md` is the one doc target still off main, which is what the corrected docblock now says. sdk 352 pass, tsc clean. three things:

**`workers: 1` costs about four minutes a run.** Playwright on this PR: 9m40s serialized, against 5m42s on 1043 with the same specs running parallel — so ~1.7x, on every PR, against a 30-minute timeout. plenty of headroom, but it's a real standing cost for contention among four spec files. the narrower fix is the pattern 1043 itself introduced: `package-dashboards.spec` registers its own temp environment in `beforeAll` and removes it after. if the three storefront writers did the same, the other twelve wouldn't need serializing. also, the comment singles out `packages.spec` as the environment churner but `environments.spec.ts:59` and `package-dashboards.spec.ts` both create and delete one too — worth listing, since this comment is what someone will read before relaxing it.

**`external={false}` on the api-spec link is inverted relative to the risk the comment right above it is reasoning about.** `Home.tsx:288` opens `${serverBaseUrl(server)}/api-doc.html` in the same tab. that's right for the Console, but `Home` is on the public surface, and the comment at :283 says outright that "an SDK consumer's page origin need not be that server" — which is the case where same-tab unloads the customer's app instead.

**`styles.ts` tells a brand retuner the 3:1 bar isn't theirs, and the new spec says it is.** the docblock says "the bar is stated for these three only" and that `MALLOY_BRAND` is "a brand decision rather than a component one" — but `CONTENT_TINT` assigns `MALLOY_BRAND.orange` to dashboard and `.darkBlue` to model, and `GRANDFATHERED` lists only `report`, so both are asserted at 3:1. orange is 3.09. retuning the logo orange from `styles.ts`, on the comment's own invitation, reds a suite in a file you never opened. either grandfather the two by type the way `report` is, or say in `styles.ts` that two of the three are now bound.

mlennie:
Housekeeping on the basing note, leaving the body as written since it was accurate at submit time.

**The base is now `main`.** This slice was opened against `monty/pr935-slice8-dashboard-rendering` because it extends that slice's files. #1043 merged as a squash, which left slice 8's commits in this branch's history as duplicates of content that had already landed under a different SHA, so the branch was rebased with `git rebase --onto origin/main` to lift only slice 10's own commits clear. The diff is slice 10's work alone and nothing in it duplicates #1043.

One specific thing to ignore in the body above, because following it now gives the wrong answer: it says that if GitHub shows a `main` base then "slice 8's work is bundled in". That diagnostic inverts after the rebase. A `main` base is now the correct state and slice 8's work is *absent* from this diff, not bundled into it. `git log --oneline origin/main..HEAD` carries no slice 8 commits.

The rebase was verified by content rather than by the rebase exiting cleanly: all files blob-hashed before and after, 16 of 17 byte-identical, the one change being `docs/dashboards.md` adopting slice 8's `baseURL=".../api/v0"` fix. A pre-rebase backup ref is retained locally. Gate green on the rebased tree.

Sha-Bang:
correcting my own number above: i used 1043's 5m42s as the parallel baseline and called `workers: 1` a ~4-minute, 1.7x cost. 1020 has since run Playwright parallel against today's main with the same spec set — 7m30s. against the 9m40s here, the real cost is about 2m10s, ~1.3x, roughly half what i said. the point stands but it's a smaller number than i gave it, so weigh the per-spec-isolation suggestion accordingly.

PR #1027, opened 2026-08-14
  examples(storefront): drive the data app from givens, and let Malloy escape them

mlennie:
**Slice 9b of the ten-way split of #935.** The storefront HTML data app: it now drives its filters through givens, and lets Malloy escape them instead of doing it by hand.

**Base is `monty/pr935-slice5-given-control-spec`, not `main`.** At submit time this is 3 commits and 20 files; if GitHub shows eleven commits, the base reverted to `main` and slice 5's work is bundled in.

## Why this is not just an example refresh

`examples/storefront/public/app/format.js` carried a **second copy of the broken filter encoder**. It is the same defect #985 found and fixed on the SDK side: the encoder quoted with double quotes, and Malloy's filter grammar has no double-quote quoting, so a value beginning with a minus inverted the filter and one containing a percent removed it entirely. Copying the fix across would have made a third copy.

So the page now **delegates to Malloy's own filter printer** rather than restating its rules. That is stronger than validating against a known set, because it is correct for values nobody enumerated.

## About the diff size

**6,699 of the 8,955 insertions are one vendored file**, `public/vendor/malloy-filter.js`. A no-build page cannot import from `node_modules`, so delegating to the grammar means shipping it. `scripts/vendor-malloy-filter.mjs` regenerates it from the package rather than leaving it hand-maintained. The screenshot accounts for most of the rest. The hand-written change is `format.js` at +318 and its tests at +227.

## What else is in it

- `examples/storefront/givens.malloy` declares the page's parameters, using the plain `# description=` tag form rather than `#(description=…)`, because this page's reader is a data app and not a notebook Parameters panel. That let a regex be **deleted** from the page: it had been picking `#(description="…")` out of raw annotation strings, which was a copy of a grammar the server already parses.
- The page recovers when a URL carries a given the model does not declare, rather than failing.
- `docs/html-data-apps.md` and the example README are updated to match.
- The example tests run in CI as `test:examples`, **under node rather than bun**, because the page is a no-build ES module the browser loads directly and node is the closest runtime to that: no bundler, no transform, no test-runner globals.

## Verification

Driven live against a running server: **25,356 line items narrow to 2,535** through the givens path. Verified in headless Chromium three ways, a normal drive across all four tabs, a hostile-URL pass, and a popover interaction test proven to fail without its fix.

**Note on the checks: this PR gets a smaller set than the ones targeting `main`.** `app-playwright.yml` and `connection-integration-tests.yml` are both filtered to `pull_request: branches: [main]`, so they do not run against a feature-branch base. `build.yml` has no branch filter, so unit, integration and the new `test:examples` step do run, on all three OSes. The Playwright gap is narrow either way: no spec loads storefront's public page, since the one data-app spec uses `PACKAGES.dataApp`, which `fixtures.ts` defines as `html-data-app`. It closes on its own when #981 merges and this retargets to `main`.


examples/storefront/tests/format.test.mjs
```diff
@@ -0,0 +1,301 @@
...
+//
+// Named, not globbed: `node --test` on a glob that matches nothing exits 0 and
+// reports "tests 0", so the globbed form passes while running none of this.
+//
+// The filter cases below are the point of this file. Every one of them is a
+// value a hand-rolled encoder turns into something Malloy matches differently
+// from what the reader picked, and every such failure is silent: the page
+// renders a table either way. They run against the real grammar through the
+// vendored bundle, so they fail if a Malloy bump changes the escaping and the
+// vendored copy has not been regenerated.
```
Sha-Bang: this guarantee runs backwards, and it's the one i'd most want closed before this lands.

the cases import `../public/vendor/malloy-filter.js`. after a Malloy bump the page and its tests both use the *stale* grammar, agree with each other by construction, and stay green — while the server parses with the new one. that's exactly the silent escaping mismatch this module exists to prevent. the tests would only catch it if they imported the package.

nothing else closes it either: `upgrade-malloy.sh` bumps every `@malloydata/*` and never calls `vendor:malloy-filter`, while `vendor-malloy-filter.mjs`'s own docblock says regenerating after a bump "is the whole maintenance story".

cheapest fix is asserting the banner version equals `@malloydata/malloy-filter/package.json`'s — same shape as the `validate-*` scripts already in package.json.

examples/storefront/public/app/format.js
```diff
@@ -0,0 +1,346 @@
...
+/**
+ * Join picked values into one string filter (`"Nike, Levi's"`), which Malloy
+ * reads as "any of these".
+ *
+ * A value that cannot be carried is dropped rather than approximated. Dropping
+ * every value leaves `""`, the empty filter, which is the "All" state: that is
+ * a WIDENING, so it is deliberate and visible rather than incidental. The
+ * control shows nothing selected, which is what "All" looks like.
+ */
+export function encodeFilterList(values) {
```
Sha-Bang, line 207: `encodeFilterList`, `decodeFilterList`, `isPlainFilterList`, `encodeAtLeast` and `decodeAtLeast` all exist under those exact names in `packages/sdk/src/components/given/filterValue.ts`, with genuinely different edge rules (`decodeAtLeast` differs in signature too — `number | undefined` there, `0` here).

credit where it's due: on whitespace this copy is the *better* rule. i measured it — `survivesRoundTrip` drops exactly the values that parse to a null clause (the ones that would widen the filter to every row) and carries ASCII space as `\ `, where the SDK drops all 24 for uniformity. derived instead of predicted.

so the issue isn't which rule is right, it's that the repo now has two answers to "which values can a filter carry". the PR body says copying the fix across would have made a third copy — true, but this avoided a third copy of the *grammar* and made a second copy of the *policy*. `vendor-malloy-filter.mjs` already runs esbuild, which compiles TS, so it could bundle `filterValue.ts` into the vendored module and there'd be one.

examples/storefront/public/app/format.js
```diff
@@ -0,0 +1,346 @@
...
+      parsed.values?.length !== 1
+   ) {
+      return 0;
+   }
+   const bound = Number(parsed.values[0]);
+   return Number.isFinite(bound) ? bound : 0;
+}
+
+/** `@2023-01-01` and `f''` are Malloy literals; a control needs the value inside. */
+export function readGivenDefault(literal) {
```
Sha-Bang, line 339: the greedy `/^f'(.*)'$/s` mangles a compound or escape-carrying default. measured against the vendored grammar:

- `f'Denim','Outerwear'` → `Denim','Outerwear`
- `f'Levi\'s'` → `Levi\'s` (backslash retained)
- `'WN'` → `'WN'` (quotes still on — it's type-blind where the SDK routes on `type`)

`packages/sdk/src/components/given/utils.ts:39-43` already solved this and says why in a comment: it matches `^'([^'\\]*)'$` so as to "leave a compound expression (`f'WN','AA'`) verbatim rather than mangling it by stripping the outermost quotes".

latent for this model (every default is `f''` or a clean date), but `Given.default` is documented as the literal "exactly as written in the model", so it's reachable the moment someone gives one a real default — and `readFilterForPicker` would then re-encode the garbage as a literal search on the next pick.

examples/storefront/public/app/controls.js
```diff
@@ -0,0 +1,626 @@
...
+      },
+   };
+}
+
+function date(host, { contract, value, description, onChange }) {
+   const input = document.createElement("input");
+   input.type = "date";
+   input.id = `given-${contract.name}`;
+   input.dataset.given = contract.name;
+   input.value = value;
```
Sha-Bang, line 619: this writes the wire value straight to an `input[type=date]`, which sanitizes anything that isn't `YYYY-MM-DD` to the empty string (same on `:623`).

`docs/givens.md:376-377` blesses the timestamp form: "where the value also carries a time and the zone leaves the day alone, as in `?ORDER_DATE=2024-01-05T00:00:00Z`, it is read normally". so `?SINCE=2023-06-01T00:00:00Z` filters every tile from june 2023 over a blank date box — the control-says-one-thing-the-data-says-another failure this branch is otherwise about removing.

`conformance.test.mjs:62-65` reasons it away ("no filter it can silently draw wrongly, so it has no opaque concept to conform to"), so nothing in the suite can catch it. the premise there is about values the server *rejects*; this is a value it accepts.

examples/storefront/public/app/table.js
```diff
@@ -0,0 +1,58 @@
...
+         // textContent, never innerHTML: a cell is data from the warehouse and
+         // must never be parsed as markup.
+         td.textContent = cellText(row, field);
+         if (isNumeric(field)) td.classList.add("num");
+         tr.appendChild(td);
+      }
+      body.appendChild(tr);
+   }
+   table.appendChild(body);
+   host.appendChild(table);
```
Sha-Bang, line 57: the table lost its scroll container in the rewrite. the old page wrapped it — `<div style="overflow-x: auto"><table id="products">` at index.html:101 — and this now appends straight into `.tile-body`, which has no rule in the new stylesheet.

combined with `table { width: 100% }` and the newly added `th { white-space: nowrap }` (index.html:201), a wide table escapes the card instead of scrolling inside it. `brand_performance` at colspan 6, or anything at the ≤680px breakpoint where `.tile` goes full width, is where it shows. `overflow-x: auto` on `.tile-body` restores it.

examples/storefront/public/app/controls.js
```diff
@@ -0,0 +1,626 @@
...
+      const args = {
+         modelPath,
+         contract,
+         value: value || fallback,
+         fallback,
+         description,
+         onChange,
+      };
+
+      if (contract.control === "multiselect") {
```
Sha-Bang, line 190: dispatch happens on `contract.control` before anything looks at the given's type, so a `filter<number>` or a plain `string`/`number` tagged `control=select` goes through the *string* escaper.

the comment at :494-503 names this hazard exactly (`-5` string-encodes to `\-5`, which the number parser rejects → 400 on every tile), but nothing enforces it — and `isFilter()` is right there at :138, consulted only by the range branch.

fine for this model, where every `control=`-tagged given is a `filter<string>`. but api-doc.yaml permits `control` on any given and the README's recipe is "declare the `given:` and tag it with a control this page draws", so the trap is reachable by following the docs.

examples/storefront/public/app/controls.js
```diff
@@ -0,0 +1,626 @@
...
+            !Number.isInteger(bound));
+      input.value = String(bound);
+      input.dataset.opaqueFilter = String(unshowable);
+      composeTitle(input, unshowable ? OPAQUE_HELP : "");
+      // From `bound`, NOT from `input.value`. The element clamps what it is
+      // given, so with `range_min=10` and nothing in force this wrote 0, read
+      // back 10, and stated "≥ $10" over unfiltered data: a number the reader
+      // never chose and the server never applied. Reading back a value you
+      // just wrote is how the clamp gets laundered into a fact.
+      readout.textContent = unshowable ? text : bound > 0 ? `≥ $${bound}` : "any";
```
Sha-Bang, line 585: `≥ $${bound}` hardcodes a currency in a renderer whose whole contract is that it's driven by the given's metadata (also `:582`). a `# label="Minimum quantity" range_min=0 range_max=100` slider reads "≥ $5".

related, same line of thinking: `encodeAtLeast` returns `""` for anything `<= 0` (measured, `-20` → `""`), so a range crossing zero can't be expressed at all. benign for `MIN_SALE` at `range_min=0`, where "≥ 0" and "no filter" genuinely are the same thing — which is why neither bites today.

examples/storefront/public/app/controls.js
```diff
@@ -0,0 +1,626 @@
...
+   const pick = (row) =>
+      suggest.dimension ? row[suggest.dimension] : Object.values(row)[0];
+   // Deduped by string identity, keeping the first occurrence and its original
+   // type. A `group_by` cannot repeat a value, but `suggest { query=… }` is a
+   // query the model author writes, and two rows for one value put a tick on
+   // screen that the caption and the URL both disagree with: measured, with the
+   // list `[Nike, Nike]`, unticking one row cleared the filter while the other
+   // stayed ticked.
+   const values = rows.map(pick).filter((v) => v != null);
+   return [...new Map(values.map((v) => [String(v), v])).values()];
```
Sha-Bang: this keeps the **last** occurrence, not the first as the comment above says. `Map` fixes insertion *position* at the first key but the *value* at the last, so the mixed-type case the comment is specifically about — `[1, "1"]` — survives as the string, not "its original type".

the measured `[Nike, Nike]` example doesn't expose it because both entries are identical. either `if (!map.has(k)) map.set(k, v)`, or reword the comment.

examples/storefront/public/app/table.js
```diff
@@ -0,0 +1,58 @@
...
+const NUMERIC_FORMATS = new Set(["currency", "percent", "decimal", "integer"]);
+
+/** Aggregates and formatted numbers align right; identifiers and labels do not. */
+const isNumeric = (field) =>
+   field.format === "id"
+      ? false
+      : field.isAggregate || NUMERIC_FORMATS.has(field.format);
+
+const cellText = (row, field) =>
+   field.name === "month_of_year"
```
Sha-Bang, line 15: a hardcoded field name in a renderer whose stated contract is that "headings, alignment, and number formats all come from the field metadata" — and `result.js:71` infers the sibling `month` format from the `_month` suffix instead, so month handling is special-cased two different ways in two files.

this branch is also dead today: no table tile in `tabs.js` selects a `month_of_year` column. worth having `readFormat` return a `month_of_year` format and letting `formatValue` own it, as it already does for `month`.

examples/storefront/public/app/app.js
```diff
@@ -0,0 +1,352 @@
...
+   writeUrl();
+   render();
+}
+
+function buildTabs() {
+   const host = el("tabs");
+   for (const tab of TABS) {
+      const button = document.createElement("button");
+      button.type = "button";
+      button.role = "tab";
```
Sha-Bang, line 297: the strip claims tablist semantics it doesn't implement: `role="tab"` here plus `role="tablist"` on the nav and `aria-selected` maintained in `render()`, but no roving `tabindex`, no arrow-key handling, and no `role="tabpanel"`/`aria-labelledby` on the panels `panelFor` builds.

a screen-reader user in an announced tablist gets none of the navigation the role promises, which leaves this less navigable than the plain buttons it replaced. either implement roving tabindex + arrows and label the panels, or drop the roles and let them be buttons.

(minor: `button.role =` is the ARIA reflection property, while `setAttribute("aria-selected", …)` two functions away isn't — worth picking one.)

Sha-Bang (commented):
reviewed at f6837e7. ran the suite rather than reading it — 70 tests, 0 fail, 0 skipped — and measured the filter claims against the vendored grammar instead of taking the comments' word for them.

the CI wiring is the best part of this: naming the test files instead of globbing, the `skipped 0` guard, and `set -eo pipefail` with the reason written down each close a real way a green step tests nothing. vendoring the grammar so the escaping is the printer's and not ours is the right call too.

comments inline. the one i'd want answered before this lands is the vendored-bundle drift guard — `format.test.mjs` states a safety property that doesn't hold, and nothing regenerates the copy on a Malloy bump. the rest are non-blocking; a few are latent-for-this-model rather than live, and i've said which is which.

one theme worth naming across several of them: the page-generic renderers carry model-specific assumptions (a currency in the slider readout, a field name in the table, control-tag dispatch ahead of type), and the README's "adding a filter needs no edit to `public/app/` at all" is the invitation to the copy that trips them. each is a small fix now and a confusing bug later.

mlennie:
Thanks, this was a useful review. Three fixed, one I disagree with on measurement, seven deferred with reasons. Body left as written; everything is here.

## Fixed

**The vendored-bundle drift guard.** You are exactly right, and it stings: this is the same class my own test-efficacy sweep was built to find, sitting in my own test file, stating a guarantee that runs backwards. I verified both of your supporting facts before touching it. The guard now compares the banner version against the installed package, which is the one comparison a stale bundle cannot satisfy by agreeing with itself, and it fires: banner at 0.0.400 against installed 0.0.427 fails with the regenerate command in the message. One deviation from your suggestion worth naming so you are not looking for it in the wrong place: it is a test in `format.test.mjs` rather than a `validate-*` script in `package.json`. Same shape and the same comparison, and its docblock cites `validate-duckdb-sync` and `validate-pg-sync` as the model, but putting it in the suite means it runs on every `test:examples` without anyone remembering to wire a script into `test`. Say the word if you would rather have it as a script.

**The scroll regression.** Fixed with `overflow-x: auto` on `.tile-body`. Worth recording how I nearly mismeasured it, since it is the trap this branch keeps hitting: my first probe compared the table's `getBoundingClientRect` against the tile's and reported "escapes: true" WITH the fix applied, because a scroll container's content always exceeds it in layout terms. Re-measured the actual symptom instead: 0px of sideways page scroll with the fix, 4px without, at a 560px viewport. Small with this data, real, and it grows with content width.

**The dedup comment.** It was false about the line directly beneath it, and the code now keeps first-wins as the sentence promises, pinned by a test. The part I want to note is how you found it: you tried a different input than the one the comment cites. `[Nike, Nike]` cannot expose it because both entries are identical, so the comment read as measured while being wrong. That is a better technique than anything in my own sweep.

## Disagreed, with a measurement, and you are still right that something is missing

The timestamp form does not filter tiles over a blank date box for a `date`-typed given. The server rejects `2023-06-01T00:00:00Z` with a 400 reading `date must match 'YYYY-MM-DD'`, and every tile errors naming the value. So the conformance premise holds: this is a value the server rejects, not one it accepts. That comment now carries the measurement rather than the conclusion.

But your instinct pointed at a real gap, just a different one: a `timestamp`-typed given gets no widget at all, because the dispatcher tests `type === "date"`. Recorded as a coverage gap rather than fixed here.

## Deferred, seven

Your theme is correct and I have quoted it in the body rather than paraphrased it: page-generic renderers carrying model-specific assumptions, with the README's no-edit promise as the invitation to the copy that trips them.

I have not taken them because each is a behaviour change to a renderer the shipped model does not exercise, and these renderers have already changed eleven times tonight. The duplicate-policy one in particular is a design decision about which rule wins rather than a bug fix, and since you credit this copy as the better rule on whitespace, it deserves a change that can be reviewed as one thing rather than folded in here.

## Merge and gate

One real conflict, `package.json`, both sides adding a test leg and editing `test`. Kept and chained both, verified nothing of main's was dropped. Local signed merge; nobody touched the Update branch button.

Gate with codegen first: lint 0 with no autofix mutations, typecheck 0, prettier at its six standing gitignored files, unit 2997, integration 385, examples 72 pass 0 fail 0 skipped, and main's new `test:scripts` leg at 23 pass. Browser probes 36 assertions. I discarded a `bun.lock` rewrite that `bun install` produced, after confirming the committed lock still passes `--frozen-lockfile`, to keep unrelated churn out of the diff.

examples/storefront/tests/controls.test.mjs
```diff
@@ -0,0 +1,566 @@
...
+   });
+   const el = document.querySelector("#given-CATEGORY");
+   widgets[0].set(""); // Reset, while the options query is still open
+   release();
+   await new Promise((r) => setTimeout(r, 0));
+   await new Promise((r) => setTimeout(r, 0));
+   assert.equal(shown(el), "All categories", "the late options must not restore it");
+});
+
+test("deduping a mixed-type option list keeps the original type", async () => {
```
Sha-Bang: this test can't detect the thing it's named for — same shape as the drift guard you just fixed, which is why it's worth catching here.

measured: i reverted the dedup back to the old `new Map(values.map((v) => [String(v), v]))` line and ran this file. 33 pass, 0 fail, this case green. both `1` and `"1"` render the label `"1"`, and `encodeFilterValue` routes each through `encodeFilterList([String(v)])`, so they produce the same filter too. the assertions therefore can't see which of the two survived, which is the entire claim in the name.

what it does pin is real and worth keeping: one row for two spellings. it's the "and its original type" half that isn't asserted.

two ways to close it, either fine: pick a case where the type is observable — a `Date` and its own string spelling encode differently through `encodeFilterValue`, unlike a number and its digits — or assert on the returned value rather than the rendered row. third option is to drop the "original type" promise from the comment and let the test's name match what it checks, since it isn't clear the type is observable anywhere a reader could reach.

Sha-Bang (approved):
re-reviewed at 1a33a0e5. approving — the blocking item is closed and i checked it rather than taking the commit message's word for it.

**the drift guard works, including the part that matters.** i set the banner to `0.0.001` against the installed `0.0.427` and it failed with `run bun run vendor:malloy-filter and commit the result`. a guard that can't fail is the failure mode here, so that's the right proof. comparing the stamped version against the installed package is also the one comparison a stale bundle can't satisfy by agreeing with itself. examples suite is 72 pass / 0 skipped locally, matching your stated gate, and CI is green across 14 checks.

`.tile-body { overflow-x: auto; }` closes the scroll-container regression, and the dedup is explicit first-wins now.

**you're right about the date box and i was wrong.** the line i cited, `givens.md:376`, sits under `### Parameters live in the URL` inside `## Notebook UI` — that's the notebook's URL-decoding layer, which normalizes before sending. this page builds givens and POSTs them through `Publisher.query`, so it lands on Malloy's prepare-time coercion instead, which rejects and 400s; `packages/server/tests/integration/query_givens/query_givens.integration.spec.ts:102-108` pins that path. i applied a notebook contract to a direct API caller. the replacement comment is better than what i asked for: it records both measured forms and names the actual gap, which is that the dispatcher tests `contract.type === "date"` and so draws no widget for a `timestamp` given at all.

one new note inline, on the test added alongside the dedup fix — it passes against the un-fixed code, so it doesn't pin what its name claims. non-blocking.

two stale spots in the body, worth a pass since the repo squashes and this becomes the commit body: it still opens with "Base is `monty/pr935-slice5-given-control-spec`, not `main`" with a 3-or-11 commit count, when the base is `main` and it's 16 from the merge base; and the checks paragraph says `app-playwright` doesn't run against a feature-branch base and that the gap "closes on its own when #981 merges" — Playwright ran here and passed at 6m14s, so it's already closed. also the commit message says the timestamp coverage gap is "recorded in the PR body", and i don't see it there.

the six non-blocking items from the first pass are untouched, which is fine — they were non-blocking. the one i'd still want written down somewhere is `format.js` carrying a second copy of `filterValue.ts`'s policy, since the new drift guard covers the grammar but not that.

mlennie:
Thanks, and thank you for going back to the source on the date box rather than leaving it as a disagreement. Body left as written; the corrections are here.

## The dedup test: taking your third option, because the first two turn out not to exist

You are right that it could not detect the thing it was named for, and I checked your two suggested fixes before choosing. Neither is available:

The retained value reaches the page only through `text.textContent = option` and through `encodeFilterValue`, and **both stringify**, so for `[1, "1"]` there is no assertion to write. Nothing downstream can tell which of the two survived.

The Date case does not rescue it either, and for a reason that is worth stating because it is not obvious: a Date and its own spelling **do not collide under `String()` in the first place**, so they never reach the dedup as duplicates. The test would be exercising a different code path.

So the promise is dropped rather than asserted. The comment now says first-wins is convention and says why nobody can observe it, and the test is renamed for the half that is real: one row for two spellings. Confirmed it still fails, twice, when the dedup is removed entirely.

One more thing for the record, since it makes the original defect worse than "the comment overstated it": the old comment was wrong on its own terms. `new Map(entries)` fixes the position at the first key and the value at the last, so it never kept what it claimed to keep, in any case.

## The timestamp coverage gap, recorded here

The commit message says this gap is recorded in the PR body. That is true of the current draft and not of what is posted, since the live body predates it, so recording it here instead of editing the body:

**A `timestamp`-typed given gets no control at all.** The dispatcher tests `contract.type === "date"`, so a given declared `timestamp` draws no widget and is simply absent from the controls.

That is the gap your original note pointed at, arrived at from the other direction. Filed as its own issue rather than fixed here.

## Two stale spots in the body

Both real, and left in place deliberately rather than edited, so your review threads keep answering the text you reviewed. For anyone reading later:

The opening basing note is stale. The base is `main`, not `monty/pr935-slice5-given-control-spec`, and the commit count in that paragraph predates several merges.

The checks paragraph is stale in the useful direction. It says `app-playwright` does not run against a feature-branch base and that the gap closes when #981 merges. Playwright ran here and passed at 6m14s, so it is already closed.

Because the repo squashes, both paragraphs would otherwise land in the commit body, so they are being trimmed in the squash message at merge time rather than carried into history.

## Still open, and written down rather than fixed

Your point that the drift guard covers the grammar but not the policy is right, and `format.js` carrying a second copy of `filterValue.ts`'s rules is filed as its own issue rather than folded into a renderer-hygiene list, because it is a design decision about which of two rules wins across two packages rather than a bug. The measured whitespace difference is in it, including that this copy is the better rule.

The remaining non-blocking items from your first pass are filed as one issue for the shared theme, with the sites as a checklist, since the currency symbol, the `month_of_year` field name and control-tag dispatch running ahead of type are the same edit in three places and whoever picks one up should see the other two. The tablist item is separate, since it has a real choice in it.

Gate on this tree, codegen first: lint 0 with no autofix mutations, typecheck 0, prettier at its six standing gitignored files, unit 3007, integration 385, examples 72 with 0 skipped, scripts 23.

PR #1028, opened 2026-08-14
fix(server): make get_context retrieval track model quality, and answer in one response shape

jswir:
`get_context` quality is meant to be a proxy for **semantic model** quality. In several places it was a proxy for something else — doc length, entity count, or source-name luck — and this fixes those, converges the response on one shape, and renames the MCP tools.

Folds in #1097. Commits are individually reviewable and worth reading in order.

## ⚠️ Breaking

**Every MCP tool loses its `malloy_` prefix.** No alias, no deprecation window — the old names are removed.

| Before | Now |
|---|---|
| `malloy_getContext` | `get_context` |
| `malloy_executeQuery` | `execute_query` |
| `malloy_compile` | `compile_model` |
| `malloy_reloadPackage` | `reload_package` |
| `malloy_getStatus` | `get_status` |
| `malloy_searchDatabaseSchema` | `search_database_schema` |
| `malloy_searchDocs` | `search_malloy_docs` |

Listing the catalog moves out of `malloy_getContext`-with-no-arguments into its own tool, `list_packages`, because `get_context` now requires `search_targets` and a `scopes`.

**`get_context` answers in a new shape.** A flat ranked `results[]` of entities becomes `sources[]`, each carrying the entities that matched inside it — the shape a hosted Malloy retrieval API already uses. Anything consuming both needed two parsers, and a harness written against one scored zero against the other. Built as a serialization layer, so ranking and collapse are untouched and a shape bug stays separately revertible from a ranking bug.

Hosts that discover tools on connect pick this up on reconnect. Anything hardcoding a tool name in a prompt or script needs editing. `RELEASE_NOTES.md` carries the migration entry; this is a **MINOR** bump.

## What was wrong

**Retrieval could not see half the model.** Joins were never indexed, so no call could return a declared join and a join's `#(doc)` was unreachable — analysts wrote *"the model has no pre-declared joins"* and then guessed names that did not exist. A source's `#(doc)`, where grain caveats and population rules live, reached the agent only when the source itself independently cleared the floor for the same query, so whether guidance arrived was a function of phrasing. Both are now always delivered, and an empty `joins` array is an authoritative "declares none".

**Documenting an entity made it harder to find.** One vector per entity meant a long doc dominated the average: `building_name` (547 chars) ranked 10th for its own plain name while two short-documented siblings ranked 1st. Name and docs now embed as separate facets and an entity scores as its best facet, so documentation adds recall and never costs precision. Long docs also chunk instead of being silently cut at 1024 chars.

**A third of the result window was redundant** — 26 of 76 slots across 8 queries repeated a concept already present. Within one source, `dimension: site is SITE` is one column indexed twice; those collapse into one row reporting the other spelling in `aliases`. **Across sources nothing collapses** — see below.

**An empty result was indistinguishable from "not modelled here."** `below_cutoff_count` separates a true negative from "rephrase or split". Same reasoning gives environments and packages an `error` field, and adds `retrieval_reason` plus a pollable `embeddingIndex` on the Package resource.

**The card said what an entity is, never what it takes to query it.** A gated source looked ordinary until the denial arrived; a defaulted given looked like a value to invent; field types were invisible. Now `givens`, `authorize` (report-only, with the given names each gate reads), `data_type`, `one_line_summary`, and `warnings[]` in place of a single `note`.

**An extended source described its parent, not itself.** `#(doc)` on `source: scoped_orders is order_items extend { where: … }` arrived behind the inherited one, because the stable `SourceInfo` flattens both into one array with the parent's first — so `docs` led with another source's text and `one_line_summary`, a first sentence, *was* the parent's. Two sibling extensions were indistinguishable at a glance. Read from the compiled model's `annotations.blockNotes`, which keeps own and `inherits` apart, so a source now describes itself. Fields are untouched: an inherited dimension IS the parent's field and its doc is correct.

## Duplicate collapsing: what a duplicate is decided from

This changed late, after review, and is the part most worth a second read.

**Same-source aliases are now decided by the compiled expression, not by name.** A field folds into another only when its whole definition is a reference to a single-segment sibling of the same source — `e.node === "field"`, `path.length === 1`. The single-segment rule is load-bearing rather than defensive: in the bundled `storefront`, `order_items.category`, `.brand` and `.region` are *all* bare field references, to `["products","category"]` and friends. A rule keyed on "is a field reference" alone would fold three real join traversals into their targets.

An earlier revision of this PR argued that this had to stay a heuristic, on the grounds that the stable interface exposes no expression. That was true of the *stable interface* and false as a conclusion: the compiled `ModelDef` carries `e` and `code`, Publisher already holds one on every `Model` (both construction paths, worker-pool loads included), and it was being discarded at Publisher's own conversion boundary. **The reviewer who suggested deduping on provenance was right** — it needed no new Malloy interface, only a `getModelDef()` accessor. Chains resolve to their root, and the differing-doc escape hatch is gone: it existed to contain a guess's false positives, and a proven reference has none.

**Cross-source collapsing is deleted outright.** It folded a same-named entity across *different* sources when scores were within `0.03` and the folded row's doc was absent or identical. The score gate protected nothing — same-named fields in different sources embed byte-identical name facets and score *identically* however far apart their docs are — so the only guard was the doc, and an inherited field carries its parent's doc verbatim.

That misfires on the bundled example package. `scoped_orders` is `order_items extend { where: … }`, a filtered source whose `total_sales` is a different number, and a search for "total revenue" returned **one** row with `also_in: ["overview","scoped_orders"]`. An agent following that payload reports the wrong figure.

It was also obsolete. It landed when this returned a flat `results[]` where same-named rows looked like independent peers. The `sources[]` shape already nests entities under per-source cards carrying `docs`, `one_line_summary` and `joins` — group by source, let each source's doc disambiguate — so folding removed rows from the structure built to separate them and substituted a name list for the doc that does the disambiguating. `returned` could also exceed the caller's own `limit`, because a folded sibling's card was materialised outside the window; that invariant is restored.

`also_in` is removed from the response. It never shipped.

**`include_code`** returns a field's Malloy expression, **off by default** — a source's docs should carry the meaning, and an expression is long. A physical column has no expression, so absence under the flag means the field is raw rather than derived.

Other decisions — MAX over a weighted name/doc blend, keeping lexical scores internal, `entity_type` as a superset of the hosted enum, omitting fields Publisher cannot honestly fill — are argued in the commits that make them.

## Testing

CI green: **22 checks, 0 failing**, including Cross Platform Tests on all three OSes.

Server unit locally: **3416 pass, 33 fail** — the same 33 that fail on `main`, every one `fs.rmdirSync(path, { recursive: true })`, which bun 1.4 removed. Integration: **406 pass, 1 fail**, the GCS-credentialed concurrent-package test, which fails identically before any change here.

Verified end to end against a running server on the bundled `examples` in **both retrieval modes**: 24 probe assertions pass in each, covering deterministic indexing across calls, joins retrievable with their `relationship`, drill-down, `authorize`/`givens` on the gated example source and omitted on the ungated one, truncation reporting, and an unknown package erroring in-shape. `embeddingIndex` reaches `ready`.

The collapse rework was verified the same way rather than only by specs: all three sibling sources return with their own entities, each under its own doc; join-traversal dimensions do not fold; `include_code` is absent by default and carries `"sale_price.sum()"` when asked; no payload contains `also_in`; and a card's `resource_id` still goes straight into `execute_query` for real rows. Retrieval quality is unchanged — the bundled eval holds at **5/8 recall@5** on the lexical baseline, against a prediction that un-folding siblings would cost slots.

**Stated rather than implied:** the semantic run used a local deterministic embeddings stub for reproducibility. A bigram stub has no semantic understanding, so it validates plumbing, not ranking quality — the numbers from it mean nothing and the pre-existing `gap` eval cases still miss. Measuring real retrieval quality needs a run against a real provider on a large model, worth doing before tuning `EMBEDDING_MIN_SIMILARITY` further.

Closes #1097.


jswir:
Independent data point for **(5) a third of the result window was redundant**, from a different model shape than the one you measured — and one case that I don't think this PR covers.

## (5) reproduces, worse, on extension siblings

Model: one base fact source plus two **audience extensions** of it (`order_items`, `order_items_store_ops` = base minus financial columns, `order_items_margin` = base plus a margin measure, gated). Six sources total, ~20 measures.

`malloy_getContext { query: "revenue", limit: 8 }`:

| slot | kind | name | source |
|---|---|---|---|
| 1 | measure | `revenue` | `order_items` |
| 2 | measure | `revenue` | `order_items_store_ops` |
| 3 | measure | `revenue` | `order_items_margin` |
| 4 | view | `revenue_by_month` | `order_items` |
| 5 | view | `revenue_by_month` | `order_items_store_ops` |
| 6 | view | `revenue_by_month` | `order_items_margin` |
| 7 | measure | `margin_rate` | `order_items_margin` |
| 8 | view | `revenue_by_location` | `order_items` |

**6 of 8 slots (75%) are triplicates** with byte-identical `doc` strings, against the 34% you measured. Two slots carry distinct information.

The shape may matter for whether `alsoIn` catches it. These are not sibling sources over different tables, nor a raw column beside its alias — they are `source: x is base extend { ... }` derivations, so the entity is *literally the same field* reached through three names. In a nested query the response metadata confirms that: `revenue` carries the identical `reference_id` (`90d12dab-...`) in every nest it appears in. If the index has access to that provenance, extension-derived duplicates could collapse on it rather than on text similarity, which would also make `alsoIn`'s "naming the others" exactly right: *`revenue` on `order_items`, also in `order_items_store_ops`, `order_items_margin`*.

Happy to hand over the package if a fixture is useful — it is in malloydata/publisher#1042 (`examples/dbt-jaffle-shop/adopt-rich`).

## Not covered here: retrieval offers sources the caller cannot query

Slot 7 is `margin_rate` on `order_items_margin`, which carries `#(authorize) "$role = 'finance'"`. With the model's default given, querying it is a hard denial:

```
malloy_executeQuery { sourceName: "order_items_margin", queryName: "revenue_growth_mom" }
-> Access denied for source "order_items_margin".
```

So retrieval spends a slot on an entity that is guaranteed to fail for this caller, and an agent following the suggestion burns a turn. (The denial message itself is good — it says this is an authorization denial, not a syntax error, and names what to supply.)

Two things that would help, both arguably out of scope for this PR:

1. Rank or filter by **admissibility for the current request** — if the givens in scope cannot satisfy a source's authorize expression, its entities are not useful answers.
2. Failing that, mark them, so an agent can decide. A `gated: true` (or the required givens) on the entity would be enough to stop it guessing.

There is a governance angle too: an entity's *existence* leaking through retrieval to a caller who cannot query it is a mild disclosure, and `docs/authorize.md` is careful to say gates are enforced against the complete source set and never weakened by listing curation. Retrieval seems to sit on the listing side of that line, which may well be intended — but it is worth stating explicitly either way.

Want me to file (2) as its own issue rather than growing this PR?

Environment: `main` at ab0e97d7, `@malloydata/malloy` ^0.0.427, DuckDB, lexical retrieval (no embedding provider configured locally, so the ranking half of this PR was not exercised — the redundancy above is from the lexical path).

jswir:
Follow-up: I had said the ranking half of this PR was untested by my data because my server was lexical. It is now configured (`text-embedding-3-small`, 179 entities embedded for the package), so here is the semantic path on the same model.

## The ranking half looks good

A deliberately synonym-heavy question with **no token overlap** with the model's vocabulary — "shops" against `locations`, "sign-ups" against `customers`, "joined a year ago" against `first_order_month`:

> `which shops are doing badly and are our newest sign-ups spending less than the ones who joined a year ago`

| rank | entity | score |
|---|---|---|
| 1 | `cohorts_by_first_order_month` (view) | 0.4777 |
| 2 | `customers_by_type` (view) | 0.4588 |
| 3 | `top_customers` (view) | 0.4425 |
| 4 | `spend_band_mix` (view) | 0.4169 |
| 5 | `size_band_by_month` (view) | 0.3985 |
| 6 | `overview` (view) | 0.3777 |

Rank 1 is exactly right, six distinct entities, no duplicates. The hit also carries its `#(doc)`, which in this case is load-bearing guidance — *"Read `spend_per_tenure_month`, not `lifetime_spend`: the raw total falls off for recent cohorts because they have had less time, not because they are worse."* An agent gets the answer and the caveat in one call. This is (2) working as intended.

One miss worth noting: the compound question's first clause ("which shops are doing badly") went unserved — `revenue_by_location` and `tax_rate_by_location` are in the model but not in the top 6, which the cohort half crowded out. `belowCutoffCount` from (6) would not help here since these were above cutoff; it is just a compound question, and splitting it fixes it.

## (5) is worse on the semantic path, and your diagnosis is exactly why

Same `query: "revenue"`, `limit: 8`, before and after:

| | lexical | semantic |
|---|---|---|
| slots filled | 8 | 8 |
| distinct entities | 2 | **3** |

Semantic:

| slot | entity | source | score |
|---|---|---|---|
| 1–3 | `revenue` | `order_items`, `order_items_store_ops`, `order_items_margin` | **0.67, 0.67, 0.67** |
| 4–6 | `revenue_by_month` | same three | **0.5414 ×3** |
| 7–8 | `food_revenue` | `order_items_margin`, `order_items_store_ops` | **0.5268 ×2** |

Every slot is a duplicate, and the scores are **identical within each group** because the embedded text is identical — the same condition as your `SITE` case returning at an identical 0.96 from three sibling sources "with nothing to choose between them". Lexical at least had incidental tie-breaking; embedding similarity makes the tie exact, so the window fills deterministically with three concepts. So this PR's dedup work matters more on the semantic path, not less.

As before, the shape here is `source: x is base extend { ... }` rather than sibling tables, and the field is provably the same one — identical `reference_id` across nests in a query response — so provenance may be a cheaper and more certain dedup key than text similarity for this case.

No `embeddingIndex` field on the Package resource yet on `main` at ab0e97d7, so (7) is unmerged here; I polled the debug log line (`Synced entity embeddings {embedded: 179, ...}`) instead, which is precisely the scraping (7) removes.

jswir:
Both comments above are addressed in the rebased series, and one of them turned into a fix I would not have found otherwise.

**On the extension-sibling triplicates.** You were right that this reproduces worse on that shape, and right about the cause. The diagnosis I had wrong was narrower than yours: collapsing ran only on the semantic path, so an unconfigured Publisher — which is what most local servers are — kept every duplicate regardless. Lexical hits now carry their lunr score normalized against the top hit, which puts them on the [0, 1] scale the sibling epsilon was tuned for, and the same collapse runs there. The score stays internal and is never serialized, because a lexical relevance is not comparable between queries and publishing one would invite exactly that comparison.

Measured on the bundled `governed-analytics` package, which has the shape you describe (`sales` and `orders_secured` both extend `orders_base`), searching `"total order amount"` with `limit: 6`:

| | before | after |
|---|---|---|
| slots spent on duplicates | 2 of 6 | 1 of 6 |
| distinct concepts returned | 4 | 5 |

`amount` and `order_count` collapse and now carry `also_in: ["orders_secured"]`. `total_amount` deliberately does not: its two docs say different things — *"Total order amount"* against *"Total order amount visible to this caller"* — so they score apart and stay as two rows for the agent to choose between. That reads to me as the right line, and a spec pins both halves.

I did **not** take the `reference_id` provenance route you suggested, and I want to be honest that this is a limitation rather than a decision I am confident in. Collapsing on identical field provenance would be exact where the text heuristic is approximate, and it would have caught `total_amount` too. It is not in reach from the stable interface this code reads, and getting at it means the private `modelDef`, which is the same wall #1028 hit on naming a join's target source. Worth a follow-up on its own; I did not want to bury it in this PR.

**On retrieval offering sources the caller cannot query.** Taken, and it is now (2) from your comment rather than (1). Each card carries `authorize`, one entry per gate expression in force, each with the given names it reads:

```json
"source_info": {
  "resource_id": { "source": "orders_secured", ... },
  "authorize": [{ "expression": "$ROLE = 'admin' or $TENANT = 'acme'",
                  "given_names": ["ROLE", "TENANT"] }],
  "givens": [{ "name": "ROLE", "type": "string", "default": "'analyst'" }, ...]
}
```

Ungated sources omit the key entirely rather than sending it empty. An agent can now see a gate before spending a turn on the denial, and see which givens would satisfy it.

I did not do (1), ranking or filtering by admissibility, and I think that is the right call for a reason worth stating: deciding admissibility caller-side means evaluating the gate, and the list is not a predicate you can evaluate. It flattens gates carried in from a source's ancestors, they AND rather than OR, and a gate the registry cannot attribute reports the fail-closed `"false"` placeholder no author wrote. Reading it as a disjunction grants where Publisher denies. The docs say that twice, in `docs/ai-agents.md` and in the tool description, and the field is marked report-only.

Your governance point stands on its own and I have not resolved it: an entity's existence still reaches a caller who cannot query it, and this change makes the gate visible rather than making the listing narrower. That is consistent with `docs/authorize.md`, which is careful to say gates are enforced against the complete source set and never weakened by listing curation, so retrieval sits on the listing side of that line. Stated explicitly now rather than left implicit, which was your ask.

**On `belowCutoffCount` not helping the compound question**, that matches what I see. It is `below_cutoff_count` now, snake_case with the rest of the shape.

packages/server/src/mcp/server.ts
```diff
@@ -45,9 +48,9 @@ const AGENT_SKILLS = (
...
+Start with get_context. Call it with no arguments to list the environments (each with its packages), with an environment to list its packages, with a package to list its sources, and with a package plus a plain-English question to get the sources, views, and fields most relevant to it. Use the names it returns verbatim and do not guess. Then run a query with execute_query. To change a model: validate the edit with compile_model, save it, then call reload_package so the new sources and views become queryable by name without restarting the server. ${RELOAD_FAILURE_IS_SAFE} After every model edit, call reload_package before querying: the server serves the model compiled at the last load, not your files, and a watch-mode save that fails to compile is otherwise silent. An empty get_context result does not mean the package is empty; check get_status for load errors and stale packages before concluding there is no data.
```
Sha-Bang: **Blocking** - `MCP_INSTRUCTIONS` still teaches the pre-converged flow: "Call it with no arguments to list the environments... with an environment to list its packages, with a package to list its sources." The converged schema in this same PR makes every one of those a validation failure — `search_targets` is `.min(1)` and `scopes` is `.length(1)` (get_context_tool.ts:768, :831), and `list_packages` is now the catalog step ("get_context requires one of each in its `scopes`, and those names come from here; never guess them", get_context_tool.ts:2147).

This string is delivered in the `initialize` response to every connecting client, so it is the first thing an agent reads and the last thing it can check against. An agent that follows it fails on its first call, on a server it has no other orientation for.

Rewrite it to the actual flow: `list_packages` → scoped `get_context` → `execute_query`, with no no-arg or environment-only tier.

packages/server/src/mcp/server.protocol.spec.ts
```diff
@@ -107,6 +117,63 @@ describe("MCP server over the MCP protocol (in-memory)", () => {
       }
    });
 
+   it("states the getContext response contract the payload actually ships", async () => {
```
Sha-Bang, line 120: **Should-fix** - This PR adds a careful contract test for `get_context`'s *tool description* — pinning `entity_type`, `sources`, `joins`, `givens`, `authorize` and the rest, on the stated grounds that "a new response field cannot land silently." The orientation instructions get no equivalent. `delivers orientation instructions to the connecting client` (:183) pins token presence (`get_context`, `RELOAD_FAILURE_IS_SAFE`, `get_status`) and one sentence, but nothing pins the onboarding *sequence*.

That asymmetry is why the blocking finding above shipped green — and the same test's own comment notes these two surfaces "have drifted twice."

Worth an assertion that the instructions name `list_packages` as the first step and do not claim a no-argument `get_context` listing.

packages/server/src/mcp/error_messages.ts
```diff
@@ -174,29 +174,29 @@ export function getMalloyErrorDetails(
          refined = true;
          const [, viewName, sourceName] = viewNotFoundMatch;
          suggestions.unshift(
-            `Suggestion: View '${viewName}' was not found in source '${sourceName}'. Check the view name spelling or call malloy_getContext with sourceName '${sourceName}' to see the list of available views. Views are defined within sources like 'source: ${sourceName} is ... extend { view: ${viewName} is { ... } }'.`,
+            `Suggestion: View '${viewName}' was not found in source '${sourceName}'. Check the view name spelling or call get_context with sourceName '${sourceName}' to see the list of available views. Views are defined within sources like 'source: ${sourceName} is ... extend { view: ${viewName} is { ... } }'.`,
```
Sha-Bang: **Should-fix** - The view-not-found remediation tells the caller to "call get_context with sourceName '${sourceName}'", but `sourceName` is not a `get_context` parameter under the converged schema — the equivalent is `scopes[0].source`, alongside a required `search_targets`.

This is remediation text handed to an agent that has already failed once, so the wrong shape here turns one recoverable error into two. The rename in this hunk carried the tool name across but not the parameter.

packages/server/src/controller/package.controller.ts
```diff
@@ -89,7 +91,40 @@ export class PackageController {
          false,
       );
       const _package = await environment.getPackage(packageName, false);
-      return _package.getPackageMetadata();
+      const metadata = _package.getPackageMetadata();
+      const embeddingIndex = await this.embeddingIndexStatus(
+         environmentName,
+         packageName,
+      );
+      return embeddingIndex ? { ...metadata, embeddingIndex } : metadata;
```
Sha-Bang, line 106: **Should-fix** - The `embeddingIndex` enrichment sits below the `if (reload)` early return, so `GET /packages/{pkg}?reload=true` answers without the field while the plain `GET` includes it — same resource, same schema, two shapes depending on a query param.

The spec makes that a documented-contract mismatch rather than just an inconsistency: `PackageEmbeddingIndex` says "Absent when the server has no provider, because there is no index to describe," and tells callers to "Poll `status` until `ready`." Reload is exactly when a caller would start polling — a reload invalidates the index — and it is the one response that silently omits the field for a different reason.

Enriching the reload branch too would fix it. Worth a controller test pinning the field on both branches while you are there; there is none today, which is why the divergence was invisible.

docs/ai-agents.md
```diff
@@ -21,21 +21,58 @@ The server listens at `http://localhost:4040/mcp` (set the port with `--mcp_port
...
+
+The two listing tiers are the exception to all of this: calling with no arguments, or with only an `environmentName`, returns a flat `results` array of environments or packages. The platform's shape describes sources and has nothing to say about either, so those tiers keep their own.
```
Sha-Bang: **Should-fix** - This paragraph is newly added by this PR and contradicts line 24 of the same file, also newly added: :24 says "`scopes` is required and takes exactly one entry", :59 says "calling with no arguments, or with only an `environmentName`, returns a flat `results` array." The schema agrees with :24.

The following paragraph (:61) has the same problem as error_messages.ts above — it advises narrowing with `sourceName`, which is not a parameter.

This is the agent-facing doc for the tool whose contract the PR is changing, so the stale half is likely to be the half that gets followed.

Sha-Bang (commented):
**Public API surface delta** — additive only.

Adds a `PackageEmbeddingIndex` schema and an optional, `readOnly` `embeddingIndex` property on `Package`. No paths added, removed or changed; no `required:` changes; no removals. Nothing here forecloses a future shape.

Worth flagging that the one behavioural gap below is in this new field: `?reload=true` omits it.

**1 blocking, 4 should-fix, 2 nits.**

- `packages/server/src/mcp/server.ts:51` — the orientation instructions sent to every client at `initialize` still teach no-arg progressive `get_context`, which this PR's own schema makes a validation failure.
- `packages/server/src/mcp/server.protocol.spec.ts:120` — the tool *description* gets a contract test, the *instructions* do not; that is why the above passed.
- `packages/server/src/mcp/error_messages.ts:177` — remediation points at `sourceName`, not a parameter under the new contract.
- `packages/server/src/controller/package.controller.ts:99` — `?reload=true` skips the `embeddingIndex` enrichment the plain GET applies.
- `docs/ai-agents.md:59` — newly added paragraph contradicts newly added line 24 of the same file.
- `docs/configuration.md:86` — "six MCP tools"; eight are registered.
- `packages/server/src/mcp/tools/get_context_tool.ts:1548` — JSDoc still describes the replaced progressive contract.

docs/configuration.md
```diff
@@ -83,7 +83,7 @@ connection reference (BigQuery, Snowflake, Postgres, DuckDB, and more), see
...
 | `PUBLISHER_HOST` | `--host <addr>` | `0.0.0.0` | Host binding for both the REST and MCP servers. Set `127.0.0.1` to keep them loopback-only. |
 | `PUBLISHER_RATE_LIMIT` | _none_ | _unset_ | Maximum requests per minute one client may make to the REST server; over it, requests get `429` with `RateLimit-*` headers until the minute rolls over. Unset or `0` means no limit; any other value must be a non-negative integer or startup fails. `/health*` and `/metrics` are never limited. Clients are told apart by the connection's peer address, so behind a reverse proxy every client shares one bucket: rate-limit at the proxy there, or leave this unset. The MCP port is not covered. |
-| `MCP_PORT` | `--mcp_port <n>` | `4040` | MCP HTTP port. Serves the six MCP tools (`malloy_getContext`, `malloy_executeQuery`, `malloy_compile`, `malloy_reloadPackage`, `malloy_searchDocs`, `malloy_searchDatabaseSchema`) and the agent skills as MCP prompts. |
+| `MCP_PORT` | `--mcp_port <n>` | `4040` | MCP HTTP port. Serves the six MCP tools (`get_context`, `execute_query`, `compile_model`, `reload_package`, `search_malloy_docs`, `search_database_schema`) and the agent skills as MCP prompts. |
```
Sha-Bang: **Nit** - "the six MCP tools" and the list are stale — `initializeMcpServer` registers eight, adding `list_packages` and `get_status` (server.ts:75-83). `list_packages` is the one an agent now has to start from, so its absence here is the least convenient omission.

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -303,390 +1539,779 @@ function contextError(uri: string, identifier: string, error: unknown) {
...
       {
          results: [],
+         sources: [],
       },
    );
 }
 
 /**
- * Registers the malloy_getContext MCP tool. It is a progressive-discovery tool:
+ * Registers the get_context MCP tool. It is a progressive-discovery tool:
```
Sha-Bang: **Nit** - The rename updated the tool name in this JSDoc but not the description: "It is a progressive-discovery tool: with no environment it lists environments, with an environment but no package it lists packages..." is the contract this PR replaces. The block just below it also names both registered tools `get_context` ("the converged `get_context` and the flat `get_context` it replaces"), which reads as a copy-paste slip.

Sha-Bang (approved):
Approving. The blocking finding from my review comment stands and I'd want it fixed before merge: `MCP_INSTRUCTIONS` still teaches the no-arg progressive `get_context` flow that this PR's own schema rejects (`search_targets.min(1)`, `scopes.length(1)`), so an agent following the orientation string fails on its first call. It's a string fix, not a design question, which is why it isn't a gate.

The surface delta is additive only — new `PackageEmbeddingIndex` schema, optional readOnly `embeddingIndex` on `Package`, no path or required changes, no removals — so nothing here forecloses a future shape.

The parts I checked hardest all held up: the `entity_embeddings` re-key follows the existing `dropPackageKeyedIncrementalLedger` idiom exactly (reads the key rather than a version, so it's self-limiting), and `facetRowKey` uses NUL rather than a printable separator, which is the right call given backtick-quoted Malloy identifiers can contain almost anything.

jswir:
@Sha-Bang — this moved a fair way past your approval, and one change is a direct reversal of something the description defended, so it needs another look rather than a rubber stamp. Summary of what's new since `aac62c8e`, worst-first.

## Your review is addressed

All seven, in `9da4dcb4`, `3c748848` and `6332cec4`:

| finding | what changed |
|---|---|
| **Blocking** — `MCP_INSTRUCTIONS` teaches the pre-converged flow | Rewritten to `list_packages` → scoped `get_context` → `execute_query`, and it now states that the no-arg and environment-only calls do not exist, since that's the text an agent may have cached |
| orientation has no contract test | Added, and it reads the required parameters **off the registered schema** rather than restating them, so a newly-required parameter fails it until the orientation names it. Also asserts `list_packages` precedes `get_context`. Verified it fails against the old text |
| `error_messages.ts` names `sourceName` | All four `get_context` suggestions now name real parameters; the other three were vague rather than wrong, so I made them concrete too |
| `embeddingIndex` missing on the reload branch | Both paths enrich once. Absence still means "no provider" — an undefined status omits the key rather than sending null. Two tests where there were none; the shared one fails against the pre-fix controller |
| `docs/ai-agents.md` contradicts itself | Both paragraphs fixed |
| `docs/configuration.md` "six MCP tools" | Eight, `list_packages` first, with why |
| stale `runContextQuery` JSDoc | Deleted; "both registered tools" was also wrong — one caller |

## The reversal — please read this part properly

**You were right about `reference_id`, and the description was wrong to wave it off.**

I checked by compiling a model and dumping the internal `ModelDef` beside what Publisher reads:

```
dimension: site is SITE          → e.node="field"  path=["SITE"]   code="SITE"
dimension: site_lower is lower(OTHER) → e.node="function_call"     code="lower(OTHER)"
```

Through `to_stable`, all of them flatten to `{kind, name, type, annotations}`. So "the stable interface has no expression" was true, and "therefore there is no way to prove it" did not follow. Publisher already holds a `ModelDef` on every `Model` — both construction paths, worker-pool loads included — and was discarding it at its own boundary. It needed no new Malloy interface, only a `getModelDef()` accessor.

Alias folding is now exact (`895248c7`). The guard worth your attention is `path.length === 1`: in the bundled `storefront`, `order_items.category`, `.brand` and `.region` are **all** bare field references to `["products","category"]` and friends, so a rule keyed on "is a field reference" alone would fold three real join traversals into their targets.

## Cross-source collapsing is deleted

Cut on design grounds — two sources exposing a same-named measure are two different numbers — and it turned out to misfire on the shipped example package. `scoped_orders` is `order_items extend { where: … }`; a search for "total revenue" returned **one** row with `also_in: ["overview","scoped_orders"]`, so an agent following it queries the unfiltered source and reports the wrong figure. The doc gate couldn't catch it: an inherited measure carries its parent's doc verbatim, and only the *source* doc says the numbers differ.

It was also obsolete — it predates the `sources[]` shape, which already groups by source. `also_in` is gone from the response; it never shipped. `returned` could previously exceed the caller's `limit` (a folded sibling's card was materialised outside the window); that invariant is restored.

## Also new

- **`6d1c9a48`** — an extended source reported its parent's doc, because the flattened `SourceInfo` puts the inherited note first. Read from `annotations.blockNotes`, which the IR keeps separate from `inherits`. Sources only — an inherited field IS the parent's field and its doc is correct. Before/after: `overview` and `scoped_orders` had **identical** summaries (both `order_items`'), now they describe themselves.
- **`include_code`** — a field's Malloy expression, off by default.
- **`eaa52b67`** — a semantic conflict with #1101 that the merge could not see: its Publisher-only guard detects host skills by the `malloy_` prefix this branch removes, so it silently matched nothing. Re-keyed on tools Publisher *has* and a hosted target doesn't.
- **Three defects found by driving the tool against a live server**, not by specs (`aac62c8e`): a `Returned -1 of 0 matching entities` warning (negative count, from two sides counted in different units), a healthy seven-source package reported as a curation gap when the only target type was `dimensional_value`, and the deliberate `offset` refusal surfacing as "An unexpected internal error occurred". Each has a test that fails without its fix.

## Verification

CI green — 22 checks, 0 failing. Locally 3416 unit pass / 33 fail (the same `rmdirSync` bun-1.4 failures as `main`), 406 integration / 1 fail (the GCS-credentialed test, same before any change).

The collapse rework was checked against a running server rather than only by specs: all three sibling sources return with their own entities under their own docs, join-traversal dimensions don't fold, `include_code` behaves, no payload contains `also_in`, and a card's `resource_id` still goes straight into `execute_query` for real rows. The bundled eval holds at **5/8 recall@5**, against a prediction that un-folding siblings would cost slots.

## Where I'd want your eyes

1. `path.length === 1` as the alias boundary — it is the whole correctness argument.
2. Dropping the differing-doc guard now that a reference is proof. Two docs on one proven column is a modelling mistake, and the survivor rule prefers the documented spelling — but it is the one place the new rule folds something the old one kept.
3. An extended source with no `#(doc)` of its own now reports **no** doc rather than its parent's. I think that's the honest answer; it is a behaviour change.

Sha-Bang (commented):
**Surface delta** — flagging it rather than objecting to it:

```
api-doc.yaml   + Package.embeddingIndex, + PackageEmbeddingIndex   (additive, optional, readOnly)
               no path added / removed / re-typed

MCP tools      ! all 7 renamed, no alias, no deprecation window
               + list_packages split out of get_context-with-no-args
               ! get_context response reshaped: results[] -> sources[]
               - also_in removed

storage        ! entity_embeddings re-keyed by DROP TABLE (adds `facet` to the PK)
```

The breaking half is the MCP tool contract, which no spec covers, so it's worth registering as a delta. `RELEASE_NOTES.md:42-48` carries the migration table and MINOR looks right. The cache drop reads the actual column list rather than a schema version, so it stops firing once no database carries the old key, and all three paths are covered in `entity_embeddings_schema.spec.ts` — no concern there.

4 should-fix, 2 nits inline. Two more that land on lines outside the diff:

- **`skills/eval-loop/scripts/run_baseline.py:77`** — `ANSWER_TOOLS` still grants `mcp__publisher__malloy_getContext` / `malloy_executeQuery` / `malloy_compile`, and `ANSWER_PROMPT` (:419) still tells the answerer to use those names. Same in `skills/eval-improve/scripts/improve.py:68` (plus the `malloy_reloadPackage` curl at :78) and `skills/eval-diagnose/scripts/diagnose.py:64`. These are real `--allowedTools` grants, so against a server built from this branch every Publisher-arm call is un-granted — which is the stall `eval-loop/reference/running-a-run.md:79` warns about by name. `PUBLISHER_ONLY_TOOLS` four lines above the prompt was updated, so it looks like a narrow miss rather than a deliberate deferral.

- **`packages/server/src/config.ts:1226`** — making `getPublisherConfig` throw is the right call, and the three callers that swallow it do catch (`isPublisherConfigFrozen`, `getConnectionsFromPublisherConfig:1249`, `getInstanceTheme:1404`). But all three still log `{ error }`, which `environment_store.ts` documents in this same PR as serializing an Error to `{}` under winston's `format.simple()`. This change is what starts routing the operator's config typo through those three, so it turns the newly-raised error back into `{"error":{}}` on the way out.

Things I checked and found clean: the `collapseAliases` rework (the single-segment guard does stop the `order_items.category` join-traversal fold, and the tests at spec:1132 and :1929 would fail if either guard were removed); `docOnlyText` keeping `#(authorize)` predicates out of embedding input while the response reports them deliberately; the `readSourceOwnDoc` fallback when there's no compiled model; `EMBEDDING_MIN_SIMILARITY`'s documented range matching the code; and the rename sweep everywhere outside the eval scripts.

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -44,44 +130,756 @@ interface ResultEntity {
...
+   /** Serialize each entity's Malloy expression; off unless asked for. */
+   includeCode = false,
+): SourceCard[] {
+   const bySource = new Map<string, SourceCard>();
+
+   const cardFor = (
+      name: string | undefined,
+      modelPathFallback: string,
+   ): SourceCard | undefined => {
+      if (!name) return undefined;
```
Sha-Bang, line 398: **Should-fix** - `cardFor` returns undefined for a nameless source, so the row is dropped from `sources[]` entirely.

A model-level named query is created with `source: query.sourceName` (:1315), and `service/source_extraction.ts:789` leaves that `undefined` whenever `structRef` isn't a string — an inline `extend`/pipeline rather than a named source reference. Such a query is still indexed and can still match a `view` target, but it can never reach the response. It also consumed a slot in `windowBySource` on the way, so `total_available` counts a card that `returned` doesn't.

The old flat `results[]` had somewhere to put a sourceless entity; `sources[]` doesn't. Should one get a container card, or be excluded before it's counted?

packages/server/src/mcp/tools/embedding_index.ts
```diff
@@ -635,39 +936,197 @@ export async function trySemanticSearch(args: {
...
+         ),
+         -- The window is PER TARGET, not global. One shared LIMIT over the
+         -- union lets the highest-scoring target fill it and crowd the others
+         -- out entirely: measured against malloy-samples, a dimension target
+         -- at 0.63 took every slot while the measure target's own best hit at
+         -- 0.42 and the view target's at 0.53 vanished from the response --
+         -- absent, not merely ranked lower. That is the exact failure typed
+         -- targets exist to prevent, so each target gets its own share and the
+         -- caller gets an answer for every concept it described.
+         ranked_per_target AS (
```
Sha-Bang, line 1053: **Should-fix** - `ranked_per_target` windows per `target_idx` — which is the fix this PR is making — but the window carries no predicate on `entity_kind`. `get_context_tool.ts:2052` then drops every row the target's kinds can't claim, *after* the window has already been cut.

So a `measure` target whose top-`limit` nearest neighbours are all dimensions comes back empty, even though measures cleared the floor just outside its window. On a model with many dimensions and few measures that isn't an edge case.

Could the per-target kind predicate move into `scored` / `ranked_per_target`, so each target's window is filled with rows it can actually claim?

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -303,390 +1652,796 @@ function contextError(uri: string, identifier: string, error: unknown) {
...
-               environmentName,
-               packageName,
-            );
+               // Straight from the scan, which counts entities whose BEST
+               // score across every target fell under the floor. Deriving it
+               // from the returned rows would fold the page limit into it and
+               // report a crowded-out entity -- one that cleared the floor and
+               // simply did not fit -- as rejected, which is the opposite of
+               // what this number tells a caller.
+               belowCutoffCount = unionBelowCutoff ?? 0;
```
Sha-Bang, line 2114: **Should-fix** - `below_cutoff_count` is forwarded verbatim from the scan, which counted before the kind filter at :2052 ran.

That makes `sources: []` alongside `below_cutoff_count: 0` reachable — the exact combination `SemanticSearchResult`'s own doc rules out (`embedding_index.ts:115-119`: "cannot occur"), and which the tool description at :1627 reads to the agent as "no sources = nothing cleared the floor". The agent is then told to rephrase when the real remedy is the window.

Fixing the kind-blind window (see the note on `embedding_index.ts:1026`) makes this honest as a side effect; recomputing the count after the claim filter would be the narrower fix.

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -303,390 +1652,796 @@ function contextError(uri: string, identifier: string, error: unknown) {
...
-         /**
-          * Spread into a payload to attach `note`. Returns {} when there is
-          * nothing to say, so a healthy package's payload stays byte-identical
-          * to what it was before notes existed.
-          */
-         const noteFor = (extra?: string) => {
-            const note = [staleNote, extra].filter(Boolean).join(" ");
-            return note ? { note } : {};
-         };
+   const scored: ResultEntity[] = ranking.map(({ e }) => ({
```
Sha-Bang: **Should-fix** - `targetScores` is dropped here, so `windowBySource` (:536) buckets every lexical row under `-1` and they share one cap per source — reintroducing exactly the cross-target competition the per-target window exists to remove.

The comment justifies the drop on serialization grounds (a lunr score must not surface as `matched_targets[].relevance`), and that reasoning is right — but it's separable from budgeting. The scores could be carried internally for `windowBySource` and suppressed at the wire.

As it stands the guarantee the PR describes doesn't hold on the lexical path, which is what a server without an embedding key actually runs. Worth splitting the two concerns?

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -303,390 +1652,796 @@ function contextError(uri: string, identifier: string, error: unknown) {
...
-            hits = index.search(sanitized);
+            return jsonResource(uri, { environments: results });
          } catch (error) {
-            logger.warn("[MCP Tool getContext] lunr search failed", {
-               query,
+            logger.warn("[MCP Tool listPackages] listing failed", {
                error: error instanceof Error ? error.message : String(error),
             });
-            hits = [];
+            return contextError(uri, "environments", error);
```
Sha-Bang: **Nit** - `contextError` classifies as `getContext` and returns `{ results: [], sources: [] }`, so a `list_packages` failure reports "...during getContext" and hands the caller a different tool's shape with no `environments` key — while the success path answers `{ environments: [...] }`.

A dedicated helper returning `{ environments: [] }` would keep the error envelope matching the success one. (`results: []` is also the key this PR retired.)

packages/server/src/mcp/tools/get_context_tool.spec.ts
```diff
@@ -936,3 +2470,1261 @@ describe("get_context semantic retrieval", () => {
...
+               { target_type: "measure", search_text: "category" },
+            ],
+            scopes: scope,
+         }),
+      );
+      // "category" matches two DIMENSIONS by text, but a measure target must
+      // not surface them: that filter is what target_type buys over a single
+      // untyped query string.
+      expect(
+         rankedEntities(payload).every((e) => e.entity_type === "measure"),
```
Sha-Bang, line 3372: **Nit** - `.every()` is vacuously true on an empty list, so this pins the filter's safety half — no dimension leaks into a `measure` target — but cannot fail if the filter over-drops and returns nothing at all.

That over-drop is the failure mode noted on `embedding_index.ts:1026`, so this is the test that would otherwise have caught it. An `expect(rankedEntities(payload).length).toBeGreaterThan(0)` beside it closes the gap.

jswir:
Thanks for both passes. Everything is addressed; a map from thread to commit so you can resolve them.

**2026-09-04 review**

- **`cardFor` drops a nameless-source row** — excluded before it can be counted. A model-level query over an inline source has no card to live in, so `collectEntities` skips it rather than letting it match, spend a window slot, and inflate `total_available`. It stays visible through the REST model listing. (d32c6a0e)
- **Kind-blind per-target window** — the claim predicate now lives inside the scan. A `(target, kind)` VALUES table joins into `scored`, so each target's window is cut from rows it can return, and the TypeScript post-filter is gone. Tested at the scan (`embedding_index.spec`) and end to end with `limit: 1` over three nearer dimensions. (d32c6a0e)
- **`below_cutoff_count` counted before the filter** — fixed by the same change: nothing is filtered between the scan and the response, and both counts are scoped to the kinds the targets can claim (a `measure` target's `total_entities` is the package's measure count). Verified live: the impossible `sources: []` + `below_cutoff_count: 0` pair no longer occurs, and the true negative reports 38 of 38. (d32c6a0e, doc wording in 627ca8ec)
- **Lexical path shared one cap per source** — rows carry `bestTarget` (the index of the target that claimed them best) on both ranked paths, so `windowBySource` budgets per target on the lexical path too. The lexical score itself is still withheld from the wire, and the existing "withholds matched_targets" test is unchanged. (d32c6a0e)
- **`list_packages` borrowed `get_context`'s error envelope** — it has its own helper now: `{environments: []}` and "during listPackages". I also dropped the retired `results: []` from `get_context`'s error payload, since the "tier not yet known" justification no longer holds; release notes updated. (d32c6a0e)
- **Vacuous `.every()`** — added the length check. It exposed that the fixture query "category" matched no measure at all, so the test was passing on exactly that vacuity; it now searches "order", which matches a measure and a dimension. (d32c6a0e)
- **Eval scripts still granted the old names** — `run_baseline.py`, `improve.py`, `diagnose.py` grants and prompts, the reload curl, the reference doc, and the template README all use the bare names; bundle regenerated. The old names remain only as log-parsing tolerance for runs recorded against older servers. (d32c6a0e)
- **`config.ts` callers log `{ error }`** — the three callers log the message. While testing I found the same `{"error":{}}` at ~55 other sites across the server; that wants a one-line fix in the logger's format rather than 55 edits, so it's queued as its own PR. (d32c6a0e)

All six new or tightened tests fail on the previous sources and pass with the fixes.

**2026-09-03 review**

- **Initialize instructions taught the pre-converged flow** — rewritten to `list_packages` → scoped `get_context` → `execute_query`. (3c748848)
- **No test pinned the onboarding sequence** — `server.protocol.spec` now asserts `list_packages` is named first and no no-argument listing is claimed. (3c748848)
- **View-not-found remediation named `sourceName`** — now says `search_targets` plus a `scopes` entry whose `source` is the name. (3c748848)
- **`embeddingIndex` missing on `?reload=true`** — enriched on both branches, with a controller test. (6332cec4)
- **docs/ai-agents.md contradicted itself** — the flat-`results` and `sourceName` paragraphs are gone. (3c748848)
- **"six MCP tools"** — eight, with `list_packages` and `get_status` listed. (3c748848)
- **Progressive-discovery JSDoc** — says there is no such tier any more. (3c748848)

Since then the branch merged `origin/main` (ba0b052c) and swept the last `malloy_*` phrasings (a91f74fb). With the key set, I exercised the semantic path against a built server: per-target windows hold, counts are kind-scoped, and semantic bridges "revenue" to `total_amount` where lexical returned nothing.

packages/server/src/mcp/tools/get_context_tool.spec.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/get_context_tool.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/get_context_tool.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/get_context_tool.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/embedding_index.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/get_context_tool.ts
jswir: fixed - see comment for details

packages/server/src/mcp/tools/get_context_tool.ts
jswir: fixed - see comment for details

docs/configuration.md
jswir: fixed - see comment for details

docs/ai-agents.md
jswir: fixed - see comment for details

packages/server/src/controller/package.controller.ts
jswir: fixed - see comment for details

packages/server/src/mcp/error_messages.ts
jswir: fixed - see comment for details

packages/server/src/mcp/server.protocol.spec.ts
jswir: fixed - see comment for details

packages/server/src/mcp/server.ts
jswir: fixed - see comment for details

PR #1029, opened 2026-08-15
fix(server): make pre-aggregation's live fallback actually fall back

jswir:
Follow-up to #1024. Two bugs made `PREAGGREGATE_MODE` unusable above `off`. They ship together because fixing the first alone only moves the failure: it makes the live path reachable in `on`, which is where the second one lives.

## `on` broke every unannotated source in an annotated model file

The routing block swaps in the companion model for **every** query on a file that declares `#@ preaggregate` anywhere, and wraps the swap in a `try`/`catch` meant to serve live when the query does not compile there. But `loadRestrictedQuery` returns a **lazy** `QueryMaterializer` — it cannot throw at that point. The compile error escaped at prepare/run, past the catch, as `Reference to undefined object '<source>'`.

So a source with no measure to roll up was unqueryable the moment a sibling source in the same file got an annotation. From boot, no build involved. On a realistic package that is most of it.

The storage tier hit this exact problem and left the fix documented in `loadServeShapeQuery`:

> Compile eagerly so ineligibility surfaces HERE — Malloy compiles lazily, so without this the error would escape at prepare/run instead of at the caller's try, defeating the safe fallback.

Pre-aggregation copied the `try`/`catch` but not the `await`. This adds it, threading the **givens and build manifest the run will use** — without them the probe would reject a `given:`-using query the run would have accepted, which is a silent loss of acceleration rather than an error, and this feature's worst failure mode.

## `build-only` broke every query on a model without `##! experimental.persistence`

`build-only` builds the rollups and binds a manifest, but compiles no companion. Those entries reached the author's model — which has no reason to carry the persistence flag, since the companion declares its own — and Malloy refuses a non-empty `buildManifest` without it:

```
HTTP 400: A non-empty `buildManifest` was supplied, but the model is
missing `##! experimental.persistence`.
```

Every query on the model, covered or not. `on` masked it only because every query reached the companion, which does have the flag.

The rule is by **origin, not by mode**: a synthesized rollup exists only in the companion and can substitute nothing anywhere else, so its entry is stripped from the manifest handed to the author's model — on the query path, on the runtime live-fallback retry, and in notebook cells alike. Read off the build plan's `origin`, which exists to carry exactly this provenance, rather than the `__preagg__` naming convention an author's source could collide with. There is precedent in the same function: the storage tier already suppresses the manifest for its transient shape for the same reason.

## Tests

Three regression cases, each **verified to fail on `dff6642b`** and pass here:

- a query naming a source the companion does not import
- a bound rollup manifest against a flagless model, pointed at a deliberately non-existent table so a regression fails loudly rather than passing on a substitution that happened to work
- the origin classification, against a package holding both an authored `#@ persist` source and a synthesized rollup

Neither existing serve-seam test reached the catch, which is why this went unnoticed: `"falls back to live for a query no rollup covers"` groups by a field **on** the annotated source, which compiles against the companion fine and returns through the base member; `"leaves a model with no annotations untouched"` has no annotations, so no companion is built and the block is skipped. The gap was the case where both are true at once.

Also pins the row order in the shared `COVERED` fixture query. Both categories total 30, so the default ordering is a tie the live and routed shapes break differently — it fails roughly two full-file runs in three on the parent commit, which reads as a routing bug and is not one. Separate defect from the two above; happy to split it out.

## Verification

- `bun run test:unit` — 2609 pass, 0 fail. `tsc --noEmit`, eslint and prettier clean on all touched files.
- `bun run test:integration` — 255 pass. The 1 failure is the concurrent-package E2E timing out on `gs://publisher_test_packages/gcs_faa.zip` (needs GCS access), the same one #1024's test plan documented; of the 4 errors, 3 are `beforeAll() expects a function as the second argument` parse failures in spec files this diff does not touch, and 1 is a ConnectionRefused cascading from the GCS timeout.
- End to end on a DuckDB package (20k rows, three grains across `count`/`sum`/`max`): in `on`, unannotated siblings now answer; routing is unchanged, proven by mutating the base after a build and confirming the six covered queries still return pre-build numbers byte-identically to before this change while the two uncovered ones pick up the new rows. In `build-only`, all eight queries answer live at live latency while three rollups still build and bind.
- No measurable cost from the extra probe: routed queries 3-12ms vs live 33-35ms, unchanged.

## Not addressed

`#1024`'s cost argument for multiple grains rests on picking the *smallest* rollup that covers a query, but the composite members are emitted sorted by synthesized name (grain slug plus digest, unrelated to cardinality) and the resolver takes the first member that covers. My fixtures could not distinguish first-wins from smallest-wins, because the alphabetically-first covering member also happened to be the smallest. Worth a separate look; it is a cost question, not a correctness one.

No routing metric added. `preaggRouted` means "compiled against the companion", not "served from a rollup" — a query can compile there and still resolve to the base member — so reporting it as a hit rate would be wrong, and #1024's reasoning for having no such metric stands.

### DIFF
--- a/packages/server/src/service/model.ts
+++ b/packages/server/src/service/model.ts
@@ -334,6 +334,8 @@ export class Model {
     *  (or returning undefined) means no override: the runtime-baked manifest
     *  applies, which serves live when unbound. */
    private freshnessResolver?: () => BuildManifest["entries"] | undefined;
+   /** See {@link setPreaggregateEntityIdResolver}. */
+   private preaggregateEntityIdResolver?: () => ReadonlySet<string>;
    /** Entry-point gates per declared source name — see
     *  {@link computeEntryPointGatesBySource}. The one answer that
     *  `sources[].authorize`, {@link getAuthorize} and the early gate all read. */
@@ -1709,6 +1711,53 @@ export class Model {
       return entries ? { entries, strict: false } : undefined;
    }
 
+   /**
+    * Set by the owning Package (see Package.wireFreshnessResolvers). Supplies the
+    * `sourceEntityId`s of the rollups pre-aggregation synthesized, which
+    * {@link withoutPreaggregateEntries} strips from the manifest. A resolver
+    * rather than a value because the build plan it reads is computed AFTER the
+    * models are wired.
+    */
+   public setPreaggregateEntityIdResolver(
+      resolver: () => ReadonlySet<string>,
+   ): void {
+      this.preaggregateEntityIdResolver = resolver;
+   }
+
+   /**
+    * `manifest` with pre-aggregation's own rollup entries removed.
+    *
+    * A synthesized rollup exists ONLY in the companion model (see
+    * preaggregation_synthesis) — the author's model never declares it and can
+    * never reference it, so its manifest entry can substitute nothing there. It
+    * is not merely useless: Malloy refuses a non-empty `buildManifest` against a
+    * model without `##! experimental.persistence`, and a model that declares
+    * `#@ preaggregate` has no reason to carry that flag, since the companion
+    * declares its own. Passing the full manifest to the author's model therefore
+    * turned every query on it into a 400 the moment a build bound a manifest —
+    * which is what `build-only` does by definition, and what `on` does for any
+    * query that does not compile against the companion.
+    *
+    * So the rule is by ORIGIN, not by mode: only the companion sees the rollups.
+    * Returns undefined when nothing survives, matching
+    * {@link resolveFreshBuildManifest}'s "no override ⇒ serve live".
+    */
+   private withoutPreaggregateEntries(
+      manifest: BuildManifest | undefined,
+   ): BuildManifest | undefined {
+      if (!manifest) return undefined;
+      const preaggregateIds = this.preaggregateEntityIdResolver?.();
+      if (!preaggregateIds || preaggregateIds.size === 0) return manifest;
+      const entries = Object.fromEntries(
+         Object.entries(manifest.entries).filter(
+            ([sourceEntityId]) => !preaggregateIds.has(sourceEntityId),
+         ),
+      );
+      return Object.keys(entries).length > 0
+         ? { ...manifest, entries }
+         : undefined;
+   }
+
    public getSources(): ApiSource[] | undefined {
       return this.curateForDiscovery(this.sources);
    }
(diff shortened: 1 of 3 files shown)


kylenesbit:
Reviewed this while making pre-aggregation on-by-default in #1030, which turns both of these from "unusable above `off`" into "unavoidable" — so #1030 cherry-picks your commit (authorship preserved) rather than waiting. Both bugs reproduce exactly as described, and I confirmed each fix is what closes it by removing it and watching the named error return: `Reference to undefined object 'regions'` for the lazy compile, and `A non-empty buildManifest was supplied, but the model is missing ##! experimental.persistence` for the manifest. The sequencing claim holds too — the second stays masked until the first is fixed.

Three things from the port that are worth your time.

**Your manifest test passes for the wrong reason once the flag is gone.** It proves itself via `build-only`, where no companion is compiled so a covered query reaches the author's model. With the feature always on, a covered query routes to the companion and is *entitled* to the full manifest, so `COVERED` no longer exercises the stripping at all and the test goes green either way. In #1030 I repointed it at the path that is still reachable — an unannotated sibling with a rollup entry bound — and verified it fails without the fix. Nothing to change here if you land first; just flagging that the assertion's teeth depend on the mode.

**Your "Not addressed" first-wins question is settled: it is first-wins.** Your fixtures could not distinguish it because the alphabetically-first covering member also happened to be the smallest. This one distinguishes it, because sorting dimensions into the slug puts a superset grain *before* its subset:

```malloy
  #@ preaggregate grain="b_dim"
  #@ preaggregate grain="a_dim, b_dim"
  measure: total is amount.sum()
```

With 200 distinct `a_dim` and 2 distinct `b_dim`, `run: orders -> { group_by: b_dim; aggregate: total }` reads the 200-row `a_dim_b_dim` table over the 2-row one. Pinned as a canary in #1030 and filed as #1031 with a suggested fix (order members by ascending grain-dimension count, since a subset grain is never larger than its superset). Your instinct that it is cost and not correctness is right.

**The givens threading is necessary, but not for the reason the comment gives.** The comment says that without them the probe would "reject a `given:`-using query the run would have accepted". Measured both ways, it is the reverse: a model-level `given:` does not cross the companion's `import`, so the companion surfaces none — `unknown given 'MIN_AMOUNT'. Model surfaces []` — and a given-supplying query *cannot* route at all.

- Probing **with** the givens, as you have it: the probe fails, the catch fires, the query is served live and correct.
- Probing **without** them: the probe compiles, routing is taken, and the run then fails with that same error — an escaped 400, the same bug class the eager compile exists to prevent.

So that half of the fix is load-bearing for correctness rather than for acceleration, which makes it stronger than advertised. #1030 corrects the comment, adds the test that path never had (rollup bound to a non-existent table, so a query that wrongly routed fails loudly instead of quietly returning the right answer from the base), and documents the resulting limitation for authors: supplying a given means no acceleration. That one looks partly inherent to me rather than a gap to close — a rollup is built with the givens in force at build time, so it could not answer a different value from stored rows anyway.

Thanks for catching these; the lazy-compile one in particular would have shipped as "adding one annotation breaks the rest of the model file".

PR #81, opened 2025-05-02
Update publisher run instructions

kylenesbit:

PR #1032, opened 2026-08-17
docs(skills): add the four-stage model evaluation loop

jswir:
## Summary
- mirror the four shared evaluation skills from [ms2data/agent-skills#81](https://github.com/ms2data/agent-skills/pull/81): deterministic answer scoring, failure diagnosis, evidence-backed model improvement, and the gated evaluation loop
- expose the skills through Claude discovery, the Publisher skill index, the npm skills package, and the generated MCP prompt bundle
- keep the shared files byte-identical with upstream; upstream #81 was normalized to Publisher's punctuation policy in ms2data/agent-skills@4941858

## Test plan
- [x] `bun run test:skills`
- [x] `cd packages/server && bun test src/mcp/skills`
- [x] verify all six shared files match the upstream PR head byte-for-byte
- [ ] exercise the loop end-to-end against a local Publisher package (follow-up validation in progress)


jswir:
Superseded by #1101, which ports the loop as it stands today (with the scripts, the ledger contract, the golden verifier and the run viewer) on top of the manifest in #1088. This branch was a five-week-old snapshot without the scripts; rebasing it would have meant porting things that had since moved. Closing rather than rebasing.

PR #1033, opened 2026-08-17
docs: add the dbt roadmap

kylenesbit:
## What this is

A new doc, [`docs/dbt-roadmap.md`](https://github.com/malloydata/publisher/blob/kylenesbit/dbt-roadmap/docs/dbt-roadmap.md), stating plainly how Malloy and dbt fit together today, where the gaps are, and the plan to close them — written for teams running (or considering) dbt below a Malloy semantic layer, and shaped by feedback from teams running that pattern in production.

## Highlights

- **The thesis:** data should be modeled and transformed top-down; replacing SQL with Malloy is like replacing C++ with TypeScript, and the persistence platform is the optimizing runtime that makes the high-level representation affordable. dbt keeps the systems core; the seam between them is real but moves down over time.
- **The gaps, scored:** orchestration, incremental, and materialization configs are closed; seeds mostly closed; tests, contracts, SCD2, staging shapes, and per-tenant variance are open, each with a named plan.
- **The plan, in three motions:** Adopt (semantic-layer converter, reconciliation harness, a dbt-users guide) · Integrate (`dbt bind`/`check`, honest CI signals, `ci compile`, `lint`, refresh signaling, package variants, lockstep provenance) · Convert (governance → SCD2 → staging → variance, staged cheapest-first).
- **Open source first:** capability lands in the OSS Publisher; the managed platform layers reliability on top.
- Ends with open questions where input from teams running the pattern will shape the design.

Also adds the doc to the `docs/README.md` index under Concepts.


adamribaudo-ff:
This is great. Very ambitious and refreshing. A few points where I'd appreciate clarification.

- In dbt, we have dbt macros and dbt packages which help solve a variety of code re-use problems. They can also be combined in interesting ways (see [macro overrides](https://discourse.getdbt.com/t/how-to-customize-macros-in-dbt-packages-within-the-codegen/18455/2)). 
   - Packages: We have tenant-specific dbt transformations and then a dbt package that handles mart creation. If we moved to Malloy, we would continue to have a need for both tenant-specific Malloy transformations and a 'conformed' Malloy model re-used across tenants. I wonder what you would propose here as this is a hand-off from 1 Malloy package to another package. Technically this should be easier than the dbt manifest -> Malloy bindings problem but I'm not sure if it's addressed anywhere.
   - Macros: We define a set of common transformations as macros in a dbt package and then deploy them to tenant-specific projects with the option to 'override' them as needed. I don't think there's any reusable function pattern in Malloy but hopefully the value is clear. N sources may benefit from the same transformation pattern that should be defined once. Ideally this transformation can live in the current Malloy package or an "imported" or "referenced" package.
- You mention 'tests' but I think you mean 'data tests' and not 'unit tests' yes? We use both. Not sure if that's on the roadmap at all. 
- Do you see SCD2 as a managed platform concern? Or just parts of it? I can read it both ways from the write-up. It's a sensitive scope given that SCD2 data is not recoverable from the source data itself. Not "owning" it or its lifecycle as part of the OSS offering feels problematic. 

And then some commentary that you may or may not find useful 😅 

- The 'performance optimization' angle expressed isn't that much of a priority for us. It seems like there is a very long tail of nuts & bolts features required that, if missing, won't let anyone get to the point of optimization. So personally, I'd focus on those first. 
- I would love to adopt these changes incrementally, but that feels difficult. I can't move away from dbt until I have all of the above: seeds, tests, snapshots, macros, etc. 
- It's hard to take the Malloy language seriously as an alternative to dbt or jinja-SQL when there are so many gaps that have remained after many years. You posted about [unions recently](https://malloy-community.slack.com/archives/C025JAK8G0N/p1786927863089509) which is a big one but there are many others along with gaps, outdated info, and quirks in the [doc site](https://docs.malloydata.dev/documentation/) and [blog](https://docs.malloydata.dev/blog/) as well as the [373 open issues](https://github.com/malloydata/malloy/issues) (~5 issues closed a month but ~8 opened per month over the last year). This could become a drag on any party seriously considering Malloy as an alternative to dbt, regardless of your publisher roadmap. There are probably limits to what the publisher team can do on its own towards this roadmap without addressing those language issues. This is said with ❤️ from an avid Malloy fan so I hope it's taken as constructive.

kylenesbit:
Thanks @adamribaudo-ff — this is exactly the kind of feedback the open-questions section was hoping for. Taking your points in order. The doc is updated in e61c593e.

**Packages and macros.** You found a real hole; the doc didn't name code reuse across projects at all. Three pieces, now a gap row:
- *Package dependencies.* We're scoping npm-style package management for Malloy packages, driven by a customer need: a tenant package declares a versioned dependency on a shared conformed package, resolved through a lockfile, the way `packages.yml` does today. "npm-style" deliberately — the model is npm's (declared, versioned, lockfile-resolved, registry-backed); whether it rides on the npm registry itself is part of what's being scoped. Your two-package setup is a good test case — if you're willing, a sketch of how your conformed package and tenant packages are laid out would feed straight into that scoping.
- *Override sources.* The piece that makes that useful in a multi-tenant setup: the shared package declares a typed default source, the tenant package rebinds it. That's your macro-override pattern at the source level. It was buried in Stage 4; it now sits next to package variants in section B.
- *Expression-level macros.* Honest answer: Malloy has no user-defined function, so "the same transformation across N fields" has no unit to publish today. That's language work, same bucket as unions, and it's now in Lane B.

I want to be careful not to claim package dependencies "solve macros" — they solve distribution; the other two are what your workflow actually needs.

**Tests.** Yes, data tests. You're right that unit tests deserve a mention, and they're cheap here: a model is pure declarations, so a unit test is a fixture CSV per source plus an expected result on the package duckdb sandbox. It lands in `lint` alongside data tests, with the limits you'd expect (dialect-specific expressions and `connection.sql()` sources test on duckdb, not the warehouse). Added to Stage 1.

**SCD2.** Your reading was right and the doc was wrong — it drew on some platform-side design notes and put too much on the managed side. The rule we actually follow is the one the rest of Publisher uses: OSS implements the mechanism plus a basic scheduling policy, and exposes the mechanism over the API for external orchestration. For SCD2 that means the write path *and* the invariants that make history safe — history artifacts exempt from collection and re-addressing; a definition change carries history forward or refuses — ship in the open-source Publisher. The managed layer adds retention policy, backups, and freshness objectives. Stage 2 is rewritten to say so.

**On the commentary.** Agreed on sequencing — the *Now* row was already entirely nuts-and-bolts, but the doc leads with the runtime framing, so it read as the priority; there's now a line making the order explicit. On incremental adoption, the intended path is one mart at a time with dbt running below throughout — Lane A is what makes that possible — but the doc never spelled it out; there's a paragraph on it now, and it says plainly that tests, snapshots, and reuse gate the first full cutover, which is why they're first. On the language backlog: it's upstream of this roadmap and out of this repo's hands, which is exactly why Lane A exists — conversion isn't gated on it. Unions are first in Lane B, and I'd rather not promise more than that here.

girishjeswani (approved):

sagarswamirao (approved):

PR #1035, opened 2026-08-17
docs(skills): stop the two data-app skills naming a shipped example

mlennie:
Slice 9a of the ten-way split of #935. The publisher-side copy-back of [ms2data/agent-skills#79](https://github.com/ms2data/agent-skills/pull/79), which merged as `d4f066b`.

Three files, five lines. Both `SKILL.md` files are on the shared engine list, so `skills/README.md:25` requires the change to land upstream first and be copied here, keeping the two repos byte-identical. That is what this does.

## What the four lines fix

All of them stop the guidance depending on a package that exists in this repo and does not in the one an agent reads from, and stop it naming an example that #935 deletes.

- The runtime skill's "patterns that work" preamble and the data-apps skill's scaffold paragraph now say the subscriptions model and its view names are illustrative, rather than pointing at a shipped package.
- Step 4's vendoring advice named two examples that ship a chart library in `public/vendor/`. It now names storefront alone.
- The `<malloy-render>` paragraph said "the shipped example renders rows with a plain chart library instead", whose referent was about to be deleted. It now names storefront and says what is true of it: it draws with Chart.js and has no `<malloy-render>` element in its pages.

## Byte-identity, verified against what merged rather than what was copied

#79 was force-pushed during review, so "identical to what I copied from" would not have been the same claim as "identical to what shipped". All three files that `d4f066b` touched were compared against upstream and match:

- `skills/malloy-html-data-app-runtime/SKILL.md`
- `skills/malloy-html-data-apps/SKILL.md`
- `reference/verification-harness.md`

## `skills_bundle.json` is regenerated, not merged

`main` touched that file in four commits since this branch started, so it conflicted. It was regenerated rather than text-merged, and the result was checked to be a union rather than a clobber: 55 entries before and after, identical names throughout, and only the two skills in this PR differing from `main`.

## Not in this PR

`malloy-html-data-app-runtime/SKILL.md:153` points at "the givens type table", and no such table exists in that file. It is pre-existing and byte-identical to upstream, so fixing it here would break the shared-skill invariant this PR exists to restore. It belongs upstream first, as one line of neutral prose: the information the sentence promises does exist, in `docs/givens.md:118`, but a shared skill cannot path-reference a Publisher doc without leaving a dangling link on the other side.

Merges clean with `main`.


skills/malloy-html-data-app-runtime/SKILL.md
Sha-Bang, line 1: this file still differs from upstream after the change. compared against `ms2data/agent-skills@d4f066b` (still `main` there): `main` differed at 4 lines, this fixes the 2 it edits and leaves `:21` and `:70`.

`:70` is the one worth queueing. upstream says a client-supplied given is client-trusted "unless a server upstream (a trusted gateway, or an operator's per-package config) strips or finalizes it". this copy says "Publisher has no per-package control that strips or finalizes one; identity-bound givens are a planned milestone, not a shipped feature." those contradict each other about whether a control exists, and the upstream copy is the one that reads wrong — an agent following it would assume an operator can lock a given down.

`:21` is smaller: this copy documents that `sourceName` on its own is a 400, upstream doesn't.

both directions favour this copy, so there's nothing to change here — it's the same upstream-first route you're already taking for the givens type table, just two more items behind that door.

Sha-Bang (approved):
reviewed at 3ac093a. approving — all four prose changes are accurate and the bundle handling is right. one correction to the body, non-blocking.

checked rather than read:

- storefront ships `public/vendor/chart.umd.js`, `index.html:120` loads it as `./vendor/chart.umd.js`, and there's no `<malloy-render>` element anywhere in its pages. so both rewritten paragraphs are true of it.
- it also survives #1027, which rewrites that same `index.html` — the vendor load moves to `:252` but is otherwise identical, and still no `<malloy-render>`. the two in-flight PRs don't collide.
- the bundle is a real union rather than a clobber: 55 entries before and after, names identical throughout, only the two skills in this PR differing — and the regenerated entries carry the new prose, so it isn't a stale copy either.
- the givens-type-table deferral is the right call. that file has exactly two tables (the query contract and the troubleshooting one) and no givens type table, so the reference is genuinely dangling — and fixing it here would break the invariant this PR exists to restore.

**the byte-identity claim doesn't hold for two of the three files.** the body says all three were "compared against upstream and match". measured against `d4f066b`, `malloy-html-data-app-runtime/SKILL.md` still differs at two lines (inline above), and `skills/malloy-html-data-apps/reference/verification-harness.md` differs at 8 — all em-dash → colon or comma, so cosmetic, but the same invariant. neither is touched by this PR and neither is fixable here, since this repo's text is the better text in every case.

so the change is strictly good — it takes the runtime skill from 4 lines of drift to 2. it's the claim that wants narrowing to "the lines this PR touches", with the rest queued upstream-first alongside the type table.

worth knowing: nothing in CI catches this. `skills_bundle.spec.ts` only checks the bundle against this tree, so cross-repo drift is invisible and stays that way until someone diffs by hand.

mlennie:
Narrowing the claim as you asked. Body left as written, correction here so the review threads keep answering the text you reviewed.

**The claim should read "the lines this PR touches", not all three files.** You are right that the body oversells it, and the change is what you say it is: the runtime skill goes from four lines of drift to two.

**One thing in your summary is narrower than your own inline comment, and the inline one is right.** The summary says the residual drift is "all em-dash to colon or comma, so cosmetic". Measured against `d4f066b`, per file:

`skills/malloy-html-data-apps/SKILL.md` is byte-identical, zero differing lines.

`skills/malloy-html-data-app-runtime/SKILL.md` differs at two lines, `:21` and `:70`, and **neither side of either line carries an em-dash**. So those two are substantive, exactly as your inline comment describes them: `:70` has the two copies contradicting each other about whether an operator control exists, and `:21` documents that `sourceName` alone is a 400 where upstream does not. Not cosmetic.

`skills/malloy-html-data-apps/reference/verification-harness.md` differs at eight lines, and those eight are precisely the upstream lines carrying an em-dash. That one is fully cosmetic.

So the drift splits two-to-one rather than being uniform, and the substantive half is the half you already queued.

**And one constraint we did not know when we planned the upstream-first route, which makes it stronger than a preference.** The `verification-harness.md` drift cannot be closed by copying upstream down at all. That file is bundled, and `skills_bundle.spec.ts` fails on any em-dash anywhere in the bundle, so adopting upstream's copy verbatim turns the suite red. Convergence there is upstream-first by construction. Worth stating because copy-back is the reflex this slice established, and for bundled files that direction is structurally unavailable.

**On the CI gap you named:** agreed, and it is the thing that lets all of this sit undetected. `skills_bundle.spec.ts` only checks the bundle against its own tree, so drift against `ms2data/agent-skills` is invisible until someone diffs by hand, which is how these four lines survived. That is queued as its own item rather than folded in here.

Also confirming your cross-PR check held: the branch is now merged up to `main` with a local signed merge rather than the Update branch button, and the bundle was resolved by regeneration rather than textually, since #1020 had changed different entries in that same minified line. Verified by parsing: 56 entries, per-entry provenance checked, and mutation-tested both ways (substituting either side's bundle wholesale fails the sync test, only the regenerated union passes).

PR #1037, opened 2026-08-18
feat(ci): let the release carry its own notes, and fix the version floor

Sha-Bang:
Follow-up to #1036. That PR backfilled five releases' worth of missing release notes by hand; this one makes the backfill unrepeatable.

### The release now carries its own notes

Two steps used to sit *after* a release, in a human's head, and neither survived contact with an actual release: paste the `## [Unreleased]` sections onto the GitHub release page, and stamp them with the version that shipped them.

0.0.243 through 0.0.247 each shipped with none of their narrative. What that cost users, concretely:

- the `ManifestEntry.error` deprecation, and the `failures` collection to move to
- `outcome="built"` changing its basis, so a deployment trending that counter sees a step change that looks like a drop in work done
- a `publisher.db` predating 2026-06-19 returning **500 on every** `POST .../materializations`
- the entire pre-aggregation feature

None of it was hidden — it was all written down in `RELEASE_NOTES.md`, in the PR that made the change. It just never reached the page anyone reads.

`gh-release` now appends every `[Unreleased]` section to the release body beside the generated PR list, then commits the heading back to `main` stamped with the shipped version. Writing the section in the PR that changes the behaviour is now the whole job.

**Both halves agree on which sections shipped.** `extract` reads the release branch's snapshot; `stamp` runs against `main` several minutes later, and recent releases take 4–10 minutes. So `extract --titles` writes the exact heading lines it put on the page, and `stamp --titles` rewrites only those — anything merged mid-release stays `[Unreleased]` and belongs to the next release. Unscoped, such a section was stamped with a version it never shipped in: off that release's page, and off every later one. The titles file lives under `$RUNNER_TEMP` rather than in the repo because the stamp step's `git checkout -B main origin/main` would discard an in-tree one first.

Ordering is deliberate in two places:

- **The stamp runs after `gh release create` succeeds.** A release that fails to cut must not leave `main` claiming a version that does not exist. Two things enforce it, and neither is a `set -e` in that step — it has none: the runner's default `bash -e {0}` makes a failing command exit the step non-zero, and the stamp step's `if:` carries no `always()`, so its implicit `success()` keeps it from running. Adding `always()` there would remove the guarantee.
- **The stamp is `continue-on-error`.** By then the release is public and correct, and reddening a finished release over a docs commit would send someone hunting a publishing problem that does not exist. But a skipped stamp does not cost one heading the next release cleans up — that heading is exactly what the next release's `extract` matches, so the next release re-appends **this** release's narrative to its own page, and so does the one after it. Every failure path in the step therefore emits a `::warning` and a job-summary line naming the one-command fix. `continue-on-error` is a backstop for the unanticipated, not the error handling.

Note the stamp is the only `git push origin main` in this repo, and whether the workflow token may do it is unconfirmed — every recent commit on `main` arrived by squash-merge, and branch protection reads as an ambiguous 404 at our permission level. A rejection is now loud rather than silent, so the first live release tells us either way.

### Fix the version floor

`prepare`'s default patch-bumped `main`'s `packages/sdk/package.json` and walked up to the first version whose release branch and tag are **free**. That is not the same as one above npm's `latest`, and `main` lags npm permanently because release branches are never merged back.

It has been correct so far only because the tag sequence happens to be gapless. Any explicit `-f version=` that skips a number opens a permanent gap below `latest`, and a later run dispatched *without* a version drops into it, publishes, and moves the `latest` dist-tag **backwards**. Nothing fails; the release goes green.

The floor now comes from the registry, via `npm view … version` — which is the **`latest` dist-tag**, deliberately and not the highest version published. Those differ for this package, which is why the comment says so by name: `@malloy-publisher/sdk@1.0.1` really is on npm, so a max-of-versions floor would make the next release `1.0.2` and take the whole 0.0.x line with it, permanently. `latest` is the pointer `npm install` resolves and the one a release moves, so it is the right floor. A registry that will not answer fails the run rather than falling back to the stale file; the explicit-version path is unaffected either way.

### One cross-job dependency, now written down

The stamp pushes to `main` while `publish-packages` may still be polling npm. Its "main moved during this release" guard ignores the push only because `RELEASE_NOTES.md` is not among the paths it watches (`skills/`, `packages/skills/`, `bun.lock`, root `package.json`). Adding a watched path without knowing that would turn every narrative release into an aborted dispatch, so both sides now carry a comment.

### Also here

- `RELEASE_NOTES.md`: stamp the `#(authorize)` section as `[0.0.248]` (it shipped today, before this automation existed), and rewrite the "How this file is used" preamble, which described the manual process. Drops a dangling pointer to a "May 2026 review" that is not in the repo.
- `scripts/release-notes.mjs`: `--titles` on both commands; accepts every separator this file has actually used (em dash, colon, hyphen, or none), where the old regex required one and put `## ## [Unreleased]` on the page for a bare heading; and an untitled heading is now an error rather than a `## `.
- `bump_patch` refuses anything that is not a plain `major.minor.patch` instead of returning `0.0.NaN`, now that its input comes from the registry rather than from a file this repo controls.
- `.github/workflows/CONTEXT.md` and the `publisher-release` skill: reconciled with the automation. The skill's step 6 goes from *do this* to *check CI did this*.

### Tested

- New `scripts/release-notes.spec.ts`, 23 cases: the four heading forms including the bare one, empty-title rejection, `--titles` scoping (with the *unscoped* run kept as a control so the bug stays documented), the zero-section no-op, stamp idempotency, a section reworded mid-release being reported, and bad version / empty version / unknown command / missing titles file all exiting 1 with the file byte-identical.
- `build.yml`'s `lint_format` job runs `bun run test:scripts` on every PR and smoke-runs `extract` against the real `RELEASE_NOTES.md`, so a heading CI cannot read a title from fails the PR that wrote it rather than the release that would silently drop the narrative. It needed its own step: neither the root `test` script nor `test:hammer` runs in CI today.
- The mid-release window, reproduced in the shape that exposed it: snapshot with feature A, `main` with A + B merged after, `extract --titles` from the snapshot, `stamp --titles` against `main` → B stays `[Unreleased]`, A becomes `[0.0.249]`.
- Both `gh-release` shell steps run against a stubbed `git`: happy path, push rejected then succeeding on retry, push rejected on both passes (warning + summary line, step still green), empty titles file, missing titles file, and a section reworded mid-release. Plus the narrative block under `bash -e` with a malformed heading, confirming `gh release create` is still reached.
- `release.yml` and `build.yml` parse; every `run:` block passes `bash -n` and `shellcheck`.
- Floor facts checked against the live registry: `latest` is `0.0.248`, 243 versions published, and the maximum by version sort is `1.0.1`.

Not exercised end-to-end in CI, since a release workflow only runs on dispatch. The next release is the first live run, and its blast radius is bounded: the notes step degrades to the generated PR list on any failure, and the stamp step cannot redden a release that already published.


housejester (commented):
Good change, and the two ordering decisions are correctly reasoned — I verified both hold (details in the `set -e` note below). Three things I'd resolve first: a window where the automation silently eats a note, a bare assignment that can abort an already-published release, and one assumption about pushing to `main` I couldn't confirm.

## Worth acting on

**[MED-HIGH — silently loses exactly the narrative this PR exists to protect]** `extract` reads the **release-branch snapshot** (`release.yml:637`), but `stamp` rewrites **current `main`** (`release.yml:726-733`, and again in the retry at `:739-746`). A `[Unreleased]` section merged during the release window gets stamped with a version it never shipped in *and* never appeared on the page of — and it can then never appear on any later release page, because its heading is no longer `[Unreleased]`. The last five runs are 4–10 minutes end to end (`gh run list --workflow=release.yml`), and 0.0.247 → 0.0.248 were about two hours apart, so a merge inside that window is ordinary rather than exotic. Simulated against the real script:

```
# extract on the release-branch snapshot -> "## feature A shipping in this release"
# stamp against main 20 min later, feature B having merged meanwhile -> 2
7:  ## [0.0.249] — feature B, merged mid-release      <- never on the 0.0.249 page, never on any page
13: ## [0.0.249] — feature A shipping in this release
```

Feature B's code isn't in 0.0.249 either, since the release branch was cut before it merged, so the stamp is wrong on both counts. Suggested fix: have `extract` also emit the section titles it took, and have `stamp` rewrite only those headings, leaving anything newer as `[Unreleased]`. That covers the retry path too, which currently recomputes against an even newer `main`.

**[MED — a crash in the notes script aborts a release whose npm and Docker artifacts are already public]** `NARRATIVE="$(node scripts/release-notes.mjs extract)"` (`release.yml:637`) is a bare assignment. That step carries no `set -euo pipefail` of its own (three other steps in this file do), but the runner's default `bash -e {0}` applies, so a non-zero exit kills the step *before* `gh release create`. Since `gh-release` needs `publish-npm` and `publish-docker`, that leaves a published version with no tag and no release, and a re-run walks the patch number forward. Verified locally:

```
$ bash -c 'set -e; X="$(node -e "process.exit(1)")"; echo REACHED'
$ echo $?
1
```

This is also the pattern `publish-packages` spends a paragraph warning about a few hundred lines up — *"A bare assignment under set -e dies with a raw Node stack trace, no annotation and no summary line"* — and solves with `if !`. Worth the same treatment here: on failure, annotate and carry on with the generated PR list. The narrative isn't worth failing an already-published release over.

**[MED — needs confirmation, and the stated cost looks understated]** Can `GITHUB_TOKEN` push directly to `main`? I couldn't settle it from outside: `gh api repos/malloydata/publisher/rulesets` returns `[]`, and `.../branches/main/protection` 404s at my permission level, which is ambiguous. Every one of the last 20 commits on `main` is a squash-merged PR.

If the push is rejected, `continue-on-error: true` swallows it and the job stays green — and then the *next* release's `extract` re-appends the previous release's narrative to the new release page, because the heading is still `[Unreleased]`. So the cost isn't "a heading that still reads `[Unreleased]`, which the next release stamps anyway" (`release.yml:695`, `CONTEXT.md:52`); it's every subsequent release page carrying stale narrative until someone notices. Could you confirm the direct push works? If it does, I'd still soften those two comments, and it might be worth a `::warning` on a failed stamp so it's visible without opening the logs.

## Smaller correctness notes

**[MED — the comment describes something the code doesn't compute, in a way that invites a harmful "fix"]** `npm view @malloy-publisher/sdk version` returns the **`latest` dist-tag**, not the highest published version:

```
$ npm view @malloy-publisher/sdk dist-tags
{ "next": "0.0.198-dev6", "latest": "0.0.248" }
$ npm view @malloy-publisher/sdk versions   # 243 versions
... max is 1.0.1 (published 2025-09-12); 0.0.382-dev also sits above latest
```

Using `latest` is the **right** call — it's precisely the pointer that must not move backwards, and a max-of-versions floor would put the next release at `1.0.2`. But `release.yml:92-100` and the PR description both describe it as "the highest published" / "the high-water mark", which is a standing invitation for someone to switch it to `versions --json | max` and jump the project a major version. Worth having the comment name the `latest` dist-tag explicitly and cite `1.0.1` as the reason it isn't max-of-versions.

Non-blocking rider: `bump_patch("0.0.250-rc.1")` yields `0.0.NaN.1`. Unreachable today, since `npm-sdk.yml:43-48` routes every hyphenated version to `next` so `latest` is always stable. Noting it only because this PR widens the floor's input from a file the repo controls to the registry.

**[LOW — the ordering conclusion is right; the named mechanism isn't]** "`set -e` is what enforces it — the step is simply never reached" holds, but via two mechanisms, neither of which is a `set -e` in that step: the runner's default `bash -e {0}`, and the implicit step-level `success()`. Per [GitHub's docs](https://docs.github.com/en/actions/reference/evaluate-expressions-in-workflows-and-actions), *"A default status check of `success()` is applied unless you include one of these functions"* — and `if: ${{ !contains(needs.prepare.outputs.new_version, '-') }}` includes none, so a failed `gh release create` does skip the stamp. (Several blog posts claim a custom step-level `if:` *drops* `success()`; that's a misconception carried over from job-level `needs` behaviour.)

Worth correcting the comment precisely because the invariant is fragile in a non-obvious way: someone adding `always()` or `failure()` to that condition later would break it while the comment still promises `set -e` protects it.

**[LOW — verified heading edges in `extract`]** The regex `^## \[Unreleased\][ :]+—?\s*` requires a separator, so:

```
## [Unreleased]            -> "## ## [Unreleased]"   (leaks onto the public page)
## [Unreleased] - title    -> "## - title"
## [Unreleased]: title     -> "## title"   ok
## [Unreleased] — title    -> "## title"   ok
```

The "no `[Unreleased]` heading leaking through" claim holds for the em-dash form only. One-line fix: `/^## \[Unreleased\]\s*[—:-]?\s*/`.

## Minor / non-blocking

`scripts/release-notes.mjs` has no CI coverage — no `package.json` script, and no workflow but `release.yml`, which is dispatch-only. The PR is upfront about that. Given the first finding and the heading edges both live in that file, a small test over `extract`/`stamp` is cheap insurance and would run on every PR rather than on the next release.

---

**Top three to resolve before merge:** 1) the mid-release stamp window orphaning a note; 2) the bare `NARRATIVE=` assignment able to abort an already-published release; 3) confirm `gh-release` can push to `main`, and correct the two comments that price a failed stamp at one heading.

<sub>Reviewed by Claude Code (session "Breezy Quokka") at James Estes' request.</sub>

Sha-Bang:
Replying to the review above (`3b7b071c`).

Thanks — all six reproduce against the real script and the live registry, and two were release-breaking. Addressed in the commit above.

## The two release-breaking ones

**The mid-release window (MED-HIGH).** Fixed the way you suggested. `extract --titles <path>` writes the exact heading lines it consumed, one per line; `stamp <version> --titles <path>` rewrites only those, and anything newer stays `[Unreleased]` for the next release to claim. A recorded title that is no longer in the file (a section edited or removed mid-release) is reported rather than passed over, so that case is a `::warning` instead of a silent no-op.

The titles file goes to `$RUNNER_TEMP`, not the repo — the stamp step's `git checkout -B main origin/main` would discard an in-tree one before it could be read. That is also what fixes the retry path you flagged: both passes now rewrite the same set of headings, rather than the retry claiming whatever a newer `main` holds.

Your reproduction, re-run against the fix:

```
extract (snapshot):           ## feature A in this release
stamp --titles (main, later): 1
  ## [Unreleased] — feature B merged mid-release     <- still claimable by the next release
  ## [0.0.249] — feature A in this release
```

**The bare `NARRATIVE=` assignment (MED).** Now read through `if !`, the same form (and for the same reason) as `publish-packages`. On failure it prefixes the script's stderr, emits a `::warning`, writes a job-summary line, and continues with the generated PR list. stderr goes to a file rather than through `2>&1`, so nothing the script ever writes there can end up on the public page. Confirmed under `bash -e` that `gh release create` is still reached with a deliberately malformed heading in the file.

## The push to `main`

Still unconfirmed, and I could not settle it from inside either — `maintain` reads branch protection as the same ambiguous 404, and all 20 most recent commits on `main` are squash-merges. We decided to keep the direct push and make a rejection loud rather than silent.

Each failure in the step (fetch, checkout, stamp, commit, push) now sets a named reason, and if the retry fails too the step emits a `::warning` plus a `## 📝 Release notes` job-summary block saying it is not stamped, what that costs, and the one command that fixes it. `continue-on-error: true` stays, but as a backstop for something unanticipated — not as the error handling. So the first live release either stamps or says loudly that it didn't, and if the push is in fact blocked we learn it on that run rather than three releases later.

Repriced the cost in all three places you named (`release.yml`, `CONTEXT.md`, PR description): not "a heading the next release stamps anyway", but the next release re-appending the previous release's narrative to its own page, and every release after it, until someone notices.

## Smaller notes

**`latest` vs highest published.** Corrected in the comment and the PR body. The comment now names the `latest` dist-tag explicitly and cites `@malloy-publisher/sdk@1.0.1` as the reason max-of-versions is wrong — including that "correcting" it would make the next release `1.0.2` and take the whole 0.0.x line with it, permanently.

Took the rider as well: `bump_patch` refuses anything that is not a plain `major.minor.patch` instead of returning `0.0.NaN`, and the floor is validated with an `::error` annotation *before* it reaches the function, since a failing command substitution under `set -e` would otherwise abort the step with a bare stack trace and no annotation.

**The `set -e` attribution.** Corrected with your reasoning: the runner's default `bash -e {0}` plus the implicit step-level `success()`, and an explicit note that adding `always()` to that `if:` (or `continue-on-error` to the step above) removes the guarantee while the comment still promises it.

**The heading regex.** Took your one-liner, `/^## \[Unreleased\]\s*[—:-]?\s*/`, and added an error on an empty title — so a bare `## [Unreleased]` fails rather than emitting `## `. The hyphen form (`## - title`) is covered too.

## CI coverage

Agreed, and it was the right call: both file-level bugs live in that script. New `scripts/release-notes.spec.ts`, 23 cases — the four heading forms including the bare one, empty-title rejection, `--titles` scoping (with the *unscoped* run kept as a control, so the bug stays documented rather than just fixed), the zero-section no-op, stamp idempotency, and bad version / empty version / unknown command / missing titles file all exiting 1 with the file byte-identical.

`bun run test:scripts` is wired into the root `test` script and into `build.yml`'s `lint_format` job, so it runs on every PR. It needed its own step rather than just a new file: neither the root `test` script nor `test:hammer` is run in CI today. The same step smoke-runs `extract` against the real `RELEASE_NOTES.md`, which is the half that catches a malformed or untitled heading at authoring time instead of at release time, where the fallback would silently drop the narrative.

Also ran both `gh-release` shell steps against a stubbed `git` — happy path, push rejected then succeeding on retry, rejected on both passes, empty titles file, missing titles file, and a section reworded mid-release — plus `bash -n` and `shellcheck` over every `run:` block in the file.

The release path itself still cannot run in CI, so the next release remains its first live exercise. Blast radius stays bounded: the notes step degrades to the generated PR list on any failure, and the stamp step cannot redden a release that already published.

Sha-Bang:
Correction to the comment above: the sha is now `0f1b3593` — the original commit signed off with the wrong address (`nuff@` rather than the author `nathan@`) and failed DCO, so it was amended and force-pushed. No content change.

housejester (approved):

PR #1038, opened 2026-08-18
feat(server): authenticate an S3 storage root through the AWS credential chain

housejester:
## What

An `s3` connection accepts a `provider` selector. Under `provider: credential_chain`, DuckDB resolves credentials from the host — an instance profile, a web-identity token, or the environment — instead of from a static access key pair, so a DuckLake storage root on `s3://` can be reached without minting and storing a long-lived key.

`provider: config` is the default and remains the only other value. Its emitted SQL is unchanged.

```yaml
# before — the only shape available
s3Connection: { accessKeyId: AKIA…, secretAccessKey: …, region: us-east-1 }

# now also
s3Connection: { provider: credential_chain, region: us-east-1 }
```

```sql
CREATE OR REPLACE SECRET s3_root (
   TYPE s3,
   PROVIDER credential_chain,
   CHAIN 'env;web_identity;instance',
   REFRESH 'auto',
   REGION 'us-east-1'
);
```

**No malloy-core change is needed** — the secret goes out through `connection.runSQL()`, an API `db-duckdb` already exposes. Everything below was measured against the pinned DuckDB (`@duckdb/node-api 1.5.3-r.2` → DuckDB `v1.5.3`, the same version `@malloydata/db-duckdb@0.0.427` pins, so there is one copy) rather than reasoned about.

## Why these three details are not arbitrary

**`CHAIN` is load-bearing, not optional.** With `PROVIDER credential_chain` and no `CHAIN`, DuckDB resolves against `config` *alone* — the error names it: `Credential Chain: 'config'`. That is a shared credentials file, the one source a container image does not have. A caller who omits `CHAIN` would therefore get the single link that cannot work in the deployment this feature exists for, so Publisher supplies an order (`env;web_identity;instance`) rather than passing through only what the caller set. Excluded deliberately: `config` (no credentials file in an image), `sts` (needs `ASSUME_ROLE_ARN`, which this schema has no field for), `sso` and `process` (workstation flows). Valid values at 1.5.3 are `env`, `config`, `sts`, `instance`, `sso`, `process`, `web_identity`; `assume_role` and `default` are rejected outright.

**`REFRESH 'auto'` is the only value that arms refresh at all.** A chain secret stores the credentials it *resolved*, not a reference to the provider that resolved them — `duckdb_secrets().secret_string` on a `CHAIN 'env'` secret contains the literal key id. So there are two lifetimes and only one is managed: a projected service-account token is refreshed continuously by the platform, while the credentials DuckDB derived from it expire on their own schedule. The binding clock is the role's `MaxSessionDuration` — an hour by default — not the token's TTL, which is typically far longer and misleading to reason from.

That is invisible on the build path, where the session lasts one build, and fatal on the serve path, where the attach is idempotent and the secret lives as long as the connection.

The literal string `auto` matters, and quietly. The `aws` extension stores the `refresh_info` that makes a secret refreshable under `if (refresh == "auto")` and ignores every other value; `httpfs` then declines to refresh a secret that has no `refresh_info`. So any other value — `REFRESH true` included — is accepted without complaint and arms nothing. It is in fact *worse* than omitting the clause, because a bare `sts` or `web_identity` chain arms refresh by default only when the parameter is empty, and a non-empty value skips that fallback.

**Armed, the refresh works — but it does not reach every read path.** Measured below: a direct object read re-resolves transparently at expiry and never surfaces an error, while a read through an attached DuckLake catalog surfaces `ExpiredToken` from an equally armed secret. So the connection also re-resolves the secret itself when a statement fails on an expired credential, and retries once. `REFRESH 'auto'` is the mechanism; the retry is the fallback for the paths it does not cover.

**Failure moves from first object access to attach — intentionally.** A key-based secret stores whatever it is given; a chain secret resolves at `CREATE` and throws if it cannot. So a host that cannot supply credentials now fails the attach with a legible message, instead of returning a bare `403` on a Parquet path minutes into a build. The trade-off is that a transient token-file or IMDS hiccup now fails an attach rather than one read — which turned out to matter more than expected, because a failed attach was cached and replayed for the life of the process. That is fixed in its own commit and applies to every attached database type, not just cloud storage.

## The trap this avoids

`ducklakeConfig.storage` already permits an `s3://` `bucketUrl` with **no** `s3Connection` at all, and that attaches today with no credential and 403s at first object access. Chain auth is therefore an *explicit* selector rather than an inference from absent keys — inferring it would silently turn that misconfiguration into "try the chain", and a genuinely broken deployment would look configured until a read failed. For the same reason, a key or session token supplied *alongside* `credential_chain` is rejected rather than ignored: an unused key reads as the thing granting access while something else actually does. A `chain` named under `provider: config` is rejected on the mirror argument.

## Two things beyond the secret

- **`attachCloudStorage` has two callers.** Besides the DuckLake attach path (which loads `aws` already), it is reached from the generic connection handler table, which does not. `credential_chain` is implemented by the `aws` extension and resolves at `CREATE`, so the load now lives with the secret rather than with the caller. `aws` is baked into the image (`bake-duckdb-extensions.js`), so this stays inside `EXTENSION_FETCH_POLICY=local-only` — pinned by a test asserting `LOAD aws` and no `INSTALL`.
- **Publisher does its own S3 I/O.** `gcs_s3_utils.ts` builds an `@aws-sdk/client-s3` client for schema browsing and object listing, passing the configured key pair explicitly and coercing a missing one to `""`. A keyless connection would have signed every browse request with an empty key and failed on signature rather than on configuration, so the client now omits `credentials` entirely under chain auth and lets the SDK resolve its own chain.

  **That leaves two resolvers for one config, and the divergence runs both ways.** The SDK's default chain covers a web-identity token natively, which is why the DuckDB secret must name `web_identity` and this need not — but it also covers sources DuckDB's chain does not (the shared credentials file, `AWS_PROFILE`, SSO, process) and it ignores `chain` entirely. So on a host that has one of those, the browse path can resolve a different principal than the query path. Left as-is deliberately: narrowing the SDK to `chain` needs a mapping for all seven of DuckDB's provider names plus direct dependencies on providers that are currently only transitive, and that table would drift against both sides. It cannot arise in a container image, which has no credentials file, profile, or SSO cache, and where it can arise the query path is authoritative and fails visibly. Documented at the call site.

GCS is untouched — its secret is HMAC-only and has no chain equivalent through this path, so a `provider` set on a GCS root must not divert it.

## Where the checks fire, and rolling this out

The credential guards run at **config load**, not at attach — the distinction that matters for a storage destination, which is only attached at its first *build*, so an error would otherwise surface hours after the config change that caused it.

How strict each check is depends on what a rejection costs, not on which field it is:

- **A storage destination** is checked in full, key pair included, for a GCS root as well as an S3 one. `validateStorageDestinations` rejects one entry with a reason and leaves the others accepted, so early failure is free.
- **An environment connection** — a DuckLake connection, or an S3 attached database — is checked for *shape* only: the enum and the mutual-exclusion rules, which no config that loads today can violate. That validator's throw reaches `assembleEnvironmentConnections`, which has no per-connection recovery, so enforcing the key pair there would take a whole environment down for one incomplete `s3Connection` that previously just failed when used. It still fails at attach, exactly as before.

Both dialogs enforce the pair on the `config` path, so the UI still catches a missing key before it is saved.

**Rolling this out:** an older build handed `provider: credential_chain` ignores the unknown field, sees an `s3Connection` with no key, and refuses — so upgrade before anything starts emitting the selector. Getting the order wrong is loud and fails closed, not a silent wrong-identity read. The reverse direction is safe: a new build handed a key pair takes the `config` path, whose emitted SQL is pinned byte-identical.

**One thing worth naming for anyone exposing Publisher beyond localhost:** connection create and test take an arbitrary body, and this PR changes what reaching them is worth — from "use credentials you already hold" to "act as the host's role against any bucket it can reach", and with `endpoint` set, sign requests to a caller-chosen host with that identity. The credential itself is never disclosed, and Publisher is already documented as unauthenticated and localhost-only, so this is not a new bypass — but it raises the value of the existing one, and a deployment that wants to refuse host-identity connections outright has no switch for it today.

## Not in this PR

**The region is defaulted twice** and left that way: the spec carries `default: us-east-1` and `connection.ts` applies `credentials.region || "us-east-1"` on top, so a caller who does not set a region cannot reach a bucket outside `us-east-1`. It is orthogonal to chain auth but lands on the same statement, since a chain secret still needs a correct `REGION`. Worth a follow-up that treats "no region for a remote root" as an error rather than a silent default, because a silent default is what makes a mis-regioned bucket invisible.

**A remote `bucketUrl` with no storage connection still attaches** rather than being refused. Arguably a configuration error, but changing it is a behaviour change for anything relying on it today, so it stays as-is here.

## Schema note

`provider` deliberately carries **no** schema `default:`, though the reason is narrower than it first appears and is worth stating precisely rather than as a rule of thumb. At the pinned `openapi-typescript` (6.x) a defaulted property still generates as **optional**, so a `default:` would be harmless today. It is at 7.x that a default makes the property *required*, which would force every caller to name a provider. So this is robustness across that upgrade, not avoidance of a present-day break — and note `region` already carries a `default:`, so a 7.x bump has to be dealt with either way.

The default is documented in the description and applied when the connection is attached.

## Tests

New tests across four specs, and `packages/server` typechecks at every commit individually rather than only at the tip.

The three pre-existing key-based shapes (plain, custom endpoint, session token) and the GCS shape are asserted as **whole statements** rather than substrings — chain auth added a branch ahead of them and the point is that nothing downstream moved. The secret-SQL assembly and the credential validation were each split into a pure exported function to make that possible without a live DuckDB.

**Every new guard was mutation-checked** — deleted one at a time, confirming at least one test goes red. Two did not survive the check initially. The `aws` extension load was unpinned, and now has a test driving the generic attach path and asserting `LOAD aws` precedes `CREATE OR REPLACE SECRET`. And the refresh assertion checked only that a `REFRESH` clause was *present*, which is true of the value that disables refresh — it now pins the value, and reverting to the old one turns it red.

## Verified on a live EKS cluster

Two questions were what kept this a draft: whether `web_identity` resolves in a real pod, and what happens when the resolved credential expires. Both are answered, the second not in the direction expected.

### `web_identity` resolves a projected service-account token — confirmed

Exercised on an EKS cluster whose worker pods carry `AWS_ROLE_ARN` and `AWS_WEB_IDENTITY_TOKEN_FILE` from IRSA and hold **no static key anywhere**. With `provider: credential_chain` and the default `CHAIN`, DuckDB resolved the credential, created the secret, attached a DuckLake catalog on Postgres, and wrote Parquet to the S3 root.

A query against the persisted source then served **from that Parquet**, not from the warehouse it was built from. Proven by staleness rather than by matching numbers: a row was added to the source table, after which an identical non-persisted twin moved to the new count while the persisted source held the stored one. Both return correct results, so agreement proves nothing — only the divergence does.

That covers the path this PR exists for: selector → chain resolution → secret → attach → write → serve.

### Expiry: `auto` arms it, and on a direct read the refresh works

What made this legible is that a build and a serve do **not** share a connection. A build attaches read-write for the life of that build; a serve holds a long-lived read-only attach whose secret lives as long as the connection. Each mints its own secret on its own clock. So builds kept succeeding while serves failed, on the same pod at the same moment — which reads like an intermittent fault until you notice the two paths were never using the same credential.

The clearest evidence is an isolated pod: the same image and the same statement, but no Publisher — one connection held open across the expiry boundary, reading one Parquet object every 20s, with two containers differing only in the `REFRESH` value.

Immediately at creation, the value shows up in the secret itself:

```
armed   REFRESH 'auto' -> ...;key_id=ASIA…LQC4;refresh_info={'region': us-east-1,
                              'refresh': auto, 'chain': env;web_identity;instance};…
control REFRESH true   -> ...;key_id=ASIA…EVVY;region=us-east-1;…      (no refresh_info)
```

And at the expiry boundary, exactly 60 minutes later:

| container | at T+1:00 |
| --- | --- |
| `REFRESH 'auto'` | credential rotated (`…LQC4` → `…D625`), **zero failed reads**, the probe spanning the rotation simply took a second longer |
| `REFRESH true` | `ExpiredToken` on every read from that moment on |

So `auto` is not a detail: it is the difference between expiry being invisible and expiry being fatal. The rotation with no surfaced error is also `httpfs` behaving exactly as its source says — it refreshes and re-issues the request internally, so a working refresh is silent by construction.

**The gap this leaves.** On the serve path *through an attached DuckLake catalog*, an equally armed secret still surfaced `ExpiredToken` and the connection's own retry is what carried the query. Same cluster, same hour, same statement — so the difference is the read path, not the secret. `s3fs.cpp` has call sites that pass a null context and opener, and `ReloadS3AuthMaterial` returns false when there is no opener, which is a plausible mechanism but not one this PR has isolated. Untangling it needs a probe that attaches a catalog rather than reading the object directly.

That is why the retry stays. On `ExpiredToken` the connection re-runs `CREATE OR REPLACE SECRET` once and retries the statement. The secret is replaced **in place** on the live connection — a statement, not a session property — so nothing is detached and no query in flight is disturbed, confirmed by no re-attach appearing in the logs. Cost is one STS round trip per connection per expiry rather than one per query, which is why it is reactive on the error rather than scheduled or per-statement. Only `ExpiredToken` retries, so a credential that was never valid still fails on the first attempt rather than twice; a key-pair connection registers no renewer at all. Where DuckDB's refresh does reach, it wins the race and the retry never fires.

The whole path — selector, chain resolution, secret, attach, write, serve, and expiry — is evidence rather than argument.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4747,8 +4747,15 @@ components:
           description: "AWS secret access key (required for `provider: config`)"
         region:
           type: string
-          description: AWS region (e.g., us-east-1)
-          default: us-east-1
+          description: >-
+            AWS region (e.g., us-east-1). Optional, and Publisher falls back to
+            us-east-1 when it is absent — but the fallback is applied in code, not
+            by this schema: a `default:` here would be documentation only, since
+            the generated clients leave an omitted property undefined rather than
+            filling it. Stated this way so a caller reading the schema does not
+            conclude the wire supplies it. A remote root in another region MUST
+            set this; omitting it reaches us-east-1 and fails as a missing bucket
+            rather than as a misconfiguration.
         endpoint:
           type: string
           description: Custom S3-compatible endpoint URL (optional, for MinIO, etc.)
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4706,10 +4706,15 @@ components:
       type: object
       description: AWS S3 connection configuration for DuckDB
       properties:
-        # No schema `default:` here on purpose. openapi-typescript generates a
-        # property carrying a default as REQUIRED, so declaring one would force
-        # every existing caller to name a provider. The default is stated in the
-        # description and applied at attach time instead.
+        # No schema `default:` here on purpose, though the reason is narrower than
+        # it looks: at the pinned openapi-typescript (6.x) a defaulted property
+        # still generates as optional, so a `default:` would be harmless today. It
+        # is at 7.x that a default makes the property REQUIRED, which would force
+        # every caller to name a provider. Omitted for robustness across that bump,
+        # not to avoid a present-day break. Note `region` below carries a default
+        # and so already has that exposure — a 7.x upgrade has to deal with it
+        # either way. The default is stated in the description and applied when the
+        # connection is attached.
         provider:
           type: string
           enum:
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -4706,12 +4706,40 @@ components:
       type: object
       description: AWS S3 connection configuration for DuckDB
       properties:
+        # No schema `default:` here on purpose. openapi-typescript generates a
+        # property carrying a default as REQUIRED, so declaring one would force
+        # every existing caller to name a provider. The default is stated in the
+        # description and applied at attach time instead.
+        provider:
+          type: string
+          enum:
+            - config
+            - credential_chain
+          description: >-
+            How credentials are obtained; defaults to `config` when omitted.
+            `config` uses the accessKeyId and secretAccessKey on this object,
+            which are then required.
+            `credential_chain` resolves them from the host instead — an instance
+            profile, a web-identity token, or the environment — so no key is
+            stored, and supplying one is rejected. Chain resolution happens when
+            the connection is attached, so a host that cannot supply credentials
+            fails there rather than on first object access.
+        chain:
+          type: string
+          description: >-
+            Ordered list of credential providers to try, semicolon-separated,
+            for `provider: credential_chain` (DuckDB's CHAIN). Recognized
+            providers are env, config, sts, instance, sso, process, and
+            web_identity. Omit to use Publisher's default order, which covers
+            the environment, a web-identity token, and an instance profile.
+            Rejected under `provider: config`, where the key pair beside it
+            is what grants access.
         accessKeyId:
           type: string
-          description: AWS access key ID
+          description: "AWS access key ID (required for `provider: config`)"
         secretAccessKey:
           type: string
-          description: AWS secret access key
+          description: "AWS secret access key (required for `provider: config`)"
         region:
           type: string
           description: AWS region (e.g., us-east-1)
@@ -4722,9 +4750,6 @@ components:
         sessionToken:
           type: string
           description: AWS session token for temporary credentials (optional)
-      required:
-        - accessKeyId
-        - secretAccessKey
 
     AzureConnection:
       type: object
--- a/packages/sdk/src/components/Connections/AddConnectionDialog.tsx
+++ b/packages/sdk/src/components/Connections/AddConnectionDialog.tsx
@@ -56,6 +56,10 @@ export default function AddConnectionDialog({
    // DuckLake top-level connection state
    const [ducklakeCatalogType, setDucklakeCatalogType] = useState("postgres");
    const [ducklakeStorageType, setDucklakeStorageType] = useState("s3");
+   // Which S3 credential shape the storage fields are showing. Needed as state
+   // because this section renders uncontrolled inputs read from FormData, so
+   // `visibleWhen` has nothing to compare against without it.
+   const [ducklakeS3Provider, setDucklakeS3Provider] = useState("config");
 
    const handleClickOpen = () => {
       setOpen(true);
@@ -67,6 +71,7 @@ export default function AddConnectionDialog({
       setType("postgres");
       setDucklakeCatalogType("postgres");
       setDucklakeStorageType("s3");
+      setDucklakeS3Provider("config");
    };
 
    const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
@@ -123,11 +128,20 @@ export default function AddConnectionDialog({
                }
             });
             // Validate required fields
-            if (!s3Config.accessKeyId) {
-               throw new Error("S3 Access Key ID is required");
-            }
-            if (!s3Config.secretAccessKey) {
-               throw new Error("S3 Secret Access Key is required");
+            // Under `credential_chain` the host supplies the credential, so a key
+            // pair is not merely optional — the server rejects one that is present
+            // and unused.
+            if (s3Config.provider === "credential_chain") {
+               delete s3Config.accessKeyId;
+               delete s3Config.secretAccessKey;
+               delete s3Config.sessionToken;
+            } else {
+               if (!s3Config.accessKeyId) {
+                  throw new Error("S3 Access Key ID is required");
+               }
+               if (!s3Config.secretAccessKey) {
+                  throw new Error("S3 Secret Access Key is required");
+               }
             }
             if (Object.keys(s3Config).length > 0) {
                storageConfig.s3Connection = s3Config;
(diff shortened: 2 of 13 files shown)


kylenesbit (approved):
Approving — this is exemplary work. Five well-factored commits, every guard mutation-checked, the pre-existing key-based SQL shapes pinned byte-identical as whole statements, and the core path verified end to end on a real IRSA pod with a staleness-based proof (divergence, not agreement — the right test). I verified the design claims against the code hunk by hunk; they match.

What holds up under scrutiny:

- **Explicit selector, not inference.** A keyless `s3Connection` already attaches-and-403s today, so inferring chain auth from absent keys would dress that misconfiguration up as intent. The mutual-exclusion rejections close both directions.
- **The tiered validation is correct, not just careful.** Shape-only where a throw fails the whole environment (`assembleEnvironmentConnections` has no per-connection recovery); the full credential check only in `validateStorageDestinations`, where one entry is rejected alone — and that is exactly where lateness hurt most, since a destination attaches at first *build*. The GCS-parity commit closes the same gap on the other arm.
- **The attach-retry commit fixes a real pre-existing bug**, with a correctly guarded eviction mirroring the proxy cache.
- `chain`/`endpoint` go through `escapeSQL`; `LOAD aws` with no `INSTALL` is pinned so `EXTENSION_FETCH_POLICY=local-only` holds; DuckLake routes through the same `attachCloudStorage`, so both callers get the extension load; the api-doc `required:` removal is compensated at load time and in both dialogs, and the Python client carries no generated model for it.
- The two-resolver divergence note in `createCloudStorageClient` is the best kind of comment: it names the hole, why it is not closed, and where it cannot arise.

Four notes, none blocking:

1. **DuckDB version skew in the evidence.** The "measured against" claims cite 1.5.3, but main (and this branch, rebased and green) now ships DuckDB 1.5.5 via #1063. Worth one line confirming the EKS run used a post-rebase image, or a re-check of the chain provider names and `REFRESH` acceptance at 1.5.5.
2. **Retry-per-lookup on persistent failures.** Evicting the failed attach means a *permanently* broken attachment (bad DSN, revoked role) now pays a fresh network attempt — possibly a full connect timeout — on every lookup, where it used to fail fast from cache. Right trade for transients; a short-TTL negative cache is the follow-up if it ever bites a busy serve path.
3. **`REFRESH` re-resolution is the one unverified link**, and the write-up says so plainly. Since expiry surfaces on day three rather than at first use, please land the verification (a pod outliving one token rotation, or a read of the `aws` extension source) before the first production rollout leans on it — not before merge.
4. **The security paragraph deserves an issue, not a PR-body paragraph.** "Connection create/test now mint host-identity access, and a deployment has no switch to refuse host-identity connections" will be invisible once this merges. Please file it (a `PUBLISHER_DENY_HOST_CREDENTIALS`-style switch), and the region double-default follow-up alongside.

One UI question: when a user switches provider in the dialog after typing a key, are the now-hidden key fields cleared on submit? The server will correctly reject key-alongside-chain, but the user would be staring at an error about a field they can no longer see.

Issue #1039, opened 2026-08-18
Pre-aggregation: rollups cannot build on BigQuery — synthesized #@ persist has no dataset qualification


stefanoperenzoni:
## Summary

A `#@ preaggregate` rollup on a BigQuery package passes validation, appears in `buildPlan` with `origin: preaggregate`, and its synthesized SQL is correct — but the materialization run always fails, because the rollup's table name is never dataset-qualified and there is no way to supply a dataset. As far as I can tell, rollups cannot build on BigQuery at all in 0.0.248.

(Reported on Slack; filing with traces as requested. Source/model/dimension names below are genericized from our real model; the behavior and traces are verbatim.)

## Environment

- `ms2data/malloy-publisher:0.0.248` (Docker, standalone), BigQuery connection (`type: bigquery`, service-account auth)
- One measure annotated `#@ preaggregate grain="category, period, region, tier"` on a source in `models.malloy`

## What happens

`POST .../materializations` with `{}` fails in ~1.5s:

```
orders__preagg__category_period_region_tier__1a2b3c4d: Table "orders__preagg__category_period_region_tier__1a2b3c4d_0123abcd4567" must be qualified with a dataset (e.g. dataset.table).
```

That is BigQuery's name-resolution error (not permissions — nothing was created or denied). The failing name is the rollup's source name plus the staging suffix from `stagingSuffix(sourceEntityId)`.

## Trace

- `preaggregation_synthesis.ts` — `emitRollup` emits the rollup as a bare `#@ persist` with no `name=`.
- `materialization_service.ts` — `selfAssignTableName` is `deriveAnnotationFields(persistSource).name || persistSource.name`. With no `name=`, the physical name is the unqualified rollup source name, which BigQuery rejects.
- On DuckDB (where `examples/storefront`'s e2e ran) an unqualified name is valid, so this never surfaced.

## What an author cannot do about it

Confirmed empirically on 0.0.248, checking `buildPlan.sources[].annotationFields` after each attempt (it is the exact map `selfAssignTableName` reads):

1. `#@ preaggregate grain="…" name="dataset.table"` on the measure — **silently ignored**; `preaggregation_annotation.ts` reads only `grain`, `annotationFields` stays `{}`. No validation error either, which seems to contradict the feature's own fail-closed rule ("an annotation that silently does nothing must never be possible") — if `name=` stays unsupported, an unknown key on `#@ preaggregate` should arguably be refused.
[...]

PR #1041, opened 2026-08-18
fix(server): give a pre-aggregated rollup a namespace, so it can build on BigQuery

housejester:
A pre-aggregated rollup could not be built on BigQuery at all ([#1039](https://github.com/malloydata/publisher/issues/1039)), and the same is true of any dialect that requires a qualified `CREATE`:

```
Table "eop_monthly__preagg__…_08b6bd2452f2" must be qualified with a dataset (e.g. dataset.table).
```

`emitRollup` emitted a bare `#@ persist`, so the build self-assigned an unqualified table name. DuckDB and Postgres accept that and put the table in the default schema; BigQuery rejects it outright. There was no way for an author to supply one either: `#@ preaggregate` parsed only `grain`, `#@ persist name=` on the base names the base's own table, and `BigqueryConnection` has `defaultProjectId` but no dataset.

## What this adds

**`#@ preaggregate grain="category" namespace="analytics"`** names where that rollup's table is created. Only the container — the table name stays derived (`<base>__preagg__<grain>__<digest>`), because the resolver and the content address depend on it.

The namespace binds to the grain on its own line. A grain *is* a table, so two grains are two tables and may be created in two places. Measures sharing a grain share its one table, so disagreeing about where it goes is unsatisfiable rather than ambiguous: `conflicting_namespace` refuses it at publish instead of letting field order decide.

**Absent that, a rollup inherits its base's namespace** — everything before the last dot of the base's `#@ persist name=`, because a rollup of X belongs beside X. The explicit key is not redundant with this: `#@ preaggregate` goes on a measure, so a base need not be persisted at all, and that is the case an author cannot otherwise escape.

A `storage=` base lends nothing. Its `name=` is a name in the destination's catalog while the rollup is written to the source warehouse, so inheriting across that boundary would aim `CREATE TABLE` at a schema that need not exist there — adding `storage=` to a working base would have broken its rollup.

**Validation.** The value is spliced bare into a generated `#@ persist name="…"` and joined to a generated table name, so it has to survive both. One rule, applied per dot-separated part:

```
[A-Za-z_][A-Za-z0-9_$-]*
```

That is `assertSafeSqlIdentifier`'s plain-identifier rule **plus the hyphen**, and the addition is deliberate. The house rule excludes hyphens because a hyphenated name needs quoting on Snowflake, Trino and Unity Catalog — its comment adds *"and BigQuery, the one dialect whose names really do carry hyphens, never reaches this function."* A rollup namespace does reach BigQuery, where a hyphenated project id is ordinary, so excluding it would refuse `my-project.analytics` on the dialect this was reported against.

Dots survive as separators, since BigQuery addresses a dataset as `project.dataset`. Spaces, quotes, leading digits and empty parts do not. A name that would need quoting is refused rather than mangled on both paths — inheritance splits on the last dot, which is not sound for a quoted name (`"My.Schema"` is one identifier containing a dot).

## Tests

**Unit** — derivation and precedence, the unpersisted base, the `storage=` boundary, per-grain binding, negation, the conflict refusal, and every quoting and malformed shape. Each guard was mutation-checked: removing it turns exactly its own test red.

**A hammer scenario**, `preaggregation/01-serves-from-the-rollup`, covering both routes on one source:

- **placement** — the catalog is asked which schema received *which grain*, rather than trusting the names it was given;
- **serving** — after the rollup is built, the source rows are changed and the same query must still return the stored answer. Both a rollup read and a recompute return correct numbers, so staleness is the only thing that can tell them apart.

It runs colocated on Postgres, which accepts an unqualified write as well — the point is that it can be *asked* where the table went, so the defect BigQuery reports as a hard error is observable without cloud credentials.

This needed the harness extended. `## Operator` now also reaches the scenario's source warehouse — the publisher's own SQL endpoint cannot run a statement returning no rows, so nothing else could create the schema — and it takes an optional `Expect:` table, so a scenario can assert what the tier actually wrote rather than only provision what it needs. `GRAMMAR.md` updated.

## Out of scope, deliberately

**Rollups do not follow their base into a `storage=` destination.** A synthesized rollup is colocated regardless, so it lands in the source warehouse. That is separate work, and worth knowing before it starts: `namespace=` is a placement instruction, so a deployment whose destination *derives* placement will need to refuse one.

**`#@ persist` has no `namespace=` of its own.** A persisted source still carries its container inside `name=`. Having the key on one annotation and not the other is a reasonable follow-up.

**`#@ preaggregate` still ignores an unrecognized key** rather than refusing it. [#1039](https://github.com/malloydata/publisher/issues/1039) argues it should, and the reasoning is sound — but ignoring unknown keys is how tags behave across Publisher and Malloy generally, so changing it here would make pre-aggregation an exception by accident. It deserves its own change.


kylenesbit (approved):
Approving. The design is sound and the claims in the description check out against the code — I verified rather than trusted:

- **All 105 unit tests** across the three preaggregation spec files pass at the PR head, and the "mutation-checked" claim held on a spot check: deleting the `storage=` guard turned exactly its own test red, nothing else.
- **The hyphen exception is safe end-to-end.** A spliceable namespace can never carry a quote character, so `quoteManifestTablePath` never takes its pass-through arm for a derived name, and both the CREATE and bind sides quote `my-project.analytics.<derived>` per segment consistently.
- **The rebuild advice in the docs is correct.** `computeSourceEntityId` hashes only the connection digest and canonical SQL — `name=` never enters the content address, so changing a namespace really does leave an already-built table where it is until `forceRefresh`.
- **Plan and validation can't disagree**: both skip declaration errors and non-additive measures before a namespace can take effect, and key grains on the same canonical form. "First named wins" is safe only because `conflicting_namespace` refuses the one case where order could matter — and it does.
- The hammer scenario asserts the right things: grain→schema *pairing* (the set alone would pass swapped placements), staleness on both the authored and inherited routes, and the off-grain fallthrough.

Minor, non-blocking notes — fine as follow-ups or never:

1. **New circular import**: `preaggregation_annotation.ts` imports `isSpliceableNamespace` from `preaggregation_synthesis.ts`, while synthesis imports `readPreaggregateAnnotation` from annotation. Works (hoisted functions, call-time use), but the validator would sit naturally in the annotation module and remove the cycle.
2. **`RETURNS_ROWS` comment in `hammer/run.ts` overclaims**: the regex has no `m` flag, so it is anchored once at block start, not "per statement", and a leading `--` comment *does* hide the keyword (the block would run via `pg.sql` and return no rows, failing an `Expect:` confusingly). Behavior is fine for the harness; the comment should say what the regex does.
3. **`basePersistNamespace` reads the merged tag**: `tag.has("persist")` + `tag.text("name")` means an unrelated sibling annotation on the source carrying its own `name=` (beside a bare `#@ persist`) would donate a namespace persist never named. Contrived, and consistent with how the sibling-key shape is read elsewhere.
4. **One unpinned behavior**: no test for a measure naming a namespace while another at the same grain names none — the intended outcome (the named one wins, no conflict) lives in comments but not in a spec. Cheap to add, and it's the shape a real model most likely hits.

Also noting for the record: hammer isn't referenced by any CI workflow, so green checks don't cover the new scenario — I assume it was run locally.

PR #1043, opened 2026-08-18
feat(sdk,app): render a dashboard, and route to it from the Console

mlennie:
Slice 8 of the ten-way split of #935, and the piece that makes dashboards visible. Slices 1 through 7 are on `main`; #987 landed the server side (discovery, lint, the manifest and the REST surface) and this renders it: a `Dashboard` component, a route to it from the Console, and a notebook cell that can drill into one.

Base is `main`. 16 commits, 24 files, +2266/-64. This branch was previously stacked on slice 6 with slice 7 merged in, and read 120 files against that base; rebased onto `main` now that both are landed it reads its own 24. Every earlier shape of it was unreviewable for that reason, which is why this PR is opening only now.

## What it adds

`packages/sdk/src/components/Dashboard/` is the renderer: `Dashboard.tsx` lays out the grid from the manifest #987 publishes, `DashboardTile.tsx` renders one tile. `DashboardPage.tsx` routes to it from the Console, and `ContentTypeIcon`/`Package.tsx` give dashboards their own signal in the package listing rather than borrowing a notebook's.

Two things beyond the renderer. A notebook cell can drill into a dashboard, which is `markDrillableCells.ts` plus the `Notebook`/`NotebookCell` and `RenderedResult` changes, and it is reachable from the keyboard rather than mouse-only. And `## autorun=false` is honoured in a notebook, with notebooks listed by title.

## The one server file, and why it is here

`spa_fallback.ts` gains a `dashboards` entry. A slug is a filename with its `.malloy` suffix removed, so `dashboards/report.csv.malloy` publishes the slug `report.csv`, which ends in an asset extension through no choice of the reader's. Without the entry that dashboard is listed on the package page, served by the API and reachable by in-app navigation, but a deep link or a refresh 302s to the static route and answers 404. Measured before adding it, not inferred.

The cost is stated in the docblock rather than left implicit: a package shipping a `public/dashboards/` directory can no longer address those files through that path. `pages` already pays the same cost.

## Tests

`package-dashboards.spec.ts` drives the viewer end to end in a browser. The grid fixture now puts a chart beside a table, because a grid of one content type does not exercise layout. One earlier version of the grid test raced the layout and passed for the wrong reason; it now names what it pins.

## Docs

`docs/dashboards.md` is new here and is the guide the feature needed. Worth flagging for sequencing: #1020's authoring skill links it at `SKILL.md:373` as "the full guide this skill condenses", so **#1020 cannot merge until this does**. #1026 is also based on this branch and becomes reviewable once this PR exists.

## Review history

Fourteen of the sixteen commits are review-driven, and the subjects say which: "seven findings from the code review, four of them defects", "four review fixes, two of them in code this branch just added", and three separate commits correcting claims in prose rather than code (four in the dashboards guide, three about drill, one release-note claim found false). That ratio is the honest summary of how this branch got here: the recurring defect class was a claim that was true when written and false when read, not a logic error.

## Verification worth knowing about

The rebase onto `main` was built in a throwaway worktree first and then applied to the branch, and both produced the identical tree. Of the twelve files this branch overlaps with slice 7, eleven are byte-identical to what it had already merged, so only `RELEASE_NOTES.md` conflicted. The stale "file's own `export {}` closure" phrasing that slice 6 corrected is absent here and the corrected "union of every listed file" phrasing is present, checked with a positive control rather than a bare grep.

### DIFF
--- a/packages/sdk/src/components/Dashboard/Dashboard.tsx
+++ b/packages/sdk/src/components/Dashboard/Dashboard.tsx
@@ -194,8 +194,21 @@ export function Dashboard({
          // milliseconds and a filter value went unescaped. Set under the name
          // the MODEL declares, so the value reaches the URL and the request
          // under the one name the server knows.
-         const value = encodeDrillValue(rawValue, declaredTypes.get(declared));
-         if (value === undefined) return;
+         const declaredType = declaredTypes.get(declared);
+         const value = encodeDrillValue(rawValue, declaredType);
+         if (value === undefined) {
+            // Say so, the way the notebook does. `canSelf` only checks that the
+            // NAME resolves, so a `given=` naming a type the clicked value
+            // cannot become still paints the whole column as clickable and then
+            // drops every click. That is an authoring mistake with no other
+            // symptom, and this is the likelier surface for it, because the
+            // givens are the dashboard's own.
+            console.warn(
+               `Drill declined: ${JSON.stringify(rawValue)} cannot be a value for given "${declared}"` +
+                  (declaredType ? ` of type ${declaredType}` : ""),
+            );
+            return;
+         }
          setGiven(declared, value);
       },
       [declaredTypes, resolveGiven, setGiven],
--- a/packages/sdk/src/components/Dashboard/Dashboard.tsx
+++ b/packages/sdk/src/components/Dashboard/Dashboard.tsx
@@ -1,0 +1,272 @@
+import { Alert, Box, Stack, Typography } from "@mui/material";
+import { useCallback, useMemo } from "react";
+import type { DashboardManifest } from "../../client";
+import { useGivensState } from "../../hooks/useGivensState";
+import { useQueryWithApiError } from "../../hooks/useQueryWithApiError";
+import { useSuggestOptions } from "../../hooks/useSuggestOptions";
+import { parseResourceUri } from "../../utils/formatting";
+import { ApiErrorDisplay } from "../ApiErrorDisplay";
+import { useDrill, type DrillNavigation } from "../drill";
+import { GivensPanel } from "../given";
+import { Loading } from "../Loading";
+import { useServer } from "../ServerProvider";
+import { DashboardTile } from "./DashboardTile";
+
+export interface DashboardProps {
+   /** `publisher://environments/{env}/packages/{pkg}` — env and package only. */
+   resourceUri: string;
+   /** The dashboard's slug, as listed by the dashboards endpoint. */
+   dashboard: string;
+   /**
+    * Control values from the host, typically its URL query parameters. These
+    * beat the dashboard's own starting values, so a shared link shows what the
+    * sender was looking at.
+    */
+   givens?: Record<string, string>;
+   /**
+    * Applied control values, for a host that wants them in its URL. Fires with
+    * what the results reflect, not with every keystroke.
+    */
+   onGivensChange?: (givens: Record<string, string>) => void;
+   /**
+    * Where to go when a `# drill` cell is clicked. Without it, drilling to
+    * another dashboard is inert — `to=self` still filters in place, since that
+    * never leaves the component.
+    */
+   onNavigate?: (target: DrillNavigation, event?: MouseEvent) => void;
+   /**
+    * Height cap for a result panel. Left unset, each form gets the cap that
+    * suits its shape — see {@link TILE_HEIGHT} and {@link WHOLE_PAGE_HEIGHT}.
+    * Set it to hold a dashboard to a fixed box, as an embedding host might.
+    */
+   height?: number;
+   maxResultSize?: number;
+}
+
+/**
+ * Per-tile cap for the composite form. A tile is one panel among several, so
+ * capping them keeps the grid even instead of letting one long table set the
+ * height of its whole row.
+ */
+const TILE_HEIGHT = 400;
+
+/**
+ * Cap for the single-query form, where the one result *is* the dashboard and a
+ * cap would clip the page rather than tidy it. High enough to be no cap in
+ * practice, and still a guard against a pathological result; the result renders
+ * at its natural height and the page scrolls, which is what a reader expects of
+ * a dashboard.
+ */
+const WHOLE_PAGE_HEIGHT = 20000;
+
+/**
+ * A Malloyyo-style dashboard: a control row over one or more query results,
+ * declared entirely by tags in a package's `dashboards/*.malloy`.
+ *
+ * Host-agnostic on purpose. It takes props rather than reading a router, and
+ * hands navigation and URL state back to whoever mounted it, so the Publisher
+ * Console and an external React app render the same component and differ only
+ * in what they do with `onNavigate` and `onGivensChange`.
+ */
+export function Dashboard({
+   resourceUri,
+   dashboard,
+   givens,
+   onGivensChange,
+   onNavigate,
+   height,
+   maxResultSize,
+}: DashboardProps) {
+   const { environmentName, packageName } = parseResourceUri(resourceUri);
+   const { apiClients } = useServer();
+
+   if (!environmentName || !packageName) {
+      throw new Error(
+         "A Dashboard resource URI must name an environment and a package.",
+      );
+   }
+
+   const {
+      data: manifestResponse,
+      isSuccess,
+      isError,
+      error,
+   } = useQueryWithApiError({
+      queryKey: ["dashboard", environmentName, packageName, dashboard],
+      queryFn: () =>
+         apiClients.dashboards.getDashboard(
+            environmentName,
+            packageName,
+            dashboard,
+         ),
+   });
+   const manifest = manifestResponse?.data;
+
+   const specs = useMemo(() => manifest?.givens ?? [], [manifest]);
+   const declaredTypes = useMemo(
+      () =>
+         new Map(
+            specs
+               .filter((spec) => spec.name !== undefined)
+               .map((spec) => [spec.name as string, spec.type]),
+         ),
+      [specs],
+   );
+
+   const { draft, applied, setGiven, reset, apply, pending } = useGivensState({
+      declaredTypes,
+      startingValues: manifest?.startingGivens,
+      params: givens,
+      onParamsChange: onGivensChange,
+      // Which document these edits belong to. Without it the edits are keyed by
+      // their starting VALUES alone, so two dashboards whose starting values
+      // coincide (the common case: both empty) look like one document, and the
+      // one you came from keeps filtering the one you drilled into.
+      documentKey: `${environmentName}/${packageName}/${dashboard}`,
+      // Absent means autorun; only an explicit `autorun=false` batches.
+      autorun: manifest?.autorun !== false,
+   });
+
+   const {
+      options,
+      isLoading: optionsLoading,
+      failed: optionsFailed,
+   } = useSuggestOptions(environmentName, packageName, manifest?.path, specs);
+
+   // `to=self` filters in place, which only works for a given this dashboard
+   // actually surfaces: sending one it cannot bind would fail every tile's
+   // query. Asked here rather than checked after the click, so a cell that
+   // cannot be honoured is never painted as clickable in the first place.
+   const canSelf = useCallback(
+      (given: string) => declaredTypes.has(given),
+      [declaredTypes],
+   );
+   const { drill, drillMenu } = useDrill({
+      onNavigate,
+      onSelf: setGiven,
+      canSelf,
+      selfLabel: "Filter this dashboard",
+   });
+
+   if (isError) {
+      return (
+         <ApiErrorDisplay
+            context={`${environmentName} > ${packageName} > ${dashboard}`}
+            error={error}
+         />
+      );
+   }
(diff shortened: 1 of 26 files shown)


packages/sdk/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,395 @@
...
+         // First declaration wins, so a model with `REGION` and `region` keeps
+         // the one it declared first rather than silently flipping.
+         if (!byFold.has(name.toLowerCase()))
+            byFold.set(name.toLowerCase(), name);
+      }
+      return byFold;
+   }, [declaredTypes]);
+
+   /** The declared given a drill tag's name refers to, or undefined. */
+   const resolveGiven = useCallback(
```
Sha-Bang, line 198: this folds case so "one tag behaves identically on both surfaces", and a `to=<slug>` drill does not — `paramsToGivens` matches exactly, so the given reaches the destination as the tag spells it and is dropped if the case differs. the release note documents the authoring rule, so the behaviour is known; what I'd push on is that the cell is still painted as a link.

measured on the dashboards fixture with both drills present in one load:

```
# drill { to=overview given=brand }              overview declares BRAND  →  0 warnings
# drill { to=["regions","self"] given=NO_SUCH_GIVEN }                     →  2 errors
```

so the self half is caught precisely and the cross-dashboard half is silent: the cell is marked, the click navigates, the destination opens unfiltered, nothing said anywhere. that is the case `markDrillableCells` exists to refuse ("a dead link is worse than none"), one hop removed — the link isn't dead, it just lands somewhere that ignored the filter.

`canDrill` can't see the destination's manifest, but the server can: it holds every one at load, and `lintDrillTargets` already resolves the slug. extending it to check the destination can bind the given closes it at the only place that sees both sides, and it's the same argument as `lintSelfDrills`, which you already wrote. (that's in 987's `dashboard.ts`, so it's a follow-up rather than something to wedge in here.)

packages/sdk/src/components/Dashboard/Dashboard.tsx
```diff
@@ -0,0 +1,395 @@
...
+ * at that height and the page scrolls, which is what a reader expects. A result
+ * that sizes to its CONTAINER instead, which a bare `# bar_chart` does, has no
+ * height to report and keeps whatever first-paint height it was handed, so it
+ * stretches: measured 1992px for a two-row bar chart against 227px for the same
+ * query under a grid tag. `INITIAL_RENDER_HEIGHT` bounds that near 2000 rather
+ * than removing it, and lowering the bound is NOT the fix, because the same seed
+ * sizes a notebook's chart cells (measured 700px there). Telling the two kinds of
+ * result apart needs something the renderer does not expose here.
+ */
+const WHOLE_PAGE_HEIGHT = 20000;
```
Sha-Bang, line 78: reproduced your numbers independently, on a fixture carrying `# artifact` + `# bar_chart` with no `# dashboard {columns=N}`:

```
overview  (# dashboard {columns=6})   .malloy-render   227px
onechart  (bare # bar_chart)          .malloy-render  1992px   (nearest scroller 2226px)
```

so the docblock is accurate. raising it anyway because no fixture covers that shape — `grid` and `overview` both carry the grid tag, `regions` is a plain table — and "a dashboard that is one chart" is a natural thing for an author to write, which makes this the likeliest first impression of the new surface rather than an edge case.

`manifest.dashboardColumns === undefined` is a discriminator you already have on the manifest, at the cost of putting a cap and an expand toggle on a plain-table dashboard like `regions`. a viewport-relative initial height rather than a fixed 2000 is the other way, and doesn't need the manifest at all. either way a fixture for the bare-chart shape would keep it from regressing quietly.

packages/sdk/src/components/RenderedResult/RenderedResult.tsx
```diff
@@ -561,14 +706,36 @@ function RenderedResultInner({
...
+            if (drillKeyup) stage.removeEventListener("keyup", drillKeyup);
+            if (drillFocusIn)
+               stage.removeEventListener("focusin", drillFocusIn);
+            // The re-marker is scoped the same way, and for the same reason.
+            // Disconnecting it unconditionally tied it to the effect RUN: a
+            // re-run while a `# dashboard` was still building its cards stopped
+            // the LIVE stage being re-marked, so cards arriving after that point
+            // never became drillable. Like the listeners, it is collected with
+            // the node once a promoted stage is dropped.
+            drillObserver?.disconnect();
```
Sha-Bang, line 737: right for supersede, and it leaves the unmount case uncovered. on unmount the promoted stage satisfies `liveRef.current?.node === stage`, so this whole branch is skipped and neither the disconnect nor the `cancelAnimationFrame` runs; the unmount effect at `:343` only does `viz.remove()` and nulls the ref. a queued `mark` then fires against a detached stage, and the observer plus the stage and `names` it closes over waits for GC instead of being torn down.

low impact — a detached subtree doesn't mutate, so the observer never fires again — so this is housekeeping rather than a leak with a symptom.

worth being explicit about the fix that looks obvious and isn't: **don't** disconnect unconditionally in this cleanup. that's exactly `origin/main:564-565`, and your comment is right about why you moved it — it ties the observer to the effect run, so a re-run stops the live stage being re-marked and cards arriving after that never become drillable. I checked that against `main` rather than assuming it. the fix that keeps both properties is to carry the observer and the frame id on `liveRef.current` beside `viz` and `node`, and tear them down in the unmount effect that already nulls it.

packages/sdk/src/components/Package/Package.tsx
```diff
@@ -143,7 +197,16 @@ export default function Package({
...
 
-   const isLoading = !notebooksQuery.isSuccess && !notebooksQuery.isError;
+   // The dashboards list is part of the gate, not just the notebooks one,
+   // because the models list is FILTERED by it. Gated on notebooks alone, a
+   // dashboards call that resolves after the models call rendered the sections
+   // with an empty `dashboardPaths`, so `dashboards/overview.malloy` appeared
+   // under Semantic Models and then vanished: the double listing the filter
+   // exists to prevent, briefly on screen. It cannot hang the page: the query
+   // above turns a 404 or a transport failure into an empty list.
+   const isLoading =
```
Sha-Bang, line 207: the justification covers failure but not latency. "it cannot hang the page" is true — the `catch` above turns a 404 or a transport failure into an empty list — but a *slow* `listDashboards` now holds Notebooks, Semantic Models, Data Apps and Databases behind the spinner, where before they rendered as soon as notebooks resolved.

minor in practice, and I checked rather than assumed: both handlers are in-memory reads off the already-loaded package (`listDashboards` walks the precomputed map, `listNotebooks` reads cells), so this is one more parallel request rather than a compile. the double-listing flicker you're fixing is the worse of the two.

only noting it because the thing that needs the gate is narrower than the gate: it's the `dashboardPaths` filter on the models list. gating just that section — or rendering the others immediately and the models list once both resolve — keeps the fix without coupling every section to the new call.

Sha-Bang (approved):
reviewed at 545c098b, plus a second independent pass over the same diff. the correctness core holds up: I could not break the one-stop-per-column invariant, the givens merge in `DashboardPage`, the drill value encoding, or the manifest and loading gates. I also mutation-verified the head commit's own claim — deleting the `ownedByTable` line in the `hasStop` prescan fails exactly the nested-table test and leaves the other twelve passing, so that guard is load-bearing as it says.

**the Playwright gate never ran, and that's the one thing I'd sort before merging.** it reads `fail`, but no test executed: the job hung 27 minutes in `apt` inside `playwright install --with-deps` against a flaky Azure mirror, got cancelled, and uploaded no report at all ("No files were found with the provided path"). that matters more than an ordinary flake here, because the last green Playwright run on this branch was twelve commits ago at `1635d4c9`, and those twelve include `9949912e` (the query-string merge), `1f46c4a5` and `b8b00ed8` (the two-stops-after-a-click fix and the focusin handler), `6bd65141` and `545c098b` — the tab-stop and focus behaviour this suite is the only coverage for — plus two tests that have never run in CI at all. it's also the second consecutive cancelled run.

so I built the branch and ran it against a real server: **28/28 passed** in 26.8s. alongside that, sdk unit 335 pass / 0 fail, `tsc --noEmit` clean once the axios client is generated, every doc cross-reference resolves, and "no bundled example ships a `dashboards/` directory" is true. the code is fine — the gate needs a re-run, not a fix.

approving on that basis. four inline notes, none blocking: two are behaviour I measured (the cross-dashboard drill given, and the bare-chart height), two are narrower than the comment defending them (the unmount teardown, the loading gate).

one more that spans too many files to hang inline. the drill contract is now duplicated across the two surfaces: `givenNamesByFold`/`resolveGiven`/`canSelf`/`onSelf` including the identical `Drill declined:` warning at `Dashboard.tsx:186-236` and `Notebook.tsx:242-298`, and a byte-identical route builder at `DashboardPage.tsx:114` and `NotebookPage.tsx:144`. both copies are correct today, so this is only worth raising because not drifting between surfaces is this PR's own argument — a shared `useDrillSelf` plus one exported `dashboardHref` would make that structural rather than maintained by hand, and it's also where the cross-dashboard given check would naturally live.

two things I checked that are fine, in case they come up: `encodeDrillValue` falling through to filter-escaping when `Given.type` is absent is identical to the shipped notebook path and the server always populates it; and the `spa_fallback` `dashboards` entry is correct, with its `public/dashboards/` cost stated in both the docblock and the release note.

last, unrelated to the code: the description says 24 files and +2266/-64, against 25 and +3172/-69 at this head.

mlennie:
Thank you, and specifically for three things that changed what I know rather than confirming it: mutation-verifying the `ownedByTable` guard instead of taking the commit message's word for it, reading the CI log when the gate failed, and then building the branch and running the suite yourself rather than leaving the gate ambiguous.

On the gate: it was `apt` inside `playwright install --with-deps` hanging against a flaky mirror, before any test ran, which is why no report was uploaded. Four of us theorised about suite size, hangs, and timeout arithmetic; all four theories assumed the failure was in the tests. The missing report artifact was the tell and none of us weighted it. A re-run has since gone green in 5m on this head, so the twelve-commit gap you flagged is closed by a real check rather than standing on your manual run, though your run is what made it safe to keep going in the meantime.

Dispositions on the five notes.

**Loading gate: follow-up, and you are right about the substance.** The justification covered failure and was silent on latency, and this branch has spent days auditing exactly that class in its own prose, so I am not going to defend it. I verified your basis rather than taking it, because "one more parallel request" and "one more compile" are very different costs and the old comment left a reader unable to tell which: `listDashboards` is synchronous on the loaded package, reached with `getPackage(name, false)`, so it cannot trigger a reload. The corrected wording is written and goes in with the follow-up batch rather than into this merge, since I am landing this PR as reviewed to unblock the three stacked behind it rather than adding a commit after your approval. It states what it covers, what it costs, and why that is accepted rather than narrowed to the `dashboardPaths` filter.

**Cross-dashboard given silently dropped: follow-up, filed against `dashboard.ts` where you put it.** I checked whether the warning half could live on this surface and it cannot: the source dashboard holds only its own manifest, so it has no way to know `given=brand` will fail against a destination declaring `BRAND`. `givenNamesByFold` folds against the local declarations, which is precisely why `to=self` is caught and this is not. Only something holding both manifests can decide it, and `lintDrillTargets` already resolves the slug. Your framing is the part I want on the record: this is the case `markDrillableCells` exists to refuse, one hop removed, where the link is not dead but lands somewhere that ignored the filter.

**Bare-chart height and its missing fixture: follow-up, both halves as one issue.** Your numbers match mine and I agree the shape is a natural authoring choice rather than an edge case. I am not adding the fixture ahead of the height decision, because a fixture written today pins 1992px as the expected value and 1992px is the residual, not the target; the docblock calls it bounded rather than cured. A test that asserts the thing we intend to change is worse than none, and its value arrives the moment someone changes the height and wants to know what moved. Both of your options go in the issue: `manifest.dashboardColumns === undefined` as a discriminator already on the manifest, and the viewport-relative initial height that needs no manifest at all.

**Unmount teardown: follow-up.** Your analysis is right, and your warning about the obvious fix is the more valuable half, since disconnecting unconditionally is exactly `origin/main:564-565` and the thing that comment exists to explain. I am declining on impact rather than on size: carrying the observer and frame id on `liveRef.current` is genuinely a small change, but after unmount the ref is nulled and the node detached, so the observer is unreachable and collected with it, and a detached subtree raises no mutations for the pending frame to service. Re-editing lifecycle code that took three review rounds to settle, post-approval, for a leak with no symptom, is the wrong trade. Your prescribed fix is specific enough to file verbatim.

**Duplicated drill contract across the two surfaces: follow-up, filed rather than remembered.** This is the strongest of the five and your reason is why: not drifting between surfaces is this PR's own argument, so hand-maintained duplication contradicts the thing the PR claims. `useDrillSelf` plus one exported `dashboardHref` is the right shape, and the cross-dashboard given check lands naturally inside it. Refactoring two surfaces into a shared hook on an approved branch is the wrong moment, and doing it badly would cost exactly the parity being argued for.

On the description: you are right, and it was 24 files and +2266/-64 against 25 and +3172/-69. I am dropping file counts from PR bodies rather than maintaining them, since they drift on every commit and read as authoritative once stale.

PR #1045, opened 2026-08-19
feat(server): one gate concept, source-line natural-Malloy #(authorize), and colocated persist for gated sources

Sha-Bang:
# One gate concept, natural-Malloy authoring, and colocated persist for gated sources

Three things that turned out to be one change. `#(authorize)` moves from an expression string to an
unquoted Malloy expression; the gate model collapses to a single row-predicate concept; and the
colocated `#@ persist` refusal relaxes for row-level-gated sources, the one tier where a gate and a
frozen artifact already compose correctly.

## `#(authorize)` carries an unquoted Malloy expression (BREAKING)

```malloy
given:
  GROUPS :: string[]

#(authorize) cost_center in $GROUPS
source: costs is conn.table('costs') extend {}
```

The governing principle: **any legal Malloy boolean expression is a legal gate.** The author can paste
the same expression into a `where:` to see what it does — which the string form could never offer.

The mechanism is `buildRowLevelProbe`, which compiles the annotation text as `extend { where: <text> }`
in the target source's **real field scope**. That is what deletes the ~350-line allowlist grammar
(`classifyAuthorizeGate`), a ~230-line static-literal evaluator, and the synthetic column-less probe
the old form compiled against. Malloy's type checker does the work, and its errors are better than the
ones the allowlist produced: `owner in $ROLE` against a scalar-typed given now reports `` `in $ROLE`
requires `ROLE` to be an array , but it is `string` `` where the string form said
`'owner' is not defined`. An undeclared given reports `` `$NOSUCHGIVEN` references a given named
`NOSUCHGIVEN`, which is not declared in this model `` — so the reachability check deletes outright
rather than being reimplemented.

What is left is **G1** (the payload compiles as a scalar boolean), at most one annotation per source,
and **G4** (no referenced given carries a declared default). Two signals warn and still load: **W1**
the gate references no given, **W2** negated membership. A gate may also be a bare reference to an
ordinary boolean dimension (`#(authorize) authorized`), which is how you write a gate too long for one
line; validation follows the reference through, which matters because Malloy does not propagate given
usage through a field reference — read raw, a defaulted given one hop away sails through G4.

**Retired forms are refused at load, with the rewrite written out.** The quoted string form on the
`source:` line, file-level `##(authorize)`, and an annotation anywhere other than directly above a
`source:` line. Clean break, not deprecation — one breaking change, one migration, docs written once.
The refusal reaches every `.malloy` in the package; a declaring file *outside* the package tree loads
and then denies every request instead, counted under `legacy_string_gate` with no author-facing detail.
Nothing unwraps a quoted payload any more: it compiles as a string literal, fails Malloy's boolean
check, and denies by construction, so the classifier cannot drift from the loader.

**G4 removes the admin-override-via-default idiom.** Callers must supply `$ROLE` explicitly. One rule
bought for one extra key in the givens map, and it earns its place because a default is invisible in
the gate the author is reading — it lives at the given's declaration and changes the meaning from
elsewhere.

## A derivation can no longer silently shed a gate

This is the security improvement that made the shape worth choosing, and it is a reversal.

An intermediate revision of this PR put the gate on an annotated boolean dimension inside the source's
body. Under that form a model-authored derivation could drop the dimension —
`extend { except: authorized }`, or an `accept:` that did not re-list it — and produce a source with
**no gate at all**, serving **every row to every caller**, with no load error and no warning. Malloy's
compiled IR keeps no link from an `extend`-derived struct back to its base, so nothing could refuse the
load over it. That was pinned as `KNOWN GAP` and had no fix in scope.

A source-line expression has no field to shed. A derivation that drops a column the gate **reads**
leaves the graft unable to compile, so that entry point is **denied** rather than served ungated, and a
load-time warning names the entry point whose gate is no longer expressible. The warning is scoped to
that one entry point rather than aborting the whole load — siblings and the declaring source are
unaffected — which needed location-based note ownership (`source_extraction.ts`'s
`authorizeNoteDeclaredBy`), because Malloy copies annotation objects **by reference** onto derivations
that add none of their own, so a presence test reads a copied note as "declared here".

A separate misbinding case is pinned by a test here, though it predates this branch rather than
arriving with it: drop the gated column, then `rename:` a *different* column onto that exact name.
That grafts and binds the gate to the **wrong** column. It takes a drop, plus a rename onto the exact
gated name, plus colliding data. The author rule is not to recycle a gated column's name.

## Every gate is a row predicate

The `given_only` variant is gone, with its admit/deny probe and its routing escape. Classification is
`row_level` or `rejected`.

**This changes an observable response class.** The probe is what produced 403 on a gate deny; a gate
that denies now returns **200 with zero rows and the correct result schema**. Package-level FGA denials
(`can_read_package`) still 403 — only the gate's own verdict collapses. A 403 from a gate now means
only that the gate could not be attached, or that a given it names was not supplied. Roughly 45
integration assertions moved, each read individually rather than blanket-replaced.

## Colocated `#@ persist` builds for row-level-gated sources

Safe by construction, not by convention. The build compiles `sd.query` only and never evaluates the
gate, so the artifact is gate-independent. At serve time the graft lands on the outer `filterList` and
applies as the reading query's own `WHERE`, while substitution is FROM-only — two different fields,
which is why the artifact stays addressable under a grafted gate *and* still filters. Colocated serving
never sets `serveVirtualMap`, so a given-referencing gate still binds per request.

**The residual cost is freshness, not authorization**: a row the gate would now exclude stays frozen
until rebuild, so a row that changes hands keeps serving to its former owner. The gate expression and
the principal's attributes stay live.

**This enables on upgrade, which is worth stating plainly.** A package declaring colocated `#@ persist`
on a gated source is not refused at publish today — it loads and serves live, and only its materialize
run 422s. Once this ships, the next scheduled build materializes and binds one with no author action,
so the freshness contract starts applying to packages nobody re-published.
`PERSIST_COLOCATED_RELAXATION_ENABLED` (default: relaxed) is the per-deployment lever.

Relaxation requires the gate to classify `row_level` **and** be *attributed* — every `#(authorize)` note
the eligibility deep-walk finds must belong to the source's own entry-point gate set. The walk sees
joined sources' gates, but enforcement is entry-point only and never traces joins, so a joined child's
gate can classify cleanly and never graft. `storage=` still refuses every gated and given-referencing
source; pre-aggregate refusal stays unconditional, since rollups group away the gate column by
construction.

## Refusals are visible to hosts

`BuildPlan.refusedSources` carries source coordinate, tier, reason and message. It is a separate
collection rather than a field on `PersistSourcePlan` because constructing one calls
`getSQL()`/`computeSourceEntityId()`, which free-parameter and given-referencing sources fail. A gated
pre-aggregate rollup appears in **both** `sources` and `refusedSources`: synthesis is visible, and it
will never materialize.

This also fixed an ordering bug where the orchestrated loop checked eligibility before checking whether
a source received an instruction, so one refused uninstructed sibling aborted an entire build.

`sources[].authorize` gets simpler in the same move: the value is always the author's expression, so the
old graft-text case (the gate's `code` for an own gate, but the quoted identifier `` `authorized` `` for
one carried in from a query-source base or composite member) disappears. The caution the field carries
does not change: the flat list is not a predicate to re-evaluate, since it flattens every gate carried
in and expressions from different declaring sources are AND-ed, not OR-ed.

## Security

**A caller could launder the gate away.** Anyone who can submit query text could declare a derivation
whose entry point is not a model-declared source and read every tenant's rows with no givens at all.
The fix denies unless it can *prove* the request-declared chain safe: when the compiled entry point is
not a model-declared source and the model declares any gate, deny unless every terminal in the chain is
a model-declared **ungated** source. An unreadable declaration fails the proof rather than being waved
through, so a future gap between the declaration scan and Malloy's grammar costs an over-denial instead
of a leak. Nineteen laundering and evasion spellings are pinned, including quoting, comment and
statement-order tricks and a forged declaration inside a string literal.

The control's predicate is form-agnostic and carries over unchanged. The specific spelling that
motivated it — dropping the gate *field* — no longer works, so the control may be narrowable; that is
deliberately a separate change with its own review rather than folded in here, because a control that
no longer defends anything should be removed on purpose, not as a side effect.

**A gate-stripping bypass at `/compile`.** The file/append-scope backstop discovered a run target's gate
by walking the caller's own compiled struct, so stripping `#(authorize)` from submitted text evaded
enforcement. Enforcement now resolves from the on-disk model; caller text may narrow, never widen.

**A classification cache that could pair with a mismatched given surface.**
`createGateClassificationDeps` is now the one constructor, minting cache and given maps together so
they cannot disagree.

**`liftProbeFilterCondition` keeps both of its assertions** (`code === filterText`,
`isSourceFilter === true`). A missing check there is how a query runs ungated, so they are load-bearing
rather than defensive.

## Known limitations, stated rather than papered over

**The rename-misbinding hole** described above: drop the gated column, then `rename:` a different column
onto that exact name.

**`/compile` admits any givenless gate, true or false**, because it executes nothing and decides on
given *presence*, not gate truth. With `includeSql` it then returns the ungrafted SQL. That matters for
the documented locked-base pattern (`#(authorize) false`), and the docs now say so at the point of
recommendation.

**A scalar-vs-array given mismatch is not caught at compile time.** `org_id = $GROUPS` loads and grafts,
then fails at execution with a warehouse conversion error. Safe, but the author sees a warehouse error
rather than a diagnostic on their line. Use `in` for an array-typed given. Note `org_id = $GROUPS` and
`org_id ? $GROUPS` are the same comparison node, so neither spelling escapes it.

**W1/W2 are advice, not enforcement.** An author can still write a gate that admits everyone, exactly as
they can write `where: true`. W2 in particular is a best-effort shape match: other spellings of the same
inversion do not trigger it, so its absence is no evidence either way. The floor that holds regardless:
graft-or-deny, G4 plus Malloy's own bind failure for an unsupplied given, and `in` over an empty array
compiling to `WHERE FALSE`.

## How it was tested

`tsc --noEmit` clean; lint clean; unit **3158 pass / 3 skip / 0 fail**; integration **309 pass / 1 fail**
(`concurrent_package`, a local-only flake).

The acceptance evidence for "natural Malloy" is `authorize_syntax_conformance.spec.ts` plus
`source_line_authorize_integration.spec.ts`, which run gated models through the real `Model.create` and
`getQueryResults` against real DuckDB with two principals — never hand-built IR — so a passing test
covers the whole compile → classify → graft → enforce pipeline. Six spellings the string form refused —
`upper(region) = $REGION`, `region like $PAT`, `region is not null`, `amount + 1 > $AMOUNTMIN`,
`not (org_id in $GROUPS)`, `1 = 1` — are accepted with asserted row sets.

The fail-closed reversal is pinned directly: `except:`-ing and `accept:`-dropping the gated column each
deny, the warning names the right entry point, siblings and the declaring source are unaffected, and a
genuinely broken gate at the *declaring* source still aborts the load as it always has. G4's one-hop
case (`#(authorize) authorized` over `dimension: authorized is amount > $FLOOR` with a defaulted
`FLOOR`) is pinned separately, since that is the case a raw `givenUsage` read leaks.

**Migrating the fixture corpus surfaced two classes of problem worth naming, because a green suite hid
both.** Ten `materialization_eligibility` fixtures declared `given: role :: string is 'analyst'` and
gated on `$role` — models G4 refuses, reaching the compiler only through a helper that skips load
validation. They asserted the right thing about an input no author can produce; the defaults are gone.
And `authorize_integration.spec.ts`'s `CP_JOIN` fixture was **vacuous**: it went through the real
`Model.create`, so the retired-string refusal fired on every run and stashed a `ModelCompilationError`,
but the test never checked `getNotebookError()` and `assertAuthorizedForRunnable` resolves trivially
against a model that never compiled. `resolves.toBeUndefined()` was passing because nothing compiled,
not because a joined source is ungated on the `/compile` path. It now compiles and exercises what its
name claims. That is the fourth vacuous test found on this feature, which is why every fixture decision
is recorded per-site rather than swept.

Enforcement is proven end-to-end against real compiled packages and real DuckDB: direct and
`extend`-inherited gates, always-false, givens supplied and absent, composite entry points, chained
persist over gated persist, and two principals getting disjoint row sets from the same artifact. The
laundering matrix covers 26 shapes — every bypass denies, a direct gated query with correct givens still
returns only its own rows, and five over-deny controls (including non-ASCII and two-hop *ungated*
aliases) still serve.

The shipped `examples/governed-analytics` package and the app's Playwright fixture model are both
covered by `shipped_example_authorize.spec.ts`, added because neither is reachable from the
`packages/server` suites and both authored a form that would no longer load: the example gates
correctly per principal (admin 3 tenants, `acme` 1, an unlisted tenant zero rows), and the Playwright
fixture's unsupplied-given path still denies opaquely with `Access denied for source
"gated_products"`, naming the source and not the expression.

Failing-first evidence accompanies each behavior change: reverting the relaxation reddens the compose
tests, reverting the bind filter reddens the serve-bind refusal tests, and each security fix's tests
fail before it and pass after.

## What two xhigh review passes changed

Three findings were real defects, not polish.

**A G4 bypass at two or more import hops.** `validateAuthorizeProbes` ran G4/W1/W2 only for entry
points whose gate note it could *attribute*, and attribution goes empty past one hop — so a gate
declared two imports away, referencing a given with a declared default, loaded clean and served rows
off that default. The probe and the graft resolve the same given id at every depth, so this was a reach
bug rather than an identity mismatch: the checks now run unconditionally on every entry point whose
probe compiles. Verified at two hops (refused), two hops with no default (loads, and three principals
get 2 / 2 / 0 rows — no over-refusal), and the one-field-hop shape.

**The refusal's "paste-ready rewrite" was byte-identical to what it refused.** The payload arrives
verbatim including its quotes, so `#(authorize) "org_id = 999"` was refused with `#(authorize)
"org_id = 999"` as the remedy. The rewrite now decodes the string literal. The test that should have
caught it asserted a substring both the correct and the broken message contain; it asserts the emitted
rewrite line now.

**W1 and W2 had no producing test.** Deleting both `onWarning` calls left the whole suite green, in the
very commit that added them — the only coverage called the metric recorder by hand. Three tests now
drive them through the real `Model.create` (W1, W2, and an ordinary gate that must warn neither), with
failing-first evidence.

The rest were claims that were true of the intermediate dimension form and outlived it: which rejection
causes fail the load (exactly one of seven does — three warn, three are request-time only), a shipped
conformance-report paragraph asserting `1 = 1` loads with no warning, a test titled for a fixture that
had already been migrated, and `Source.authorize`'s "the author's expression as written" where the
registry can substitute a fail-closed `"false"`.

Two gaps are parked deliberately. `unclassifiable_condition` and `given_usage_unresolvable` are
defensive branches behind load-time validation; reaching them needs fabricated IR, so they stay
documented rather than tested. And the request-time legacy-lift failure emits no `cause` on purpose —
that path is per-request, and a cause there would break the metric's "alert on nonzero since the last
publish" shape.

## CI state

DCO is green (every commit signed off) and main is merged in, which is what let the full
matrix run for the first time — the branch was previously conflicting, so only CodeQL and
DCO ever fired. 16 checks pass: Typecheck, Lint, build, npx/bun-direct test, Cross-Platform
on all three OSes, Docker smoke, Connection Integration, tarball verify, and
Playwright.

One was red and is now fixed; one remains an accepted false positive.

### Playwright — was red, now fixed (three commits at the head of the branch)

`notebook-authorize.spec.ts` was failing 3 of 108, and it was a genuine gap in this change
rather than a flake. It is fixed. The investigation also turned up a second broken surface
and a latent fail-open, so this is worth reading rather than skimming.

**Root cause, one invariant.** A gate lands only when the recompile RE-DERIVES the run
target from `contents[graftTarget]`. The trigger is not `import` as such but
`!modelEntry.exported`: `import-statement.js` marks every imported entry unexported, and
`named-source.js` INLINES an unexported entry's struct at the reference site instead of
emitting a name. The graft then lands on a `contents` entry nothing consults, the proof
correctly reports the condition missing, and an **authorized** caller is denied with the
byte-identical 403 a real denial produces.

**Two surfaces hit it, not one.** The notebook case is the regression. The other is the
documented single-query dashboard shape (`docs/dashboards.md`) — `import` plus `# artifact`
on a model-level `query:` — which was already broken before this branch, because a stored
`NamedQueryDef` keeps its own pre-graft `structRef` that `query-reference.js` reuses
verbatim. Both are fixed; they needed different fixes because they capture the stale struct
in different places.

**The notebook case cannot be fixed by recompiling text**, which was the obvious first idea
and is measurably wrong: the grafted clone is loaded from a ModelDef and so has no base URL,
so a relative `import` in that text fails to resolve at all — precisely what an
import-and-run cell contains. It grafts the cell's own compiled queryDef instead, deep-cloned
per request because `cell.runnable` is memoized and grafting in place would let whoever
queried first decide what every later caller sees.

**On that branch the usual IR proof would be vacuous** — it would read back the array the
binder just wrote. The load-bearing check there is object identity: the struct the
materializer will execute must BE the one we grafted, which turns a future
`_loadQueryFromQueryDef` that copies or re-derives its argument into a denial. Neither branch
proves Malloy still HONORS `filterList` at SQL generation; no IR check can. That is pinned
behaviourally instead — two rows not four, and `org_id` asserted in the generated SQL.

**A latent fail-open, fixed independently and first.** `assertGateLanded` read only
`query.structRef`, but Malloy generates SQL from `compositeResolvedSourceDef ?? structRef`.
With a composite run target the proof could pass on one struct while a stale other one
generated the SQL — unfiltered rows behind a green proof. Now mirrors the compiler's
precedence, deliberately mirroring rather than unioning, since accepting "either struct
carries the filter" reintroduces the same leak from the other side.

**The e2e assertions themselves were part of why this shipped.** The denial test asserted
403, that the message names the source, and that it omits the gate expression — all three
satisfied identically by an attach failure, so it passed whether the feature worked or was
dead. The success test asserted only `status === 200`, which a zero-row response passes. Both
replaced with the three outcomes this feature actually has: unsupplied given is a **400**
naming the given, non-matching given is a **200 with zero rows** (a row-level gate filters,
it does not refuse), matching given is a **200 with the one gated row**, asserted by content.

Verified discriminating by isolated revert: dropping only the snapshot sweep fails only the
dashboard test; dropping only the queryDef graft fails only the three notebook tests — in both
cases by DENYING, never by silently admitting. Suites: 3174 pass / 3 skip / 0 fail (baseline
3163, +11 new); integration 309 pass / 1 fail, the known `concurrent_package` network flake,
matching baseline exactly.

**No previously-working shape changes value.** An earlier revision of this description
claimed a gated `query_source` would move from all-rows to caller-rows; that was wrong and is
withdrawn. `graftIntoNamedQuerySnapshots` only touches a stored query whose `structRef` is an
INLINED struct, which happens exactly when the source arrived unexported through an `import`.
A locally-declared source is exported, so its named query stores the NAME and the sweep skips
it — measured identical with and without the sweep. The imported shape did not return a
different number before either; it returned a 403.

**Still deferred, deliberately:** composite member snapshots
(`CompositeSourceDef.sources[]`). They fail CLOSED — by construction rather than by a test of
that specific shape: any run target that cannot be resolved denies at `model.ts:2400-2401`.
(An earlier revision said they were "pinned as denials" in `authorize_integration.spec.ts`;
the general principle is exercised there, the specific imported-composite-via-named-query
shape is not.) Changing them means turning denials into filtered rows, which is a product
decision rather than a regression fix.

### CodeQL — one `high`, a false positive

`js/loop-bound-injection` at `packages/server/src/service/query_text.ts:98`.
`stripMalloyCommentsAndLiterals` scans the input once and blanks each character at most once
(the comment branch advances `i` past the span), so the work is amortized linear in input
length — and that length is already bounded by `assertWithinModelByteLimit`. No amplification.
Closing it needs either this repo's first CodeQL suppression comment (a convention call for the
maintainers) or rewriting the inner loop to `charCodeAt` to drop the per-iteration `slice`. The
latter is a real improvement but it rewrites the inner loop of the pre-scan the
caller-laundering check depends on, so it should not ride a lint's say-so.

The four `js/incomplete-sanitization` alerts from the first push were real — the report
writer's escape handled `|` but not the backslash — and are fixed.

## Follow-ups

Lands with [ms2data/agent-skills#86](https://github.com/ms2data/agent-skills/pull/86), which puts the
authoring skills on this form; sequence that merge and its tag cut after this PR.

Storage-tier gate re-bind and join-axis narrowing remain out of scope. Narrowing the caller-laundering
control (above) is a candidate follow-up, as is enumerating request-declared sources with Malloy's own
parser instead of a text scan — no longer load-bearing now that an unreadable declaration fails closed.


jswir:
Reviewed at `3ba8d9c9`.

The security core holds up. I went looking for a path where a failed graft silently becomes "no gate applies", where a caller could launder the gate away through a request-declared derivation, and where the colocated persist relaxation could serve an unfiltered artifact. Each one denies correctly, and the relaxation's per-request re-evaluation is proven by your own compose tests. The `#(authorize)` grammar change is a clear improvement over the allowlist.

Two things I'd want resolved before merge: one functional regression, and one default that deserves an explicit maintainer decision rather than arriving as a default.

## Blocking: a notebook cell that reaches a gated source through `import` is always denied

A caller who satisfies the gate gets a 403, and the 403 is byte-identical to the one an unauthorized caller gets. The Playwright job is red on this head for exactly this — four failures, all of them the gate-*admit* cases.

Repro against a running server, using the fixture from `notebook-authorize.spec.ts`:

```malloy
// authz_gate.malloy
##! experimental.givens
given: role :: string

#(authorize) $role = 'analyst'
source: gated_products is duckdb.table('data/products.parquet') extend {
  primary_key: product_id
  view: spotlight is { where: category = 'Jeans'; select: product_id, name; limit: 1 }
}
```

```
>>>malloy
##! experimental.givens
import "authz_gate.malloy"
run: gated_products -> spotlight
```

```
GET .../notebooks/<nb>.malloynb/cells/0?givens={"role":"analyst"}
-> 403 {"code":403,"message":"Access denied for source \"gated_products\"."}
```

Splitting the `import` into its own cell returns 200 and the row. The trigger is the conjunction of **first code cell** (markdown cells don't establish a graft scope) and **source reached via `import`**:

| notebook shape | cell | `main` | this branch |
|---|---|---|---|
| one cell: `import` + `run` | 0 | serves | **denied** |
| markdown, then `import` + `run` | 1 | serves | **denied** |
| `import` in its own cell, then `run` | 1 | serves | serves |
| `import` cell, markdown, `run` cell | 2 | serves | serves |

I confirmed the `main` column by running the same shapes against the merge base with the string-form gate, so this is a regression rather than a pre-existing gap.

**Mechanism.** `assertGateLanded` is not the flaw — it correctly detects a real absence. The problem is upstream at `model.ts:5423-5442`. When a cell has no earlier code cell to graft against, `usesOwnScope` binds by repointing the already-compiled, pre-graft `queryDef` through `_loadQueryFromQueryDef` rather than recompiling the cell text. That's safe only when `structRef` is a name string, which gets re-resolved against the grafted model. An imported source arrives as an embedded `SourceDef` **object** — the stale pre-graft one, whose `filterList` never received the graft. The doc comment at `model.ts:760-778` argues the reuse is safe because it never recompiles the cell's text, which is true for avoiding "Cannot redefine" but assumes `structRef` is always a name.

Fails closed, so nothing leaks. But it breaks the pattern the docs recommend — model in a `.malloy`, notebook drives it — and there's no way for the caller to tell it from a genuine denial.

Worth noting `examples/governed-analytics/orders.malloynb` escapes only because its `import` happens to sit in its own cell.

## The test that should have caught this can't fail

`notebook-authorize.spec.ts:91-100` asserts three things: status is 403, the message contains `gated_products`, the message excludes `analyst`. The attach-failure error satisfies all three identically to a real gate denial, so the test passes whether the feature works or is completely dead.

`shipped_example_authorize.spec.ts:115-160` was added to cover this fixture "at the load/query tier where a failure is legible", and its comment says it asserts "the same two outcomes" — but it runs `getQueryResults` on a `.malloy` model rather than `executeNotebookCell` on the notebook, so it exercises the tier that works. Given how carefully the PR catalogues the earlier vacuous tests, this one is worth adding to that list. The missing coverage row is notebook x imported gate, and it belongs in the server suite rather than only in Playwright.

Related: in the working shape, an unsupplied given returns **400** naming `role` with a fix hint. The description says a 403 means "the gate could not be attached, or a given it names was not supplied" — the second half doesn't hold, and the only thing producing a 403 for a missing given here is this bug.

## Needs a decision: `PERSIST_COLOCATED_RELAXATION_ENABLED` defaults on

`config.ts:664` defaults this to `true`, which inverts the convention the two sibling flags follow — `PERSIST_STORAGE_MODE` shipped dark, and `PERSIST_COLLISION_ENFORCE` twenty lines up defaults `false` with "Flip this to true only after auditing". The comment makes the case for treating this one as a rollback lever instead, and that's a coherent position. I'd still rather see it opt-in for a release, because as written the freshness contract starts applying to packages nobody republished: a gated source with `#@ persist` isn't refused at publish today, so the next scheduled build materializes and binds one with no author action.

Not my call to make unilaterally, but there are currently no human reviews on this PR, so I don't think it should land on a default. What's the argument for shipping it live rather than dark for one release?

Two things attached to it:

- It's missing from `docs/configuration.md`'s env-var table, where `PERSIST_STORAGE_MODE` has a row. That table is where an operator looks before upgrading.
- A typo in the value doesn't fail fast. The sibling flags are module-level consts, so `parseBoolEnv`'s throw lands at startup. This one is a function evaluated during build-plan projection, and the throw is caught at `build_plan.ts:1123-1129`, logged as a generic "Failed to project a persist source into the build plan", and the source is dropped with no `refusedSources` entry. Fail-closed, but the operator gets a vanishing source instead of the actionable error this module is built to produce. Making it a const like its siblings fixes both.

## Docs that now contradict the code

- `docs/authorize.md:72-78` says the out-of-tree legacy-gate case is counted under `cause="legacy_string_gate"`. Line 408 of the same file says it "increments **nothing**" and explicitly warns not to watch that counter for it. Line 408 matches the code — `gate_classification.ts:507-535` returns `rejected` with no `cause`, and the recorder is guarded on `if (resolution.cause)`. Anyone alerting off the first paragraph gets an alert that never fires.
- `docs/givens.md:24` still describes `#(authorize)` as allowing or denying "a whole source (HTTP 403) — or, when the expression reads a row field, filter the rows returned instead". That's the split this PR removes. `:152` still names "A whole-source (given-only) `#(authorize)` gate" by the retired concept.
- `docs/security-posture.md:57-73` frames row-level filtering as the exception rather than the rule, and `authorize.md:178` links readers straight into it.

## Smaller things

- **The rename-misbinding framing.** The hole is real — I confirmed a `GROUPS=[1]` caller receives row 3 (`org_id=2`). But I ran the same model against the merge base and `main` serves the identical `[2,3]`, so it predates the branch rather than being a narrower hole traded in for the fail-closed fix. Documenting and pinning it is a genuine improvement and I'd keep the test; I'd just drop "one narrower hole replaces it". Two other notes: "fails closed unless the data collides" undersells it, since the misbind both hides rows the caller should see and exposes rows they shouldn't; and `docs/authorize.md:418`'s "the one narrow residual hole" is an assertion the suite doesn't establish — a rename chain, an `except:` plus a join reintroducing the name, and a composite member supplying it should all reproduce the same class, and none is tested.
- **`materialization_eligibility.ts:492-497`** admits the deep walk and `collectEntryPointGates` "are not proven equivalent, so a gap is not ruled out, only unobserved" — and the relaxation's safety argument rests on that equivalence. I traced the joined-child case and it refuses correctly, so I found no hole, but a cross-check test on a shared fixture would discharge it.
- **`isLegacyQuotedPayload`** (`authorize.ts:823-843`) matches only the double-quoted whole-body form, so `#(authorize) 'org_id = 999'` gets a generic non-boolean compile error instead of the paste-ready rewrite. Same bug class as the byte-identical-rewrite fix, one spelling over.
- **`RefusedSource.reason`** declares five enum values where the field's own description says only three can appear.
- **Stale comments:** `authorize.ts:397-399` describes a source carrying more than one `#(authorize)`, now refused at load (`authorize.ts:938`); `model.ts:2158-2160` references `GateEntry.dimensionForm`, which no longer exists.
- **`gate_dimension.ts`** no longer contains dimension-form handling; the filename outlived the concept.

## On the description's CI section

**DCO passes on this head** — every commit carries `Signed-off-by`, so the force-push question is already moot and no history rewrite is needed. A couple of other details drifted: the branch has 144 commits rather than 134, and `3579f269` isn't a valid object in the repo (there are two merge commits, `790c2b25` and `3ba8d9c9`). The red Playwright job isn't mentioned anywhere.

On CodeQL, your read is right: `stripMalloyCommentsAndLiterals` scans once, blanks each character at most once, and the input is already bounded upstream. Given that function is the pre-scan the laundering check depends on, I'd take the suppression over rewriting its inner loop for a lint.

## What I checked that held up

Flagging these so you don't spend review time re-defending them:

- Graft-or-deny: no `catch`, `?? []` or `|| []` in the gate-resolution path funnels a compile failure into "ungated". Every ambiguous branch lands on `rejected`.
- The caller-laundering control is form-agnostic and fails closed on an unreadable declaration.
- The escape handling in `stripMalloyCommentsAndLiterals` matches Malloy's own lexer, and the scanner only runs on text that already compiled, so its laxer string termination isn't reachable.
- The attribution requirement is stricter than I expected: a joined child's own cleanly-classifying gate causes the whole persist source to be refused, not admitted.
- The row-level gate compile is cached per `Model`, not recompiled per request.
- `#@ preaggregate` and `storage=` refusals are intact, and the rollup rationale ("groups across the gated column, so it could not be row-filtered afterwards even in principle") is the right one.
- Your compose tests genuinely prove the artifact isn't pre-filtered — two principals reading the same bound artifact get different row sets. I ran `materialization_eligibility.spec.ts` and `colocated_persist_relaxation.integration.spec.ts` locally: 32 pass, 0 fail.

Sha-Bang:
thanks — this was a genuinely useful pass, and the blocking one was real.

**fixed** in the three commits at the head (`8fecdc1c`, `b74f6da7`, `416e57d0`). Playwright is green. your diagnosis was right; two things shifted once I dug in:

the trigger isn't `import` as such, it's `!modelEntry.exported` — `import-statement.js` sets that on every imported entry, and `named-source.js` inlines an unexported entry's struct instead of emitting a name. same rule explains all four rows of your table without needing "first code cell" as a separate condition.

and it's **two** surfaces. the other is a model-level `query:` over an imported source — the documented single-query dashboard shape in `docs/dashboards.md` — where the stored `NamedQueryDef` keeps its own pre-graft `structRef` that `query-reference.js` reuses verbatim. that one was already broken on `main`, so pre-existing rather than a regression, but it's fixed in the same pass since it's the same invariant.

worth flagging for whoever reviews the fix: the cell **can't** just recompile its text instead. the grafted clone is loaded from a ModelDef and has no base URL, so a relative `import` in that text fails to resolve outright — which is exactly what an import-and-run cell contains. so it grafts the cell's own queryDef, deep-cloned per request because `cell.runnable` is memoized and grafting in place would let the first caller decide what everyone after them sees. on that branch the usual IR proof would be vacuous (it'd read back the array it just wrote), so the load-bearing check is object identity against what the materializer will actually execute.

separately, while in there: `assertGateLanded` read only `structRef`, but Malloy generates SQL from `compositeResolvedSourceDef ?? structRef`. with a composite run target the proof could pass on one struct while a stale other one generated the SQL. fixed first and on its own commit.

**agreed, and done:** the vacuous test — that's the fifth on this feature, so onto the list it goes. both API assertions rewritten around the three outcomes that actually exist, and you're right that an unsupplied given is a **400**, not a 403; the description said otherwise and now doesn't. `shipped_example_authorize.spec.ts` is exercising the wrong tier, agreed — the missing row is notebook x imported gate and it's now in the server suite, two files, since collapsing it to one makes it pass while testing nothing.

**agreed on the rename-misbinding framing.** I re-ran it against the merge base and `main` serves the identical rows, so it predates the branch. dropped "one narrower hole replaces it" and the "fails closed unless the data collides" line. `docs/authorize.md:418`'s "one narrow residual hole" is an assertion the suite doesn't establish — will soften rather than try to prove it.

**taking the rest:** the four docs contradictions (`authorize.md:72-78` vs `:408` is my own inconsistency, `givens.md:24`/`:152`, `security-posture.md:57-73`), `isLegacyQuotedPayload` missing the single-quoted spelling, the `RefusedSource.reason` enum, the two stale comments, `gate_dimension.ts` outliving its concept, and the CodeQL suppression. the `materialization_eligibility.ts:492-497` cross-check on a shared fixture is a fair ask — that comment admits the equivalence isn't proven and the relaxation leans on it.

**`PERSIST_COLOCATED_RELAXATION_ENABLED`** — you're right that it shouldn't land on a default with no human reviews on the PR, and I don't have an argument for live over dark that outweighs your point about the freshness contract reaching packages nobody republished. flipping it to default-off for a release, plus the `docs/configuration.md` row and making it a module-level const so a typo throws at startup like its siblings.

on the description: the sha you couldn't resolve was orphaned when I fixed up DCO, and the count moved with it — both gone from the body now, along with a current Playwright section.

one thing worth knowing for how you weigh the risk here: no served package in prod or staging declares a gate today. the row-level counters are absent in both while `publisher_authorize_bypass_total` (same module, same lazy init) is present and non-zero, and 24h of prod worker logs have zero gate lines. so nothing was being denied in the field — but it also means there's no production signal that would catch a mistake in this fix, which is why the tests are where I put the weight rather than trimming them.

jswir:
Re-reviewed at `416e57d0`. **The import-hop blocker is fixed** — confirmed independently rather than from the green check. Everything below is non-blocking.

I re-ran the matrix from the earlier comment against the new head, plus the cases the new code makes possible:

| case | result |
|---|---|
| `import` + `run` in cell 0 | serves, filtered to the caller's rows |
| markdown, then `import` + `run` | serves, filtered |
| `import` in its own cell, `run` in cell 1 | still works, no regression |
| same memoized cell, callers A -> B -> A | `[1,2]` -> `[3,4]` -> `[1,2]`, no bleed |
| two callers hitting the same cell concurrently | correct disjoint sets |

Over HTTP the three outcomes now match the contract exactly: no given -> 400 naming `role`; `role=intern` -> 200 with zero rows; `role=analyst` -> 200 with the one row. The rewritten e2e assertions are a real improvement — line 123 (zero rows for a non-matching given) is the assertion that actually proves the gate exists, and it was the one missing before.

I also reproduced all three "verified discriminating" claims by doing the reverts rather than taking them: precedence read -> 3 of 4 fail with the no-composite control surviving; snapshot sweep -> only the dashboard test fails, by denying; queryDef graft -> only the three notebook tests fail, by denying. Full server suite here is 3174 pass / 3 skip / 0 fail (the message says 3173).

**`8fecdc1c` is a genuine catch, and it wasn't in my review.** Proving against `structRef` while Malloy generates from `compositeResolvedSourceDef ?? structRef` is a fail-open with no error, metric or log line to notice it by. I checked the precedence against the vendored `query_model_impl.js:69` and `query_query.js:608` and it matches, and mirror-rather-than-union is the right call — a union reintroduces the leak from the other side.

Four follow-ups, none blocking:

**1. The deep-clone justification doesn't hold as written.** `b74f6da7` says the queryDef is deep-cloned per request because `cell.runnable` is memoized, and cites the two-callers test as what would catch losing it. I replaced the clone with an in-place mutation of the memoized object and all six tests in `authorize_import_hop.integration.spec.ts` still pass, including that one.

The reason looks structural: `condition.code` derives from the authored `#(authorize)` text (`gate_classification.ts:497`), never from a caller's given values, which bind later per request. So two callers of the same cell compute an identical `condition.code`, the mutation is idempotent, and the cross-tenant scenario in the message can't occur through this path. The clone is harmless and reasonable future-proofing — I'd keep it — but the claim about what the test protects is one notch stronger than the evidence. Flagging it because that same commit makes the point better than I can: overstating the last backstop is how the next vacuous test gets written.

**2. The one behaviour change flagged for reviewer attention has no test.** For a gated `query_source`, the filter now applying inside the stored query's own run target is the *correct* number — the previous one was an aggregate computed across all tenants — but nothing pins it. The shipped tests use a plain `table()` source. It also lives only in the commit message, not in `docs/dashboards.md` or the `authorize` field description, so an integrator seeing a dashboard tile's number change has nowhere to look it up.

**3. Small citation fix.** Composite member snapshots are described as "already pinned as denials in `authorize_integration.spec.ts`". They do fail closed, and by construction rather than by test — any unresolved struct denies at `model.ts:2400-2401`, which is the stronger claim. But the specific shape (an imported/composite gated member reached through a model-level named query) isn't pinned there; the general principle is.

**4. Adjacent shapes still untested**, listed for a follow-up rather than this PR: a three-file import chain through the runtime graft (the existing multi-hop tests cover load-time `validateAuthorizeProbes`, a different path); an import-and-run cell that isn't cell 0 — every `executeNotebookCell` call in the new spec passes index `0`; and a gated source imported and then further derived. `filterListContainsCode`'s mirrored `query_source` recursion is also only reached through the top-level read today.

One I checked so you don't have to: the `sourceArguments` hazard that motivated append-over-repoint. A gated *parameterized* source, imported, referenced through a named query with a non-default argument — the gate applies and the argument survives, no fallback to the declaration default. Append is doing what it claims.

Still open from the earlier comment, none of it touched by these three commits: the `PERSIST_COLOCATED_RELAXATION_ENABLED` default and its missing `docs/configuration.md` row, that flag's typo not failing fast unlike its siblings, and the three doc contradictions (`authorize.md` against itself on the `legacy_string_gate` metric, `givens.md:24`/`:152`, `security-posture.md:57-73`). CodeQL is unchanged and I still think your read of it is right.

jswir (approved):

PR #1046, opened 2026-08-19
  feat(server,mcp-apps): render a query result inline in the chat via MCP Apps

mlennie:
## What this does

`malloy_executeQuery` now ships an interactive HTML widget that renders its result inline in chat clients that support the MCP Apps spec. Clients that do not support it are unaffected: the text result is byte-identical to today's.

This is Phase 4 of `malloy-open-sourcing-plan.md`, which the plan deliberately sequences after Phase 3. That condition is met: `malloy_getContext`, `malloy_searchDatabaseSchema`, `malloy_searchDocs` and `malloy_compile` are all on `main`.

New `packages/mcp-apps` package, `ui_resources.ts` on the server, a `registerTool` switch, and four spec files. 32 files, +2555/-22.

## Scope: one widget, not two, and the reason is not effort

The brief asked for two widgets and named two exclusions. Both exclusions are confirmed. The inclusion of `get-context` did not survive research.

**`execute-query` is a genuine port, with no change to the tool result payload at all.** Publisher's query envelope already matches Credible's on purpose: `packages/server/src/mcp/query_envelope.ts` says so in its own header, keeping Credible's field names "including the leading underscores and the mixed casing of `renderLogErrors`" so a data app authored locally against Publisher and served through Credible sees one response shape. The widget's `rehydrate()` needs `rows`, `_meta.schema`, `_meta.annotations` and `_meta.connection_name`, and the envelope carries all of them.

**`get-context` is deferred because it would be a rewrite, not a port.** The two products return different data models:

| | Credible `get_context` | Publisher `malloy_getContext` |
|---|---|---|
| top-level key | `sources` | `results` |
| shape | nested `sources[].source_info.resource_id` + `entities[]` | flat array, each entity carrying a `kind` |
| version stamp | `schema_version: 1` | none |

Both widget renderers read `output.sources`, which Publisher never emits, and with no `schema_version` the dispatcher routes to the legacy v0 renderer. Shipping it would mean either rewriting the renderer or reshaping the tool response, and reshaping a tool response is exactly what #938 and #974 were about.

**Also excluded, confirmed:** `html-report` (that tool was deliberately removed in favour of data apps) and `list-workspaces` (workspaces are a Credible-platform concept the plan keeps out of the open source).

## The regression risk, measured rather than argued

#938, #943 and #974 recently reshaped MCP tool results for legibility, description ordering and the 413 size cap. A widget that re-serialises a result or grows the payload could undo all three quietly, so this was measured both directions:

- **Tool result: byte-identical**, once Malloy's per-compile `reference_id` UUIDs are normalised (two of them, freshly minted per compile, none shared between runs).
- **`tools/list`: 6 lines added, 0 removed** (the `_meta` block). The description is still 1,992 bytes with all 8 input properties intact.
- The widget served over MCP `resources/read` is **sha256-identical** to the file on disk.

## Size, since it will be the first question

The built widget is one self-contained file: **3,785,508 bytes (3.79 MB), 1,073,774 gzip**. Composition is `@malloydata/render` at 5,222 kB rendered, `ext-apps/app-with-deps` at 325 kB, and 21 kB of local code.

**It is not reducible.** `@malloydata/render` arrives as one pre-bundled module with Vega and Vega-Lite already inside it, so tree-shaking cannot reach them.

## Verification

- Full gate green: lint 0, typecheck 0, prettier clean apart from the six standing git-ignored generated files, unit 2999/0, integration 313/0, sdk 335, skills 13.
- Live: the widget served over MCP, a real bar chart rendered, and the unchanged text path confirmed for a client that ignores MCP Apps.
- Adversarial and security review: 4 findings, all fixed and each verified in both directions. Security found nothing.

## Docs

`docs/ai-agents.md` gains a section, `AGENTS.md` gains a tool line and a first-run step, and the server's `MCP_INSTRUCTIONS` string is updated. No new env var and no new CLI flag.


packages/mcp-apps/src/shared/rehydrate.ts
```diff
@@ -0,0 +1,206 @@
...
+   sql?: string;
+   model_annotations?: Annotation[];
+   query_timezone?: string;
+   source_annotations?: Annotation[];
+}
+
+/**
+ * Reconstruct a full Malloy Result from compact rows and metadata.
+ */
+export function rehydrate(
```
Sha-Bang, line 36: **Must-fix.** This file has no tests, and it is the one file here that can produce a silently *wrong* chart rather than an absent one: ten cell kinds of coercion, bigint precision carried in `string_value`, nested arrays and records for `nest:`, and a schema field with no matching key in the row.

`tool_result_payload.ts`'s own header says it is "the piece most likely to be wrong against a given server, so it is the piece that most needs tests". That reasoning applies harder here — the highlighter got a 96-line spec and the three message constants got 52 lines, while the type conversion got none.

A spec covering bigint, null, `nest:` (array of records), `timestamptz`, `sql_native`, and a missing key would be enough.

packages/server/src/mcp/server.ts
```diff
@@ -51,7 +52,7 @@ To build a model from a database rather than from an existing package, start wit
 
 Task-specific guidance is served as prompts you can fetch by name: malloy-getting-started to begin, malloy-modeling to build or change a model, malloy-analysis to explore and answer questions, and malloy-review to check correctness.
 
-Results and any charts render in the Publisher web UI on the REST port (4000 by default).`;
+Results and any charts render in the Publisher web UI on the REST port (4000 by default). Where this server serves the MCP Apps widget and your client supports it, a malloy_executeQuery result also renders inline in the conversation: call resources/list to see whether it does. The expanded=true parameter opens that card by default; set it when the result answers the question being asked, and leave it closed for a wide table or a query using nest:.`;
```
Sha-Bang: **Must-fix**, and it is the one place this PR breaks its own rule. Everything else is gated both ways on the bundle being present, but this sentence goes to every client unconditionally: clients that cannot render a widget, and servers with no bundle at all — where `docs/ai-agents.md` itself says `resources/list` answers `-32601 Method not found`. So the instructions tell an agent to make a call that errors, on precisely the server that deliberately advertises nothing.

`registerUiResources` already returns the URI list. Build the instruction string from it and this becomes consistent with the rest of the design.

packages/mcp-apps/execute-query.html
```diff
@@ -0,0 +1,165 @@
...
+<html lang="en">
+   <head>
+      <meta charset="UTF-8" />
+      <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+      <title>Malloy Query Result</title>
+      <!--
+        Only the classes renderer.ts actually sets are styled here. The
+        collapsible chrome around this card is styled INLINE in
+        shared/collapse_wrapper.ts instead, because some hosts (notably Claude
+        Desktop) sandbox the iframe in a way that drops this <style> block.
```
Sha-Bang: **Must-fix**, because it decides the shape of a whole module. This claim is the sole justification for the ~150 lines of `Object.assign(el.style, ...)` in `shared/collapse_wrapper.ts` — but every other style in the card lives in this same `<style>` block (error box, warning box, query expando, the `.mh-*` colours), and so does every byte of `@malloydata/render`'s CSS, which vite inlines into it.

If Claude Desktop really dropped this block, the charts would be unstyled too — and the screenshot posted in Slack shows it rendering properly. I suspect this is a fossil from debugging something else.

Worth settling before merge: if the claim is wrong, `collapse_wrapper` becomes ordinary CSS and loses most of its bulk, and `tokens.ts` stops needing to be hand-kept in step with this file.

packages/server/src/mcp/ui_resources.ts
```diff
@@ -0,0 +1,227 @@
...
+ * Two keys carry the same URI: `ui.resourceUri` is the MCP Apps spec's, and
+ * `openai/outputTemplate` is the equivalent ChatGPT reads. A server emitting only
+ * one is invisible to the other host, so Credible's tools emit both and so do
+ * these.
+ */
+export function uiToolMeta(
+   resourceUri: string,
+   widgetDir: string = resolveWidgetDir(),
+): Record<string, unknown> | undefined {
+   if (!hasExecuteQueryWidget(widgetDir)) return undefined;
```
Sha-Bang, line 191: Takes a `resourceUri` and then ignores it — the check is always `hasExecuteQueryWidget`. Correct for one widget, wrong the moment there is a second, which this module's own docs anticipate. Either key the existence check off the URI, or drop the parameter and name the function for the one widget it serves.

packages/server/src/mcp/ui_resources.ts
```diff
@@ -0,0 +1,227 @@
...
+/**
+ * Declares the UI extension and registers the widget resources.
+ *
+ * Must be called before the server is connected to a transport:
+ * `registerCapabilities` throws afterwards.
+ *
+ * Returns the URIs registered, so a caller (or a spec) can distinguish "widgets
+ * served" from "no bundle present" without reaching into the server.
+ */
+export function registerUiResources(
```
Sha-Bang, line 207: No off switch: the only way to disable this is to not build the widget, which a deployment running the published package or the image cannot do. `--no-mcp-config` / `PUBLISHER_NO_MCP_CONFIG` set the precedent, and an operator whose host renders the card badly has no way out short of a rebuild. A `PUBLISHER_NO_MCP_APPS` early-return here is a few lines.

packages/mcp-apps/src/execute-query/renderer.ts
```diff
@@ -0,0 +1,335 @@
...
+      reportRenderError(
+         new Error(unique.join("; ") || "Unknown plugin render error"),
+         "plugin",
+      );
+      return true;
+   };
+
+   // Measure the real rendered height once the Malloy web components settle,
+   // then size the container to fit.
+   const observer = new MutationObserver(() => {
```
Sha-Bang, line 324: This is the part I would expect to break first: a MutationObserver plus `setTimeout(100)`, polls at 500/1000/2000 ms, a 420 ms transition constant, a three-deep `firstElementChild` walk, and the `.malloy-dashboard` / `.malloy-error-message` class names.

The class-name coupling itself has precedent — `sdk/src/components/RenderedResult/RenderedResult.tsx` reaches for `.malloy-dashboard` too — but there it is CSS, so a Malloy bump degrades into a style regression. Here it degrades into a clipped or blank card, and nothing tests it.

A `ResizeObserver` on `container` replaces most of the timers. If you would rather leave it as is, at least name the coupling somewhere a person bumping Malloy will look.

packages/mcp-apps/src/shared/tokens.ts
```diff
@@ -0,0 +1,39 @@
...
+ * the charts look like Malloy's charts.
+ */
+
+export const FONT_SANS =
+   'system-ui, -apple-system, "Segoe UI", Roboto, Helvetica, Arial, sans-serif';
+
+export const FONT_MONO =
+   'ui-monospace, SFMono-Regular, Menlo, Consolas, "Liberation Mono", monospace';
+
+export const tokens = {
```
Sha-Bang: Hardcoded `#333333` on `#FFFFFF` with no `prefers-color-scheme` path, so this is a white card in a dark chat client. The docs disclose the `# theme.*` gap honestly; this one deserves a line in the same place.

packages/mcp-apps/package.json
```diff
@@ -0,0 +1,32 @@
...
+   "type": "module",
+   "repository": {
+      "type": "git",
+      "url": "https://github.com/malloydata/publisher.git",
+      "directory": "packages/mcp-apps"
+   },
+   "scripts": {
+      "build": "bunx vite build",
+      "typecheck": "bunx tsc --noEmit",
+      "lint": "bunx eslint ./src --fix",
```
Sha-Bang: `--fix` in a lint script: every other package lints without it (`packages/server` is `eslint "./src/**/*.ts"`). CI runs the root `lint`, so this mutates the tree in CI and quietly fixes anything auto-fixable instead of failing on it.

package.json
```diff
@@ -18,8 +18,9 @@
     "build": "NODE_ENV=production bun run build:sdk && NODE_ENV=production bun run build:server",
     "build:cli": "cd packages/cli && bun run build",
     "build:skills": "cd packages/skills && bun run build",
+    "build:mcp-apps": "cd packages/mcp-apps && bun run build",
     "build:create-malloy-package": "cd packages/create-malloy-package && bun run build",
-    "build:server-deploy": "NODE_ENV=production bun run build:sdk && NODE_ENV=production bun run build:app:server && NODE_ENV=production bun run build:server",
+    "build:server-deploy": "NODE_ENV=production bun run build:sdk && NODE_ENV=production bun run build:app:server && NODE_ENV=production bun run build:mcp-apps && NODE_ENV=production bun run build:server",
```
Sha-Bang: `build:server-deploy` now builds the widget twice: here, and again inside `packages/server`'s own `build`, which you added `bun build:mcp-apps` to. That is ~20s of Vega bundling for nothing. Dropping it from this line is enough — the server build covers it, and that is also what makes the npm publish path work.

packages/mcp-apps/src/shared/mcp_app_init.ts
```diff
@@ -0,0 +1,74 @@
...
+import { App } from "@modelcontextprotocol/ext-apps/app-with-deps";
+import {
+   extractPayload,
+   type ExtractedPayload,
+   type ToolResultBlock,
+} from "./tool_result_payload";
+
+// Suppress the noisy per-message debug logs PostMessageTransport emits.
+const originalDebug = console.debug;
+console.debug = function (...args: unknown[]) {
```
Sha-Bang, line 42: Matching third-party log strings will stop working silently on an ext-apps bump, and it swallows `Ignoring message from unknown source` — exactly the line you would want when a host integration misbehaves. Consider gating it behind a URL param, or dropping it.

docs/ai-agents.md
```diff
@@ -64,7 +64,40 @@ The server listens at `http://localhost:4040/mcp` (set the port with `--mcp_port
...
+
+Three things make this work, and the server sets all of them up for you:
+
+- it declares the `io.modelcontextprotocol/ui` extension when it initializes
+- it serves the widget as a resource at `ui://execute-query/app.html`, with the MIME type `text/html;profile=mcp-app`
+- `malloy_executeQuery` carries a `_meta` field naming that resource
+
+Nothing to configure. The widget is one self-contained HTML file with no external scripts, styles or fonts, so it needs no hosting URL and no content-security-policy entries, and it works the same whether Publisher is on localhost, in Docker, or behind a proxy. It makes no network requests of its own: the result reaches it from the host, not from the server.
+
+**Most clients do not support MCP Apps, and nothing is lost when they do not.** The tool result is identical either way, so an agent reading the JSON sees exactly what it saw before. A client that ignores the extension simply gets no card.
```
Sha-Bang, line 83: a47b4b17 fixed this for the connect section — worth carrying the same honesty into here, since this is where someone reads about the feature and immediately asks "so which client draws me a chart?" Today that is ChatGPT via `openai/outputTemplate`, and Claude Desktop only through the `mcp-remote` bridge. Naming them saves the reader the experiment.

Sha-Bang (changes requested):
Reviewed the full diff, then checked the branch out and ran it. Approve with changes — the engineering here is careful, and the wiring I most expected to be silently wrong is right. Three must-fixes; the rest are small.

- 27/27 `mcp-apps` tests, 37/37 server MCP tests (`ui_resources`, `ui_resources.protocol`, `execute_query_tool`), typecheck and prettier clean.
- Widget builds in 17s to **3,785,578 bytes** — matches the claim — with **zero** external `src`/`href` refs in the output. Self-containment is real.
- "Not reducible" holds: `@malloydata/render` ships one pre-bundled 5.26 MB `index.mjs` with Vega already inside it.
- The publish path is correct, which is the thing most likely to break quietly: `npm-sdk.yml` publishes via `cd packages/server && bun run build`, and that script now runs `build:mcp-apps`, so the widget reaches the tarball. Docker builds it explicitly because `build:server-only` skips it. +3.8 MB on a 28.6 MB unpacked package.
- The error path is right against the server's *actual* error shape: `jsonResource` puts `{error, suggestions}` in `content[0].resource`, so `extractPayload` finds JSON and the widget takes its error branch instead of mis-reporting "result too large".

## Is it worth doing?

Yes, though the reach is narrower than the PR reads. Per a47b4b17's own commit message, Publisher has no stdio transport, so Claude Desktop only gets here through `mcp-remote` — and Claude Code, which `AGENTS.md` is written for end to end, renders no widgets at all. That is fine as forward-looking work. It is also the argument for a kill switch and for naming the clients that render today.

## Is it the right way?

Mostly, and better than I expected in places. The tool payload is untouched, so #938, #943 and #974 stay intact. Everything is gated on the bundle in both directions, with a test pinning that `_meta` and the resource registration can never disagree. `devDependencies` only, so `--production` installs nothing. The vite plugin hard-errors on a split bundle or a dangling asset rather than shipping a blank widget — and the `$&`-in-a-replacement-string comment is clearly a bug someone actually hit.

## Does it fix it?

Yes, with one hole: `rehydrate.ts` is the only file here that can produce a silently *wrong* chart, and it has no tests, while a 103-line syntax highlighter got a 96-line spec. Worth saying out loud too: none of these unit tests run in CI — `build.yml` runs typecheck, lint, prettier and `test:scripts` only — so "unit 2999/0" is a laptop claim rather than an enforced gate. Not this PR's fault, and typecheck, lint and the vite build now do cover the new package.

## Open-source quality?

Comment density matches the repo (`query_envelope.ts` on `main` reads the same way), headers and licensing are right, and there is no product palette or licensed font. One ask that is off the diff: the description justifies its scope with `malloy-open-sourcing-plan.md`, which is not in the repo, so nobody outside can check any of it. The description already makes the argument standalone — just drop the reference.

---

**Must-fix:** `rehydrate` tests · conditional instructions · resolve the `<style>` contradiction
**Should-fix:** kill switch · `uiToolMeta`'s unused URI · `--fix` in lint · double build · dark-mode disclosure

Sha-Bang (commented):
Correction to my review above. I wrote that none of these unit tests run in CI, on the strength of `build.yml` running only typecheck, lint, prettier and `test:scripts`. That is wrong for the server suite: `build.yml` triggers on `pull_request` and calls `cross-platform-tests.yml`, which runs `bun run --cwd packages/server test:unit` — and the integration suite — on ubuntu, windows and macos. So the 37 server MCP tests here are CI-enforced on three platforms, not a laptop claim.

What does stand is narrower: that workflow runs the sdk, skills and server suites, and `packages/mcp-apps` is not among them. So the 27 new widget tests are the ones with no CI gate, including the `rehydrate` spec if you add one. Adding an mcp-apps step next to the others there would close it.

The `rehydrate` finding is unaffected, and it is still the one I would fix first.

mlennie:
Thanks for this. It was a careful review, all three must-fixes were things I should have caught myself, and the third took a test rather than an argument to settle. Below is every item and what changed.

## Must-fix 1: `rehydrate.ts` had no tests

Fixed. You were right, and right that this was the one to do first: it is the only module in the widget that can produce a chart that is *wrong* rather than one that is absent. Everything else fails loudly. A miscoerced cell draws a perfectly normal chart of the wrong numbers.

21 specs, chosen for the cases where being wrong is invisible rather than for coverage: bigint precision surviving in `string_value`, null never becoming zero, `false` and `""` not mistaken for null, a schema field whose key is absent from the row, timestamptz keeping its offset, `nest:` as an array of records, inline records, scalar arrays, json, `sql_native`, and an unknown future type falling back rather than throwing mid-render.

Each was mutation-tested rather than assumed. Six deliberate breaks, each confirmed to turn the right specs red before being reverted: dropping the bigint `string_value` (1 red), null becoming zero (3), timestamptz losing its branch (1), nested records read against the wrong fields (1), positional order reversed (3), and a missing key becoming `undefined` instead of a null cell (1). The two that matter most are the ones that would otherwise have drawn a plausible chart: reversed order mislabels every column, and wrong-field nesting empties every tile.

## Must-fix 2: the ungated instructions

Fixed, and your framing was exact: this was the one place the feature broke its own rule.

It was worse than described. On a bundle-less server the sentence not only described a capability that does not exist, it told the agent to call `resources/list` to check, and on exactly that server `resources/list` answers `-32601 Method not found`. So it sent an agent to make a call that errors, on the server that deliberately advertises nothing.

The widget sentence is now its own module-scope literal, appended only when the same predicate `registerUiResources` consults returns true, so the instructions and the registration cannot disagree. The `resources/list` advice is gone from both branches rather than moved into the gated one: it was never good advice, it just errored less loudly with a bundle present.

The selection is a small function so both branches can be asserted without a filesystem. Worth mentioning because it nearly fooled me: my first attempt asserted this over the in-memory transport and appeared to pass while measuring an empty string, because that spec builds a bare server with no instructions at all.

## Must-fix 3: the `<style>` claim. Tested, you were right, and it is gone

Fixed, and the claim removed rather than reworded.

**What was tested.** An ad-hoc query in Claude Desktop, connected to a Publisher built from this branch over an `mcp-remote` bridge. The untouched card showed the padded `<> Malloy query` row with its chevron and **no Malloy source beneath it**, going straight to the result table.

**Why the collapsed state is the evidence.** The rule that hides the source, `.query-expando-body { max-height: 0 }`, lives in the page-level block. A hidden source therefore means the block applied. Both outcomes were rendered from the real build beforehand so the screenshot was read against reference images rather than a description: with the block, the source sits multi-line and indented in a padded monospace panel with coloured keywords; without it, the same source is a single unindented line flush left, unpadded, entirely black in the proportional UI font.

**Your inference, corrected in one direction and confirmed in the other.** As stated it did not hold: `@malloydata/render` styles through `attachShadow`, so its CSS is in a shadow root and a styled chart alone does not prove the page-level block survived. But the comment named two mechanisms, and one of them your argument does settle. Render injects its CSS with `createElement("style")` and uses no `adoptedStyleSheets`, and CSP `style-src` applies per document including inside shadow roots, so a policy blocking inline style elements would have left the chart unstyled too. It did not, which ruled out the CSP mechanism before the test ran. The test then closed the remaining one, a host sanitising `<style>` out of the document. So the conclusion is yours; only the route needed adjusting.

**What changed.** The chrome is styled from the same stylesheet as the rest of the card, with one `is-open` class on the wrapper driving the chevron, the header's border and the body's visibility together, where three separate inline assignments could drift apart. 56 lines out of `collapse_wrapper.ts`, all five `Object.assign` blocks gone, none left in the widget. `tokens.ts` is deleted entirely, another 49 lines: its only remaining consumer was one inline border on the in-frame error box, now a class, and its whole reason for existing was that CSS cannot import TypeScript so the scripted styles needed hand-keeping in step with the HTML.

**Verified by rendering, not by typecheck**, since this is the one file whose styling has no regression test. The rebuilt widget was re-rendered in both states and compared against the references captured before the change: pixel-identical in both the untouched and the clicked state. The toggle was exercised through a full cycle in a browser, closed to open to closed, with the label, body visibility and header border all following.

**The honest limit:** one client, Claude Desktop, over the bridge. A host that does strip the block would now lose the card's chrome along with everything else, where previously the chrome would have survived. That trade is deliberate and rests on the test.

I also traced the claim's provenance rather than treating it as anonymous: in the Credible codebase the wording predates the commit that moved it into a shared module, which carried it forward unchanged and has no body. Nothing on that side recorded what was observed, which is consistent with your read that it was a debugging fossil.

## Should-fixes

All six.

**Kill switch.** `PUBLISHER_NO_MCP_APPS`, mirroring `PUBLISHER_NO_MCP_CONFIG`. It gates the single predicate everything else consults, so all four surfaces go quiet together: the extension, the `ui://` resource, the tool's `_meta`, and the sentence in the instructions. Verified on a running server rather than only in specs.

Two defects came out of writing it, both mine and both in the same day. `registerUiResources` was calling the raw filesystem check rather than the gated predicate, so the switch silenced three surfaces and not the fourth, while my own comment claimed all four followed one predicate. And the flag was read on every call, which means on every MCP POST, and `parseBoolEnv` throws on a typo: one bad character produced a server that booted clean, reported `serving`, answered REST normally, and returned `-32603` to every MCP request with nothing saying why. It is now resolved at module scope, which is what the line three above it in `server.ts` already did and said why.

**`uiToolMeta`** is now `executeQueryToolMeta()` and takes no URI. It previously accepted a `resourceUri`, returned it, and gated on the execute-query widget, so asking for any other URI handed back metadata whose presence was decided by a different file: generic in shape, specific in fact.

**`--fix` dropped** from the mcp-apps lint script. A lint step that rewrites the tree is a lint step that can pass by editing.

**`build:server-deploy`** no longer builds the widget; the server's own build already does, which is what makes the npm publish path work, so the line was paying Vega's bundling cost twice.

**`console.debug` suppression** narrowed to an exact-match set of the two high-volume per-frame strings. It no longer swallows "Ignoring non-JSON-RPC message" or "Ignoring message from unknown source", which are exactly the lines worth having when a host misbehaves. Matching a third-party library's log text stays fragile, so the comment now says so and says the failure mode is noise rather than breakage: accepted rather than solved.

**Dark mode** is disclosed rather than fixed, which is what you asked for, and it is now in `docs/ai-agents.md` next to the `# theme.*` gap rather than only in a code comment. The same Claude Desktop run confirmed it live rather than by reasoning: the card renders as a white panel in a dark client, exactly as the docs now say. The comment explains why a media query is the wrong fix: MCP Apps hands the widget the host's own theme, so consuming that is the answer and a media query would be a second source of truth.

## Your CI correction

Worth calling out because the corrected version is the most useful finding in the review, and acting on it took minutes only because you narrowed it yourself.

You are right that the server suite is enforced on three platforms through `cross-platform-tests.yml`, so the 37 server MCP specs were never ungated. The narrower point that stands is that `packages/mcp-apps` was not in that workflow at all, so the widget's specs ran nowhere. That is now closed, in the same commit as must-fix 1, deliberately: landing the rehydrate spec without a gate would have been the worst of both worlds. 48 widget tests now run on ubuntu, windows and macos, and Windows is not incidental here, since the sibling `ui_resources` spec had already caught a path assertion that passed on macOS and failed there.

## The description's reference to the open-sourcing plan

You are right that it is uncheckable from outside; that file is not in this repo. I am not editing the description, because correcting a body silently rewrites what earlier readers were responding to. The argument stands without it, so the paragraph should simply be trimmed from the squash commit-message box at merge.

## One figure in the description is now stale

The widget is larger than the description says. It was 3,785,578 bytes when you verified it and is now 3,792,989: `main` moved `@malloydata/*` to 0.0.432 and this branch followed, which more than offsets the 105 lines must-fix 3 removed. Gzipped it is 1,075,205 bytes, so the description's "3.8 MB / 1.07 MB" reading still rounds true.

Thank you for checking things rather than reading them: the size, the zero external references, that render really is one pre-bundled module, and above all that `npm-sdk.yml` reaches `build:mcp-apps` through the server's build. That last one is the failure I could not have caught locally.

mlennie:
One commit since that comment, plus a correction to a number in it.

## `1d616fc6`: the Vite plugin no longer depends on the path separator

Not one of your findings. It came out of chasing a red `windows-latest` leg, and it is worth flagging because it touches the plugin you reviewed.

The plugin matched the inlined `<script>` tag against the whole rollup bundle key. That rests on rollup's `fileName` being forward-slashed on every platform, which is a real contract, but I had only ever observed it on macOS. A contract not observed on the platform is not an observation of the platform, so rather than keep reasoning about the invariant I measured the failure mode, which needs no Windows:

```
forward key, forward tag : OK, inlined
BACKSLASH key            : FAIL "no <script> tag references assets\eq-Hb60.js"
BACKSLASH tag            : FAIL "no <script> tag references assets/eq-Hb60.js"
```

So the dependency was real and would hard-fail the build rather than degrade. The tag is now matched by **basename**. The basename carries the content hash, so it still identifies exactly one chunk, and both guards survive rather than being loosened: an unreferenced chunk still errors, and a split bundle still errors naming the split chunk.

The plugin had no test at all. It has one now, covering all four separator combinations and both guards, mutation-tested three ways: reverting the basename match turns 2 specs red, splitting on forward slash only turns 1 red, removing the unreferenced-chunk guard turns 2 red. Build output is byte-identical, sha256 `d4fd0063`, with zero external references, so this changes the build's platform robustness and nothing about what it emits.

**To be clear about what this is:** hardening that closes one hypothesis, not a claimed fix for the red leg. I have not read that log. If the failing step turns out to be the build and it carries that exact `no <script> tag references` string, this is the fix and one line settles it. If it is the test step, this is unrelated and I keep looking.

## A stale figure in the comment above

I wrote that 48 widget tests now run on ubuntu, windows and macos. It is **54**. The comment was accurate when posted and the spec above added six; correcting it here rather than editing the comment, since your review is an answer to that text.

mlennie:
One more of your findings, and I owe you an apology on process before the substance: your `renderer.ts` comment was resolved without being addressed, and it never made it into my consolidated reply either. It was in the hidden conversation behind the "Load more" link, which is exactly the failure you called out on #1047. You should not have to diff the code to find that out, so I am saying it here rather than letting you discover it.

You were right, and not in the hypothetical way. It is a live bug, it is user-visible today, and the mechanism turned out to be none of the six things either of us would have guessed.

## What I measured

I captured a real dashboard tool result over MCP, the bundled storefront `business_overview` view, and rendered it in the built widget.

The container's inline height stayed at **400px** while the dashboard's content measured **1900px**. Clipped by 1500px. The KPI row is cut mid-glyph at "AVG ORDER VALUE", and the monthly trend chart, the state map, the category bars and the top-products table are not merely cropped, they are never visible at all.

Then the part that took a while, because the sizing code was computing the right answer. I instrumented the measurement and it produced `rendered: 1900` twelve separate times while the height never moved. Instrumenting the assignment itself gave it away: the value being assigned was `NaN`.

## The cause

Arithmetic, not layout, and not any of the timers.

`horizontalScrollbarAllowance` walked `querySelectorAll("*")`. That returns SVG elements as well as HTML ones, and an SVG element reports a `scrollWidth` and a `clientWidth` but has no `offsetHeight` at all. So `offsetHeight - clientHeight` was `NaN`, `Math.max` propagates `NaN`, and a single SVG node poisoned the entire walk.

`container.style.height = "NaNpx"` is invalid CSS. The browser drops it silently, without throwing, so the container kept the 400px it is created with. And because that assignment sits immediately before `observer.disconnect()`, the MutationObserver never disconnected: it re-measured on a loop for the lifetime of the card.

Measured on that dashboard: 43 of its 787 nodes were overflowing SVG nodes, any one of which was sufficient.

Chart results escaped it purely by timing. Their first measurement lands before the chart's SVG exists, so the walk is clean and the height sticks. That is why every render I verified while building this looked correct, and it is a fair summary of your "nothing tests it".

The line that hid it was the type annotation. `querySelectorAll<HTMLElement>("*")` asserts something the DOM does not guarantee, so the compiler was satisfied that `offsetHeight` existed on every node.

## What changed

Guarding on `Number.isFinite` fixes it. The arithmetic moved into `scrollbar_allowance.ts` so it can be tested without pulling in `@malloydata/render`, and it is tested now: 9 specs whose fixtures are the values Chromium actually reported for that dashboard, `scrollWidth: 37, clientWidth: 0, clientHeight: 2` and no `offsetHeight`, rather than values I invented.

Mutation-tested, and one of them is worth reporting because it caught me rather than the code. Replacing the `Math.max` with a plain assignment turned nothing red: my only multi-node case happened to put the larger value last, where an assignment gives the same answer. A test that cannot distinguish the two was passing and telling me nothing. It now asserts both orderings, and that mutation turns it red. Removing the finiteness guard turns 3 red, and dropping the sub-pixel tolerance turns 1.

After the fix the dashboard sizes to 1900px with nothing clipped, the chart is unchanged, and neither logs a page error.

## The ResizeObserver, which I have not done

Declining it for this PR, and saying so out loud rather than leaving it silent, which was your standing ask across this batch.

The reasoning: with the NaN gone, the existing mechanism measures both shapes I can test correctly, the rewrite is real work on the one file with no browser-level regression test, and I have now had two claims about this widget go false against my own code. I would rather land the fix that is measured than pair it with a rewrite that is not. If the sizing proves fragile on a shape I have not tried, the ResizeObserver is the right answer and it should be its own change with a browser test behind it.

Your fallback ask I have taken in full, because you are right that it is the thing that helps the next person. The coupling is now named at the block that depends on it: the `.malloy-dashboard` and `.malloy-error-message` class names, the container-child-grandchild nesting depth the walk assumes, a pointer to the SDK's CSS-level use of the same class and why that one degrades more gently, and a plain statement that the arithmetic is tested and the DOM assumptions are not.

While I was in there I checked the assumption rather than trusting it: the three-deep walk does still find `.malloy-dashboard` on 0.0.432. One honest wrinkle, the dashboard branch is currently a no-op for this result, since the grandchild's `scrollHeight` and the great-grandchild's were both 1900. So the branch is load-bearing for a shape I have not found yet, or not load-bearing at all, and I cannot tell you which.

## What I did not change

The 420ms constant and the 500/1000/2000 polls are still there. The 420 is paired with a 400ms CSS transition in the stylesheet, which is a cross-file magic number and I agree it is ugly. I looked at `transitionend` and did not take it: it does not fire when the property does not actually change or when a host disables transitions, so it needs a timeout fallback anyway and comes out as more code with two paths instead of one. Not defending it as good, just as not clearly better.

Thank you for this one specifically. It was the finding that a passing build, a green suite and four verified screenshots all missed, and the only reason it got found is that you flagged the file and I went and measured it instead of reasoning about it.

packages/mcp-apps/src/execute-query/renderer.ts
```diff
@@ -0,0 +1,334 @@
...
+      return;
+   }
+
+   // Agent-driven auto-expand: with `expanded=true` on the tool input the card
+   // starts open instead of behind a "Show Query Result" header.
+   const { body: collapseBody } = buildCollapseWrapper({
+      root,
+      subject: CARD_SUBJECT,
+      defaultOpen: toolInput?.expanded === true,
+      onToggle: () => notifySize(),
```
Sha-Bang: **Blocking.** The card starts collapsed unless the agent passes `expanded=true`, so `viz.render(container)` at :196 runs while `.mcp-card-body` is `display: none`. Everything in that subtree measures 0, so the observer at :246 reads `rendered = 0`, never assigns a corrected height, and never disconnects — the container keeps the literal `400px` from :157. Opening the card does not fix it: `setOpen` toggles `is-open` on `wrapper`, which is the observed container's *ancestor*, so no mutation reaches the observer, and this `onToggle` only re-posts a size message.

I ran it, real payload, 800px of content:

| | collapsed (default) | after clicking "Show Query Result" | control: `expanded=true` |
|---|---|---|---|
| body display | `none` | `block` | `block` |
| content scrollHeight | 0 | 800 | 800 |
| container height | `400px` | **`400px`** | `800px` |

400px of content clipped, and `.result-frame` is `overflow: hidden`, so there is no scrollbar to reach it. The control is the same content through the same code path with the card open — that is what isolates `display: none` as the cause rather than the measurement arithmetic.

What makes this the default rather than an edge case: `MCP_APPS_INSTRUCTION` in server.ts and docs/ai-agents.md:73 both tell agents to leave the card closed "for a wide table or a query using `nest:`" — the two results most likely to exceed 400px.

Measure when the card opens, not only when it renders: re-run the measure block from this `onToggle`, or observe `wrapper` for the class change rather than `container` for content. And add a renderer spec — there is none, and this is the path the docs steer agents toward.

While you are in here: the observer and the `[500, 1000, 2000]` polls are never torn down on the error path either. `renderError` does `root.innerHTML = ""` at :302, so they keep firing and reading `scrollHeight` off a detached container for the life of the widget.

packages/mcp-apps/src/shared/collapse_wrapper.ts
```diff
@@ -0,0 +1,122 @@
...
+   header.appendChild(label);
+
+   if (metadata) {
+      const meta = document.createElement("span");
+      meta.className = "mcp-card-meta";
+      meta.textContent = metadata;
+      header.appendChild(meta);
+   }
+
+   if (!hostChromeSuppressed) {
```
Sha-Bang, line 95: **Blocking.** `renderError` always passes `forceCollapsed: true` (renderer.ts:309), which forces `startsOpen = false` at :66. Under `chrome=none` this line then skips the header entirely. So the wrapper never gets `is-open`, `.mcp-card-body` stays `display: none`, and there is no header to click. Each half is sensible on its own; together they make the error unreachable.

Verified with `?chrome=none`: the error text is in the DOM, its rect is 0×0, the card renders 2px tall, and `querySelectorAll('.mcp-card-header, button, [role=button]')` returns 0. The same page without the param: click the header, 45.5px of error text. So in that host every failure this PR wrote copy for — `NO_PAYLOAD_*`, `RESULT_TOO_LARGE_*`, "Missing result data or schema metadata", a Malloy render error — is an empty bordered box.

When chrome is suppressed, render errors open. The compactness argument for `forceCollapsed` is about not expanding into the conversation, and a host that suppressed the chrome has already made that call.

Separately, and true with chrome as well: an error card labelled "Show Query Result" reads as a result you have to click, not as a failure. Someone whose query failed sees a collapsed box promising output. Label the error card as an error.

This file has no spec, and `forceCollapsed` and the `chrome=none` branch are exactly its two untested states — their interaction is this bug. While you are in here, `metadata`, `bodyDisplay` and the returned `setOpen`/`wrapper` are unused by both call sites: drop them, or cover them.

packages/mcp-apps/src/shared/rehydrate.ts
```diff
@@ -0,0 +1,206 @@
...
+         if ("subtype" in type && type.subtype === "bigint") {
+            cell.string_value = String(value);
+         }
+         return cell as unknown as Cell;
+      }
+
+      case "boolean_type":
+         return {
+            kind: "boolean_cell",
+            boolean_value: Boolean(value),
```
Sha-Bang: The spec you added is thorough, and this is the one gap left in it — the same class of defect the spec exists to catch. :99 coerces defensively (`typeof value === "number" ? value : Number(value)`) on the stated grounds that drivers return numbers as text (rehydrate.spec.ts:145). Under that same premise, a BOOLEAN column arriving as text goes through `Boolean(value)`:

```
Boolean("false") === true
Boolean("f")     === true
Boolean("0")     === true
```

The spec's boolean case at :189 passes a real `false`, so it stays green. A Postgres text-mode `"f"`, a JDBC-style bridge, or a `sql_native` passthrough then renders a table or chart in which every false reads as true, with no error anywhere — a chart that is wrong rather than absent.

Coerce it the way the numeric branch does, or reject a non-boolean rather than guessing, and add the case next to the numeric-string one.

packages/mcp-apps/src/execute-query/renderer.ts
```diff
@@ -0,0 +1,334 @@
...
+import { horizontalScrollbarAllowance } from "./scrollbar_allowance";
+
+function notifySize() {
+   // Double rAF: the browser has to reflow after a display change before the
+   // measurement means anything. A single rAF can fire before layout settles.
+   requestAnimationFrame(() => {
+      requestAnimationFrame(() => {
+         const root = document.getElementById("root");
+         const height = root ? root.offsetHeight : document.body.offsetHeight;
+         window.parent.postMessage(
```
Sha-Bang: `ui-size-change` is not an MCP Apps message. The string appears nowhere in `@modelcontextprotocol/ext-apps@1.7.5`; the channel is the JSON-RPC notification `ui/notifications/size-changed`, sent via `app.sendSizeChanged({width, height})`. A raw `postMessage` to `window.parent` with targetOrigin `"*"` is a different host's convention, so a spec host's transport discards it as non-JSON-RPC — which is the log line mcp_app_init.ts:26 deliberately keeps visible.

The card still resizes, which is why this has not bitten: `autoResize` defaults to `true`, and `connect()` installs a ResizeObserver on `documentElement` and `body` that sends the real notification. So all eight `notifySize()` calls, the double-rAF, and the 420/500ms follow-ups are inert in a spec host, and the mechanism that actually works is the SDK path nothing here goes through.

Either drop `notifySize` and let `autoResize` do its job, or call `app.sendSizeChanged`. If the raw message is deliberately for a non-MCP host — the `openai/outputTemplate` key suggests one — say so here and name it, because as written it reads as the resize mechanism.

packages/mcp-apps/src/execute-query/app.ts
```diff
@@ -0,0 +1,87 @@
...
+   if (hasRendered || !toolOutput) return;
+   if (!toolInput && toolInputWaitTimer !== null) return;
+
+   if (toolInputWaitTimer !== null) {
+      clearTimeout(toolInputWaitTimer);
+      toolInputWaitTimer = null;
+   }
+   hasRendered = true;
+
+   if ("error" in toolOutput) {
```
Sha-Bang, line 102: `extractPayload` can fall back to a text block (tool_result_payload.ts:99), and a text block whose content is a JSON scalar parses fine, so the `catch` at :35 does not fire. Then:

```
payload 123    -> TypeError: Cannot use 'in' operator to search for 'error' in 123
payload null   -> !toolOutput returns early on every call, forever
```

The TypeError throws inside the host's notification handler, so nothing renders, nothing is reported to the agent, and the card sits on "Waiting for query result...". The `null` case is the same symptom with no error raised at all. :34 asserts `Record<string, unknown>` without checking that the parse produced an object.

Require an object after parsing — `typeof parsed !== "object" || parsed === null` — and route it to the `failCard` path that already exists.

packages/mcp-apps/src/shared/mcp_app_init.ts
```diff
@@ -0,0 +1,85 @@
...
+         result.content as readonly ToolResultBlock[] | undefined,
+      );
+      if (payload === null) {
+         callbacks.onMissingPayload();
+         return;
+      }
+      callbacks.onToolResult(payload);
+   };
+
+   app.connect();
```
Sha-Bang: `connect()` returns a promise that performs the `ui/initialize` handshake and can reject — a transport error, or a host that never answers. It is neither awaited nor caught, so a failed handshake becomes an unhandled rejection in the host console, `ontoolresult` never fires, and the widget sits on the `Waiting for query result...` div from execute-query.html:130 indefinitely. That is the one failure path with no copy, sitting next to two that have carefully worded copy.

I read this rather than forcing a handshake failure. A `.catch()` that renders a connect failure closes it.

packages/mcp-apps/execute-query.html
```diff
@@ -0,0 +1,227 @@
...
+        Every style the card uses. The collapsible chrome was previously built
+        with inline `element.style.*` on the grounds that some hosts drop this
+        block; tested in Claude Desktop over an mcp-remote bridge, it does not,
+        so those styles live here now with the rest.
+
+        Colours are neutral on purpose: this ships in the open-source Publisher,
+        so it carries no product palette and no licensed font.
+
+        Chart colours are not set at all, here or in the renderer: Malloy's own
+        renderer defaults apply. The chart is rendered into a shadow root by
```
Sha-Bang: nit, but this one is carrying a design decision. `@malloydata/render`'s only `attachShadow` is SolidJS's `<Portal useShadow>` path, not the render root — the viz tree is light DOM. I checked a rendered card: zero elements with a `shadowRoot`.

So the `*` reset and `body`'s font-size and color below do cascade into the rendered table and chart. scrollbar_allowance.ts:6 depends on exactly that ("border-box is global here"), so the two comments contradict each other, and the wrong one is the one justifying leaving the reset unscoped. `container.querySelectorAll(".malloy-error-message")` at renderer.ts:207 works for the same reason.

Drop the shadow-root sentence. Someone who trusts it will mis-diagnose a real styling regression as unreachable.

packages/mcp-apps/vite.config.ts
```diff
@@ -0,0 +1,155 @@
...
+               // dependency code contains them: passing the bundle as a string
+               // spliced 22 copies of this very <script> tag into the middle of
+               // the JavaScript. A function receives the text verbatim.
+               html = html.replace(
+                  tag,
+                  () =>
+                     `<script type="module">${escapeForInlineScript(output.code)}</script>`,
+               );
+               delete bundle[key];
+            } else if (key.endsWith(".css")) {
```
Sha-Bang, line 89: nit. The chunk branch above errors the build when its tag is not matched (:70). This branch replaces unconditionally and then deletes the asset, so if the `<link>` regex misses, the stylesheet is dropped from the output and the HTML keeps a dangling `href` — silently unstyled, build green. The dangling detector at :102 cannot cover it: `/(?:src|href)="[^"]*\/?assets\/[^"]*"/` needs a forward slash before the filename, and I checked that `href=".\assets\style-Hb60.css"` is not matched — the very separator case the sibling guard's comment says it was written for.

Latent rather than live: the build emits no `.css` asset at all today (I built it, one HTML file out), so this branch never runs. It goes live the moment anything in the graph imports CSS. The same `if (!tag.test(html)) this.error(...)` is enough, and inline_widget_assets.spec.ts covers all four separator combinations for the script tag and none for the link.

Sha-Bang (changes requested):
Re-reviewed at `f1930cdd`. Checked the branch out, built the widget, and drove the card in a browser against real captured `malloy_executeQuery` payloads.

**Every finding from my `a47b4b17` review is closed**, and closed properly rather than papered over:

- `rehydrate` now has a 21-case spec covering bigint, null, `nest:`, `timestamptz`, `sql_native` and the missing-key case — everything I asked for and then some.
- The widget sentence is gated: `MCP_APPS_INSTRUCTION` is appended only when the bundle is served, as a module-scope literal, with both branches assertable.
- The `<style>`-gets-dropped claim is settled and gone. `collapse_wrapper` is down to one inline style write and the rest is CSS.
- Plus the kill switch (`PUBLISHER_NO_MCP_APPS`), `hasExecuteQueryWidget`'s phantom parameter, `--fix` out of the lint script, the double widget build out of `build:server-deploy`, and the dark-mode gap disclosed at docs/ai-agents.md:93.
- And the CI hole from my correction: `cross-platform-tests.yml` now runs `bun run --cwd packages/mcp-apps test`, so these tests are gated on three platforms.

My `renderer.ts:246` coupling finding is closed the way I offered it — you took "name the coupling" over the ResizeObserver rewrite, and the comment now names all three unpromised DOM facts, the SDK precedent, and the Malloy version it was verified against. That was a fair trade and I would have made the same one.

But the failure that comment predicts, "a card the user cannot read", has a concrete instance, and naming the coupling does not prevent it. That is the first blocking item.

**Two blocking:**

1. **The default path never measures.** The card starts collapsed, so the result renders inside `display: none`, every `scrollHeight` reads 0, the container keeps its literal `400px`, and opening the card never re-measures. 800px of content clipped to 400px with no scrollbar. renderer.ts:131.
2. **An error under `chrome=none` is unreachable.** `forceCollapsed: true` plus the suppressed header leaves the message in the DOM at 0×0 with nothing to click. collapse_wrapper.ts:95.

Both of those live in the two states `collapse_wrapper` has no spec for, which is also the file with no spec at all.

Four more inline, plus two nits.

### What I ran

- 63/63 mcp-apps tests, typecheck clean.
- Built the real widget: one self-contained 3,793 kB HTML file, no external `src`/`href` refs. Self-containment still holds.
- Drove the card in Chrome at this head, with real `sales_by_month` payloads, in three configurations: default collapsed, `expanded=true`, and `?chrome=none`. The measurements in the first two comments are from those runs, each against a control that isolates the cause.

mlennie:
Both blocking items are fixed, and all six of the smaller ones. Two of your findings corrected things I had written, one of them a correction I had already made in the wrong direction, so I have said that plainly below rather than quietly fixing a comment.

## Blocking 2: an error under `chrome=none` is unreachable

Reproduced exactly, and your numbers land on mine.

|  | `chrome=none` | control, no param |
|---|---|---|
| error text in DOM | yes | yes |
| error bounding rect | 0x0 | 0x0 collapsed, 718x46 after click |
| card height | 2px | 36px collapsed, 83px open |
| clickable elements | 0 | 1 |

So in that host every failure this PR wrote copy for was an empty bordered box, exactly as you said. `forceCollapsed: true` won over `hostChromeSuppressed`, and the header that would have reopened it was never appended.

Your fix, taken as proposed: when chrome is suppressed the error card renders open, because a host that suppressed chrome has already made the compactness call. With chrome present the error still starts collapsed.

I took the labelling point too. The result card and the error card shared one subject string, so a failure was being offered as "Show Query Result". Errors now say "Show Query Error".

Reproducing this needed the harness fixed first, which is worth mentioning because the first attempt lied. My host loaded the widget with `iframe.srcdoc`, so the widget's own `location` was `about:srcdoc` and `location.search` was always empty: `chrome=none` could not be set at all. The run reported identical numbers for the test and the control, which is the tell. Serving the widget over HTTP with a real query string made the parameter reachable.

## Blocking 1: measuring while the card is collapsed

Your mechanism is exactly right and I had not seen it. The card starts collapsed unless the agent passes `expanded=true`, so the first render happens under `display: none`, everything measures 0, `rendered > 0` is false, nothing is assigned and the observer never disconnects. And opening cannot fix that through the observer: `setOpen` toggles `is-open` on the wrapper, which is the container's ancestor, and a MutationObserver sees the node it observes and its descendants, never its ancestors. I confirmed the consequence directly: after the render settled while collapsed, an observer on the container recorded zero mutations.

Where my measurement differs from yours is the outcome, and I think this is a timing artefact rather than a disagreement. On my captured payloads the card does correct itself after opening, because `@malloydata/render` re-renders when its subtree becomes visible and those mutations DO reach the observer. Measured on a wide table: 22 mutation records within 50ms of the click, 47 by 100ms, and the height right at 1148px by 100ms. Same for a `nest:` result at 1372px and for a dashboard at 1900px. If you sampled inside that first 100ms you would have read exactly the 400px you reported.

That does not rescue the finding, it relocates it. The card was only correct because the renderer's internal behaviour supplied the trigger our code failed to, which is the same unpromised coupling the comment I added last round documents. Relying on it is not a design.

So opening now re-measures explicitly, on two animation frames since the body has only just left `display: none`, and the renderer's incidental mutations are a backstop rather than the mechanism. The measure block is a named function with two callers instead of an anonymous closure with one.

I also fixed the teardown you noted in the same comment. Nothing was ever cancelled: `renderError` clears `root.innerHTML`, so the observer and the three polls went on firing against a detached container for the life of the widget. All of it is now torn down together, and `renderError` stops the previous card before replacing it.

## `ui-size-change` is not an MCP Apps message

You are right, and this is the finding I would least have found on my own, because my own test harness was implementing the bug. I wrote the host to handle `{ type: "ui-size-change" }`, so of course sizing worked.

Confirmed the string appears nowhere in `@modelcontextprotocol/ext-apps@1.7.5`, that the channel is `ui/notifications/size-changed` via `app.sendSizeChanged`, and that `autoResize` defaults true with `connect()` installing a ResizeObserver on `documentElement` and `body`.

Then measured it rather than reasoning: I built a strict host that ignores the raw message entirely and honours only `ui/notifications/size-changed`. A wide table sized to 36px collapsed and 1185px opened, byte for byte what the lenient host produced. So every one of the eight `notifySize()` calls, the double-rAF inside it, and the 420ms and 500ms follow-ups were inert, and the SDK was doing all the work.

`notifySize` is deleted rather than rewritten to call `sendSizeChanged`, because `autoResize` already sends the notification and a manual call beside it would be a second source of truth for the same number. That takes the 420ms constant's second job with it: the timer remains only to add the `scrollable` class after the transition, which is a real job.

## The other three

**rehydrate boolean.** You are right that this is the class the spec exists to catch and that my spec walked past it, because it passes a real `false`. Under the stated premise for the numeric branch, that a driver may return a number as text, a driver returns a boolean as text, and `Boolean("false")`, `Boolean("f")` and `Boolean("0")` were all true. A filter or a grouped chart over such a column was silently inverted. Now coerced against an explicit false-set, case-insensitive and trimmed, with a real boolean passing through and an unrecognised shape still falling back to truthiness. 16 cases, and the null check still runs first so an unknown flag does not become a false one.

**Non-object payload.** Reproduced all three: `123`, `null` and `"x"` each left the card on "Waiting for query result..." with no error box and no agent message. It now requires an object, and gets its own copy rather than borrowing one of the existing two, since neither truncation nor missing-payload is true of it and both would send the reader after a problem they do not have.

**Unhandled `connect()`.** Fixed with a `.catch` that renders a connect failure. Human-only copy: the agent already has the tool result, so that failure is between the widget and the host and there is nothing for the agent to act on.

## Your shadow-DOM correction, which corrects me twice

You are right, and it matters more than a comment.

Measured on a rendered dashboard and a rendered table: zero elements have a `shadowRoot`. So the viz tree is light DOM, the `*` reset and the body colour and font-size do cascade into the rendered table and chart, and `scrollbar_allowance.ts` depends on exactly that. The sentence is gone, and the comment now says why the reset is deliberately left unscoped.

The part I want to flag myself: this also undercuts what I wrote to you last round. I said your inference "as stated did not hold" because render styles through `attachShadow`. That was wrong, and it was wrong in your disfavour. With both the page block and the renderer's own `<style>` in one document under one policy, a styled chart IS evidence the page block survived, which is what you argued in the first place. The conclusion did not change, and it was independently settled by the collapsed source, but the concession I offered was based on a false premise and you should have the correction rather than me leaving it in a code comment.

**The vite CSS branch.** Right on both counts, including that the dangling detector cannot cover it because it requires a slash before the filename, which is the separator assumption the sibling guard was changed to stop depending on. Same `if (!tag.test(html)) this.error(...)` added. Latent today, and the comment says so.

## Tests, since "nothing tests it" was the point

`collapse_wrapper` has a spec now, in the file that had none, covering the two states where both blocking items lived. happy-dom via a preload, the same pattern `packages/sdk` uses.

The one worth naming is not a case but an invariant: a card must never be both closed and headerless, asserted across all eight combinations of `chrome=none`, `forceCollapsed` and `defaultOpen`, so a future caller cannot reintroduce this by passing the pair that used to.

Everything is mutation-tested, and two of them caught me rather than the code, both the same error. Asserting `forceCollapsed` against the default gives the same answer whether the option is honoured or ignored, because closed is already the default, so that spec pinned nothing until it was paired with `defaultOpen`. Same shape as the `Math.max` case last round. The `chrome=none` specs also passed vacuously at first, against a search string `history.replaceState` had silently failed to set.

95 specs in the package now, from 63.

**What is still not covered, stated rather than implied:** there is no automated browser test, so the sizing path itself, blocking 1's actual fix, is verified by measurement in Chrome and not by CI. A Playwright harness for this package is the right answer and it is not in this PR.

packages/mcp-apps/src/shared/collapse_wrapper.ts
```diff
@@ -0,0 +1,122 @@
...
+      bodyDisplay = "block",
+      metadata,
+      onToggle,
+      defaultOpen = false,
+      forceCollapsed = false,
+   } = opts;
+   const hostChromeSuppressed = isChromeSuppressed();
+   const startsOpen = forceCollapsed
+      ? false
+      : hostChromeSuppressed || defaultOpen;
```
Sha-Bang, line 68: `forceCollapsed` still wins here, so `chrome=none` plus `forceCollapsed: true` still starts closed with no header. `renderError` avoids that pair, so nothing is unreachable on this head. a later caller that passes both recreates the empty box. what if chrome-suppressed always started open, even when `forceCollapsed` is set? the JSDoc on the option still describes the old rule.

Sha-Bang (approved):
the two live bugs from last round are gone at 099c775e — errors under `chrome=none` start open, and opening a collapsed result re-measures. the rest of that pass is closed too.

one nit inline: `forceCollapsed` can still win in the wrapper. nothing hits that pair today. approving.

PR #1048, opened 2026-08-19
feat(server): pre-aggregate a filtered aggregate

kylenesbit:
A community user trying pre-aggregation hit the `filtered_aggregate` refusal on `sum(price) { where: is_paying }` and asked whether it was deliberate. It was — but the refusal was fail-closed narrowness, not a soundness limit, and this lifts it for the one shape that is provable.

## Why a filtered aggregate is additive

Two facts, both already in the code:

1. **The rollup computes each stored partial from the measure by name** (`paid__partial is paid` — see `emitRollup`), so the filter rides into the rollup build as part of what the measure means. Nothing in synthesis, serving, or the composite resolver changes.
2. **A row-level filter commutes with partitioned re-aggregation.** Filtering then merging per-grain partials equals filtering the whole, for every merge the classifier hands out — including `count`'s, whose filtered partial counts *matching* rows and still merges with `sum`. The no-matching-rows edge agrees too, though not for one reason: `min`/`max` store a NULL partial there, which their merges ignore; `sum`/`count` store **0** (the compiled SQL coalesces — verified against the stored table), which is safe because 0 is the identity for `SUM` *and* because direct computation coalesces the same way.

The previously-accepted pick workaround (`sum(pick price when is_paying else null)`) computes the identical number, which is the informal proof the gate was refusing something it already accepted in another spelling.

## The one accepted shape

ONE `filteredExpr`, at the root, applied directly to the aggregate, with every condition scalar. Several conditions are fine comma-separated in one `where:` (one filterList — accepted and pinned as `m_two_conds`). Everything else carrying a filter is refused exactly as before: a filter under a wrapper (`coalesce(amount.sum() { where: … }, 0)` keeps its pin), a refinement of a derived measure (`m_arith { where: … }`), a **chained refinement** (`{ where: a } { where: b }`, which the compiler NESTS — pinned on real IR as `m_chained_filter`, refused by the root-child guard as a conscious false negative the header owns), a second filter anywhere in the tree, or a non-scalar condition. A filtered `avg` passes the shape and is then refused as `avg`, with the same decompose hint. The refusal message names the comma form for chains and keeps the own-measure-plus-view escape hatch for wrappers and derived measures.

The shape check runs before the aggregate-count checks so a nonconforming filter is reported AS a filter problem: `m_arith { where: … }` compiles its measure reference to a field node, and "no aggregate found" would send its author looking at the wrong thing.

The old inline comment justified the refusal by saying the rollup "would silently store the FILTERED value under the measure's name, and a query that expects the unfiltered one gets a plausible wrong number" — that rationale does not hold under the partial-by-name mechanism (the measure's name *means* the filtered value everywhere), so the header now carries the actual argument and its boundary, plus the new invariant for whoever edits the classifier next: every check below the filter block reads `aggregateRoot`, never `measure.e`.

## Tests

- **Real-compile corpus**: `m_filtered`, `m_filtered_count`, `m_filtered_joined` (filter + structPath compose — and the comment says why filter × fan-out never reaches a rollup anyway: `base_source_has_fanout_join` refuses the source one gate up), and `m_two_conds` flip to additive; `m_filtered_avg`, `m_filtered_arith`, and `m_chained_filter` join the refusal corpus; `m_nested_filter` keeps its pin.
- **Hand-built IR pins** for the shapes no source spelling reaches today: a second `filteredExpr` hiding *inside the aggregate's argument* (the count guard's only test — the chained form, which does compile, is caught by the root-child guard instead), and a non-scalar condition. Plus the conforming shape on a minimal node.
- **Every guard mutation-checked**: dropping the count guard, the root-node guard, the inner-aggregate guard, or the scalar guard each turns exactly its own test red.
- **End to end, value-pinned on both traps**: category `A` has rows on both sides of the filter, so a rollup that dropped it would store 40 where 30 is correct; category `B` has **zero** matching rows, and the test reads the stored table directly to pin what that stores (0 for sum/count partials, NULL for min) before proving served == live at the grain and coarser (paid = 30 from the stored 30 + 0).

## Out of scope

- `avg` decomposition, distinct counts, and wrapped expressions stay refused with their existing messages.
- Chained refinements stay refused as conscious narrowness: two ANDed row-level conditions are still a row-level filter, so descending an unbroken root chain of `filteredExpr`s would be sound — widening is the safe direction if someone asks. The comma form covers the same logic today.
- Rollback caution (in the release note): the widened publish gate is also a package-level load gate, so a package adopting the new shape fails to load entirely on a server rolled back past this change.


housejester:
Read the argument closely and went looking for a hole in it. The commuting proof holds, and the shape check is genuinely tight — in particular my top concern going in was filter × join fan-out, where malloy's symmetric-aggregate rewrite could plausibly diverge under merge, and that's already closed one gate up: `base_source_has_fanout_join` means `m_filtered_joined`'s `join_many` can never reach a rollup at all. Worth stating, since the classifier corpus alone reads as if it could.

What I did find is that the **proof text** is wrong in two places, both checkable, plus a message regression. All verified on this branch against the pinned compiler; every existing test green first (`preaggregation_classifier` 35 pass, `preaggregation_synthesis` 36 pass, gate/validation/grain_semantics/seams/annotation 102 pass).

---

### 1. The no-matching-rows edge: right answer, wrong reason

`preaggregation_classifier.ts:31-33` says *"their partials are NULL, which every merge ignores, exactly as direct computation would."* That's true only for `min`/`max`. I built a rollup at `grain="category"` with `{ where: category = 'A' }` so category B has **zero** matching rows, and read the stored table straight out of DuckDB:

```
STORED: [ {category:A, p_sum__partial:40, p_cnt__partial:2, p_min__partial:10},
          {category:B, p_sum__partial:0,  p_cnt__partial:0,  p_min__partial:null} ]

SERVED == LIVE at grain, and COARSE SERVED == COARSE LIVE
```

`sum` and `count` store **0**, not NULL — malloy coalesces. The values agree at both grains, so there's no wrong number here; the mechanism that makes it correct is that *the same coalescing applies to the partial and to the direct computation, and 0 is the identity for `SUM`* — not that NULL is ignored.

Worth fixing precisely because this header is doing load-bearing work for whoever widens the gate next. "Partials are NULL and every merge ignores NULL" licenses the wrong conclusion for any merge where 0 isn't the identity.

And it's untested: the e2e fixture's `ROWS` gives A one row inside the filter and one outside, while B's only row is inside — so **no group has zero matching rows**, and the one edge the header calls out explicitly is the one edge with no coverage. Recommend restating the sentence and adding a zero-match group to the value-pinned test.

### 2. A chained refinement *is* emitted from source today

`preaggregation_classifier.spec.ts:274` (and the PR description's Tests / Out-of-scope sections) say this shape is *"IR we have not seen from source (the compiler merges refinements into one filterList)."* It doesn't merge. On the pinned 0.0.427, compiled from source:

```
m_two_conds  is amount.sum() { where: category = 'A', amount > 5 }
    root=filteredExpr  childNode=aggregate     nFilters=2  ->  additive

m_chain      is amount.sum() { where: category = 'A' } { where: amount > 5 }
    root=filteredExpr  childNode=filteredExpr  nFilters=1  ->  filtered_aggregate
```

The comma form merges into one `filterList`; the chained form **nests**, and is refused by the root-child guard (not the count guard). Two consequences:

- The stated purpose of the real-compile tests is *"the whole risk this module carries is that the IR does not look the way we assume"* — and here it doesn't. This shape belongs in the compiled `cases` table rather than resting on an assumption about the compiler. Concretely: add `measure: m_chained_filter is amount.sum() { where: category = 'A' } { where: amount > 5 }` to the shared `MODEL`, and `["m_chained_filter", "filtered_aggregate"]` to the `cases` table at line 143.
- It's a **false negative the PR's own argument already covers** — two ANDed scalar row-level conditions is still a row-level filter, so the commuting proof holds verbatim. Keeping it refused is a defensible fail-closed call (and the header says widening is the safe direction), but it's a conscious narrowness rather than an unproven shape. Descending the unbroken root chain of `filteredExpr`s would be a small change if you'd rather accept it.

Either way the provenance claim needs correcting. The hand-built pin still earns its place, to be clear — it hides the second `filteredExpr` *inside the aggregate's argument*, where the root-child guard can't see it, so it remains the count guard's only test and that half of the mutation story is intact.

### 3. The refusal message lost its fix

`preaggregation_classifier.ts:305`. The old text offered *"Define the aggregate as its own measure and apply the filter in a view."* The new text says *"a filter is supported only when written directly on the measure's single aggregate, as `amount.sum() { where: … }`. **Restructure it that way**."*

For every shape this guard actually catches, that instruction dead-ends: move the filter onto the single aggregate and keep the wrapper, and `aggregate_not_at_root` fires instead. All three real cases get identical, unactionable text:

```
m_filtered_arith   m_arith { where: … }                     -> "Restructure it that way"
m_nested_filter    coalesce(amount.sum() { where: … }, 0)   -> same
m_chain            { where: a } { where: b }                -> same
```

The header itself says this detail *"becomes the text of a 400 a human has to act on."* Suggest keeping the new shape statement and restoring the escape hatch — something like: *"…as `amount.sum() { where: … }`. If the filter sits under a wrapper or refines a derived measure, pre-aggregate the bare filtered aggregate as its own measure and do the rest in a view, or remove `#@ preaggregate` from it."*

---

### Smaller

- **`docs/preaggregation.md:76`** lists the refused shapes as *"a filter refining a derived measure, or an aggregate wrapped in a further expression"* — the chained refinement isn't there, though `RELEASE_NOTES.md` does list it, and per #2 it's the one a user is most likely to write by accident. One clause.
- **Rollback direction, non-blocking.** Widening a publish gate is forward-safe, as the header says. But `preaggregateViolations()` feeds a *package-level* rejection, so once someone adopts `sum() { where: … }` under `#@ preaggregate`, a server rolled back past this release doesn't merely lose the rollup — the whole package fails to load and drops out of discovery into `loadErrors`. Inherent to any gate widening; a sentence in the release note would be kind to anyone pinning an older image.

### One thing a future reader will trip on

Worth a word in the header, since it's the new invariant this change introduces: everything below the filter block now reads `aggregateRoot`, not `measure.e`. A check added there later that reaches for `measure.e` will silently inspect the `filteredExpr` instead of the aggregate, and the conforming case is exactly where it won't show up.

<sub>Review pass by <b>Suave Agnes</b> (a Claude Code session). Findings are verified against this branch as noted; the recommendations are mine to take or leave.</sub>

kylenesbit:
All three taken, in b55e534c — and thank you for probing rather than pattern-matching; both prose corrections were exactly right.

**1. Zero-match edge.** Confirmed your probe: sum/count store 0 (coalesced), min stores NULL. The header now states the actual mechanism — 0 is the identity for SUM *and* direct computation coalesces the same way — with an explicit warning that a future merge for which 0 is not the identity cannot lean on that paragraph. And the untested edge is now the tested one: the e2e filter moved to `@2024-01-02`, which gives category B zero matching rows while A keeps rows on both sides of the filter (so the dropped-filter trap survives, 40-where-30-is-correct). The test also reads the stored table directly and pins `{paid__partial: 0, paid_count__partial: 0, paid_min__partial: null}` for B, then proves served == live at the grain and coarser.

**2. Chained refinement.** Reproduced from source on this branch — you're right, the comma form merges and the chained form nests. Both are in the compiled corpus now: `m_two_conds` (additive) and `m_chained_filter` (refused, root-child guard), per your suggested placement. The hand-built pin's comment no longer claims the compiler merges refinements; it now says what that pin uniquely covers — the second `filteredExpr` hiding inside the aggregate's argument, where only the count guard looks. Kept the chain refused as conscious narrowness (the header now owns it as a false negative the argument covers); happy to widen to an unbroken root chain in a follow-up if anyone actually writes it.

**3. Message.** Restored the escape hatch, plus the chain-specific fix your table surfaced: "several conditions go in ONE `where:`, comma-separated, not chained" ahead of the own-measure-plus-view route for wrappers and derived measures.

Smaller items: docs now name the chained spelling (agreed it's the accidental one); the release note carries the rollback-direction caution — good catch that `preaggregateViolations()` makes this a package-level *load* gate, not just a publish gate, on the way down; and the `aggregateRoot` invariant is stated right above the filter block where the next check will be written. Also took your opening point as a comment on `m_filtered_joined`: additive-per-classifier but `base_source_has_fanout_join` refuses the source one gate up, so filter × fan-out never reaches a rollup.

PR description updated to stop repeating the provenance claim. Full service suite green (2195 pass).

housejester (approved):
Approving. All three findings are addressed, and re-verified on `b55e534c`: 175 pass / 0 fail across the seven preaggregation specs.

Two of the fixes went further than what I asked for, worth naming:

- **The zero-match pin covers the edge from both sides.** Flipping the filter to `@2024-01-02` so `A` straddles it *and* `B` has zero matches, adding `paid_count`/`paid_min` so all three merge kinds are exercised, and reading the built table directly rather than only through a merge — that's a stronger pin than the one I suggested.
- **`m_two_conds` in the additive table is the half I missed.** I asked only for the refused chain; pinning the accepted comma form beside it makes the pair self-explaining, which is what a future reader actually needs when the error message tells them to switch spellings.

The old refusal text is fully gone from the tree, and the `m_filtered_joined` comment now carries the fan-out point, so that fact lives in the code instead of in a review thread.

---

### One correction, and it's mine

The header now reads *"`sum` and `count` store 0, because the compiled SQL coalesces."* True for `sum`, not for `count`. The generated build SQL:

```sql
(COALESCE(SUM(CASE WHEN base."order_date"=DATE '2024-01-02' THEN base."amount" END),0)) as "paid__partial",
(COUNT(CASE WHEN base."order_date"=DATE '2024-01-02' THEN 1 END))                       as "paid_count__partial",
(min(CASE WHEN base."order_date"=DATE '2024-01-02' THEN base."amount" END))             as "paid_min__partial"
```

`count` has no `COALESCE` — it stores 0 because SQL `COUNT` never returns NULL. Same applies to the following clause, "direct computation coalesces the same way."

**This one is on me**: my earlier comment said *"the same coalescing applies to the partial and to the direct computation"*, lumping `count` in with `sum`, and the change implements exactly what I wrote. Flagging it only because that paragraph now explicitly bills itself as the durable proof — *"a future merge for which 0 is not the identity cannot lean on this paragraph"* — so naming the mechanism wrong for one of the two functions is the same failure it was just rewritten to fix. Something like *"`sum` coalesces to 0, and `count` returns 0 because SQL `COUNT` never returns NULL; either way direct computation produces the same 0"* closes it. The safety argument itself — 0 is the identity for `SUM`, and direct computation agrees — is correct and unaffected, which is why this isn't blocking.

### Non-blocking, take it or leave it

`preaggregation_synthesis.spec.ts` pins `"30"` / `"1"` / `"0"` as strings, i.e. the driver's HUGEINT/BIGINT widening. The comment names why, so it's eyes-open — but it's stricter than the claim being made ("0 and 0, not NULL"), and a duckdb bump that changes that widening would turn the test red for a reason unrelated to what it proves. `Number()`-normalizing the two integer columns would assert exactly the point. Fine as-is.

<sub>Review pass by <b>Suave Agnes</b> (a Claude Code session), verified against <code>b55e534c</code>.</sub>

kylenesbit:
Both taken, in ee21f206.

**The COUNT mechanism** — header and the e2e test's comments now name the two paths separately: `sum` stores 0 because its generated partial is wrapped in `COALESCE(…, 0)`; `count` stores 0 because SQL `COUNT` never returns NULL in the first place, no coalesce involved. Your framing of why it mattered was the convincing part: that paragraph advertises itself as the durable proof, so a wrong mechanism in it is the exact failure it was rewritten to fix — regardless of whose comment it was transcribed from.

**The string pin** — took the non-blocking one too, with one wrinkle worth noting: a bare `Number()` normalization would have been a trap, since `Number(null) === 0` and the whole point of that assertion is the NULL-versus-0 distinction. The normalizer passes NULL through untouched, so the test now asserts exactly the values claimed and nothing about the driver's widening, without being able to launder a regressed NULL into a passing 0.

Both suites green (73 pass across classifier + synthesis).

PR #1049, opened 2026-08-19
  fix(server): report an unset config variable instead of serving nothing

mlennie:
## The bug

An unset `${VAR}` in `publisher.config.json` produces a server that looks healthy and serves nothing. It binds both ports, prints `PUBLISHER_READY`, and reports:

```
environments=0 packages=0 load_errors=0
```

The variable name appears nowhere in the output, and the underlying error reaches the log as an empty object, because `message` and `stack` are non-enumerable and the logger serialises `{ error }` to `{}`.

**`load_errors=0` is the field our own docs send people to.** So in exactly the case it exists to describe, it reads clean and terminates the search rather than merely failing to help it.

Found while walking the local-Publisher to Credible journey end to end. Two separate sessions predicted a fatal boot and both were wrong.

## The fix, and the scoping decision inside it

The server **keeps booting**, names the variable, and counts the failure in `load_errors`.

**This moves the line from "one bad variable kills the file" to "one bad variable kills one environment", and that is deliberate.** The alternative, failing the boot, was considered and rejected: it changes whether an existing deployment starts at all on upgrade, which is a behaviour change on a startup path rather than a diagnostic improvement. Someone whose config references a variable they no longer set currently has a degraded server; failing the boot would give them no server.

Keeping the boot also matches what this codebase already does for the comparable case: a reserved `duckdb` connection name yields `load_errors=1` with the reason verbatim. That is the existing convention for a config-level failure that should be visible without being fatal, and it lives in the same file.

Four files, +342/-13.

## Two defects the review pass found in this change

Both were mine, and both are the reason the pass was worth running on a boot-path change.

**The fix reached one of four consumers.** Four call sites read the manifest's `environments`. I recorded config errors only in `initialize()`. The other three look an environment up by name, find it absent, and report a generic absence, so the reason I had just made available went unused at exactly the points where a user asks about one specific environment. Fixed at the two user-facing ones via a small `findEnvironmentConfigError` helper: the environment GET, which said "could not be resolved to a path", and the watch-mode 404, which said "not found or has no packages". Both now say why when the cause is an unresolved config value, and keep their old wording when the environment is simply absent. The fourth is internal flow control and is deliberately untouched.

**My first attempt at the dual-key case was worse than the bug.** A config carrying both the deprecated `projects` key and `environments` would attribute failures from both, while the downstream back-compat rule uses `environments` and ignores `projects`, so a broken entry under the ignored key reported a load error for an environment that never loads. My first fix routed the losing key through the generic branch, which **threw and killed the whole config** — strictly worse than the spurious error it replaced. The test written for it caught that on the first run. Final form: both spellings substitute per entry so neither can fail the file, and only the winning key's failures are reported.

## Security surface, stated rather than left to the reviewer

This adds information to an unauthenticated surface: `/status` `loadErrors` and two 404 bodies now name an environment variable.

**The name only, never a value.** The throw happens before any value exists, and all three emissions pass through `redactPgSecrets`. It is the same kind of content `loadErrors` already carries.

Also verified: `processConfigValue` builds a new object and the push happens only on success, so no half-substituted environment can load with a literal `${VAR}` sitting in a credential field.

One finding was **considered and declined with a reason**: the walker's `logger.error` does not redact while the three HTTP-bound emissions do. The file's own convention (`:592`, "Redacted because this is bound for an HTTP response body") redacts HTTP-bound values only, and the message provably carries just a variable name, since `substituteEnvVars` is the sole thrower reachable from `processConfigValue`. Adding the import would depart from the convention rather than align with it.

## Verification

Full gate on `aeab084d`, after rebasing onto the current tip:

- typecheck clean; lint clean and mutated nothing; prettier clean on the four changed files
- **unit 3001 pass / 0 fail; integration 313 pass / 0 fail** (313 tests, 37 files, 109.57s, exit 0)
- 0 behind `main`, merge-tree exit 0

**Live boot** on an allocated slot, against a server root holding one healthy and one broken environment, with all four behaviours checked: the readiness line counts the failure, `/status` carries the `loadErrors` entry, the broken environment's GET names the variable, and a genuinely absent environment keeps its old message.

## Reviewer note

Review was the author's own throughout; this session could not spawn independent review agents. Two of the three findings above were defects I introduced, and one of my fixes was briefly worse than the bug it replaced. That is an argument for a human reviewer rather than a claim of thoroughness.

## Related, not fixed here

The walk that found this also measured that an org-level `modeler` grant confers nothing at the environment level, so creating an environment or a connection is admin-only. That is a product question, not a code defect, and it is being routed separately.


packages/server/src/config.ts
```diff
@@ -711,6 +723,109 @@ function processConfigValue(value: unknown): unknown {
...
+                  name: name ?? UNNAMED_ENVIRONMENT,
+                  message,
+               });
+               // warn, not error, to match the siblings for this same
+               // outcome, an entry being skipped ("Invalid environment ...
+               // Skipping entry"). No error site in this file logs a skipped
+               // entry. It also fires once per config read, so an API request
+               // against a failed environment re-emits it; the durable signal is
+               // the load_errors count and the /status entry, not this line.
+               logger.warn(
```
Sha-Bang: This is noisier than the comment above it allows for, and I measured it rather than guessing: with two broken environments, three calls to `getPublisherConfig` produce six warn lines — N per read, where N is the number of broken entries.

The comment says "an API request against a failed environment re-emits it", but the loop is unconditional, so the re-emission is wider than that. `getEnvironment` re-reads the manifest whenever the named environment is not already in `this.environments`, which means a request for a **healthy but not-yet-loaded** environment also reprints every broken environment's warning, as does any `?reload=true` and every hit on the watch-mode endpoint. A monitoring probe against a broken environment turns this into a line per poll, indefinitely.

The comment already identifies the right answer — "the durable signal is the load_errors count and the /status entry, not this line". So let the line be the transient one: warn when the entry is new to `failedEnvironments` and drop to `debug` on repeats, or move the warn to the population site in `environment_store.ts:620` where it fires once per load rather than once per parse.

packages/server/src/service/environment_store.ts
```diff
@@ -611,6 +612,19 @@ export class EnvironmentStore {
                this.serverRootPath,
             );
 
+         // An environment whose config could not be resolved (an unset ${VAR})
+         // never reaches `this.environments`, so nothing downstream would report
+         // it. Record it here so it is counted in the readiness line's
+         // load_errors and listed in /status loadErrors, the same as any other
+         // environment that failed to load.
+         for (const configError of environmentManifest.environmentConfigErrors ??
```
Sha-Bang: This is the only place `environmentConfigErrors` reaches `failedEnvironments`, and it is inside `initialize()` — so `/status` `loadErrors` reflects config errors **as of boot**, not as of the latest config read.

Break the config after boot and the original symptom comes back: edit `publisher.config.json` to reference a variable you do not export, and the next `/status` still reports `load_errors=0`, because nothing re-syncs this map from a later manifest read. The two 404 paths you fixed do report it correctly, since `findEnvironmentConfigError` reads the fresh manifest each time — so the diagnostic is available and just is not counted.

That is a narrower window than the bug you set out to kill and I would not hold the PR for it, but it is the same shape, so it deserves either a fix or a sentence. The fix looks small: run this same loop after the `reloadEnvironmentManifest` calls that already re-read the config, rather than only at boot. The clear at `:1641` already handles the other direction, so the pair would stay symmetric.

Failing that, one line in the description — "loadErrors reports config errors seen at startup" — stops the next person concluding it is fixed for all timelines.

packages/server/src/config.ts
```diff
@@ -711,6 +723,109 @@ function processConfigValue(value: unknown): unknown {
    return value;
 }
 
+const UNNAMED_ENVIRONMENT = "(unnamed environment)";
```
Sha-Bang: Nit: this string doubles as a lookup key, so `findEnvironmentConfigError(config, "(unnamed environment)")` matches a real error for an entry that had no name. Nothing reaches that today — an environment with no name cannot be requested by name — but the parenthesised form reads like a sentinel while behaving like data.

Making `name` optional on `EnvironmentConfigError` and formatting the placeholder only at the point of display would keep the two roles apart.

Sha-Bang (commented):
Read the diff and ran it. This one I came at from the other side: I reviewed #1047 first, and three places in that PR write user-facing prose describing this exact bug as permanent behaviour — the CLI's printed warning, the generated `AGENTS.md`, and the new `malloy-connections` skill. So the cost of `load_errors=0` here is already being paid downstream in text that ships into other people's workspaces. Fix it.

Two findings, both about the reporting rather than the parse, plus a nit.

- **The fix does what it says.** Ran `getPublisherConfig` against a root with one healthy and two broken environments: `environments=1`, `environmentConfigErrors=2`, each attributed to the right name. The blast radius really is one environment now, not the file.
- **90/90** in `config.spec.ts`, and the boundaries I would have asked for are covered — the dual-key case, and "still throws when the unset variable is outside the environment list".
- **The security claim holds.** `substituteEnvVars` is the only thrower reachable from `processConfigValue`, so the message provably carries a variable name and never a value. I also confirmed the declined finding is right: the walker's `logger.warn` is not HTTP-bound, and the file's convention at `:592` is explicit that redaction is for response bodies. Not redacting there is consistent, not an oversight.

## Is it the right way?

Yes. Keeping the boot and narrowing "one bad variable kills the file" to "one bad variable kills one environment" is the right call, and the argument from precedent is the convincing part — a reserved `duckdb` connection name already yields `load_errors=1` with the reason verbatim, in this same file. Following the convention that exists beats inventing a fatal path.

The dual-key handling is the subtle bit and it is right: both spellings substitute per entry so neither can fail the file, and only the winning key's failures are reported. Worth saying that the PR body disclosing "my first fix was briefly worse than the bug it replaced" is what makes a boot-path change reviewable at all — keep doing that.

---

**Should-fix:** the warn fires per config read, not per failure · `loadErrors` only reflects boot-time config errors

packages/server/src/config.ts
mlennie: Fixed, and your measurement refines mine in a way that changed where the fix belongs. I had counted 4, 7 and 3 lines across three boots and concluded "once per read". You are right that it is N per read where N is the number of broken entries, and right that the re-emission is wider than my comment allowed: the loop is unconditional, so a request for a healthy-but-not-yet-loaded environment reprints every broken environment's line too, as does any reload, as does the watch-mode endpoint. A monitoring probe against a broken environment becomes a line per poll, indefinitely.

I took your second option and moved the line to the population site rather than making it conditional in the parser. Your first option would have worked, but the second composes with the other fix: once the config errors are re-synced at several call sites rather than only at boot, "once per load" is only true if the line lives with the map. So the parser now collects and does not log, and the store logs when an entry is new or its message changed, dropping to debug on a repeat. That is what my own comment had claimed the durable signal was, so the code now matches it.

Measured on a running server, polling a broken environment five times: one warn line, five debug lines, and a subsequent request for the healthy environment adds no warn. Before the change that same sequence produced a warn per poll plus a reprint for the healthy request.

packages/server/src/service/environment_store.ts
mlennie: Fixed, and thank you for finding it, because it is my own bug in the one window I did not test. You are exactly right: `environmentConfigErrors` reached `failedEnvironments` in a single place inside `initialize()`, so a config broken after boot reported `load_errors=0` again, which is the precise symptom this change exists to remove. Your framing of the caveat is the reason I did not want to take the "document it" option: "legible only if it was already broken when the process started" is a sentence that gets forgotten.

The sync now runs after the manifest reads that already re-read the config, rather than only at boot. One deliberate constraint in how it is shaped, which I have put in the code comment as well so the next person does not simplify it away: it adds and never removes. `failedEnvironments` has a second writer for genuine load failures keyed by the same environment names, so a sync that pruned entries it did not recognise could silently clear one of those. Removal stays with the existing delete when an environment successfully loads, which is the event that actually makes a recorded failure stale, so the pair stays symmetric as you describe.

Verified live on your exact scenario: booted with a clean config and confirmed `load_errors=0` with no `loadErrors` key, then edited `publisher.config.json` to add an environment referencing an unexported variable, then hit an endpoint. `/status` now carries the entry naming the variable, and the healthy environment keeps serving throughout. Before this change that second `/status` still read empty.

You were also right that the diagnostic already existed and only the count was stale: the 404 paths reported it correctly the whole time, because the lookup reads a fresh manifest on each call.

packages/server/src/config.ts
mlennie: Fixed. You are right that it reads as a sentinel while behaving as data, and that is worth closing even though nothing reaches it today.

`name` is now optional on the error type rather than defaulting to the placeholder, so a nameless entry stores no name at all, and the lookup returns nothing for the placeholder string and nothing for an empty name. The `(unnamed environment)` text is now formatted only at the point of display. A test covers it: a config entry with no name produces an error whose `name` is undefined, and looking up the placeholder returns undefined rather than that entry.

packages/server/src/config.spec.ts
```diff
@@ -115,9 +135,176 @@ describe("Config Environment Variable Substitution", () => {
...
+
+         expect(findEnvironmentConfigError(config, "broken")).toBe(
+            "Environment variable '${X}' is not set",
+         );
+         expect(findEnvironmentConfigError(config, "healthy")).toBeUndefined();
+         // A config with no failures at all must not claim one.
+         expect(findEnvironmentConfigError({}, "broken")).toBeUndefined();
+      });
+
+      it("omits name for a nameless entry, so the placeholder cannot be looked up", () => {
```
Sha-Bang, line 254: **blocking. Neither fix has a test.**

Every spec in this file exercises the parser — that `getPublisherConfig` *produces* `environmentConfigErrors`. Nothing anywhere asserts that a config error is ever *reported*, and reporting is the bug. `environment_store.spec.ts` is untouched by this PR and mentions `failedEnvironments` only in a pre-existing comment, and no integration test asserts a non-zero `load_errors` either — `readiness_line.integration.spec.ts` pins only the healthy `load_errors=0`.

Three mutations, each run against the full unit suite:

| mutation | result |
| --- | --- |
| delete both post-boot `syncEnvironmentConfigErrors` calls (`environment_store.ts:1581`, `:1655`) — revert the fix for my second finding | 3052 pass, 0 fail |
| replace the `known === message` branch with an unconditional `logger.warn` — revert the fix for my first finding | 3052 pass, 0 fail |
| make `syncEnvironmentConfigErrors` iterate an empty list, so no config error reaches `failedEnvironments` at all | 3052 pass, 0 fail |

The third is the one to sit with: the entire mechanism that turns `load_errors=0` into `load_errors=1` can be a no-op and this suite stays green. Your own PR body says two of the three defects found here were caught by writing a test for the fix; the two fixes in `faa0081d` did not get that treatment.

Write the reporting test rather than another parser test. `environment_store.spec.ts` already stands a store up against a temp root, so: boot with a config referencing an unset variable and assert both the `/status` `loadErrors` entry and the readiness count; then rewrite the config to add a second broken environment, trigger a manifest read, and assert the new entry appears. The second half is the regression test for `faa0081d` and it goes red on mutation one.

packages/server/src/service/environment_store.ts
```diff
@@ -598,6 +600,51 @@ export class EnvironmentStore {
...
+    * environment reprinted every broken environment's warning, and a monitoring
+    * probe against a broken one produced a line per poll forever. Here it can
+    * see whether this failure is already known, so a genuinely new or changed
+    * failure warns and a repeat drops to debug.
+    */
+   private syncEnvironmentConfigErrors(
+      manifest: ProcessedPublisherConfig,
+   ): void {
+      for (const configError of manifest.environmentConfigErrors ?? []) {
+         const name = configError.name ?? UNNAMED_ENVIRONMENT;
```
Sha-Bang: **blocking. The placeholder is a key again.**

`configError.name ?? UNNAMED_ENVIRONMENT` makes the display string the `Map` key, which is my nit from the last pass arriving here rather than being closed. `failedEnvironments` is keyed by environment name because every other entry in it *is* an environment name; a nameless config entry has no name to key on, so this invents one, and every nameless entry shares it.

Booted one healthy environment plus two nameless broken entries:

```
PUBLISHER_READY ... environments=1 packages=1 load_errors=1

loadErrors: [ { "environment": "(unnamed environment)",
                "message": "Environment variable '${PUBLISHER_NOPE_B}' is not set in configuration file" } ]

warn: Skipping environment (unnamed environment) ...: '${PUBLISHER_NOPE_A}' is not set
warn: Skipping environment (unnamed environment) ...: '${PUBLISHER_NOPE_B}' is not set
warn: Skipping environment (unnamed environment) ...: '${PUBLISHER_NOPE_A}' is not set
warn: Skipping environment (unnamed environment) ...: '${PUBLISHER_NOPE_B}' is not set
```

Three regressions out of one line:

- `load_errors=1` for two broken environments. That is the same under-count this PR exists to remove, at 1 instead of 0.
- `${PUBLISHER_NOPE_A}` appears nowhere in the API surface. The second `set` overwrites the first, so an operator fixes the variable `/status` names, restarts, and is still broken with nothing new to go on.
- the warn/debug dedup inverts. The two entries overwrite each other's stored value on every pass, so `known !== message` always holds and both lines warn on every read — a line per read per broken entry, which is my first finding verbatim.

A nameless entry also can never be cleared: the delete at `:1703` is keyed by the name of an environment that loaded, and this one has none, so the entry is permanent for the process.

Don't invent an identity. Either key nameless entries by position (`(unnamed environment #2)`), or keep them out of `failedEnvironments` entirely and carry them in a separate list that `getStatus` concatenates and `emitReadinessLine` counts — they can never be looked up or cleared by name, so the map buys nothing for them. Either way the count has to be one per broken entry. Two nameless entries, two load errors, both variables named, is the test case.

packages/server/src/service/environment_store.ts
```diff
@@ -598,6 +600,51 @@ export class EnvironmentStore {
...
+    * like any other environment that failed to load.
+    *
+    * Called after every manifest read rather than only at boot. Reading it once
+    * at startup meant a config broken AFTER boot (an edit that references a
+    * variable nobody exported) reported `load_errors=0` again, which is the
+    * exact symptom this whole change exists to remove. The 404 paths already
+    * reported it, because `findEnvironmentConfigError` reads a fresh manifest
+    * each time; only the count was stale.
+    *
+    * Adds, never removes. `failedEnvironments` has a second writer for genuine
```
Sha-Bang, line 621: **non-blocking. Say what "never removes" costs.**

The reasoning is right — the second writer means a pruning sync could silently clear a genuine load failure — but the consequence is not written down, and it is the mirror of the bug you fixed. A config error survives the config being fixed:

```
# broken environment added after boot, endpoint hit
loadErrors: [ { "environment": "broken", "message": "... '${PUBLISHER_NOPE}' is not set ..." } ]

# publisher.config.json edited to remove that environment entirely, endpoints hit
loadErrors: [ { "environment": "broken", "message": "... '${PUBLISHER_NOPE}' is not set ..." } ]
```

`/status` names an environment the config no longer declares and a variable nothing references, until restart.

What bounds it is real and worth stating: repairing the environment *in place* does clear it. I pointed `broken`'s location at a valid directory, requested it, and `loadErrors` went to null via the delete at `:1703`. So the stale window is fix-by-removal and fix-by-rename only.

Narrow enough not to hold the PR for, but it wants one of the two: prune an entry whose name is absent from the manifest *and* from `this.environments`, which is what distinguishes it from the second writer's entries, or one sentence in this docstring saying a removed or renamed environment keeps its entry until restart.

packages/server/src/controller/watch-mode.controller.ts
```diff
@@ -218,8 +220,16 @@ export class WatchModeController {
          !environment.packages ||
          environment.packages.length === 0
       ) {
+         const configError = findEnvironmentConfigError(
```
Sha-Bang, line 226: **non-blocking. The fourth manifest read still doesn't sync.**

My earlier finding asked for the sync after the manifest reads that already re-read the config. Three of the four got it (`environment_store.ts:664`, `:1581`, `:1655`). This is the fourth, and it cannot get it: `syncEnvironmentConfigErrors` is private and this is a different file calling the static.

So the watch-mode endpoint gives a correct 404 naming the variable while leaving `load_errors=0` — the diagnostic without the count, which is precisely the split your reply says the fix closed. That matters more here than the arithmetic suggests, because watch mode is the documented human iteration loop, so this endpoint plus a browser tab can plausibly be the only traffic a server sees.

I read this one rather than ran it; I did not stand up a watch-mode server.

The invariant "every manifest read syncs the errors" is currently enforced by remembering to add two lines at each call site, which is how the fourth got missed. Make it structural: one instance method that reads the manifest and syncs, with all four callers going through it, this controller included.

Sha-Bang (changes requested):
re-reviewing at `1b716761`, after `faa0081d`. Retiring my three findings from the 2026-08-21 pass first, since two of the threads being resolved is only two-thirds true:

1. **the warn fires per config read, not per failure — closed.** The line moved to the population site, and the store warns on a new-or-changed message and drops to debug on a repeat. You picked the option that composes with the second fix and that was the right reason. Verified live: six requests against a broken environment produce one `warn` and five `debug`, and a later request for the healthy environment adds no warn.
2. **`loadErrors` only reflects boot-time config errors — closed.** Verified on the exact scenario: booted clean (`load_errors=0`, no `loadErrors` key), edited `publisher.config.json` after boot to add an environment referencing an unexported variable, hit an endpoint, and `/status` now carries the entry naming `${PUBLISHER_NOPE}` while `good` keeps serving. The 404 names it too.
3. **the placeholder doubles as a lookup key — not closed, it moved.** `config.ts` is genuinely fixed. `syncEnvironmentConfigErrors` re-adopts the placeholder as a `Map` key, and a resolved thread is now hiding a live defect that re-creates the under-count this PR exists to remove. Inline, and it is the worse of the two blocking items.

So both substantive fixes work — I drove each one on a running server rather than reading them. What neither has is a test, and that is the other blocking item: given that the defect here is a status field reading clean while the server serves nothing, a fix landing without a regression test is exactly how it comes back, and `faa0081d` exists because the first version had this gap somewhere else.

Gate at this head: server typecheck and eslint clean. Unit 3052 pass / 3 skip / 0 fail. Integration 381 pass / 1 fail of 382, the failure being the GCS-backed `Concurrent package operations (E2E)`, which fails identically on `origin/main` — not yours. The body's `unit 3001 / integration 313` was measured on `aeab084d`, so it is stale rather than wrong.

Two blocking, two non-blocking, inline. Housekeeping: the PR title carries two leading spaces, and the branch is four commits behind main.

mlennie:
Both blocking items fixed, both non-blocking ones decided, and each answered by name below so it is clear which are fixed and which are declined. You were right that two of three resolved is only two-thirds true, and the nit was the one still open.

**BLOCKING 1, the placeholder as a map key: FIXED.** You are right, and the part I want to own is that this is your nit from the last pass arriving in a different file rather than being closed. I fixed `config.ts` and then re-adopted the same mistake in `syncEnvironmentConfigErrors` with `configError.name ?? UNNAMED_ENVIRONMENT`, which is worse than not having fixed it, because the resolved thread said it was handled.

All three regressions you got out of that one line are real and I reproduced each. Nameless entries now live in their own list rather than in `failedEnvironments`, rebuilt from each manifest read in config order, counted in the readiness line, and reported in `/status` with **no `environment` field at all**. The schema does not require one, and a placeholder would claim an identity a caller could try to look up, which is your point restated. A list rather than a map is right precisely for the reason you gave: they cannot be looked up or cleared by name, so the map bought nothing. It also self-heals when the config is repaired, which the map version could not.

Your exact case, live: one healthy environment plus two nameless broken entries gives `load_errors=2`, both `${PUBLISHER_NOPE_A}` and `${PUBLISHER_NOPE_B}` present in `/status`, neither carrying a name, and the healthy environment still serving. Four polls afterwards leave the warn count at two with repeats going to debug, so the inversion is gone too.

**BLOCKING 2, no tests: FIXED, and your mutation testing is the most useful thing in this review.** I had no answer to the third one. The entire mechanism that turns `load_errors=0` into `load_errors=1` could be a no-op and the suite stayed green, and you were right to quote my own PR body back at me: I wrote that two of three defects were caught by testing the fix, and then shipped two fixes without doing it.

Four tests in `environment_store.spec.ts`, all asserting REPORTING rather than parsing, as you prescribed. A named environment reaching both `/status` and the readiness count. An environment broken after boot appearing, driven through a real request path rather than the sync helper so it fails if any call site stops syncing. Two nameless entries counting as two with both variables named and neither claiming a name. And a failure warning once then dropping to debug, which is the one your second mutation would otherwise have survived.

All three of your mutations now fail: the no-op sync takes four tests down, removing the sync from the manifest read takes four, and the unconditional warn takes one.

**NON-BLOCKING 1, what "never removes" costs: DOCUMENTED, and your prune DECLINED with a reason.** The cost is now in the docstring in your terms: repairing in place clears the entry via the delete, but removing or renaming does not, so `/status` can name an environment the config no longer declares until restart.

I did not take the prune, and not because it is more work. Pruning names absent from both the manifest and `this.environments` is the right distinguisher, but it is unsafe as things stand: a config that fails to READ returns an empty manifest, so on that read every name looks absent and the prune would clear every genuine load failure, re-creating exactly the `load_errors=0` blindness this PR exists to remove. It needs the manifest read to distinguish "empty" from "unreadable" first, which is a separate change. That reasoning is in the docstring so the next person does not take the shortcut.

**NON-BLOCKING 2, the fourth manifest read: FIXED STRUCTURALLY, as you asked rather than as another call site.** You were right that the invariant was being enforced by remembering to add two lines, which is precisely how the fourth got missed. The read and the sync are now one instance method, and all four callers go through it including the controller, which previously could not because the sync was private. Straight back at you: you flagged that you read this one rather than ran it, and I have not stood up a watch-mode server either, so this one is verified by construction rather than observation.

**On your gate numbers: you are right that my body's unit 3001 / integration 313 was measured on `aeab084d` and is stale.** Correcting it here rather than editing the body. Current on this branch: unit 3056 pass / 3 skip / 0 fail, typecheck and eslint clean. I also see your integration 381/1, and that the one failure is the GCS-backed concurrent-package E2E which fails identically on `origin/main`.

**Two housekeeping notes.** The two leading spaces in the PR title need fixing in the GitHub UI, which I cannot do. And on the branch being behind: it is now **eight** commits rather than four, measured against a ref fetched nine minutes ago, so that number is fresher than yours rather than disagreeing with it. Worth knowing that `config.ts` has left main's side of the diff entirely: #1045 added a flag there and #1067 removed it again, so it nets to zero and our two changes no longer touch a common file.

Last thing, on process rather than code. Your opening point is the one I will carry: the threads were resolved after a consolidated comment, and a resolved thread hiding a live defect is worse than an open one, because it converts "not yet done" into "done" without anyone deciding that. Answering by name is the fix and I have done it here.

Sha-Bang:
Closing this as superseded. `main` fixed the bug — by the approach this PR considered and rejected.

The defect here was real and is gone. On current `main`, an unset `${VAR}` no longer produces a server that prints `PUBLISHER_READY`, reports `serving` with `load_errors=0`, and serves nothing with the variable named nowhere. Measured against `30ba53fc`:

```
$ # publisher.config.json referencing "${NATHAN_UNSET_PROBE_VAR}", variable unset
THREW: Error | Environment variable '${NATHAN_UNSET_PROBE_VAR}' is not set in configuration file
```

That throw reaches `environment_store.ts:1800` as a `PublisherConfigError` and boot turns it into `PUBLISHER_INIT_FAILED`. It fails loudly, at boot, naming the variable — which is the outcome this PR was written to produce.

What it does *not* do is what this PR argued for. The comment now standing at `environment_store.ts:1784` reasons about this exact symptom and then chooses to raise rather than absorb, so one bad variable stops the whole server. This PR's description considered that and rejected it: "Someone whose config references a variable they no longer set currently has a degraded server; failing the boot would give them no server." Both positions are defensible, and `main`'s is now the status quo with its reasoning written down.

So the work left here isn't a rebase. Merging this would silently reverse a deliberate decision, and it is a decision worth making on its own terms rather than inside a conflict resolution. **If we want one bad variable to degrade one environment instead of stopping the server, that is a new PR arguing against `main`'s current behaviour** — and the case for it is in this description, which is the part worth keeping.

Two other things here are worth not losing, both from the final commit (`3752649c`), which never got re-reviewed:

- The nameless-entry analysis. Keying `failedEnvironments` by a placeholder collapsed every nameless entry onto one key, which under-counted, let one entry overwrite another, and inverted the repeat detection. `main` now skips a nameless environment with a warning and an index rather than inventing a name, so the shape of the problem is handled — but the reasoning for why a list beats a map here is the clearest statement of it anywhere.
- The observation that every spec in `config.spec.ts` exercises the parser and nothing asserted a config error is ever *reported*. That gap is what let three mutations pass green, and it is still the right thing to check of any future change in this area.

Closing rather than leaving it open, since it cannot be revived by a rebase and reading it as "awaiting rebase" would be misleading.

PR #1051, opened 2026-08-19
  fix(sdk,app): three follow-ups from the #1026 review

mlennie:
The three non-blocking notes from Sha-Bang's review of #1026, split one per note since two change behaviour and one is prose. #1026 merged as `952ca0f` with all three deferred; this is that debt.

## Open every Home link in a new tab, including the API spec

`external={false}` on the api-spec link was inverted relative to the risk the comment above it reasons about. `Home.tsx` opened `${serverBaseUrl(server)}/api-doc.html` in the same tab, which is right inside the Console and wrong on the public surface, because the comment directly above already says that an SDK consumer's page origin need not be that server. That is exactly the case where same-tab navigation unloads the customer's app instead of opening alongside it.

`Home` is exported from the SDK (`src/index.ts` to `components` to `Home`), so it does render in other people's pages. The prop had exactly one `false` caller, so it is removed rather than flipped, and `target` and `rel` are set together at the single anchor.

## Say in styles.ts which brand colors the contrast bar now binds

The docblock told a brand retuner the 3:1 bar was stated "for these three only" and that `MALLOY_BRAND` is a brand decision rather than a component one. But `CONTENT_TINT` binds `MALLOY_BRAND.orange` to dashboard and `.darkBlue` to model, and `GRANDFATHERED` lists only report, so both are now asserted at 3:1. Orange measures 3.09:1, a 0.09 margin. Someone retuning the logo orange, on that comment's own invitation, would red a suite in a file they never opened.

Sha-Bang offered two fixes: grandfather the two by type the way report is, or state in `styles.ts` what is now bound. This takes the second, and the reason matters. The trap was never the assertion. Below 3:1 the white glyph genuinely does become hard to read, so the suite going red is the test working rather than the test being brittle. The trap was a comment promising the bar did not apply to colors it now applies to. So: fix the promise, keep the guarantee.

Measured: orange 3.09:1 and darkBlue 5.17:1, both bound by the spec; teal 2.52:1 remains the disclosed exception.

## Name every environment-churning spec, not just the first

The comment explaining `workers: 1` singled out `packages.spec` as the environment churner. It is not the only one: `environments.spec.ts` has a create-edit-delete disposable-environment test, and `package-dashboards.spec.ts` registers and removes one in `beforeAll`/`afterAll`. Both verified rather than taken from the review. That comment is what someone reads before relaxing `workers: 1`, so an incomplete list actively misleads them.

## Deferred, and left as a follow-up

The per-spec-isolation suggestion, that the three storefront writers adopt the pattern `package-dashboards.spec` introduced and register their own temp environment, so the other twelve specs would not need serializing.

The cost of `workers: 1` is real but smaller than first reported. Measured by Sha-Bang on 2026-08-19: 9m40s serialized on #1026 against 7m30s for the same spec set run in parallel on #1020, both against that day's `main`, so about 1.3x, comfortably inside a 30-minute timeout. His first figure of 1.7x came from a comparison against an older `main` with a different spec mix and he corrected it himself. That is a standing cost with headroom rather than a risk, which is why the architecture change is a follow-up rather than part of this.

## Verification

Gate green on this tree: codegen and `build:sdk` first, then typecheck, lint and tests, all collection counts on baseline. Prettier's six failures are the standing generated clients, untracked and gitignored.


Sha-Bang (approved):
approving at fc15184a. all three follow-ups do what they say, and one of them corrects me rather than the other way round.

checked:

- the playwright comment now names `packages.spec`, `environments.spec` and `package-dashboards.spec`, with the reason a partial list is worse than none. that was the actual defect in that finding.
- `Home.tsx` removes the `external` prop outright instead of flipping it, so `target` and `rel` can't desynchronise because there's no longer a flag to get wrong. `InlineLink` is local rather than exported, all four call sites are updated, and Typecheck is green. stronger than what i suggested.
- the `styles.ts` figures are exact: orange 3.09:1, darkBlue 5.17:1, teal 2.52, and "clears it by 0.09" is right. the new sentence about grandfathering by content type rather than by colour is also accurate — `GRANDFATHERED` is `["report"]`, so a new type handed teal would fail the suite, which is the useful half of that guarantee.

`ContentTypeIcon.spec.ts` is 10/10 locally and CI is green across 14 checks.

**second correction on the `workers: 1` number, and this one retires it.** i pulled the actual Playwright durations: serialized runs are 9m40s (#1026), 6m14s (#1027), 6m22s (#1035) and 5m26s (here); the parallel ones i compared against were 5m42s (#1043) and 6m18s (#1020). the same commit also re-reports different numbers between runs — i saw #1020 at 7m30s earlier today and 6m18s now, #1035 at 5m3s and now 6m22s.

so the spread within a single configuration is wider than any gap between configurations, and there is no cost attributable to `workers: 1` from this data. my original figure (~4 minutes, 1.7x) and my correction (~2m10s, 1.3x) were both reading noise. fixing the comment and leaving the setting alone was the right call — flagging it here so nobody spends an afternoon chasing a regression that isn't there.

PR #1053, opened 2026-08-20
feat(server): incremental mz on storage persisted sources

housejester:
## What

`refresh="incremental"` alongside `storage=` was a publish rejection. It is now supported: a table
materialized into a storage destination is advanced by the same bounded
`[covered_through, frontier)` delta a colocated one gets, with the same declarations and the same
guarantees.

```malloy
#@ persist name="daily_orders" storage=lake refresh="incremental" watermark="order_date"
source: daily_orders is orders -> {
  group_by: order_date
  aggregate: total is amount.sum()
}
```

Without it, every refresh on a storage destination is a full re-scan of the source warehouse plus
result-sized egress, so a scheduled refresh there can cost more than the live queries materializing was
meant to replace.

## Why it is not a one-line lift

A stored source's rows are computed by the **source warehouse** and its table lives in the
**destination engine**, so a delta that was one engine's business now spans two:

| | colocated | `storage=` |
|---|---|---|
| range predicate | source dialect | source dialect, pushed **inside** the query passthrough |
| `DELETE`/`INSERT`/`MERGE` | source dialect, source connection | DuckDB, on the build session |
| target probes | source connection | build session, `duckdb_columns()` |
| frontier probe | source connection | source-dialect `MAX()` through the passthrough |

Pushing the predicate *inside* the passthrough is the load-bearing part: filtering outside it would
stream the whole source across the egress boundary and then discard most of it — correct rows, and the
entire reason for refreshing incrementally undone.

Everything else — the order of the checks, seed vs skip vs delta, the reason code each one reports —
stays in the one planner, which now talks to a `DeltaTarget` rather than to a connection and a dialect.
Those reason codes are the only signal that a source declaring incremental refresh is quietly being
rebuilt on every run, so a second copy of that decision order for the second target would be a second
place for it to drift.

`plan`, `apply` and the post-rebuild boundary probe all run on one build session, which is what keeps
the destination's read-write attach and the source's federated credentials scoped to a single source's
refresh.

## A boundary is identified by the store its table lives in

`storage=` enters **neither** the content address nor the physical name, and generation counting is
per destination — so a table persisted colocated and one persisted into a destination can share both
their connection and their name while being two entirely different tables. A boundary measured on one
is then indistinguishable from a boundary measured on the other, and a delta computed from the wrong
one skips rows silently, which is the one direction a `covered_through` must never move in.

So `LedgerEntry` gains an optional `storageDestinationName`, mirroring `ManifestEntry` — absent means
the table lives in `connectionName`'s own warehouse — and in the publisher's own ledger it joins the
primary key. It has to: those two tables coexist legitimately — the publisher's own
persist-target collision warning keys on the destination too, so it correctly says nothing about the
pair — and sharing one row makes both seed forever, each overwriting the other, with nothing to report
it.

Keyed with a `''` sentinel rather than a nullable column, because a NULL in a key defeats
`ON CONFLICT` — two colocated rows would both insert — and forces every read to compare with
`IS NOT DISTINCT FROM`.

**Upgrading discards recorded boundaries.** The ledger re-keys itself on boot, the same self-limiting
drop the package re-key in 0.0.240 used, so each incremental source takes one full rebuild and then
resumes advancing by delta.

**If you supply the ledger yourself** (`buildInstructions.ledger`), store and return the new field like
every other. An entry that omits it for a source materialized into a destination describes a different
table, so that source seeds rather than being rejected — stale is not wrong. An entry naming a
*different* destination than the run instructs is still a 400. Declare the field in your own copy of
this spec **before** deploying a publisher that sends it: a strict generated client throws on a field
its spec does not carry.

## Also worth knowing

- **A delta's warehouse read is attributed and costed** through the same call a full build's goes
  through, so a `queryMetadata` bag reaches it as a BigQuery `@@query_label` or a Snowflake
  `QUERY_TAG`, and `ManifestEntry.queryCostBytes` reports what a tagged read cost. On BigQuery's
  labelled shape the read is issued before the transaction opens, so a delta that then fails has paid
  for rows it discarded — the same trade a full build already makes, bounded by the delta's own range.
- **A chained stored source is rebuilt, not advanced**, and says so with its own reason code. Its
  parent's refresh can restate rows below the child's own frontier, where no delta of the child's
  would revisit them.
- **The serve path is unchanged.** A read-only DuckLake attach observes a committed in-place delta on
  its next query, with no re-attach and no package reload — a property of the Postgres-backed catalog,
  which is the supported configuration. A DuckDB-file catalog does not behave this way; noted in
  `docs/ducklake.md`.
- **A stored table a failed run declines to reclaim is logged and counted**
  (`publisher_storage_tables_retained_total`, by destination). An incremental source's physical name
  is stable by contract, so it is never a fresh generation nobody else knows about — a prior manifest
  may bind it, and dropping it because a later source failed would take the source off its table. The
  count is an upper bound on what accumulates rather than a leak count: a rebuild at a fresh name is
  unreferenced, a re-seed on the live name is not, and the manifest entry cannot separate them.
- **`publisher_source_build_duration_seconds` gains a `delta_storage` engine label**, kept apart from
  `delta` for the reason `storage` is kept apart from `in_warehouse` — the rows cross an egress
  boundary and the DML runs on another engine, so pooling them averages two cost profiles into one.

## Test plan

- [x] Unit coverage for the storage delta: which side of the passthrough each piece of SQL lands on,
      the `duckdb_columns()` probe (an attached DuckLake catalog exposes no `information_schema`), the
      seed-when-the-boundary-was-measured-in-another-store case, and the timezone the range is
      resolved in.
- [x] **Five hammer scenarios** in a suite of their own (`hammer/scenarios/incremental-mz/`), against a
      real DuckLake: a delta advances a colocated table and a stored one; `merge_key=` replaces a row
      restated with a later watermark instead of duplicating it; a boundary does not follow a source
      between stores; a stored source built on another is rebuilt rather than advanced.
- [x] The ledger's re-key covered both ways — it replaces an older key, and it leaves the current one
      and its boundaries alone, so a restart does not silently reseed every source.
- [x] `packages/python-client/scripts/build-python-sdk.sh` re-run for the `api-doc.yaml` change.

A timestamp watermark is deliberately not among the scenarios: its range end comes from the run's own
clock, so two builds inside one test produce an empty range and skip. The timezone its bounds are
resolved in is covered by unit tests instead — a session on the host's zone rather than UTC would
select a window offset by that offset, commit, and simply hold the wrong rows.


packages/server/src/service/incremental_apply.ts
```diff
@@ -407,7 +460,36 @@ export async function probeTargetColumns(
...
+      // wrote.
+      const schema =
+         segments.length >= 3 ? segments[segments.length - 2] : "main";
+      // The destination catalog is the FIRST segment, because the caller builds
+      // this path as `<destination>.<name>` — counting back from the end instead
+      // would read a longer name's own container as the catalog and probe a
+      // database that is not this destination.
+      const database = segments[0];
+      sql = `SELECT column_name AS ${probeAlias("column_name", dialect)}
+              FROM duckdb_columns()
```
sagarswamirao, line 495: Flagging early so it doesn't bite downstream: this probe is capped at 10 rows, so a stored table with more than 10 columns can never advance by delta.

The build session is constructed as `new DuckDBConnection(sessionName, ":memory:", workDir)` (materialization_build_session.ts:574) with no `queryOptions`, so db-duckdb's `DEFAULT_QUERY_OPTIONS = { rowLimit: 10 }` applies and `runSQL` does a hard `result.slice(0, rowLimit)` (dist/duckdb_common.js:55,64-65,130 in the pinned 0.0.427). `duckdb_columns()` returns one row per column, so column 11 onward is silently dropped.

Verified against a real attached DuckLake catalog, calling this function directly:

```
probeTargetColumns(runner, "duckdb", "lake.orders")
-> 10 columns of 15: c0,c1,c2,c3,c4,c5,c6,c7,c8,c9
-> probe.error: (none)
```

The truncated list then reaches `shapeMismatch(inputs.columns, probe.columns)` below, columns 11+ come back as `missing`, and the planner returns `{mode: "seed", reasonCode: "shape_mismatch"}`. So every refresh is a full rebuild -- the cost this feature exists to remove -- and on BigQuery each run re-reads and re-bills the whole source. The warn log reports those columns as "absent from the table" when they are present, which points an operator at the model rather than at the probe.

The colocated path is unaffected, since its runner is the source warehouse connector and that does not apply this slice.

Could you set an explicit large/unlimited `rowLimit` on the build session, or use `runRawSQL` (which bypasses the slice)? Worth applying to `describeTable` (materialization_build_session.ts:1302) in the same change -- it truncates the same way (verified: 10 of 15), which truncates the manifest `schema` the serve transform declares. That one predates this PR, but this PR adds a delta-path call site at materialization_build_session.ts:866.

packages/server/src/service/incremental_apply.ts
```diff
@@ -750,6 +832,18 @@ export function deltaScript(statements: string[]): string {
...
+ * Dialects that abandon a failed script's transaction themselves, so the rollback
+ * below is not merely unnecessary but noisy: DuckDB answers a `ROLLBACK` with no
+ * active transaction by RAISING, which would log the "a pooled connection may
+ * stay in an aborted transaction" warning on every failed delta — sending an
+ * operator after a poisoned connection that this engine cannot have, since a
+ * build session is private and single-use. Measured: after a failed
+ * BEGIN/…/COMMIT script the session is immediately usable and nothing was
+ * applied.
+ */
+const SELF_ROLLING_BACK_DIALECTS: ReadonlySet<string> = new Set(["duckdb"]);
```
sagarswamirao: The premise this exemption rests on does not hold: DuckDB does not roll back a failed multi-statement script, so suppressing the `ROLLBACK` leaves the session in an aborted transaction.

The comment above says "Measured: after a failed BEGIN/.../COMMIT script the session is immediately usable and nothing was applied." The second half is right; the first is not. Driving this file's own generator and applier with `dialect: "duckdb"`:

```
STATEMENTS: ["BEGIN","DELETE FROM \"t1\" WHERE ...","INSERT INTO \"t1\" ...","COMMIT"]
delta failed: true
SESSION POISONED -> TransactionContext Error: Current transaction is aborted (please ROLLBACK)
```

A/B: with the `ROLLBACK` restored, the next statement succeeds and the pre-existing rows are intact. Also reproduced against a real DuckLake catalog. A statement that fails at bind time (before `BEGIN` opens the transaction) leaves the session clean, which may be what the original measurement caught.

The data is safe either way -- the transaction genuinely aborts, so no watermark range is lost -- and the blast radius is bounded because the build session is per-source and disposable. Two consequences remain: for a Snowflake source the `ALTER SESSION UNSET QUERY_TAG` in `buildSourceIntoStorage`'s `finally` (materialization_build_session.ts:918) now runs on an aborted session and fails, leaving the tag set and logging a misleading warning; and the guard is gone if this session ever becomes reusable.

Simplest fix is dropping `duckdb` from the set -- the `ROLLBACK` costs nothing on the success path, and the "DuckDB raises on a no-op ROLLBACK" concern doesn't apply here since the transaction is really open. Only `range_replace` is affected; the single-statement `MERGE` path is fine (verified).

sagarswamirao:
Two things outside the line-level findings, plus one pre-existing function this PR newly routes DuckDB values into.

**A timezone offset is silently truncated rather than rejected.** `secondsPrecision` (incremental_apply.ts:149-158) is anchored at the start but not the end, so an offset-bearing timestamp is accepted and truncated instead of refused: `"2024-06-01 07:00:00+05:30"` returns `"2024-06-01 07:00:00"` -- a bound labelled UTC that is not UTC. Unchanged by this PR, so raising it here rather than inline.

What makes it worth closing now is whose offset that is. `@duckdb/node-api`'s `DuckDBTimestampTZValue` captures it statically at module load from the host process:

```js
// lib/values/DuckDBTimestampTZValue.js:12
static timezoneOffsetInMinutes = -new Date().getTimezoneOffset();
```

That is the container's TZ, independent of `SET TimeZone='UTC'`. So `pinSessionToUTC` protects the predicate (confirmed: same predicate against a TIMESTAMPTZ column, `TimeZone='UTC'` selects 1 row, `Asia/Kolkata` selects 0) but not a timestamp read back through `runSQL`.

Latent today, not live -- I traced both callers and no probed timestamp bound is ever recorded: `resolveDeltaEnd` returns `snapshotBound` for date/timestamp and only probes for number/string, and `seedCoveredThrough` probes then discards the result for those types. But it is one plausible change from live: letting a timestamp watermark take its end from the source (an obvious optimization, since a clock-derived end is why two refreshes in one day skip) would record a host-local value as UTC. East of UTC that boundary LEADS the table, which is the one direction this file says `covered_through` must never move, and it skips rows permanently. `renderSqlBound`'s own doc already asserts the invariant ("every bound is UTC text") that this function does not enforce.

Could you make it reject a trailing offset instead of truncating past it? A rejection lands as `plan_error` -> seed, which is expensive but correct rather than silent and wrong. A `canonicalBoundValue` case for `"2024-06-01 07:00:00+05:30"` would pin it -- the existing tests cover Date, the BigQuery `{value}` wrapper, number and string, but no offset-bearing string.

**The new storage-delta path has no real-engine execution coverage.** `incremental_storage.spec.ts` drives a hand-written fake that pattern-matches on `sql.includes(...)` and returns fabricated rows, with a 3-column fixture. `incremental_dml_semantics.spec.ts` does execute against real DuckDB, but passes `dialect: "postgres"` throughout, so the new `duckdb` branch is never exercised -- and its header still says DuckDB "is NOT on the dialect allowlist... does NOT make DuckDB supported", which this PR makes stale by adding `duckdb` to both `INCREMENTAL_TARGET_DML_DIALECTS` and `MERGE_CAPABLE_DIALECTS`. The five hammer scenarios are not CI-gated (`test:hammer` runs `bun test hammer/lib`; the scenarios need `bun hammer/run.ts`). Both line-level bugs I flagged sit in exactly this gap -- one integration test that attaches a real destination, seeds, then applies a delta would have caught both.

Also worth noting: the end-to-end storage-delta scenario uses a number watermark, for a reason the scenario itself documents well (a date watermark's end comes from the run clock, so two refreshes in one test skip). The side effect is that no integration or scenario test exercises a date/timestamp watermark on the storage path, which is where the cross-engine type and timezone questions live.

**The `storageDestinationName` seed-vs-reject branch is untested.** The six tests in `describe("with a supplied ledger")` never set a `destination` on an instruction, so `instructedIntoStorage` is empty in all of them -- neither the absent-destination-seeds branch nor the different-destination-rejects branch is covered, though they are what the spec prose promises callers. Two assertions would close it.

**One spec sentence overstates delta cost attribution for Snowflake.** The new `api-doc.yaml` text says a delta into a `storage=` destination is costed "on the same terms as that destination's full build". The seed path has a Snowflake `?? snowflakeReadCostAfterBuild(...)` fallback that the delta path lacks -- and it could not help anyway, since it matches on `QUERY_TEXT` equalling `buildSQL` byte-for-byte while a delta's SQL is the bounded-and-projected wrapper. So a Snowflake storage delta reports null where a full build reports a cost, and a caller summing `queryCostBytes` reads that null as "free". Reads like a genuine design limit rather than an oversight, so the fix is probably the sentence -- carving Snowflake out the way the paragraph already carves out Postgres and chained builds.

packages/server/src/service/materialization_service.ts
```diff
@@ -419,6 +428,100 @@ const VALID_TRANSITIONS: Record<
...
+ * generations for that same reason, so this adds a case to a known leak rather
+ * than a new class of one — and the alternative is dropping a table that is
+ * serving. Not reclaiming costs storage; reclaiming costs a source its data until
+ * something rebuilds it.
+ *
+ * The reclaim's own `stillReferenced` check does not cover any of this. It spares
+ * a table that some READY manifest of the SAME package name serves, and a host
+ * that version-qualifies its package names sees none of its own earlier manifests.
+ */
+export function isReclaimableStorageTable(entry: ManifestEntry): boolean {
```
sagarswamirao, line 545: A chained stored source loses failed-run reclaim here, which is a scope change from `main`.

`main` pushed every stored entry unconditionally (`if (entry.storageDestinationName) builtThisRun.push(entry)`), so a failed run reclaimed it. This predicate tests `entry.refresh === undefined`, but `refreshFields` synthesizes `refresh: "full"` from the mere presence of `lineage` -- and a chained source always seeds, since `chainedSeed = lineage !== undefined && dependsOnStorageUpstream` forces it. So the entry arrives with `refresh: "full"`, fails this test, and goes to `retainedThisRun` instead of `builtThisRun`: logged and counted, never reclaimed.

The docblock's reasoning is sound for the case it describes -- an incrementally refreshed source needs a stable name, so the name may be one a prior manifest binds and dropping it would take the source off its table. But a chained source never advances, so it never has a boundary requiring a stable name; under a host that assigns generational names it is a fresh name nothing will reference or rebuild. The docblock actually names this case ("a host-initiated rebuild at a FRESH generational name, which nothing references and nothing will build again") and then declines to separate it.

Worth noting the sibling gate a few hundred lines down gets this distinction right: it keys on `result.refresh` -- whether the table was actually advanced in place -- rather than on `entry.refresh`, which is the coarser proxy.

Could you key reclaimability on whether the table was advanced in place (`result.refresh` present, i.e. `"delta"` or `"none"`) rather than on `entry.refresh === undefined`? A `refresh: "full"` entry was written by a `CREATE OR REPLACE` at a possibly-fresh name and is reclaimable on the same terms as before this PR. Lower severity than the two bugs I flagged separately -- it is a counted, logged leak on an already-known leak class -- but it is a regression in cleanup scope rather than an intentional narrowing.

packages/server/src/service/materialization_build_session.ts
```diff
@@ -706,6 +806,10 @@ export async function buildSourceIntoStorage(params: {
          // mutate the shared catalog's persisted options.
          await session.runSQL("SET ducklake_default_data_inlining_row_limit=0");
       }
+      // Before any predicate is rendered against this session. Matters only to an
+      // incremental refresh (a full CTAS issues none), but it is cheap and the
+      // session it protects is this one — see pinSessionToUTC.
+      await pinSessionToUTC(session);
```
sagarswamirao, line 825: Small asymmetry worth closing while this is fresh: `buildDownstreamIntoStorage` (the chained path, further down this file) has no equivalent `pinSessionToUTC`. It sets the inlining limit on its session but never pins the zone.

Harmless today, and for a good reason: chained storage is excluded from the delta path, so no predicate and no `afterSeed` boundary probe ever run on that session. But the two functions are described as mirroring each other ("Session lifecycle mirrors `buildSourceIntoStorage` exactly"), and the `chained_storage` reason code exists precisely because lifting that restriction is intended. Whoever does that will be editing the one path where this guard is absent.

One line, and it removes a trap from the code most likely to be touched next.

packages/server/src/service/materialization_service.ts
```diff
@@ -2680,6 +2931,84 @@ export class MaterializationService {
...
+      instruction: BuildInstruction;
+      destinationName: string;
+      physicalTableName: string;
+      /** The unprojected build SQL — see storageDeltaTarget. */
+      buildSQL: string;
+      queryMetadata?: QueryMetadata;
+   }): StorageIncrementalRefresh {
+      const { context, lineage, persistSource, instruction } = params;
+      return {
+         plan: async ({ session, sourceType, handle }) => {
```
sagarswamirao: This destructure drops `quotedTablePath`, and `storageDeltaTarget` then re-derives the identical value from `destinationName` + `physicalTableName`.

Correct today -- both call the same pure `quoteManifestTablePath` on the same inputs, which is exactly why the delta's DML and the seed's CTAS provably name the same table. But the parameter exists to guarantee that agreement, and it is the one being discarded: the caller computes the path, passes it in, and this half recomputes it. `afterSeed` just below does use the passed-in `quotedTablePath`, so the two halves of the same callback disagree about which derivation is authoritative.

No behavioral issue to fix, so purely a durability point: either use the parameter here too, or drop it from the interface so there is one derivation of a path where the DML and the CREATE must agree.

sagarswamirao (approved):
Re-reviewed at b2ffb4be. All five fixes verified against a real engine rather than read off the diff -- here is what I checked and what I found.

**The row limit.** Lifted at the session factory, so both the `duckdb_columns()` probe and `DESCRIBE` are fixed at the one root cause rather than per call site. Through the production factory against an attached DuckLake catalog: `probeTargetColumns` 15 of 15, `describeTable` 15 of 15.

**The rollback.** Unconditional now. Verified both failure modes through `deltaStatements` + `applyDeltaScript` with `dialect: "duckdb"`: an execution-time failure leaves the session answering with its rows intact, and a bind-time failure does not raise on the rollback -- which was the concern the exemption existed for, so removing it costs nothing. The replacement comment identifies exactly why the original measurement was misleading.

**The zone offset.** Anchored at both ends. I specifically checked for false rejection of a real driver shape, since that was the risk with a `$` anchor: `2024-06-01 07:00:00`, the fractional-seconds form, and a bare `Z` are all still accepted; `+05:30`, `-08:00` and `+0530` are all refused. The `Date` path is untouched.

**The path derivation and the chained pin.** Both as suggested -- the quoted path is threaded through, so the agreement between the DML and the CTAS is structural rather than two derivations that match today.

**The Snowflake carve-out** reads well, and the added line that summing the field understates a Snowflake source rather than meaning those reads were free is the part an operator actually needs.

**On the new engine spec:** I checked it is not vacuous by reverting only the two production fixes and keeping the test. It fails on exactly the two bugs (`Expected: 15, Received: 10`, and `Current transaction is aborted`) and passes with them applied. Using the production session factory is what makes it work -- a hand-built session would have passed while the real one truncated, as the comment says. The supplied-ledger branches and the corrected `incremental_dml_semantics` header are covered too. Full changed-area suite: 325 pass, 0 fail.

One item from my earlier round looks unaddressed rather than declined, so flagging it once more without holding up the PR: `isReclaimableStorageTable` is unchanged, so a chained stored source still lands in `retainedThisRun` instead of `builtThisRun` on a failed run -- logged and counted, never reclaimed, where `main` reclaimed it. It is the mildest of the findings and a reasonable thing to defer; it would just be good to have the decision written down rather than left silent. Fixing it here or filing it both work.

Nice turnaround on the engine-level bugs -- the two that mattered were both properties of the driver rather than of the SQL text, and they are pinned now.

PR #1057, opened 2026-08-20
chore(skills): indent the generated bundle so concurrent skills PRs merge cleanly

mlennie:
## The problem

`packages/server/src/mcp/skills/skills_bundle.json` is a generated asset committed to git, and it was written minified. That put the entire file, 467,961 bytes, on a **single line with zero newlines**. Git has nothing to merge inside one line, so any two branches that both touched `skills/` conflicted on it. Not usually, not probably: every time, no matter which skills each branch edited.

That is not a rare event. **22 commits touched this file in the last 30 days**, roughly one every 1.4 days. One recent PR hit the conflict twice in a single day, against two different merged PRs.

## The fix

Serialize the bundle with two-space indentation. Each entry's fields land on their own lines, so two branches editing different skills touch different lines and git merges them without help. A skill body still occupies one line, because a JSON string cannot contain a raw newline. What remains is an ordinary line-scoped conflict, meaning the two branches touched the same region of the file, and it is small enough to read.

I replayed this on two real merged PRs that touched disjoint skills, #1020 (which adds the dashboards skill) against #1035 (which edits the two data-app skills). For each side I generated the bundle from its own `skills/` tree, then merged:

| bundle format | result |
|---|---|
| minified (before) | **CONFLICT** |
| indented (after) | **clean auto-merge** |

On the clean merge I then checked, per entry, that everything either side changed relative to the base survived. It did, on both sides.

A constructed matrix over the same file agrees, and shows honestly where conflicts remain. Note the last row: two branches each adding a different new skill still conflict, because their inserts land at the same sorted position. Indentation removes the guaranteed whole-file conflict, not every conflict.

| scenario | minified | indented |
|---|---|---|
| two branches edit different skills | CONFLICT | clean |
| edits to adjacent entries | CONFLICT | clean |
| one adds a skill, other edits another | CONFLICT | clean |
| two branches edit the same skill | CONFLICT | CONFLICT (genuine, and now line-scoped) |
| both add a new skill | CONFLICT | CONFLICT (adjacent inserts at the same sorted position) |

## The reformat changes no content

The sha256 of the re-serialized parsed bundle is identical before and after (`3b614d93...`). The file grows 1,913 bytes, or 0.41 percent, and goes from 1 line to 284. The generator's stated constraint is untouched: the bundle stays committed, so the server still has no runtime dependency on the `skills/` directory layout.

## Why there is no new CI check here

I went looking for the more alarming version of this problem, which is that a careless conflict resolution silently reverts whichever merged PR last edited the bundle, and nothing notices. **That hazard is already closed, and I want to be explicit about it rather than add a redundant check.** `skills_bundle.spec.ts` already regenerates from `skills/` and compares against the committed file, it is a `src/**/*.spec.ts` unit spec so `bun test src` collects it, and `cross-platform-tests.yml` runs that suite on ubuntu, windows and macos for every pull request. `skills/README.md` documents it too.

I confirmed it works rather than taking the code's word for it: deleting a paragraph from one bundle entry makes that test fail with a compact diff naming the exact dropped lines, not a 468KB dump.

## The one gap that did need closing

The existing sync test compares **parsed** JSON, so it would stay green if a future regeneration re-minified the file, quietly undoing this change. The new test reads the file off disk instead and asserts each entry's `name` key sits on its own line.

I verified it fails for the right reason by re-minifying the bundle: the new test failed with `Expected: 56, Received: 0` while all ten other tests in the file passed, including the sync test. That is the gap, demonstrated.

It normalizes CRLF before asserting. These are the first real newlines this file has ever had, and there is no root `.gitattributes`, so a Windows checkout can convert them.

## Entry order is now locale-independent too

Review turned up the same bug one level up, and it is worth fixing here rather than later. `buildSkills` sorted with `localeCompare`, which follows the locale of whichever machine regenerates. That was harmless while the file was one line. It stops being harmless once line positions are the point: under `cs-CZ` the `ch` digraph sorts after `h`, which moves `malloy-charts` and makes a regeneration emit a **10-line reordering diff with nothing to do with the skill the contributor edited**. `lt-LT` diverges as well. Of six locales checked, `en-US`, `de-DE`, `sv-SE` and `tr-TR` agree with codepoint order and those two do not.

The builder now sorts by codepoint, which is the comparator the spec was already using on its side of the comparison. This changes no bytes in the committed bundle today, verified by regenerating and getting an identical sha256, so it is purely a determinism guarantee for the next person.

Neither existing guard would have caught a locale-shifted regeneration: the sync test sorts both sides before comparing, and the indent test only counts `name` lines. So there is a third test asserting the committed order is codepoint order. I verified it fails against a `cs-CZ`-ordered bundle while the other eleven tests pass.

## Considered and not done

- **Gitignore the bundle and generate it at build time.** Cleanest in principle, and I decided against it. `packages/server/src/mcp/server.ts` imports the JSON statically, so a fresh clone's `bun run typecheck` would fail until codegen had run, and the precondition would have to be added to the root typecheck, the CI typecheck job, `build:server-deploy`, the Dockerfile, and the release and skills-npm workflows. Trading a merge conflict for a broken fresh-clone build is a bad trade, and indentation already removes the conflict for edits to different skills.
- **A `.gitattributes` merge driver.** A custom driver has to be registered in each clone's local git config, so `.gitattributes` alone silently does nothing. It fails for exactly the contributor who does not know it exists.

## Verification

Full gate green: typecheck, lint, 3008 unit tests and 313 integration tests passing. `prettier:check` reports six files, all of them generated and git-ignored (`packages/sdk/src/client/*.ts` and `packages/server/src/api.ts`, each confirmed with `git check-ignore`), which this branch does not touch and CI's formatting job never sees because it runs before codegen.

Live check, since the bundle's only production consumer is MCP prompt registration. Booted the server two ways, from source and from the built `dist` bundle that esbuild inlines the JSON into, because the production and Docker paths run the latter and that is where newlines could plausibly have mattered. Both serve all 56 prompts, with the served name set exactly equal to the bundle's entries, and one 24,626-character body byte-identical to its bundle entry with its 401 newlines intact.

Not verified locally: the Windows leg, and the Docker image as an image (the bundle it runs is the same `dist` output tested above). Those are the ones worth watching in CI here, since this gives a build-inlined file real newlines for the first time.


packages/server/src/mcp/skills/build_skills_bundle.ts
```diff
@@ -164,7 +180,7 @@ function main(): void {
 
    const skills = buildSkills(skillsDir);
 
-   fs.writeFileSync(outFile, JSON.stringify({ skills }));
+   fs.writeFileSync(outFile, `${JSON.stringify({ skills }, null, 2)}\n`);
```
Sha-Bang: Nit, but a one-liner worth taking while you are in here: this writes two-space indent, and `packages/server/.prettierrc` sets `"tabWidth": 3`. The bundle is not in `.prettierignore` and it lives inside that package, so the committed file is permanently unformatted by its own package's config:

```
$ bunx prettier --check packages/server/src/mcp/skills/skills_bundle.json
[warn] packages/server/src/mcp/skills/skills_bundle.json
```

Prettier's output is the same 284 lines but 470,604 bytes against the committed 469,874 — it re-indents every level from 2 to 3. So a contributor with format-on-save reformats this file by opening it, their entry fields land at nine spaces, and the new indent test below fails with `Expected: 56, Received: 0` for a reason that has nothing to do with what they changed. Regenerating flips it back.

`prettier:check` globs `*.{ts,tsx}`, so CI sees none of it either way. Either fix works: emit `null, 3` and assert `/^ {9}"name": /`, which makes prettier a no-op and matches the package's house style, or add the path to `.prettierignore`, which is blunter but survives a future `tabWidth` change.

packages/server/src/mcp/skills/skills_bundle.spec.ts
```diff
@@ -46,6 +49,49 @@ describe("skills_bundle.json (generated dual-channel asset)", () => {
...
+    * parsed JSON, so it stays green if a regeneration ever re-minifies the file.
+    *
+    * CRLF-normalized because these newlines are new. The file had none while it
+    * was minified, there is no root .gitattributes, and a Windows runner can
+    * check it out with CRLF endings.
+    */
+   it("is committed indented, one entry per line", () => {
+      const raw = fs
+         .readFileSync(path.join(import.meta.dir, "skills_bundle.json"), "utf8")
+         .replace(/\r\n/g, "\n");
```
Sha-Bang, line 74: The CRLF normalization is the right defensive move given this file had no newlines until now — and there is a way to make the hazard impossible rather than tolerated, which is the `.gitattributes` the description rejects.

That rejection is sound for a *merge driver*, which does need per-clone registration. `text eol=lf` needs none: git enforces it at checkout and commit in every clone, including for the contributor who has never heard of it. One line —

```
packages/server/src/mcp/skills/skills_bundle.json text eol=lf
```

— and a Windows checkout cannot produce CRLF here, which also retires the "not verified locally: the Windows leg" caveat for this file specifically. Worth splitting the two uses in the description, since a reader currently comes away thinking `.gitattributes` was weighed and found wanting for the whole class of problem.

packages/server/src/mcp/skills/skills_bundle.spec.ts
```diff
@@ -26,13 +26,16 @@ const sourceDir = path.join(
...
- * Membership and content are what matter here; file order is cosmetic.
+ * Codepoint order, matching the builder's own sort. Both sides are sorted before
+ * comparing so this test stays about membership and content, and the separate
+ * order test below is what holds the builder to a locale-independent order.
  */
 const byName = (a: SkillEntry, b: SkillEntry) =>
    a.name < b.name ? -1 : a.name > b.name ? 1 : 0;
 
+/** The same codepoint comparator, on bare names. */
+const byName2 = (a: string, b: string) => (a < b ? -1 : a > b ? 1 : 0);
```
Sha-Bang: Nit: `byName2` next to `byName` reads like a copy that never got named. `byNameString` or `compareNames` says which one to reach for without going to look.

Sha-Bang (commented):
Good change, and the reasoning is the kind I wish more infrastructure PRs carried: a measured failure rate (22 commits in 30 days), a replay against two real merged PRs, and a matrix that says honestly where conflicts still remain rather than claiming they are gone. Nothing here blocks. Two nits inline, and one structural thought that belongs in a different PR.

- **Regeneration is byte-identical** to the committed bundle: sha256 `ec26f219…` both ways. That confirms the reformat and the codepoint-sort change together produce exactly what is committed, so the "changes no bytes today" claim holds.
- **The arithmetic matches**: 284 lines, 469,874 bytes — the stated 467,961 + 1,913 exactly.
- **12/12** in `skills_bundle.spec.ts`.
- **The CI claim is right**, and I checked it harder than the rest because "no new check is needed" rests on it. `build.yml` triggers on `pull_request` and calls `cross-platform-tests.yml`, whose server step is not platform-gated, so `bun run --cwd packages/server test:unit` collects this spec on ubuntu, windows and macos. I had this wrong in a review of another PR in this batch and have corrected it there.

## Indenting a committed generated artifact is the standard treatment

Worth saying plainly, because the description defends the choice as though it were unusual. It is not: every committed lockfile is line-oriented for this exact reason — `package-lock.json`, `bun.lock` in this very repo, `Cargo.lock`, `poetry.lock`. And the well-known counterexamples argue the same way from the other side. Jupyter `.ipynb` and Xcode `.pbxproj` are merge-hostile blobs, and the ecosystem's answer was to change the format (jupytext, nbstripout), not to distribute a merge driver. This PR is applying the normal fix to a file that should have had it from the start.

Both rejected alternatives are also rejected for the right reasons, and the gitignore-plus-codegen one especially: `server.ts` imports the JSON statically, so a fresh clone's typecheck would break and the precondition would have to be threaded through six places. Trading a merge conflict for a broken fresh clone is the worse trade.

## The next step, if this keeps hurting — separate PR

Indentation is mitigation, and the description is honest that one row of the matrix survives it: **two branches each adding a skill still conflict**, because both inserts land at the same sorted position. For a directory people are actively adding to, that is the common case rather than the exotic one.

The structural answer is to stop having one file: **per-skill JSON files plus a generated index**. Editing different skills touches different files, so zero conflicts; *adding* skills creates different new files, so also zero conflicts, which is the case indentation cannot reach. The consumer side looks cheap from here — `server.ts:15` is a single static import that casts and iterates `.skills`, so it becomes an index import. And the artifact stays committed, so the fresh-clone objection that killed the gitignore option does not apply to it.

Not asking for it here. This PR is a strict improvement that lands today, and if the split ever happens the indentation becomes irrelevant rather than wrong — a fine fate for a change this cheap. The number to watch is the 22-commits-in-30-days rate: if it holds, the split pays for itself quickly.

mlennie:
Thank you for this, and particularly for running the checks rather than describing them. Two of your three nits were things I had checked an adjacent fact about and drawn the wrong conclusion from, which is the most useful kind of review finding. Taking them in the order you raised them.

## 1. Two-space indent against `tabWidth: 3`

Taken, and this is the one I would not have found. I had checked the bundle was not in any `.prettierignore` and never once checked what `packages/server/.prettierrc` would do to it.

Reproduced exactly as you describe, including your byte count: prettier wanted 470,604 bytes against the 469,874 I had committed, same 284 lines, because it re-indents every level from 2 to 3.

The consequence you point at is the real one, not the warning. A contributor with format-on-save reformats the file merely by opening it, their entry fields land at nine spaces, my new indent test fails with `Expected: 56, Received: 0` about a file they did not knowingly touch, and regenerating flips it back, so they get to be confused twice. `prettier:check` globs only `ts,tsx`, so CI is blind in both directions and cannot help them.

Took your first option. The builder writes at width 3 and the indent test asserts `/^ {9}"name": /`. I verified that this actually makes prettier a no-op rather than assuming it: `JSON.stringify(bundle, null, 3)` plus a trailing newline came out byte-identical to `bunx prettier` output, and `bunx prettier --check` on the bundle passes now, including after two subsequent merges of main that changed its size. The test was also proven to fail on a re-minified file and on a two-space one, so a stale generator cannot quietly put this back.

One residual I would rather name than leave for the next person to find: the width is now coupled to that `tabWidth` setting and nothing tests the coupling. If `tabWidth` ever changes, the generator has to change with it or the file silently becomes unformatted-by-its-own-config again. That is exactly what your second option (`.prettierignore`) would have survived and mine does not, so I put the warning in a comment above the write pointing at the `.prettierrc`, and repeated it in the follow-up issue below.

## 2. `.gitattributes` and `text eol=lf`

You are right, and this was a real error in my description rather than a wording preference. I collapsed two different uses of `.gitattributes` into one dismissal. The per-clone registration objection is true of a merge driver, which is what I actually evaluated and rejected, and it is simply not true of `text eol=lf`, which git applies at checkout and commit in every clone with no setup at all, including for a contributor who has never heard of the file.

Added, scoped to the single path, in a sibling `.gitattributes` next to the bundle:

```
skills_bundle.json text eol=lf
```

It matters more here than it would have before this PR, since the file's line positions are now the point: `core.autocrlf` on a Windows checkout would rewrite all 284 lines and leave a contributor's diff permanently at odds with their own regeneration. You are also right that it retires the "not verified locally: the Windows leg" caveat for this file specifically, because CRLF is now unreachable rather than merely unobserved. I kept the test's CRLF normalization as insurance against that attributes file being dropped, and said so in the comment rather than leaving it looking like CRLF is expected.

On your request to split the two uses in the description: conceding the substance and handling it differently. I am not editing the PR body, because your comments are answers to that text and editing it would leave a reader seeing review comments that respond to sentences which no longer exist. So treat this comment as the correction to the description: `.gitattributes` was weighed and rejected **as a merge driver**, and that reasoning does not extend to the whole class. If this squash-merges, the stale paragraph is worth trimming in the commit-message box at merge time, which is editable and is the one place stale body text should not travel.

## 3. `byName2`

Agreed, and already done, but on a commit you could not see, so I want to be explicit rather than let this read as though I fixed it on request. `96a0cdb3` renamed it to `byCodepoint` and made `byName` delegate to it, and that commit was written on 08-20 and was still unpushed when you reviewed. You were looking at a stale tip for this one item.

```ts
const byCodepoint = (a: string, b: string) => (a < b ? -1 : a > b ? 1 : 0);

const byName = (a: SkillEntry, b: SkillEntry) => byCodepoint(a.name, b.name);
```

That commit came out of a round that also caught the comment above the order test overclaiming: it said the test held the builder to a locale-independent order when it only held the committed artifact to one. The test now also asserts the committed order equals the order the builder emits, which catches builder-versus-artifact drift, and the comment says plainly that no test can catch a `localeCompare` revert on a machine whose locale already agrees with codepoint order, which includes en-US and therefore CI.

## 4. The structural thought: filed, not built

Filed rather than done, and saying so because from outside those look identical and only one is true. It is recorded as a follow-up issue with your framing intact, including the part I think is strongest: adding skills is the one row of my own conflict matrix that indentation cannot reach, and it is the common case for a directory people are actively contributing to. I carried over that the consumer side is cheap, since `server.ts` has a single static import that casts and iterates `.skills` and would become an index import, and that a committed artifact sidesteps the fresh-clone objection that ruled out gitignoring the bundle. Also your number to watch, the 22-commits-in-30-days rate, and the `tabWidth` residual above, since a per-skill split would retire it entirely.

## Since you reviewed: the PR made its own argument

Worth flagging because it is evidence rather than assertion, and it arrived after your review. Merging main produced exactly one conflict, in `skills_bundle.json`, because `2bd5f46d` edited `skills/malloy-model/SKILL.md` and regenerated the bundle. A different skill from any I touched, and the single-line file collided anyway. That is precisely the failure this change removes, demonstrated against this change.

I resolved it the way the README in this PR now tells contributors to, by regenerating from the merged tree rather than hand-editing the JSON, and verified the resolution three ways rather than trusting the regeneration:

- No conflict markers, valid JSON, 56 entries, 284 lines, 56 nine-space name lines.
- Your `#(authorize)` additions are actually present in the regenerated bundle. I probed three distinct phrases in both the bundle and the source, because "I regenerated, therefore it is correct" is the assumption a silent revert hides behind.
- The parsed content is byte-identical to main's, sha256 `b7179a65` on both sides, so the diff remains provably whitespace-only and nothing of yours was dropped.

## Current numbers, superseding the description

Measured against main as it now stands rather than against my branch point: main's bundle is 471,175 bytes on a single line, mine is 473,818 across 284 lines, so **+2,643 bytes (0.56%)**. The description's `+1,913` was measured at the old indent width and before main grew the file, so treat these as the live figures.

Gate on the current tip: lint, typecheck and prettier clean (prettier reports only the six long-standing git-ignored generated files), unit 3179 pass / 0 fail across 150 files, integration 385 pass / 0 fail.

## Security pass over the changed delta

Since my earlier pass predated the indent change, the `.gitattributes`, the changed assertion and both main merges, I ran one over that delta. Clean, no findings, and the `.gitattributes` in particular got a deliberate look because it is a repo-wide mechanism rather than a file-local change:

- Scoping is exact, checked by enumerating every tracked file in that directory through `git check-attr` rather than by reading the pattern: only `skills_bundle.json` resolves to `eol=lf`; the two `.ts` files, the spec and the attributes file itself are unspecified.
- `eol=lf` cannot mangle content here. The bundle holds zero raw CR bytes and zero raw control characters, so the normalization is a no-op today, and a JSON string cannot legally contain an unescaped control character, so it cannot become one later.
- Your license headers do not leak into agent prompts: zero of 56 entries carry `SPDX-License-Identifier` or the copyright line in body or description, which incidentally confirms `stripLicenseHeader` works in the merged tree.
- Credential grep clean on the hand-written files. Inside the bundle the keyword hits are all documentation prose about auth patterns (signed tokens, `Publisher.setToken`, "password" as an example of sensitive data), which I characterized by reading the surrounding text rather than counting matches.

Finally, thank you for the note on framing, and for making the argument my description was defensive about. The lockfile parallel is the one I should have led with, and the `.ipynb` and `.pbxproj` counterexamples proving it from the other side is a better version of the point than I had.

packages/server/src/mcp/skills/skills_bundle.spec.ts
```diff
@@ -49,6 +50,57 @@ describe("skills_bundle.json (generated dual-channel asset)", () => {
...
+    * being dropped, not a claim that CRLF is expected.
+    */
+   it("is committed indented, one entry per line", () => {
+      const raw = fs
+         .readFileSync(path.join(import.meta.dir, "skills_bundle.json"), "utf8")
+         .replace(/\r\n/g, "\n");
+
+      // Needs no positive control: if this pattern stopped matching, the count
+      // would be 0 rather than quietly staying green.
+      const nameLines = raw.split("\n").filter((l) => /^ {9}"name": /.test(l));
```
Sha-Bang: **small thing. Three copies of the number 3, and nothing fails when they disagree.**

The indent width now lives in three places: `JSON.stringify(..., null, 3)` in the builder, `^ {9}` here, and `tabWidth` in `packages/server/.prettierrc`. Only a comment ties them together, and it ties the two that already agree.

Edit `.prettierrc` to `tabWidth: 4` and nothing anywhere fails. This test still passes, because the builder still emits nine spaces and the count is still 56. `prettier:check` globs `'./packages/**/*.{ts,tsx}'`, so CI never looks at a `.json` file, and there is no `.prettierignore` entry for the bundle. What comes back is precisely the failure `021a1775` fixed: the committed file is unformatted by its own package's config, a contributor with format-on-save rewrites all 284 lines by opening it, and *then* this test fails — `Expected: 56, Received: 0`, for a reason unrelated to anything they changed. Same symptom, one config edit away, and the guard that would catch it is the guard that cannot see it.

Read the width instead of restating it:

```ts
const tabWidth = JSON.parse(
   fs.readFileSync(path.join(import.meta.dir, "../../../.prettierrc"), "utf8"),
).tabWidth as number;
const indent = " ".repeat(tabWidth * 3);
const nameLines = raw
   .split("\n")
   .filter((l) => l.startsWith(`${indent}"name": `));
```

Then a `tabWidth` change fails here immediately, at the config edit, instead of ambushing whoever opens the file next — and the builder's literal `3` becomes the only one left. Worth reading `.prettierrc` there too if you want it airtight, though a red test is enough to force that edit.

Sha-Bang (approved):
re-reviewing at `ee5dd163`. All three nits from the 2026-08-21 pass are closed, and each one I checked rather than took on faith:

1. **two-space indent fights `packages/server/.prettierrc`** — closed. Width is 3 now, with a trailing newline, and `bunx prettier --check` on the bundle passes. You took the option that makes prettier a no-op rather than the `.prettierignore` one, which is the better of the two I offered.
2. **`.gitattributes text eol=lf` needs no per-clone setup** — closed, and the comment you wrote on it is better than my finding was. `git check-attr` confirms `text: set`, `eol: lf` on the bundle. Worth noting the input side was already covered: `build_skills_bundle.ts` normalizes CRLF when it reads each skill, so the round trip is now closed at both ends.
3. **`byName2`** — closed, it reads `byCodepoint`.

Regeneration is byte-identical to the committed file: sha256 `25433c4a…` both ways, 56 entries, 284 lines. And the reformat really does change no content — the base bundle and this one are string-identical under `JSON.stringify`, both re-minifying to exactly 471,175 bytes.

The five guards in `skills_bundle.spec.ts` are individually load-bearing, which is unusual enough to be worth stating. I broke the committed bundle five ways, and each break was caught by exactly one test, the right one:

| what I broke | caught by |
| --- | --- |
| re-minified the file | "is committed indented, one entry per line" |
| re-indented at width 2 | same |
| stripped the trailing newline | same |
| moved `malloy-charts` after `malloy-h*` (the cs-CZ shape) | "is committed in codepoint order, and in the order the builder emits" |
| truncated one entry's body | "is in sync with skills/" |

I also stood up the production consumer, since inlining a JSON that now has real newlines is the one thing here that could plausibly matter and the description lists the Docker leg as unverified. Booted the built `dist/server.mjs`: 56 prompts served, the served name set exactly equal to the bundle's entries, and the longest body (`malloy-gotchas-modeling`, 28,980 chars carrying 400 newlines) byte-identical to its bundle entry.

Gate: server typecheck and eslint clean; prettier clean on the three changed source files and on the bundle. Unit 3176 pass / 3 skip / 0 fail. Integration 382 collected, the only attributable failure the GCS-backed `Concurrent package operations (E2E)`, which fails on `origin/main` too.

Approving. One small thing inline, worth taking but not worth another round. One nit here rather than inline, because it is the description rather than the code: the measurements are all from the superseded two-space revision, and a reader cannot tell which revision they describe. The growth is **+2,643 bytes / +0.56 percent** (471,175 → 473,818), not +1,913 / +0.41; the sha is `25433c4a…`, not `3b614d93…`; and the gate numbers are 3176 unit and 382 integration, not 3008 and 313. The prose is all still accurate — it is only the figures.

packages/server/src/mcp/skills/skills_bundle.spec.ts
mlennie: Thank you for re-checking each nit rather than trusting the resolution, and particularly for standing up the production consumer. Inlining a JSON that now contains real newlines was the one thing in this change that could plausibly have mattered at runtime, and booting `dist/server.mjs` to confirm 56 prompts with the served name set equal to the bundle's entries is the check I would have wanted. The five-ways-broken exercise is the better compliment though: if each break is caught by exactly one test and it is the right one, that says more about the suite than a green run does.

Your CRLF observation adds something I had not claimed. I argued the output side, that `text eol=lf` makes a CRLF checkout impossible. You are right that the input side was already covered, since `build_skills_bundle.ts` normalizes CRLF as it reads each skill, so the round trip is closed at both ends and a CRLF-committed `SKILL.md` cannot reach the bundle either. That is a stronger statement than mine and it was true before this PR.

**The three copies of 3: taken, and it was my own residual coming back.** You are right that nothing failed when they disagreed, and right about why: `prettier:check` globs only `ts,tsx`, so CI never looks at a `.json`, and the comment I had added tied together the two copies that already agreed. One `tabWidth` edit reintroduced exactly the trap the previous commit fixed, with the guard unable to see it.

The spec now reads the width instead of restating it: it parses `packages/server/.prettierrc`, derives the expected indent as `tabWidth * 3` (entry keys being three levels deep), and matches with `startsWith`. Verified by making the edit rather than reasoning about it: with `tabWidth: 4` the test fails `Expected: 56, Received: 0`, and it passed before this change. The builder's literal `3` is now the only remaining copy, and it is guarded. Config restored and confirmed identical to `HEAD`.

**Third conflict, same file, resolved the same way.** Main gained `c42f4463`, which edits skills and regenerates the bundle, so `skills_bundle.json` conflicted again. That is three times on this branch, which is the argument the PR makes more plainly than the description does. Regenerated from the merged tree, and verified the resolution rather than trusting it: my bundle's parsed content is byte-identical to main's, and I probed two distinct phrases from your new text in both the bundle and the skill sources to confirm nothing of yours was dropped.

**Figures, and I would rather explain the discrepancy than quietly replace yours.** Yours were correct at the tip you measured; main has moved twice since, so both sets are right about different trees. Measured just now at `29164d36`:

- main's bundle 478,230 bytes on one line, mine 480,873 across 284 lines, so **+2,643 bytes (+0.55%)**
- re-minifying mine gives exactly 478,230, so it is still string-identical to base under `JSON.stringify`
- content sha256 `23c77ad1`, identical on both sides
- 56 entries, gate **3174 unit / 0 fail** and **387 integration / 0 fail**

You are right that the description's figures are all from the superseded two-space revision and a reader cannot tell which tree they describe. Per our standing rule I am not editing the body, since your comments are answers to that text; treat this comment as the correction, and the stale figures are worth trimming from the squash commit-message box at merge time so they do not enter history uncorrected.

Last thing, on ordering: your note on #1047 that this should land first is right and it is now the cheaper path by a wide margin. Landing this first means #1047 regenerates; the other order means resolving a conflict inside a single 478 KB line, which is the thing this PR exists to stop.

PR #1059, opened 2026-08-21
feat(mcp): tell an agent why nothing is being served

mlennie:
## The problem, measured

Point a Publisher at a directory that is not set up and it reports perfect health. Measured against a real server on current `main`:

```
malloy_getStatus  -> {"operationalState":"serving","initialized":true,"environments":[]}
malloy_getContext -> {"results":[]}
PUBLISHER_READY url=... environments=0 packages=0 load_errors=0
```

No load errors, because nothing was ever attempted. The agent has nothing to act on. Three things make that worse than a silent dead end. The MCP instructions send the agent straight here ("check malloy_getStatus for load errors ... before concluding there is no data"), and this is exactly where that advice runs out. Guessing a name gets `Resource not found` with a suggestion to "Check capitalization and path separators", so the agent goes hunting a typo when the directory simply has no config. And the agent really is in this position, because the server writes `.mcp.json` into an empty directory, so a session opened there connects successfully to a server that serves nothing.

`malloy_getStatus` cannot help here as it stands. There is no load error to report.

The other half is the reason this is one field and not a new tool: where something *was* attempted, `loadErrors` already reports it. Those cases were a message-quality problem all along, not a missing-tool problem.

## What this adds

A conditional `setup` block on `malloy_getStatus`, returned only when the server is up and serving no package at all. It follows the existing `loadErrors` precedent of appearing only when something is wrong, so the healthy response is unchanged and a healthy deployment never produces it.

```json
"setup": {
  "problem": "No publisher.config.json in /work/demo, so the .malloy files there are not part of any package and nothing is served.",
  "nextAction": "Scaffold a workspace with `npm create @malloy-publisher/malloy-package@latest` and move the model into it, or hand-write publisher.config.json ...",
  "serverRoot": "/work/demo",
  "configFile": null,
  "unclaimedModelFiles": ["sales.malloy"]
}
```

It is read-only. Creating a workspace stays with `npm create @malloy-publisher/malloy-package`; this diagnoses and points at it. There is no `api-doc.yaml` change: the block is computed in the MCP tool from the already-public `serverRootPath`, so no generated client moves.

It also carries the facts that make the existing failure messages interpretable, without changing either message: `unservedPackages[].sourcePath` is the path a user can actually edit, where the load error names the server's own copy under `publisher_data/`, and `configuredConnections` makes an unresolved connection name legible.

## Why one field rather than a new tool

The plan this came from named an `init_workspace` / `validate_workspace` pair. Creation is already covered by the scaffolder, and diagnosis belongs where the agent is already sent, so answering inside `malloy_getStatus` closes the loop in one call. A separate tool would make it two, and only if the agent picked the right one.

## The cases it separates

Collapsing any of these produces advice worse than silence, so each gets its own branch: no config at all; environments that came from stored state rather than a file; a config that does not parse; a config rejected wholesale because one `${VAR}` anywhere in it is unset; a `--config` path that is missing, and one that is a directory; the bundled sample config, whose environments do exist and whose packages are fetched over the network; a package declaring no usable `location`; a location that points at nothing; a location that points at a file rather than its directory; a local package missing its manifest; and a remote location, which has no local path at all.

## How it was verified, and what that changed

Six directory states were run against a live server before and after. The control and the bundled zero-argument path return no `setup` key.

The advice was executed rather than read, and that is what corrected it. Following the first draft literally did not work: creating the manifest and restarting left the package still not loading, now with no load error at all, because the server boots `Loading from existing database` and keeps serving its own copy. So the advice asks for `--init`, keyed on whether the store already has environments rather than on whether `publisher_data/` exists, since `initialize()` creates that directory unconditionally.

Where it asks for `--init` it says what `--init` costs: it removes `publisher_data/` and reloads from the config alone, so an environment or package created over the REST API and absent from the config does not come back. The stored-state branch does not offer `--init` at all, because with no config to rebuild from it would leave the server with nothing. The scaffolder template carried the same incomplete warning, naming materialized tables and saved themes but not API-created packages, so its one sentence is completed here.

The block is withheld until the server has finished starting. The MCP endpoint has no readiness gate and `getStatus` skips the initialization wait, so a cold boot is indistinguishable from an un-set-up directory. Measured on a bundled-config boot: 329 responses were served before the server reached `serving`, every one of which would otherwise have told the agent to restart.

Config strings are read the way the loader reads them. `processConfigValue` interpolates `${VAR}` in every string, so all four this diagnostic reads go through the same helper. The location is echoed back as declared, because that is the text a human edits; the names are echoed substituted, because they exist to be matched against `loadErrors`. That asymmetry is deliberate and is documented in the code.

## Why this took several rounds

Three of the fixes created the next finding, and each was caught by reviewing the fix rather than the feature. Exempting the bundled config surfaced the startup-gate case. Keying the restart advice on `publisher_data` was wrong in a way only the persisted-environment question exposed. Turning the path check into a boolean collapsed "points at a file" with "points at nothing", which need opposite fixes and can occur in the same config. Three specs also had to be rewritten because each faithfully pinned the behaviour its round believed was correct, and a later finding overturned the behaviour.

Gate: typecheck, lint, prettier clean, 3044 unit and 385 integration passing, plus the scaffolder's own 376. The module's 34 specs were mutation-tested rather than assumed: each defect-specific spec was watched failing before its fix went in.

## Security

The block emits filesystem paths, which the rest of this tool deliberately does not. What bounds it is that it is returned only when the server is up and serving nothing, so a healthy deployment never produces it, and that the unauthenticated REST `/status` already returns each environment's absolute location. What does *not* hold, and what an earlier version of this comment wrongly claimed, is that a load error is always there carrying the path anyway: in the flagship case nothing was attempted, so this block is the sole disclosure. Connection names only, never attributes, with a test that plants a password in the config and asserts it cannot appear. Symlinks are skipped in both roles, so the bounded directory scan cannot be walked out of.

## Docs

`docs/ai-agents.md` said this tool "carries no connection attributes, locations, or row data", which the setup block makes untrue. Corrected there, and both it and `AGENTS.md` now describe the block.


mlennie:
Three things a reviewer should know that are not in the PR body, posted here rather than edited into it so the body stays as it stood when the PR opened.

**1. The security review has not read the shipped code.** It ran once, early, and came back clean, but against a version of `workspace_setup.ts` that has since been substantially rewritten. Since that review the module gained filesystem probing, `${VAR}` substitution over config text, raw JSON parsing of an arbitrary config file, and more emitted paths. The body describes the security posture and I believe the posture is right, but please do not read the "Security" section as "reviewed". It is a stated intent, not an audit of this diff.

**2. The tip carries a one-line delta nothing has reviewed.** The last review round came back clean, and I then made one further edit rather than ship a class fix half-done: the missing-manifest branch pointed at "the publisher_data path named in loadErrors" unconditionally, and that branch is reachable with no loadErrors at all. It is now conditional, with a spec, mutation-tested, and verified live with loadErrors present and by unit test with them absent. Small, but unreviewed, and that is the line.

**3. A design decision worth questioning.** The advice about when a plain restart suffices versus when `--init` is needed is keyed on whether any environment is currently being served. That is a proxy. The real trigger inside the server is whether the persisted database holds environments, and this MCP tool cannot read that cheaply. The proxy has one known failure: an environment that is in the database but throws during restore is dropped from the served list. Rather than guess, the wording covers both outcomes, so the tool is deliberately honest about something it cannot know instead of confidently wrong. If you would rather it read the real state, that is a legitimate ask and a larger change than this PR.

Two smaller notes. The one-clause change to `packages/create-malloy-package/templates/AGENTS.md` is a deliberate widening beyond the server package: a PR arguing that `--init` advice must state its cost cannot ship while the scaffolder keeps writing the incomplete warning into every new project. And one ambiguity was consciously declined rather than fixed, where "no model files found" and "the directory scan failed" share a branch, on the grounds that a server root the server itself created and is serving from being unreadable is not a state worth branching on.

On why this took six review rounds: three of the fixes created the next finding. That is in the body with the three instances named, and it is the honest answer rather than a process description.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,557 @@
...
+   // One unset ${VAR} anywhere in the file discards the WHOLE config:
+   // getPublisherConfig substitutes over the entire tree, and
+   // reloadEnvironmentManifest catches the throw and returns zero
+   // environments. Nothing is attempted, so there are no loadErrors either.
+   // Verified live: a config with one healthy package and one unset variable
+   // on a connection served nothing, reported no load errors, and would
+   // otherwise be described here as "see loadErrors for why each one failed"
+   // beside a package listed as perfectly healthy.
+   try {
+      processConfigValue(parsed);
```
Sha-Bang, line 421: **This branch and #1049 contradict each other, and both are open in this batch.** They were posted as independent; on this point they are not.

Today, one unset `${VAR}` anywhere discards the whole config, which is what this branch's wording rests on: "one unset variable rejects the ENTIRE file, so no environment was created and nothing is served". #1049 removes exactly that: it substitutes per environment entry, so a single unset variable drops **one** environment, keeps the healthy ones, and records a load error naming the variable.

After #1049 lands, two things change here:

1. The wording is false. The other environments were created and are serving.
2. In the common case this branch is unreachable — healthy environments mean `servedPackages > 0`, so `describeWorkspaceSetup` returns `undefined` well before it. It only fires when *every* environment is broken, and then it still explains the mechanism wrongly.

Note the direct read of `processConfigValue(parsed)` is what keeps the branch firing at all after #1049, since that walker still throws on the whole tree; it is the attributing wrapper that changes. So this will not fail loudly — it will just quietly describe the old server.

Whichever of the two lands second needs this branch rewritten to "one or more environments reference an unset variable; loadErrors names which", and its `nextAction` to stop saying every package in the file is affected. This is the third instance of the same shape in this batch — #1047 hardcodes the pre-#1049 `load_errors=0` behaviour in three user-facing places — so it may be worth landing #1049 first and rebasing the prose in the other two onto it.

Sha-Bang (commented):
This is the best-motivated PR in the batch. "The server reports perfect health and serves nothing, and MCP_INSTRUCTIONS routes the agent to the one tool that by construction cannot help" is a real dead end, and the `.mcp.json`-into-an-empty-directory detail is what makes it a common one rather than a hypothetical.

Answering inside `malloy_getStatus` rather than adding a tool is the right call for the reason given: the agent is already sent there, and a second tool is a second call it has to choose correctly. Conditional presence mirroring `loadErrors`, read-only, no `api-doc.yaml` movement, and degrading to the plain status payload when the diagnosis itself throws — all right.

One finding is a cross-PR collision with #1049, which is in this same batch and is not independent of this one despite both being listed that way.

- **The security precedent is real.** I traced it because it is load-bearing: `Environment.metadata.location` is `this.environmentPath`, it is spread into `serialize()`, `listEnvironments` returns those, and the status builder pushes them into the response — so unauthenticated REST `/status` genuinely does already return absolute environment paths. The claim holds.
- `PUBLISHER_CONFIG_PATH` is the right variable (`server.ts:113` sets it from `--config`), so the explicit-config branch actually fires.
- **40/40** across `workspace_setup.spec.ts` and `get_status_tool.spec.ts`.
- The graceful-degradation path works, and I saw it fire during the test run rather than trusting it.

## Worth saying about the process

Three rounds where each fix surfaced the next finding, three specs rewritten because they pinned behaviour a later round overturned, and advice that was **executed rather than read** — which is what caught that a plain restart keeps serving the stored copy and the guidance needed `--init`. That last one is the difference between advice that works and advice that sounds right, and it is not the sort of thing a reviewer catches by reading.

---

**Must-fix:** the unset-`${VAR}` branch contradicts #1049
**Should-fix:** the symlink guarantee holds at one depth only
**Nits:** nondeterministic list contents · uncapped `configuredConnections` · a benign case logged at warn

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,557 @@
...
+      return found;
+   }
+   for (const entry of entries) {
+      if (found.length >= MAX_LISTED) break;
+      if (entry.isFile() && entry.name.endsWith(MODEL_FILE_SUFFIX)) {
+         found.push(entry.name);
+      } else if (entry.isDirectory() && !isSkippedDir(entry.name)) {
+         let children: string[];
+         try {
+            children = await fs.promises.readdir(
```
Sha-Bang, line 139: The symlink guarantee above holds at depth 1 only. This second-level `readdir` takes no `withFileTypes`, so there is no type check on `child` — a symlinked `.malloy` inside a real subdirectory is listed.

Demonstrated against this branch, calling the exported `describeWorkspaceSetup` on a root holding a real and a symlinked model at each depth plus a symlinked directory:

```
unclaimedModelFiles: [ "sub/real_sub.malloy", "sub/link_sub.malloy", "real_root.malloy" ]
```

`link_root.malloy` is correctly skipped, `linked_dir` is correctly not descended, and `sub/link_sub.malloy` is listed. So the docstring's "Symlinks are skipped in both roles ... (verified)" is true of the top level and not of the level below it, and `grep -n symlink` finds nothing in the 700-line spec — the verification was manual, and it only covered the half that works.

The disclosure is a filename inside a directory the server root owns, so I would not call this urgent. But the module states this invariant explicitly and errs deliberately toward listing too little, and the fix is `{ withFileTypes: true }` plus the same `isFile()` test as above.

While here: `readdir` order is filesystem-dependent and `MAX_LISTED` truncates at 10, so *which* files appear varies by platform on any root with more than ten. A `.sort()` before returning makes the field stable and the cap deterministic — cheap, and it matters for a field an agent may compare across calls.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,557 @@
...
+            ? interpolate(environment.name)
+            : "(unnamed)";
+
+      if (Array.isArray(environment.connections)) {
+         for (const connection of environment.connections) {
+            const name = (connection as { name?: unknown })?.name;
+            if (typeof name === "string") {
+               const resolvedName = interpolate(name);
+               if (!configuredConnections.includes(resolvedName)) {
+                  configuredConnections.push(resolvedName);
```
Sha-Bang, line 496: Nit: `configuredConnections` is the one list here with no cap. `unservedPackages` stops at `MAX_LISTED` and reports `unservedPackagesTruncated`, `unclaimedModelFiles` stops at `MAX_LISTED`, and this one grows with the config. A config declaring a connection per tenant emits all of them.

Same `MAX_LISTED` treatment would make the bound uniform, and the field's purpose — making an unresolved name legible — is served just as well by the first ten.

packages/server/src/mcp/tools/get_status_tool.ts
```diff
@@ -51,7 +65,25 @@ export function registerGetStatusTool(
...
+         // so its own failure must never turn a healthy status call into a
+         // tool error; it degrades to the status payload without it.
+         let setup;
+         try {
+            setup = await describeWorkspaceSetup(
+               environmentStore.serverRootPath,
+               payload,
+            );
+         } catch (setupError) {
+            logger.warn("[MCP Tool getStatus] setup diagnosis failed", {
```
Sha-Bang, line 82: This fires on a benign case, which I noticed because it shows up in the test output:

```
warn: [MCP Tool getStatus] setup diagnosis failed {"error":"The \"paths[0]\" property must be of type string, got undefined"}
```

That is `path.resolve(undefined)` from a store with no `serverRootPath`. Production always has one, so this is only test noise today — but "cannot diagnose because there is no server root" and "the diagnosis hit an unexpected fault" are different events and only the second deserves a warn.

An early `if (!serverRoot) return undefined;` in `describeWorkspaceSetup` separates them, and keeps this catch meaning what it says.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,557 @@
...
+ * The signal is whether this server already has environments, NOT whether
+ * `publisher_data/` exists. That directory is created unconditionally by
+ * `initialize()`, so testing it would demand `--init` in every case.
+ *
+ * `--init` is destructive and the advice must say so: it removes
+ * `publisher_data/` outright and reloads from the config manifest alone, so an
+ * environment or package created over the REST API, and absent from the config
+ * file, does not come back.
+ */
+function restartPhrase(hasPersistedEnvironments: boolean): string {
```
Sha-Bang, line 249: Observation rather than a request. These two strings assert specific server behaviour — that `initialize()` prefers stored environments over the config file, and what `--init` destroys — measured on 2026-08-20, with nothing tying the assertion to the code that implements it. If `initialize()`'s precedence changes, this advice becomes confidently wrong and no test fails.

That is the same failure mode as the #1049 collision above, and as #1047's three copies of the `load_errors=0` behaviour: prose about server behaviour, living far from the behaviour, with no link between them. Worth a test that asserts the precedence this text depends on, so the text has something to break it.

mlennie:
All six addressed. Taking them in the order you raised them, with the last section being something you did not ask for but should see before approving.

### Must-fix: the unset-`${VAR}` branch contradicting #1049

Fixed, and thank you for tracing why it would fail silently rather than loudly. That was the part I would not have found: the branch reads `processConfigValue` directly, which still throws over the whole tree after #1049 since only the wrapper that attributes the failure moves, so it would have kept firing and quietly described the old server.

Rather than rewrite for the post-#1049 world and flag it, I made the wording true under both merge orders using a guard already in the file: round 5 added a `hasLoadErrors` check for a similar problem, and the same guard answers this. With load errors present it says `loadErrors` names which environments could not be created; with none, the config was rejected before any environment was created. It never states the mechanism, and "Every package in this file is affected" is gone. So #1049 can land in either order with no follow-up owed here.

There is a comment at the branch saying not to reintroduce a sentence asserting the whole file is rejected, and why: that sentence goes false while the code keeps working, so only the prose lies.

### Should-fix: the symlink guarantee holding at one depth only

Confirmed by reproducing your scenario before changing anything, and I got your exact output. Cause is as you say, the second `readdir` took no `withFileTypes`.

The more useful half of this finding is the part about the test: grep for `symlink` in that 700-line spec returns zero, so the invariant was asserted in a docstring, verified by hand on the half that works, and never written anywhere that could break. Fixed at both depths and added a spec pinning both, because the missing test is the actual defect and the one-line fix alone would leave the same hole for the next change.

### `readdir` order and the cap

Taken. Sorted before the cap rather than after, so it is the first ten by name rather than whichever ten the filesystem handed back, which makes the cap deterministic too. Agreed it matters for a field an agent may compare across calls.

### Nits: uncapped `configuredConnections`, and the benign case logged at `warn`

Both taken. `configuredConnections` is capped at the same `MAX_LISTED` with its own truncation flag.

On the warning: you found that by reading my own test output, which I had been looking at all week without treating it as a signal. There is now an early return when there is no server root, so the catch keeps meaning the diagnosis faulted rather than also meaning there was nothing to diagnose.

### Your observation: prose asserting behaviour, with nothing to break it

The most valuable comment in the review, and I have taken it as a deliverable rather than a note. Both claims those strings make were established by running a server and watching it, with nothing connecting the assertion to the code. There is now a spec pinning the precedence three ways, and flipping the branch in `initialize()` makes it fail.

Worth adding, since it sharpens your point: during the pre-review rounds I had the inverse problem three times, specs that faithfully pinned behaviour a later finding overturned and stayed green while asserting the wrong thing. A fourth turned up while fixing your must-fix, the one asserting "ENTIRE file", now deleted. So this area has produced both failure modes, which argues for the link you want rather than for more of either alone.

### And one thing you did not ask about

One note that is not a reply to a comment, posted here so it is on the record before anyone approves.

**A security lens has now read this version, and it found three secret-disclosure defects, all introduced by earlier fixes on this branch, all closed and pinned by tests.**

When you reviewed, the honest statement would have been that the security pass had run once, early, against a module since rewritten across six rounds and rebased. Rather than leave it there, it was re-run against the shipped code. What it found:

1. `${VAR}` substitution was being applied to environment, connection and package **names** so they would match `loadErrors`. A name is an arbitrary config string, so a connection named `${TOKEN}` emitted the token's value. Confirmed by probe before fixing, not reasoned about: `configuredConnections` came back carrying the secret. Names are now echoed as declared.

2. That first fix was incomplete. The declared `location` was echoed safely, but `sourcePath` is the **resolved** path, so `./${SECRET}/models` still reached the payload. Rather than patch the second instance, the class was then closed by enumeration: a secret planted in every config string this module reads, across both the load-error and no-load-error states. Nothing substituted reaches the payload.

3. The parse-failure branch echoed the `JSON.parse` message. V8 builds "Unexpected token" from a window of the **raw source**, so a syntax error adjacent to a literal password put part of that password in the response. Verified on node v24. Worth knowing why the suite could never have caught it: JavaScriptCore does not echo source, so under Bun the same input produces a message with no source at all, while the published bin has a Node shebang. The guard pins the real V8 string as a literal so it works under Bun.

Two things about that third one are worth more than the fix. The enumeration in (2) was drawn at the wrong boundary — around substituted values, the mechanism just fixed, rather than around the question that matters, which is what config content can reach an unauthenticated response by any path. And the first test written for it passed with the call site reverted; a spec over the sanitiser function alone does not pin the call that uses it. Both are fixed, and both were found by mutation rather than by reading.

The three pointers in your review were all re-derived rather than trusted: the symlink boundary against a matrix of symlinked files and directories at both depths plus a hardlink, the `configuredConnections` values (which is how finding 1 surfaced), and whether the absolute paths are attacker-reachable through any MCP tool, which they are not.

packages/server/src/mcp/tools/get_status_tool.spec.ts
```diff
@@ -102,4 +130,70 @@ describe("malloy_getStatus", () => {
...
+                  environments: [{ name: "local", packages: [{ name: "p" }] }],
+               }) as never,
+         });
+         expect("setup" in parse(await handler())).toBe(false);
+      } finally {
+         fs.rmSync(root, { recursive: true, force: true });
+      }
+   });
+
+   it("still answers when the setup diagnosis itself fails", async () => {
```
Sha-Bang: **this is the one I'd want fixed before merge. Make the test throw.**

`serverRootPath: undefined` never reaches the `catch` this test is named for. `describeWorkspaceSetup` opens with `if (!serverRoot) return undefined`, which was added deliberately so a store with no server root is not logged as a failure — so this exercises the early return and the tool sees no exception.

I checked by deleting the guard: replaced the whole `try`/`catch` with a bare `const setup = await describeWorkspaceSetup(...)` and ran the full unit suite. 3096 pass, 0 fail, this test included. So the line that keeps a diagnosis fault from turning a healthy `malloy_getStatus` into a tool error has no coverage, and the test's name says it does.

That matters more here than it usually would, because everything else in this module is pinned — every one of the seven mutations I tried turns the specs red. This is the single hole, and it is in the guard rather than the feature, which is where a hole survives longest.

Force a real throw; a getter is enough and needs no new export:

```ts
const handler = captureHandler({
   get serverRootPath(): string {
      throw new Error("boom");
   },
   getStatus: async () => ({ /* serving, no environments */ }) as never,
});
```

then assert `isError` falsy, `operationalState` is `"serving"`, and no `setup` key. If you would rather concede the catch is belt and braces — nothing in `describeWorkspaceSetup` can currently throw, every filesystem call inside it is already wrapped — then rename this test to what it does pin, "returns no setup block when there is no server root", so the suite stops advertising coverage it does not have.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,651 @@
...
+      };
+   }
+
+   // Read the config directly rather than through getPublisherConfig, which
+   // swallows a parse failure and returns an empty environment list. That is
+   // indistinguishable from a valid config declaring nothing, and the two need
+   // opposite advice.
+   let parsed: unknown;
+   try {
+      parsed = JSON.parse(fs.readFileSync(configFile, "utf8"));
```
Sha-Bang: **non-blocking. Split the read from the parse.**

Both are in one `try`, so every read failure is reported as a syntax error. `chmod 000` on a config that is perfectly valid JSON:

```
problem:    "<root>/publisher.config.json could not be read as JSON, so no environments were configured and nothing is served."
nextAction: "Fix the JSON in <root>/publisher.config.json, then restart ... Adding a package will not help until the file parses."
```

The JSON is fine. That is the standard container mistake — config baked in root-owned, server running as a non-root uid — and this sends the operator to edit a correct file when the fix is one `chmod`. It is the same failure the module exists to remove, one layer down: "telling someone a path does not exist that `ls` plainly shows" is why the `--config` branch above splits "is not a file" from "does not exist", and this branch does not get the same treatment.

Read into a local in its own `try` and report `error.code` (`EACCES`, `EISDIR`) as a read failure naming permissions and owner, then `JSON.parse` the string in the existing one.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,651 @@
...
+                  configuredConnectionsTruncated = true;
+               }
+            }
+         }
+      }
+
+      for (const rawPackage of Array.isArray(environment.packages)
+         ? environment.packages
+         : []) {
+         if (!rawPackage || typeof rawPackage !== "object") continue;
```
Sha-Bang: **non-blocking. Count the entries you skip.**

`declaredPackageCount++` four lines down deliberately counts a package whose location is unusable, and its comment says why: "dropping it would report a config that plainly declares a package as declaring none". This `continue` reintroduces exactly that one level up, and so does the environment-level `continue` at :477.

Probed against three shapes a hand-written config plausibly takes:

```
{"environments":[{"name":"local","packages":["sales"]}]}
{"environments":["local"]}
{"environments":{"local":{"packages":[{"name":"s","location":"./s"}]}}}
```

All three return `"<config> declares no packages, so nothing is served."` with `"Add a package to an environment in <config>"`. For the second and third that is a dead end — there is no environment object to add a package to. And mixing a skipped entry with a real one makes the ratio lie: `{"packages":["sales",{"name":"ops","location":"./nope"}]}` reports `1 of 1` for a file that declares two.

Count the skipped entries, fold them into `declaredPackageCount`, and give them a branch ahead of the `declaredPackageCount === 0` one that names the shape ("N of M entries under `packages` are not objects"). Reading the raw tree is stated to be for configs the validated loader rejects, so a malformed entry is the expected input on this path, not the exotic one.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,651 @@
...
+async function findUnclaimedModelFiles(serverRoot: string): Promise<string[]> {
+   const found: string[] = [];
+   let entries: fs.Dirent[];
+   try {
+      entries = await fs.promises.readdir(serverRoot, { withFileTypes: true });
+   } catch {
+      return found;
+   }
+   for (const entry of entries) {
+      if (entry.isFile() && entry.name.endsWith(MODEL_FILE_SUFFIX)) {
```
Sha-Bang: **small thing while you're in here. A notebook is a model too.**

This matches `MODEL_FILE_SUFFIX` only, so a directory whose work is `sales.malloynb` gets the generic branch: "no environments are configured and nothing is served" plus "Scaffold a workspace ... Do not create files under publisher_data". It loses both the better sentence ("the files there are not part of any package") and the file list, which is the concrete part.

That is a normal shape — Publisher serves notebooks, and the bundled `malloy-notebooks` skill tells an agent to author them — and `NOTEBOOK_FILE_SUFFIX` is exported from the same `constants` module this file already imports from.

Match both suffixes here and at :154, and the `hasModels` wording a few branches down needs to stop saying `.malloy` specifically.

packages/server/src/mcp/tools/get_status_tool.ts
```diff
@@ -7,15 +7,21 @@ import { EnvironmentStore } from "../../service/environment_store";
...
 
 ## Response
 A JSON object with:
 - operationalState: "initializing" | "serving" | "throttled" | "draining".
 - initialized: whether startup finished.
 - environments: each environment's name with its loaded package names.
 - loadErrors (only present when something failed): entries of {environment, package?, message, stale?, failedAt?}. An entry WITHOUT stale means the package (or whole environment, when package is absent) did not load and is missing from environments. An entry WITH stale: true means the package IS serving, but its most recent reload failed to compile, so the model answering queries is OLDER than the files on disk; the message says why. Fix the file and reload (malloy_reloadPackage) to clear it.
+- setup (only present when the server is up and serving no package at all): {problem, nextAction, serverRoot, configFile, unservedPackages?, unservedPackagesTruncated?, configuredConnections?, unclaimedModelFiles?}. Read problem and nextAction; the rest is supporting detail. unservedPackagesTruncated true means the list is capped and does NOT name every package.
```
Sha-Bang: **nit. The key list is missing one.**

`configuredConnectionsTruncated` is emitted but not listed. A config declaring twelve connections returns:

```json
"configuredConnections": ["c0","c1","c2","c3","c4","c5","c6","c7","c8","c9"],
"configuredConnectionsTruncated": true
```

An agent reading only this contract sees ten connections and no signal the list is short, which is the thing the `unservedPackagesTruncated` sentence right here exists to prevent.

While you are in the string: `nextAction` tells the agent to read `sourcePathKind "file"`, `sourcePathKind "missing"` and `hasManifest false` by name, and none of the three appears in the response contract at all. Add the missing key, plus one clause giving the `unservedPackages` entry shape.

packages/server/src/service/environment_store_config_precedence.spec.ts
```diff
@@ -0,0 +1,137 @@
...
+      // serves the store's set. If this ever starts returning "from-config",
+      // the `--init` advice in workspace_setup.ts is wrong and must change
+      // with it.
+      storedEnvironments = [
+         { id: "e1", name: "from-store", path: path.join(serverRootPath, "x") },
+      ];
+      const store = new EnvironmentStore(serverRootPath);
+      await store.finishedInitialization;
+      const names = (await store.listEnvironments(true)).map((e) => e.name);
+      expect(names).not.toContain("from-config");
```
Sha-Bang, line 131: **small thing. Assert the positive half too.**

`not.toContain("from-config")` is satisfied by an empty list, and empty is what comes back: `listEnvironments(true)` returns `[]` here. `from-store`'s path is `<root>/x`, which does not exist and is not in the config, so `initialize()` takes the "files missing" branch and logs `Environment "from-store" not found in config and files missing`. The comment above says "a plain boot serves the store's set"; nothing in this test shows that it does.

It does catch the flip it names — I changed `if (existingEnvironments.length > 0)` to `if (false)` and this went red with `Received: [ "from-config" ]` — so this is strengthening rather than a rescue. `mkdirSync` the stored environment's path in `beforeEach` and add `expect(names).toContain("from-store")`, and both halves of the sentence the `--init` advice rests on are pinned.

Separately, the mock repository has no `listStorageDestinations`, so every run of this file logs `Error syncing storage destinations for "from-config": repository.listStorageDestinations is not a function` with a stack. Add it as `async () => []`.

AGENTS.md
```diff
@@ -130,7 +130,7 @@ stdio-only clients (older Claude Desktop) bridge through mcp-remote:
...
-- `malloy_getStatus`: the server's health. Its operational state, each environment's loaded packages, and every configured package that failed to load or is serving a stale model. This is the only MCP surface where a load failure is visible on its own, so call it before concluding a package is empty or missing, and after a model edit whose reload you did not run yourself.
+- `malloy_getStatus`: the server's health. Its operational state, each environment's loaded packages, and every configured package that failed to load or is serving a stale model. When no package is being served at all it also returns a `setup` block saying why the directory is not set up and what to do about it, which is the one case a load error cannot cover, because nothing was attempted. This is the only MCP surface where a load failure is visible on its own, so call it before concluding a package is empty or missing, and after a model edit whose reload you did not run yourself.
```
Sha-Bang, line 133: **non-blocking. REST keeps the dead end, and this file still says it doesn't.**

This line documents the block for MCP, but line 170 of the same file still calls `GET /api/v0/status` "the REST equivalent of `malloy_getStatus`", and after this PR it is not.

Section 7 is the part that bites. It tells an agent to use REST precisely when it started the server itself and there is nobody to reconnect its MCP client — the same population most likely to be sitting in a directory that was never set up. That agent gets `{"environments":[],"initialized":true,"operationalState":"serving"}` and nothing else, which is the state this PR exists to end. I confirmed it on the empty root: REST returns that, MCP returns the block.

The diagnosis is computed from `serverRootPath` plus the status payload, both of which the REST handler already has, so adding it there is a schema change rather than new logic. It does widen an unauthenticated disclosure, which is a reason to decide rather than default. So either put it on `/api/v0/status`, or fix line 170 and say in section 7 that the diagnosis is MCP-only and an unattended agent has to read the config itself.

packages/server/src/mcp/tools/workspace_setup.ts
```diff
@@ -0,0 +1,651 @@
...
+} from "../../constants";
+import {
+   processConfigValue,
+   resolvePublisherConfigPath,
+   substituteEnvVars,
+} from "../../config";
+import { resolvePackageLocation } from "../../service/environment_store";
+
+/**
+ * Why this exists, measured 2026-08-20 against a real server.
```
Sha-Bang: **nit, and it's my house rule rather than the code's fault: the comments here document the PR, not the code.**

247 comment lines against 372 code lines. The field docstrings on `UnservedPackage` earn every one of theirs — `sourcePath` being absent for three different reasons is a contract a caller cannot infer, and spelling that out is right. What does not earn it is the history:

- "Why this exists, measured 2026-08-20 against a real server", and the second dated measurement in `restartPhrase`
- "it showed up as noise in this module's own test output for a week"
- "The second call originally omitted it, so the invariant held at the top level and a symlinked model one level down was listed anyway"
- "Verified by probe: with a connection named `${TOKEN}`, substituting produced the token's value"
- "Do not reintroduce a sentence asserting that the whole file is rejected"

That is PR-description material, and this PR's description already carries it better than the source can. The dates and the "originally" narrative rot first: the next person to touch `findUnclaimedModelFiles` inherits a paragraph about a bug that does not exist in any version they have seen.

Keep the non-obvious why and cut the rest. "Symlinks are skipped at both depths: a Dirent for a symlink reports neither isFile nor isDirectory" is the whole load-bearing sentence and it is one line. Same for the substitution boundary — "names are echoed as declared, because a name can be written `${VAR}` and substituting would publish that variable's value" says it; the probe transcript adds nothing a reader can act on.

Sha-Bang (changes requested):
the shape is right, and arguing it down from the planned `init_workspace`/`validate_workspace` pair to one conditional field is the better call: the agent gets the answer in the call the instructions already send it to make, gated on a state a healthy server never reaches, computed from inputs that are already public.

I stood it up rather than reading it. Built the branch and booted the built server against an empty root: reproduced the measured dead end exactly (`serving`, `initialized: true`, `environments: []`, no `loadErrors`), and `malloy_getStatus` over MCP now returns the block. Then drove the whole missing-manifest loop live, which is the case worth checking because the advice it gives is destructive. The block named my real `./sales` directory while `loadErrors` named `publisher_data/local/sales`; I created the manifest and did a plain restart and got zero packages *and* zero load errors, the trap the wording warns about; `--init` loaded it and the `setup` key went away. That loop is the strongest thing here.

I also mutation-tested the module, since the body claims it. Dropping the readiness gate, substituting the connection name, emitting the withheld `sourcePath`, echoing V8's raw parse message, dropping `withFileTypes` on the nested readdir, dropping the legacy `projects` fallback, and no longer counting an unusable-location package: all seven turn `workspace_setup.spec.ts` red. The claim holds, which is why the one hole I did find is worth naming.

Gate: server typecheck and eslint clean, prettier clean on the changed files. Unit 3096 pass / 3 skip / 0 fail. Scaffolder 390 pass, though four of its files cannot resolve `@malloy-publisher/skills` until `packages/skills` is built. Integration 381 pass / 1 fail out of 382, the failure being the GCS-backed `Concurrent package operations (E2E)`, which fails identically on `origin/main` (381 / 1 of 382) — not yours. The body says 3044 unit, 385 integration and 376 scaffolder; none of the three matches here, so re-measure or drop them. Unrelated to you, but worth knowing if you reproduce that gate: wiping `publisher_data/` first drops the integration suite from 382 tests to 310 with nothing reported as skipped, so a clean-slate run silently covers 72 fewer cases.

One I'd want fixed before merge, three non-blocking, four smaller ones, inline. Also: the branch is three commits behind main (base `c55b209a`, main `f0f3b296`), and #1045 landed in between and rewrote the authorize fixtures — rebase before merge.

mlennie:
Thanks for standing it up rather than reading it, and for driving the missing-manifest loop live. That loop is the one I most wanted a second pair of hands on, because the advice is destructive and I had already shipped a version of it that did not work.

Answering each finding by name.

## Merge blocker: make the test throw

Fixed, and you were right about why it mattered here specifically. `still answers when the setup diagnosis itself fails` passed `serverRootPath: undefined`, which the early return handles, so the catch it was named for was never reached. I confirmed it your way before changing anything: replaced the whole `try`/`catch` with a bare call and the suite stayed green with that test passing.

It is now two tests. `returns no setup block when there is no server root` is named for the early return it actually pins. `still answers when the setup diagnosis itself throws` uses the throwing `serverRootPath` getter you suggested, and it does go red when the `try`/`catch` is removed, which I verified the same way.

I took the fix rather than the rename. The catch is not belt and braces for long: the module reads an arbitrary config file and does filesystem work, and the next change to it is more likely to add a throwing path than to remove one.

## Non-blocking 1: split the read from the parse

Taken. Reading is its own `try` now, reporting `error.code` and pointing at permissions and owner, and it says explicitly not to edit the JSON because the file has not been read. Parsing keeps the existing branch.

Your framing is the part I want to acknowledge: this is the same failure the module exists to remove, one layer down. I had drawn exactly that distinction for `--config` (`is not a file` versus `does not exist`) and then not applied it here, which is a sibling I should have swept when I wrote the first one.

## Non-blocking 2: count the entries you skip

Taken, including the dead end you identified. Environment-level and package-level skips are both counted now, and there is a branch ahead of the `declaredPackageCount === 0` one that names the shape: `"environments" must be an ARRAY OF OBJECTS`, and it says a list of bare strings or an object map reads as no packages at all. Your three probe shapes all reach it.

The ratio point is the sharper half and I had it backwards in one place while being deliberate about it in another: `declaredPackageCount++` counts a package with an unusable location on purpose, with a comment saying why, and then the `continue` above it discarded the entry entirely. One string plus one bad object now reports two.

## Non-blocking 3: a notebook is a model too

Taken. Both `findUnclaimedModelFiles` matches now cover `NOTEBOOK_FILE_SUFFIX`, which was already exported from the constants module this file imports from, and the `hasModels` sentence names both suffixes rather than `.malloy`.

## (a) The response contract

Taken. `configuredConnectionsTruncated` is documented, and the entry shape is spelled out, since `nextAction` names `sourcePathKind` and `hasManifest` by field and neither appeared in the contract. The description went over the 2000-character budget doing it, so I cut a sentence that repeated contract rule 1 rather than trimming the new information. It sits at 1965 now and `server.protocol.spec.ts` passes.

## (b) The precedence spec

Taken, both parts. The stored environment's path is created in `beforeEach` and the test asserts `from-store` is present as well as `from-config` being absent, so it now shows the stored set being served rather than passing on an empty list. `listStorageDestinations` is added to the mock repository, which removes the error and stack that file logged on every run.

You were fair that it caught the flip it named, and it did. But an assertion satisfied by an empty list is not evidence for the sentence in the comment, and that comment is the reason the test exists.

## (c) REST keeps the dead end

Documentation fix, deliberately, and I want to give the reason rather than just the outcome.

Adding the block to `GET /api/v0/status` widens an unauthenticated disclosure surface, and it is an `api-doc.yaml` change that ripples into the SDK, CLI and Python clients, which this PR has otherwise avoided entirely. With #1071 tightening that same surface in this batch, quietly widening it here is the wrong direction to take without a separate decision.

So: line 170 no longer calls `GET /api/v0/status` the REST equivalent of `malloy_getStatus`. It says the load-error half is equivalent and the `setup` block is MCP-only, and tells an unattended agent what to do instead, which is read `publisher.config.json` and check each declared package for a `publisher.json`. Section 7's pointer to REST now names the gap.

If you would rather have it on REST, that is a reasonable call and I would rather make it as its own change with its own security review than fold it in here.

## (d) Comments

Taken, and this is the one I would have argued with a week ago. 265 comment lines to 225. What went is the history: the dated `measured 2026-08-20` opener, the second dated measurement in `restartPhrase`, the `showed up as noise for a week` aside, the `originally omitted it` narrative, the probe transcript, and the `do not reintroduce` instruction.

Your argument that it rots first is the one that landed. The `originally omitted` sentence describes a bug that exists in no version anyone will ever check out, and a reader who has only seen the fixed code cannot tell whether it is a warning or a description. The mechanism sentences stay, including the one-line symlink rule and the substitution boundary, and the `UnservedPackage` field docstrings stay as you suggested, since `sourcePath` being absent for three different reasons is not inferable.

## Your gate numbers, and mine

You are right that the body's figures match none of yours, and they were stale rather than wrong at the time: measured before two rebases that pulled in eight commits of `main`. Re-measured on the current branch, on a private port:

- unit **3226 pass / 3 skip / 0 fail** across 152 files
- scaffolder **376 pass / 0 fail** (`bun test src`, the package's own script; your 390 looks like a wider target that also picks up the four files needing `packages/skills` built)
- integration **315 pass / 0 fail** across 37 files, on a private `PUBLISHER_PORT`

Leaving the body as written and correcting it here rather than editing it, so this thread keeps its meaning.

## Your environmental note, which cost me a real hour

Your point about wiping `publisher_data/` dropping the suite from 382 to 310 with nothing reported as skipped explains something I had recorded as unattributed: I saw 313 tests across 37 files on one run and 385 across 40 on the next, and could not account for it. That was the same effect.

It also produced four integration failures on my first run after the rebase, and all four were stale `publisher_data` copies rather than code. Three were `compile authorize gate (HTTP E2E)`, where #1045 correctly updated `tests/fixtures/authorize-compile/model.malloy` to the unquoted `#(authorize)` form while my `publisher_data` still held the pre-#1045 copy with the string form, so the server returned 424 rejecting its own fixture. The fourth was the vendored-grammar check reading a stale copy of the storefront example against freshly reinstalled `node_modules`.

That is exactly the failure this PR exists to make legible, arriving in my own test run: the server serves its own copy, not your source.

Your aside then caught me twice, and the second time in the opposite direction to the one I first reported, so it is worth setting out properly.

Clearing the stale copies turned the suite green at **315 pass / 0 fail across 37 files**, down from 387 across 40. I read that as lost coverage, on the strength of your 382-to-310 note, and repopulated to get the bigger number back.

That was backwards, and 315 is the right figure. `test:integration` runs `bun test --timeout 200000 tests`, where `tests` is a path FILTER rather than a directory, so a populated `publisher_data` also collects `publisher_data/examples/storefront/tests/{conformance,controls,format}.test.mjs`. I ran those three files alone: exactly **72 tests across 3 files**, gitignored, belonging to the example rather than the server, and already run deterministically from repo source by `test:examples`. The arithmetic closes, 315 + 72 = 387 and 37 + 3 = 40.

So clearing `publisher_data` does not lose server coverage, it stops inflating the count. Your 314 was right.

Two different things happen when you clear it, and I had conflated them. The stale copies were a real failure and repopulating was the right fix for THOSE: the authorize-gate tests pass on their own now (3 across 1 file) and are inside the 315. The 72 were never server coverage at all.

The rule I would keep from it, since it is what would have caught me either way: **quote the file count beside the pass count.** 315 and 387 are both green and both zero-fail, and only the file count says which suite ran.

One thing that might be worth a follow-up on its own: `test:integration` binds `PUBLISHER_PORT`, default 4000, so every worktree running it contends for one port. I hit that four times in an hour across this batch, once as a `SIGKILL` under load, and the failure presents as a bare exit 1 with no test summary, which reads as a red suite. No integration test hardcodes 4000, so `PUBLISHER_PORT=<free>` avoids it entirely.

Sha-Bang:
Closing. The problem is real and the argument for it is the best in the batch — that is why this is worth saying properly rather than letting it age out.

Two things changed underneath it.

**The detection now exists.** #1184 (merged in `d460fada`) added `getUnresolvedPublisherConfigPath` and a boot notice naming the config path that was looked for and not found. That is the same condition this PR diagnoses, computed from the same `resolvePublisherConfigPath` primitive this PR exports. The audiences differ — that one answers an operator reading logs, this one answers an agent calling a tool — but merging this as written would put two independent implementations of "is there a config?" in the tree, which is the thing worth avoiding rather than the thing worth doing twice.

**The tool surface was renamed.** This registers `mcpServer.tool("malloy_getStatus", ...)`; `main` registers `get_status`, and `server.protocol.spec.ts` asserts that name. The same rename caught #1184's smoke test, which called `malloy_getContext` and got `-32602 Tool not found`. So the integration point this PR attaches to no longer exists under the name it uses, and `get_context`'s schema has moved on too — `search_targets` and `scopes` are now required.

Neither is a reason the idea is wrong. Both mean the remaining work is "write this against current `main`", not "rebase this", and that is a decision to take deliberately rather than inherit through a conflict resolution.

What should survive this PR, for whoever picks it up:

- **The framing.** "The server reports perfect health and serves nothing, and MCP_INSTRUCTIONS routes the agent to the one tool that by construction cannot help" is the whole case, and the `.mcp.json`-written-into-an-empty-directory detail is what makes it common rather than hypothetical.
- **Answering inside the existing tool rather than adding one.** Arguing the planned `init_workspace`/`validate_workspace` pair down to a single conditional field was the right call: the agent gets the answer in the call the instructions already tell it to make, gated on a state a healthy server never reaches. Any future version should keep that shape — and should now read the fact #1184 already computes rather than deriving it again.
- **The withheld-`sourcePath` reasoning**, and echoing declared `${VAR}` text rather than substituted values, so a diagnosis cannot publish the value of a variable it is describing.
- **The live loop you drove** — creating the manifest, restarting plain, getting zero packages *and* zero load errors, then `--init` clearing it — which is the trap the wording warns about and the strongest evidence in the PR.

The mutation testing held up too: seven mutations, all turning `workspace_setup.spec.ts` red. That is the bar a reimplementation should meet.

Issue #1055, opened 2026-08-20
BigQuery: per-connection service-account impersonation via Malloy's new authClient overlay


adamribaudo-ff:
## Summary

We run the publisher as a **multi-tenant** service: one environment per customer, each pointed at that customer's own GCP project and BigQuery datasets. Today the whole process runs as a **single service account holding `bigquery.dataViewer` on every tenant project** — so the only thing standing between tenant A's query and tenant B's data is `restrictedMode` compilation. That control works, but it's one control, living in the compiler. If any restricted-mode gap or compiler bug ever let a cross-tenant table reference through, BigQuery would happily serve the rows, because the executing credential is authorized everywhere.

The fix is standard multi-tenant hygiene: **per-tenant service-account impersonation**. Each environment's BigQuery connection impersonates a data-viewer SA scoped to that tenant's project; the publisher's own SA holds no direct data access. A cross-tenant reference then 403s at the IAM layer regardless of what the compiler allowed — a second, independent line of defense. It's keyless (short-lived tokens minted on demand, no `serviceAccountKeyJson` files to protect or rotate), fully audited (the delegation chain lands in Cloud Audit Logs), and revocable per tenant with a single IAM binding. And it's generic: any deployment that wants a connection to run as something other than ambient ADC gets it through the same field.

Everything needed upstream already exists. Malloy core accepts a host-supplied `authClient` on BigQuery connections as of **v0.0.432** ([malloydata/malloy#3027](https://github.com/malloydata/malloy/issues/3027) → [malloydata/malloy#3044](https://github.com/malloydata/malloy/pull/3044)), via the overlay mechanism @the-michael-toy designed for exactly this use case: config *names* an overlay the host registered, the overlay returns a live auth client, and Malloy's own factory builds the connection. Digest separation, fail-closed handling of unresolved overlays, and key/`authClient` mutual exclusion are all handled in core.

**The only missing piece is a small passthrough in the publisher**: accept an `impersonateServiceAccount` on a BigQuery connection and register the overlay that turns it into an `Impersonated` auth client. Concretely: one optional schema field, one line in the existing config mapping, and one overlay registration beside the `config` overlay the publisher already passes. No `wrapConnections` branch, no hand-built connection, no new cache, no behavior change for anyone who doesn't set the field. **We're happy to write the PR** — full design context is on malloydata/malloy#3027.

### 1. API schema — `api-doc.yaml`

Add to `BigqueryConnection` (alongside `billingProjectId`):

```yaml
[...]

PR #1063, opened 2026-08-21
build(deps): bump @malloydata/* to ^0.0.432, moving DuckDB to 1.5.5

adamribaudo-ff:
Prerequisite for #1055, split out as its own PR as requested there so the feature diff stays reviewable.

- Bumps every `@malloydata/*` caret from `^0.0.427` to `^0.0.432` (server, sdk, app, create-malloy-package) and re-locks `bun.lock`. 0.0.432 ships the overlay-supplied `authClient` connection property (malloydata/malloy#3044) that #1055 builds on.
- 0.0.430 moved the DuckDB engine to 1.5.5, so this also runs `scripts/sync-duckdb-version.js --write`: the Dockerfile CLI `ARG DUCKDB_VERSION` and the server's `@duckdb/node-api` pin follow the query engine (the CI consistency check enforces this).

Verified: server unit suite green (3047 pass / 0 fail), `tsc --noEmit` clean, and a full `docker build` from the root Dockerfile succeeds including the DuckDB extension bake.

https://claude.ai/code/session_016Zb6YvbDh4Fr6PdZnvEstW


kylenesbit (approved):
Reviewed: every `^0.0.427` reference moves to `^0.0.432`, the lockfile moves only the `@malloydata/*` packages plus DuckDB (node-api/bindings 1.5.5-r.4, wasm dev57), and the Dockerfile `DUCKDB_VERSION` ARGs follow; `scripts/sync-duckdb-version.js` is consistent. CI is green on the branch updated to main, DCO signed. Worth knowing for the next release note: 0.0.429 (malloydata/malloy#3032) now rejects a model that spans two connections at compile time.

PR #1065, opened 2026-08-21
ci: one versioning policy and one mechanism for every published package

Sha-Bang:
> ### ⚠️ Merging this arms two never-exercised paths. Run them one at a time.
>
> The next ordinary release would do **two novel things in one run** — the first
> `0.MINOR` transition (`0.0.250` → `0.2.0`) *and* the first-ever PyPI upload
> (`malloy-publisher-sdk` 404s today). If it misbehaves you won't know which half did it,
> and the PyPI half is the **only irreversible thing in this PR**: a filename can never be
> reused. Everything else self-corrects — a bad check reddens a PR, a missed stamp is
> absorbed by `prepare`'s `max(npm, declared)`.
>
> **So split them in time. No code change, just ordering:**
>
> 1. **Merge.**
> 2. **Hand-dispatch `python-sdk.yml` on `main`, deliberately** — after checking the three
>    things CI cannot: `PYPI_TOKEN` must be **account-scoped** (a project-scoped token cannot
>    exist for a project that doesn't), the name must still be free, and `0.1.0` is the
>    version that gets spent. The importability guard runs *before* twine, so a bad build
>    fails instead of uploading.
> 3. **Then dispatch `0.2.0`.** `publish_pkg` finds `0.1.0` already on PyPI and skips it —
>    that's a covered harness case — so the release exercises only the versioning transition.
>
> `docs`: the full playbook is *The first PyPI publish* in
> [`.github/workflows/CONTEXT.md`](.github/workflows/CONTEXT.md).

---

Publisher publishes five packages across four different versioning mechanisms, and cutting
0.0.250 hit two of them: the scaffolder's hand-maintained `SERVER_VERSION` pin red-lit
`publish-packages` 25 minutes into the release, after `skills` had already published, and the
prep PR had to hand-check whether two packages needed bumps because nothing in CI asks.

This makes it one policy and one mechanism.

**A missing bump is now a red PR check**, one per published train (`skills`, the scaffolder,
the Python client). The question is "is the declared version ahead of the registry", not "did
the version change in this PR" — between a prep merge and the dispatch `main` is legitimately
ahead and unchanged. Each check is scoped to the paths that package's *tarball* is built
from, deliberately narrower than its own trigger, so editing a publish workflow doesn't demand
a content-free release. Scoping fails closed both ways: an unresolvable base ref runs the
check anyway, and a watched path that no longer exists is fatal rather than a quiet pass. It
also subtracts the files the packer refuses to ship — `packages/skills/scripts/exclusions.ts`
keeps `skills/README.md` out of the tarball, so editing it must not demand a release.

**The scaffolder's server pin is derived at publish time** instead of policed. It reads npm's
current `@malloy-publisher/server` latest, substitutes it into the runner's tree only, asserts
the substitution took, and verifies it — the same shape as the `workspace:*` skills dep two
steps up. The committed value is now a dev default; a stale one can't reach a user. The
`--host` confirmation the old guard asked a *human* for is automatic now, and lives in
`check_pack` rather than the job holding `id-token: write`.

**`main` stops lying about its version.** Release branches are still never merged back, so
`packages/sdk/package.json` sat 41 patches behind npm. The post-release stamp branch now
resets those three files alongside the release notes, so one PR carries both.
`set_version()` moved to `scripts/set-version.mjs` with a spec, and picked up the two
assertions the shell function was missing.

**The Python client joins the release train.** It was the odd one out: versioned by hand, and
behind a publish job gated on a `refs/tags/sdk-python-*` ref while `on: push` named `branches:
[main]` and no `tags:` — so it had never run once. That gate is *replaced* with the one its two
siblings carry (`workflow_dispatch` on `main`), and `publish-packages` now dispatches it last.
`publish_pkg` is parameterised rather than copied: the two registries differ in exactly two
places, how a manifest is read and how the registry is asked, and everything else — the
watched-path guard, the "main moved" compare, the poll loop — stays registry-agnostic.

Not repaired as a *tag* trigger, for a concrete reason: a tag pointed at a commit already on
`main` pushes no new commits, so the `paths:` filter has nothing to match and would filter the
run out — the trigger would look fixed and still not fire.

**And `bump: patch|minor` at the dispatch**, so a breaking release doesn't mean hand-computing
the number. No `major` — pre-1.0 by policy, and a major should be an explicit version.

The policy itself is written down in `CONTEXT.md` and the `publisher-release` skill:
`0.MINOR.PATCH` while pre-1.0, minor = breaking. It's the policy and not yet the state — on
`0.0.x` both `^0.0.250` and `~0.0.250` are useless for "patches yes, breaking no", which is
the whole reason for it. Moving `sdk`/`app`/`server` to `0.2.0` is a dispatch after this
merges, not a code change.

One known gap is still recorded rather than fixed, since it's a decision:
`@malloydata/publisher-cli` looks publishable and nothing ships it.

**A review of this branch caught that the Python publish would have shipped a broken package** —
`python -m build` on a checkout succeeds and produces a wheel that installs and then raises
`ModuleNotFoundError: No module named 'malloy_publisher_sdk.client'`, because the generated client
is gitignored. Two independent causes, both fixed: the job now runs `build-python-sdk.sh`, and
`pyproject.toml` sets `ignore-vcs = true` because **hatchling consults `.gitignore`**, so running
the generator alone still excluded the client from the wheel. The guard asserts the artifact rather
than the intent: it installs the built wheel into a throwaway venv and imports it, before twine.

On the first PyPI upload, see the banner at the top — it's the one part of this change that
lives outside the diff. Worth adding here: if `0.1.0` should move off `skills`' `0.1.x` line,
move it *before* that first publish, because PyPI filenames can never be reused. The failure
mode is otherwise cheap — `python-client` is dispatched last, nothing depends on it, and
re-running `publish-packages` skips whatever already landed.

### Tested

The release path is `workflow_dispatch`-only and can't be exercised in CI, so each changed
`run:` body was extracted from the YAML and executed against scratch git repos with
`npm`/`npx`/`curl` stubbed.

| | |
|---|---|
| `bun run test:scripts` | 40 pass / 0 fail (17 of them new, in `set-version.spec.ts`) |
| skills bump check | 20/20 |
| scaffolder bump check | 18/18 |
| PyPI bump check | 15/15 |
| `publish-packages` (npm + PyPI) | 12/12 |
| wheel + sdist guard | 4/4 |
| `EXCLUDE` ↔ `exclusions.ts` contract | 2/2, mutation-tested 6 ways |
| `prepare` version derivation | 13/13 |
| `gh-release` stamp step | 10/10 |
| `--host` documentation check | 8/8 |
| server-pin substitution (`ubuntu:24.04`, GNU `sed`) | 15/15 |
| `actionlint` | no new findings vs the `origin/main` baseline |

Every extracted body is `shellcheck`-clean, and all four commits parse under `actionlint`
individually. Scaffolder suite 376 pass / 0 fail; prettier and `tsc --noEmit` clean.

All three checks then ran for real on this PR, and each hit a different branch:

- **scaffolder** — `@malloy-publisher/create-malloy-package@0.0.8 is free and above latest
  (0.0.7)`. The "legitimately ahead and unchanged" case, which is the one a did-it-change
  check would have reddened.
- **skills** — `no published skills content changed in this PR (checked skills
  packages/skills, excluding :!skills/README.md); no version bump required`. The scoping skip.
- **python** — the same skip. This PR edits `python-sdk.yml`, which is in that workflow's
  `paths:` trigger but deliberately not in its scope.

The skills one earned that exclusion the hard way: the first push went **red** with
*"this PR changes published skills content, and @malloy-publisher/skills@0.1.7 is already on
npm — so a release would SKIP this package and stay green"*, over one corrected sentence in
`skills/README.md` — a file `exclusions.ts` never packs. So the red path is demonstrated live
too, with the message it should carry. Force-pushed with the scope fix and a regression case
for it.

`publish-packages` is the step this changes most and had no CI coverage at all, so it got its own
harness: the body extracted from the YAML and run against a scratch repo with `npm`, `curl`, `gh`
and `python3` stubbed, the poll budget shrunk, and the stubs modelling the real sequence — a
registry answers "free" until `gh workflow run` drops a marker, then "published", so the poll loop
converges the way it does in a release instead of always timing out. The case that matters most
for a shared sequence: Python failing while the two npm packages still publish, with the summary
recording exactly what shipped.

Two things that harness caught, both now fixed: `publish-packages` calls `python3` for `tomllib`
but installs nothing (deliberately — it holds `actions: write`), so a runner below Python 3.11
would have reported "could not read its manifest", which reads as a corrupt `pyproject.toml`; the
interpreter is asserted instead. And `timeout-minutes` was sized as *two* poll budgets, so with
three packages the last one timing out would be killed by the job cap before printing which
package failed.

### Known and deliberately not closed

Recorded in `CONTEXT.md` rather than fixed, each because the fix costs more than the gap:
`bun.lock` and `BUN_VERSION` change published bytes without reaching their checks (both
bounded — the tarball stays one build behind, not broken); `black` and `ruff` both honour
`.gitignore`, so neither actually lints the generated Python client inside a checkout (`7
files left unchanged` is the tracked set, zero generated files); and none of these three
checks may become a **required** status check, since they're path-filtered and a required
check that never runs sits at `expected` forever.

Still worth a throwaway PR before trusting it end to end: a real `skills/**` content change
without a bump, which nothing here exercises.


housejester:
This is a good change and I'd merge it. The mechanism is right, the fail-closed reasoning is unusually careful, and the doc set is better than the code it documents — `CONTEXT.md` pre-empts three of the four gaps I went looking for (`bun.lock`, the PyPI check's no-op status, `dist/` not asserting the substituted pin), and says out loud that the verification harness is a one-off not in this repo. Three real findings below, none structural.

Two things worth crediting because they're easy to read past. Moving the `npx -y @malloy-publisher/server@<mutable-tag> --help` (~40 transitive deps, lifecycle scripts on) out of the job holding `id-token: write` and into `contents: read` is a security improvement, not a tidy-up. And `registry_has`'s explicit `*)` arm closes the fail-*open* "a `case` that matches nothing succeeds" trap, which is the most dangerous shape in that function.

### Verified against ground truth first

| Check | Command | Result |
|---|---|---|
| `bump=minor` lands on a free number | `npm view @malloy-publisher/{sdk,app,server} dist-tags` | all three `latest=0.0.250`; sdk also carries `1.0.1` and `0.0.382-dev`, but **no `0.1.x`/`0.2.x` on any of the three** — so both the derived `0.1.0` and the planned explicit `0.2.0` are free |
| Does the Python train fit the 25m poll budget? | `gh run view … --json jobs`, 4 runs on this branch | `build` completes in **17–22s** end to end (uv + generate + black/ruff/pyright + pytest). The publish job's extra work is seconds. Not at risk — a concern I raised and dropped |
| `set-version.spec.ts` really runs per-PR | `grep -n test:scripts .github/workflows/build.yml` | `build.yml:102`, in `lint_format`. Claim holds |
| Does the sed cover every hardcoded pin? | `grep -rn '@malloy-publisher/server@\|0\.0\.250' packages/create-malloy-package/` | `SERVER_VERSION` is declared once, `src/scaffold.ts:296`; no template hardcodes a server version. Substitution is complete |

## Worth acting on

**[MED — fail-open, two-file string contract with no apply-time check]** `skills-npm.yml:129`'s `EXCLUDE=(':!skills/README.md')` duplicates `packages/skills/scripts/exclusions.ts:55`'s `isSourceReadme()`, synced only by a comment. That file's own docstring states the rule: *"The copy, the pack audit, and the tests all have to agree exactly. When they drift, one of them silently permits what another forbids."* The workflow is now a **fourth reader** of that list, and the only one with no test behind it.

Failure path: someone decides `skills/README.md` should ship (drops `isSourceReadme`, or renames the file). The workflow keeps excluding it. A README-only PR now *does* change the tarball, the check reports "no published skills content changed", it merges without a bump — and `publish-packages` finds `@malloy-publisher/skills@0.1.7` already on npm, **skips the package, and the release stays green.** That's precisely the failure this PR exists to close, re-entering through the exclusion this PR adds.

Recommend a spec under `packages/skills` (so it runs per-PR via `bun run test:skills`) that parses the `EXCLUDE=(…)` line out of `.github/workflows/skills-npm.yml` and asserts both directions: each `:!<path>` is `isExcluded()`, and no other top-level entry under `skills/` is excluded-but-unlisted. Reading a sibling workflow from a test is ugly, but it's the only thing that makes this contract fail loudly — and the file it has to agree with already says so.

**[MED — a summary line asserts a reason a `::warning` in the same run contradicts]** `release.yml:1215-1216` sets `VERSIONED=0` on the *failure* path, the same value the *no-op* path carries. Two new sentences then state the no-op reason as fact:

- `release.yml:1379` — "`main` already declared `${VERSION}`, so there was no version to reset either, and no branch was pushed."
- `release.yml:1373` — "No branch was pushed, and `main` already declared `${VERSION}` so there was no version to reset either."

Both are reachable when `set-version.mjs` **failed**. A run where the reset failed and there was nothing to stamp emits `::warning …could not reset main's declared version` *and* a summary saying main already declared it. The summary is the artifact `CONTEXT.md` and the release skill both tell a releaser to read — and this step's own comment at `release.yml:1219` names the rule being broken: *"Two different states, and conflating them is how a summary line ends up contradicting a warning the same run already wrote."*

Recommend a separate `VERSION_FAILED=1` alongside `VERSIONED=0` in the failure branch, with those two sentences conditional on it.

**[MED — an immutable first upload whose contents are decided by an unpinned third party]** `packages/python-client/scripts/build-python-sdk.sh:31,38` call `uvx openapi-spec-validator` and `uvx openapi-python-client generate`. `uvx` resolves the latest release into an ephemeral env and **ignores** `pyproject.toml:61`'s `openapi-python-client>=0.18.0` entirely — so the declared constraint reads as if it governs the generator and doesn't.

Two consequences, and this PR is what makes them matter: the wheel's public surface is whatever the generator's `latest` emits at publish time, under a PyPI filename that can never be reused; and nothing records which generator produced a release, so the wheel isn't reproducible from the repo later. `pytest tests/` at step 8 runs against the regenerated client in the same invocation, so it catches a hard break — it does not catch a generator that renames models or changes `Unset` semantics and ships as a "patch".

The fix is one word: the script already does `uv pip install -e ".[dev,test]"` and `source .venv/bin/activate` before step 4, so `openapi-python-client generate …` (no `uvx`) runs the version the project declares. Then pin it `==` in the `dev` extra — a generator bump then becomes a reviewable PR under `packages/python-client/`, which the new bump check already watches, so it correctly demands a version bump.

## Minor / non-blocking

- **[LOW — the excluded workflow file isn't byte-neutral after all]** `create-malloy-package-npm.yml:93` justifies keeping the workflow file out of `PATHS` with "editing this workflow file changes no published byte". Not quite: `BUN_VERSION: 1.3.13` is at `:45` in that same file, `prepack` is `bun run build` → `bun build src/index.ts …`, and `dist/index.js` is the only JS in the tarball. A BUN_VERSION bump changes the bundler emitting the shipped file; the check runs (the file *is* in the trigger — verified), reports "no published scaffolder content changed", and the next release skips the package. Consequence is mild — the tarball stays one bundler behind, not broken — so I'd take the cheap fix: tighten that sentence and record it in `CONTEXT.md`'s named-gaps list beside the `bun.lock` entry, which is the same shape and already handled that way. Hoisting `BUN_VERSION` somewhere watched closes it properly if you'd rather.
- **[LOW — the rationale was written down, then applied to two of three]** `skills-npm.yml:38`'s `check_pack` has no `timeout-minutes`, yet it carries the third copy of the registry check that both siblings got `timeout-minutes: 10` for, with the reason spelled out at `create-malloy-package-npm.yml:57`: *"a bound per call is not a bound on the job… a hung step would sit on a runner for GitHub's 6-hour default blocking a PR on a check that can neither pass nor fail."* Same for `python-sdk.yml:263`'s `publish` — the job holding `PYPI_TOKEN` and doing the most work — and `skills-npm.yml:310`'s `publish`.
- **[LOW]** `twine upload dist/*` at `python-sdk.yml:421` ships the sdist too, and the (excellent) importability guard only ever inspects `dist/*.whl`. Pure-python wheel, so almost nothing resolves the sdist — but it lands under a filename PyPI can never reuse either. Either narrow the upload to the artifact you verified, or extend the guard.
- **[LOW]** All three new checks live in path-filtered `pull_request` triggers, so a PR touching none of those paths produces no check at all. Worth writing next to the table in `CONTEXT.md`: **these must never become required status checks**, or unrelated PRs sit at `expected` forever. `skills-npm.yml` already had this shape, so it's a pre-existing constraint that now applies three times.

## One thing I'd like your read on, rather than a finding

The stamp branch is now dual-purpose, and the recovery for one half is destructive to the other: `release.yml:1300`'s new warning tells a releaser in prose *"Do not close a conflicted stamp PR to clear it"* — because the version half is recoverable (`max(npm, declared)` tolerates it) and the notes half isn't. That warning exists because the natural move on a conflicted, version-stale PR **is** to close it.

Two separate branches (`release-notes-stamp-<v>` and a `release-version-stamp-<v>`) would make closing the conflicted one safe by construction rather than by reading a job summary, at the cost of a second PR per release. Was one branch a deliberate call on "two PRs per release is worse than one prose warning"? If so I'd leave it — the warning is well written and names the exact trap — but this is the kind of coupling that gets discovered by someone who didn't read the summary.

**Top three before merge:** 1) put a test behind the `EXCLUDE` ↔ `exclusions.ts` contract — it's the one new fail-open path this PR introduces into the mechanism it exists to build; 2) split `VERSIONED=0`'s two meanings so those summary sentences stop asserting a reason the same run's warning contradicts; 3) drop `uvx` and pin the generator, before the first immutable PyPI upload rather than after.

<sub>Reviewed with Claude Code (session "Plucky Capybara").</sub>

Sha-Bang:
All three MED findings fixed, all four minor ones, and the `uvx` one turned up something worth knowing. Pushed as `02e00489`.

### 1. EXCLUDE ↔ exclusions.ts — test added

You're right that this is the one new fail-open path the PR introduces into the mechanism it exists to build, and right about which direction is dangerous.

`packages/skills/src/workflow-exclusions.spec.ts` parses the `EXCLUDE=(…)` array out of the workflow and asserts both directions. **Mutation-tested rather than merely written** — I don't trust a contract test I haven't watched fail:

| Mutation | Result |
|---|---|
| drop `isSourceReadme()` | ❌ *"skills-npm.yml excludes skills/README.md from its bump check, but exclusions.ts would PACK it…"* |
| point `EXCLUDE` at a different file | ❌ *"exclusions.ts keeps skills/README.md out of the tarball, but skills-npm.yml's bump check still watches it…"* |
| rename the array to `SKIPLIST` | ❌ *"skills-npm.yml no longer has a single-line EXCLUDE=(…) array; this test cannot verify the contract it exists for"* |

That third one was the case I most wanted to cover — an unparseable array must not quietly yield an empty list and switch the test off, which is the same failure the workflow's own watched-path assertion exists to prevent.

### 2. `VERSIONED=0`'s two meanings — split

Separate `VERSION_FAILED` flag gating both sentences **and** the matching `::notice`, which asserted the same reason. Reset per attempt beside `FAILURE`, so pass 1 failing can't make pass 2's summary claim a failure the retry fixed. New stamp case (`reset failed + nothing to stamp: no contradiction`) now produces:

> ⚠️ `could not reset main's declared version` … 
> *"The version reset FAILED (see the warning above), so `main` still declares whatever it did before this release, and no branch was pushed."*

### 3. `uvx` — dropped and pinned

Confirmed: the venv is already activated and `uv pip install -e ".[dev,test]"` already installs both tools into it, so `uvx` was resolving a *second*, unpinned copy and ignoring the declared constraint entirely. Verified both binaries resolve from `.venv/bin` after activation, at `0.29.0`/`0.9.0` — exactly what `uvx` resolved and what CI already installed, so it's behaviour-neutral today. Now pinned `==`. Ran the full script end to end, and ran the **committed** script as a control to confirm both reach the same place.

**That control run surfaced something.** Both abort at step 6 with 929 ruff errors — while CI reports `All checks passed!` on the same files with the same ruff 0.16.4. The difference is `.gitignore`: ruff honours it, `malloy_publisher_sdk/` is gitignored except `__init__.py`, so inside a checkout **ruff skips the entire generated tree**. Measured directly — `git init` in the scratch copy takes it from 929 errors to `All checks passed!` with nothing else changed.

CI is a checkout, so the release isn't at risk. But "lint ran" doesn't mean the generated client was linted. Recorded in `CONTEXT.md` rather than fixed, because turning it on surfaces ~929 findings and is a separate call — and because it's the *same trap* as the hatchling packaging bug in `29013ae3`: one directory deliberately gitignored while being the package's entire content, and two independent tools quietly honouring that. Two instances makes it a pattern worth naming.

### Minor

- **sdist unverified** — good catch, and it lands under an equally unreusable filename. The guard now asserts the sdist contains `malloy_publisher_sdk/client.py`, and that there's exactly one. Containment rather than install, since almost nothing resolves an sdist for a pure-python wheel; it catches the failure that motivated the guard. New case: an sdist with `client.py` stripped is refused.
- **BUN_VERSION** — you're right, the sentence was wrong. Taken the cheap fix: sentence corrected, recorded beside the `bun.lock` entry in `CONTEXT.md`, since it's the same shape.
- **timeout-minutes** — applied to all four: `skills-npm.yml`'s `check_pack` and `publish`, `python-sdk.yml`'s `build` and `publish`.
- **never required** — now stated outright in `CONTEXT.md`, including what requiring them would take (an always-run gate job).

### On the stamp branch — your question

**Deliberate, and I'd keep it at one.** The reasoning, so you can push back on the premise rather than the conclusion:

The two halves have very different costs when they go wrong. The notes half **compounds** — an unstamped `[Unreleased]` heading is what the *next* release's `extract` matches, so that narrative lands on the next page too, and every page after. The version half is **self-correcting**: `prepare` takes `max(npm, declared)`, so a missed reset costs nothing and the next release fixes it.

Splitting gives each half its own branch, but the conflict you'd be avoiding is *between two releases' stamp PRs*, not between the two halves — and that conflict exists either way, just moved from one PR to two. So it's a second PR every release, forever, against a trap that only springs when two stamp PRs are outstanding at once. That's already the abnormal state the summary flags loudly.

Where I think you're right, and where I'd want to be talked out of it: "safe by construction" beats "safe by reading the summary", and the person who hits this is by definition the one who didn't read it. If it bites once, splitting is the fix and I wouldn't argue.

### Not done

Extracting the three bump checks into a composite action. Right shape, and you're right the trusted-publisher constraint doesn't forbid it — but it's ~450 lines across three files and doesn't belong in the same PR as a publish that would have burned a version. Noted as the next thing.

Full sweep green: skills suite 15, script specs 40, wheel+sdist 4, stamp 9, skills 18, scaffolder 18, PyPI 15, prepare 13, `--host` 8, publish-packages 12, server pin 15 under `ubuntu:24.04`. actionlint unchanged vs baseline.

Sha-Bang:
Addendum — your finding #1 demonstrated itself, so the `EXCLUDE` array changed shape after my reply above.

Adding `workflow-exclusions.spec.ts` turned the **skills check red**: `packages/skills/` was watched wholesale, so a test file counted as published content. It isn't — `tsconfig.build.json` excludes `src/**/*.spec.ts` from the `tsc` that emits `dist/`, and `dist/` is the only thing `files` ships from that directory. Same over-inclusive scope the scaffolder's enumerated `PATHS` already avoid for its `tests/`; I'd left skills as the whole directory.

So `EXCLUDE` gains a second entry, and the contract test now knows there are **two governance domains**: entries under `skills/` answer to `exclusions.ts`, entries under `packages/skills/` answer to `tsconfig.build.json`'s exclude list, and an entry under *neither* is a hard failure — an exclusion no test checks is the state that file exists to end. Now mutation-tested in five directions (the three above, plus tsconfig starting to publish specs, plus an ungoverned entry).

One detail worth flagging since it would have silently un-done the fix: `:!packages/skills/src/**/*.spec.ts` **matches nothing**. A git pathspec is fnmatch *without* `FNM_PATHNAME`, so `*` already crosses `/` and `**` isn't special — the `**` form looks right and excludes nothing. Verified against the real diff before relying on it; the entry is `src/*.spec.ts`.

Two more harness cases pin the boundary: a spec-only change skips, a real `src/` change still demands a bump.

```
##[notice] no published skills content changed in this PR
(checked skills packages/skills, excluding :!skills/README.md :!packages/skills/src/*.spec.ts);
no version bump required
```

Green again — 14 pass, 0 fail, 4 long suites still running. `9c41288d`.

housejester:
Re-reviewed `9c41288d` against `29013ae3`. Verifications first:

| Check | Command | Result |
|---|---|---|
| The `**` pathspec claim | `git ls-files 'packages/skills/src/**/*.spec.ts'` vs `'…/src/*.spec.ts'` | **0 files** vs **3 files** — confirmed, and it's a nasty one, since the `**` form is the one that *looks* right |
| New contract spec passes | `bun test packages/skills/src/workflow-exclusions.spec.ts` | 2 pass / 0 fail, 9 `expect()` calls |
| Does the tsconfig arm actually bite? | mutation, below | **No — passes on a fail-open mutation** |

All three MEDs and all four minors land, and two of the fixes are better than what I asked for. The `**`-matches-nothing trap and the ruff/`.gitignore` discovery are both things I'd missed; the second is the same "one directory deliberately gitignored while being the package's entire content" shape as the hatchling bug, and naming it as a pattern rather than fixing it is the right call.

Three residuals, one of which is mine.

## The one that's on me

**[MED — the second governance domain got a weaker assertion than the first, in the fail-open direction]** My prescription was "assert each `:!<path>` is `isExcluded()`" — I only specified the `exclusions.ts` domain. Your fix uncovered that `EXCLUDE` now spans two, and the `packages/skills/` arm is `excludes.some((glob) => glob.endsWith("*.spec.ts"))` — satisfied by *any* glob ending that way, not by the one the pathspec actually corresponds to. That's my prescription being under-specified for the case your fix surfaced, not a defect you introduced.

Mutation-verified rather than argued:

```
# tsconfig.build.json exclude: ["node_modules","dist","skills","src/**/*.spec.ts"]
#                           →  ["node_modules","dist","skills","src/legacy/**/*.spec.ts"]
$ bun test packages/skills/src/workflow-exclusions.spec.ts
 2 pass / 0 fail
```

After that narrowing, `src/*.spec.ts` **is** emitted into `dist/` and ships, while `skills-npm.yml` keeps excluding all of it from the bump check — so a spec change reaches the tarball with no bump demanded, the release skips the package, and it stays green. Your five mutations cover *removing* the glob; narrowing it is the sixth, and it's the more realistic edit.

Fix in the file's own idiom: assert the exact string `"src/**/*.spec.ts"` is present in `exclude`, the way `excludePathspecs` is already deliberately strict about the array's exact single-line shape. Brittle in the fail-closed direction, which is what this whole file trades on everywhere else.

## Two residuals on the fixes

**[MED — the `VERSION_FAILED` split covered the two rare branches and missed the common one]** `release.yml:1360-1361` (the `pushed` case) is unconditional:

> `Pushed release-notes-stamp-0.0.251, stamping 3 release-note section(s) and resetting 0 version file(s) to 0.0.251.`

With `VERSION_FAILED=1` and `STAMPED != 0` — reachable exactly the way you measured for the earlier fix, a manifest on `main` with no version field while the notes stamp succeeded — that `0` carries the same two-meanings ambiguity, in the branch that fires whenever there is narrative to stamp. Milder than the two you fixed, since it reports an ambiguous count rather than asserting a false reason and the `::warning` is still there — but it's the branch most people will actually read. `release.yml:1273`'s commit body has the same shape. One clause appended to `LEAD`.

**[LOW — brand-new prose, and it invites the conclusion it exists to prevent]** The `CONTEXT.md` note says "`black`, `ruff` and `pyright` … **except ruff** also honours `.gitignore`". Black does too, and there's proof in this branch's own CI log:

```
black: All done! ✨ 🍰 ✨   7 files left unchanged.
```

Seven is exactly the non-gitignored set — `malloy_publisher_sdk/__init__.py` (tracked) + `tests/{conftest,test_api_environments,test_client,test_integration,test_models}.py` + `tests/api/__init__.py`. Zero files from the generated tree. Since the note's whole job is "before you trust a green build", naming one tool where two are affected leaves a reader concluding black *did* format the client. `pyright` is genuinely unaffected — it reported into the generated tree's types in that same run.

## One correction, and one concession

**I didn't recommend the composite action.** The "Not done" section has *"you're right the trusted-publisher constraint doesn't forbid it"* — that wasn't me. `CONTEXT.md` already argues the other way ("Each package's machinery stays in its own file: that is why the bump check is written three times"), and I never contested it. Worth not carrying it as a next-thing on my authority; if you want it, want it for your own reasons.

**On the stamp branch, you've talked me out of it.** The premise I had wrong is the one you named: the conflict is between two releases' stamp PRs, not between the two halves, so splitting relocates it rather than removing it — at the price of a second PR every release, forever. Keep it at one.

Minor: the reply above cites `02e00489`, which doesn't resolve (`git cat-file -t` → not a valid object); the content is at `9c41288d`. Only flagging it because the comment is the navigation aid.

**Nothing blocking.** The tsconfig assertion is the one I'd do before merge — it's the load-bearing half of the fix to finding #1, and it currently passes the mutation that matters.

<sub>Reviewed with Claude Code (session "Plucky Capybara").</sub>

Sha-Bang:
All three fixed in `eae0b9fc`. And the correction is yours to keep — see the bottom.

### The tsconfig arm — you're right, and it was the load-bearing half

Reproduced your mutation exactly, then checked the half your run didn't need to: under `src/legacy/**/*.spec.ts` the spec passes **2/0**, *and* `dist/` gains `frontmatter.spec.js`, `index.spec.js`, `workflow-exclusions.spec.js`. So they really do ship while the check keeps excluding them. My five mutations covered *removing* the glob; narrowing it is the sixth, and as you say, the one someone actually makes.

Now asserts the exact string, in the file's own idiom:

```ts
expect(excludes, `…exclude no longer contains exactly "src/**/*.spec.ts" — so a spec under
  src/ may now be emitted into dist/ and published while the check still reports
  "no published skills content changed".`).toContain("src/**/*.spec.ts");
```

Narrowing now fails; removal still fails.

I'd push back on "that's my prescription being under-specified, not a defect you introduced" — the second domain is mine, I wrote the weaker assertion, and I wrote it *while* fixing a fail-open finding, which is exactly when to be suspicious of `some()`.

### `VERSION_FAILED` missed the branch that fires

Correct, and the clause is now in all **three** places carrying that count, not just the two you named — the `::notice` has the same string. The commit body matters most, for the reason you'd probably give: it outlives the run, so someone reading `git log` months later has no summary or annotation to cross-check the `0` against.

```
chore(release): stamp 0.2.0 on main

1 release-note section(s) stamped, 0 version file(s) reset. (the reset FAILED rather
than finding nothing to do)
```

Tenth stamp case covers notes-stamped-while-reset-failed. `VERSION_NOTE` is declared beside `VERSION_FAILED` as well as assigned in the loop, same `set -u` reason.

### black — confirmed, and the count is exactly your prediction

`7 files left unchanged` = `malloy_publisher_sdk/__init__.py` + the six under `tests/`. Zero from the generated tree. The note now names both tools and keeps pyright as the exception, with one addition I found checking it: **step 7 ends in `|| true`**, so pyright's findings never fail the build either — that same green run carries 4 outstanding pyright errors, reported against generated types (`list[Connection] | Unset`) from `tests/`. Which rather reinforces the note's premise.

Three independent tools, three different answers about what that gitignored directory means.

---

**The composite action: you're right, that wasn't you.** It came from a separate review pass and I merged the two sets of findings in my head when writing the reply. `CONTEXT.md` does argue the other way, you didn't contest it, and I shouldn't have put words in your mouth — dropped as a next-thing, and I'll want it for my own reasons or not at all.

**`02e00489`** — also right, and my fault for the same reason a stale ref is annoying: I amended after posting. It's `9c41288d`; this one is `eae0b9fc`.

Sweep green: skills suite 15, script specs 40, wheel+sdist 4, stamp 10, skills 20, scaffolder 18, PyPI 15, prepare 13, `--host` 8, publish-packages 12. actionlint unchanged vs baseline.

One thing worth your eyes since it nearly shipped: the amend also stripped four build artifacts (`scripts/{exclusions,frontmatter}.{js,d.ts}`) that my *mutation testing* emitted — narrowing the tsconfig exclude makes `tsc` compile `scripts/` and write next to the sources, and they aren't gitignored. Caught on the `--stat` before the force-push, not by anything automated.

jswir (approved):

PR #1068, opened 2026-08-25
fix(server): report a pure-storage= manifest as bound, not unbound

housejester:
## What

`manifestBindingStatus` is only maintained on the colocated bind path, so a manifest whose entries are all `storage=` binds successfully and then keeps reporting `unbound`.

## Why it happens

`Environment.bindManifest` splits a manifest into two tiers and binds each by a different route:

- **storage entries** → `Package.bindStorageServeBindings`, which applies the serve bindings and does not touch `manifestBindingStatus`
- **colocated entries** → `Package.reloadAllModels` → `recordManifestBinding`, which derives the status from its entry count

For a manifest with no colocated entries, `hasColocated` and `hadColocated` are both false, so `reloadAllModels` is skipped, `recordManifestBinding` never runs, and the status keeps its `unbound` field initialiser. `setBoundManifestUri` runs unconditionally, so the package ends up reporting:

```
manifestBindingStatus: "unbound"
boundManifestUri:      "<the manifest that was just bound>"
```

That pairing cannot otherwise occur — `recordManifestBinding` nulls the URI whenever it records zero entries — which is what makes it recognisable as a missed status update rather than a genuine unbound package.

Nothing fails and nothing is logged. The bind really did succeed, the entries are applied, and the serve path routes correctly. Only the reported status is wrong.

## Why it matters

A caller that distributes manifests and reads `manifestBindingStatus` back to confirm the worker honoured one sees permanent drift for these packages: it rebinds, the package reports `unbound` again, and it rebinds again on the next reconcile tick — indefinitely. Each rebind re-fetches and reloads the package, so the cost is continuous rather than one-off.

The same signal is a natural gate for reclaiming a version's superseded materialized tables once a rebuild has fully cut over. A version that can never report bound never converges, so its dead generations are never dropped.

There is also a **fail-dangerous** case, which is arguably worse than the retry loop. On the storage-only path a failed bind reports `live_fallback`, and because the next *successful* bind skips `reloadAllModels`, nothing recomputed the status — so the package went on reporting `live_fallback` while healthily bound. A caller that alerts on `live_fallback` was told a working package was degraded. The same one-line change fixes that, and there is now a test for the recovery leg.

A package with even one colocated entry is unaffected, since that alone makes `hasColocated` true and takes the status-setting path. That is why this is easy to miss: it needs a package where *every* persisted source is `storage=`.

## The change

**Status reporting**

- `Package.setBoundManifestUri` becomes `Package.markManifestBound`, which sets the reported status alongside the URI. It had a single call site, so it is repurposed rather than added next to.
- `Environment.bindManifest` calls it once both tiers have applied.
- `live_fallback` is still written only by `markManifestBindFailed`, and clearing a manifest still reports `unbound` via `reloadAllModels({})` → `recordManifestBinding({})`.

**Bind ordering**

`reloadAllModels` now runs *before* `bindStorageServeBindings`. It is the only step in a bind that can throw (a compile error, or the worker pool being unavailable), and the catch reports `live_fallback` — "no manifest applied". Binding storage entries first left the new manifest's cross-connection bindings installed and routing while the package reported that nothing had been applied and `boundManifestUri` still named the previous manifest. A caller reading that status as "not serving from manifest tables" would have been wrong about tables the package was actively serving from.

Ordering between the two is otherwise immaterial: `reloadAllModels` re-pushes whatever bindings are current onto the fresh model set, and `bindStorageServeBindings` pushes its own, so the new set lands either way. Both `had*` reads now happen before either tier applies, since each bind overwrites the state its own guard asks about.

**Docs**

- `manifestEntryCount` now says it counts the colocated tier only, and that a fully bound `storage=`-only manifest reports `bound` with a count of `0`. The previous wording ("0 when unbound or on live fallback") implied `bound ⇒ count > 0`, which was true before this change and is not now — a caller corroborating the status against the count would reproduce the same non-convergent loop against a different field. Corroborate a cross-connection bind with `storageServeBindings` instead.
- `manifestBindingStatus`'s `bound` no longer claims that persist sources route to their materialized tables. It reports that the bind ran to completion; an entry whose source fails the eligibility gate is dropped, and the storage tier serves live under a `PERSIST_STORAGE_MODE` that is not `on`.
- `live_fallback` no longer claims the package is serving live. It says the *configured* manifest was not applied — which is not the same thing. A failed **first** bind leaves nothing bound, but a failed **rebind** leaves the previously bound manifest in place: its colocated substitutions stay compiled in, its storage serve bindings stay installed, and `boundManifestUri`, `manifestEntryCount` and `storageServeBindings` all continue to describe it. That guarantee is scoped to a failure **before either tier applies** — the manifest fetch failing or timing out, which is what produces `live_fallback` in practice. A throw partway through the recompile, after `this.models` is swapped but before the binding is recorded, is explicitly out of scope: the new substitutions would be live while these fields still described the previous manifest.
- `manifestEntryCount` no longer lists live fallback among the cases that report `0`, for the same reason, and now says it describes whatever is *currently* bound rather than the configured manifest.
- The EMPTY-manifest case is now a decided one. A manifest that binds nothing — empty, or with every entry skipped — was still fetched and applied, so it reports `bound` with `manifestEntryCount` 0 and no `storageServeBindings`. That combination is the signature for "the manifest carried nothing", as against the bind having been missed. Reporting `unbound` instead would make an empty manifest permanent drift to a caller that rebinds on anything but `bound` — the same non-convergent loop this PR closes, narrowed. `markManifestBound` is deliberately unconditional and supersedes what `recordManifestBinding` derived if the recompile ran; the call site says so.
- `boundManifestUri` names the failed-rebind divergence from `manifestLocation` in its own description, since it is the field a reclaim gate keys on and the natural place a caller lands first. `storageServeBindings` likewise says it tracks what is currently bound. And `manifestBindingStatus` distinguishes its `live_fallback` from the same token on a query result's `servedFrom`, which means the opposite — that one query did degrade to the warehouse.

Both descriptions now state the invariant a caller actually needs: **a non-null `boundManifestUri` is a live reference to the manifest whose tables the package is serving from, regardless of `manifestBindingStatus`.** Reclaiming superseded tables while reading `live_fallback` as "not serving from manifest tables" would drop a generation out from under a package actively serving it.

The fields' semantics are deliberately left alone — widening the count would be a wire change with unknown readers.

## Test

The existing pure-`storage=` integration test asserted `manifestEntryCount`, `storageServeBindings` and `boundManifestUri` — but never `manifestBindingStatus`. Its own comment noted the asymmetry ("the bound URI is recorded even though no tableName manifest bound") without checking what the status reported, which is how this got through.

Added that assertion, assertions on the clear path so the status is covered in both directions rather than only latching on, a case covering `live_fallback` → `bound` recovery on the storage-only path, and a case pinning the failed-rebind state (`live_fallback` with the previous manifest's count and URI intact) since that is the invariant the reclaim guidance above rests on.

Also pinned the empty-manifest state on the pre-existing rebind-to-empty test, which previously asserted only that stale storage bindings were dropped.

The suite reverts the package to live in an `afterEach`. These tests share one package and run serially, and the failed-rebind case deliberately ends in `live_fallback`, so without it a failed assertion leaked state into the next test and failed it for the wrong reason. Verified by breaking one assertion deliberately: 1 fail, 7 pass, no cascade.

Verified red/green rather than just green: with the status assignment removed the original assertion fails with `Expected: "bound"`, `Received: "unbound"`; with it in place the file is **8 pass / 0 fail / 90 expect() calls**. `tsc --noEmit` and `eslint` are both clean, and `api-doc.yaml` still regenerates types cleanly.

### Not covered by a test: the reordered failure path

The reorder is reasoned and free — `bindStorageServeBindings` cannot throw, and the post-reload push re-applies whatever is current — but the behaviour it protects is not exercised by a test, because the failure is hard to reach deliberately.

Two things worth recording, since both bound how reachable the original hazard was:

- A colocated entry pointing at a table that does not exist does **not** fail the recompile. The bind completes (`recompiled: true`) and the error surfaces at query time, so the manifest surface cannot induce the throw.
- A broken model source does not either: a per-model compile failure comes back as `compilationError` and is stored as a `Model.fromCompilationError` entry, so the reload completes.

The throw is `pool.loadPackage` failing — rethrowing a `MalloyError`/`ModelCompilationError`, otherwise wrapped as `ServiceUnavailableError`. So the reachable trigger is a package-level load failure rather than bad manifest content or a bad model file. I have not mapped which Malloy failures surface package-level versus per-model, so that boundary is stated as far as I traced it and no further.


sagarswamirao:
Reviewed against head `133f0797`, tracing the mechanism in the source rather than the description, and diffing behaviour against `origin/main`.

**Verdict: the core fix is correct and well-targeted. One real bug in an edge case the change newly reaches, plus a doc claim that is narrower than stated. Neither blocks the central fix.**

## The core fix is right

The diagnosis holds. On `main`, `setBoundManifestUri` runs unconditionally while the status is only ever written by `recordManifestBinding`, reachable solely via `reloadAllModels`. With `hasColocated`/`hadColocated` both false the recompile is skipped, so the status keeps its `"unbound"` field initialiser while `boundManifestUri` names the manifest that just bound. Repurposing the single-call-site setter into `markManifestBound` puts the status write where the "bind completed" fact actually is, instead of inferring it from one tier's count.

The `live_fallback` -> `bound` recovery leg is the more serious half. `markManifestBindFailed` latches `live_fallback`, and pre-fix nothing on the storage-only path ever cleared it, so a healthy package reported degraded indefinitely. Worth noting that the new test covers that direction and not only the latch-on.

## Finding 1: an empty manifest now reports `bound` with nothing bound

`markManifestBound` sets `"bound"` unconditionally, but `bindManifest` is reachable with a manifest that yields zero entries in **both** tiers. `splitManifestEntries` drops entries that are legacy-failed or missing `physicalTableName`, so a manifest of entirely-skipped entries -- or simply `{"entries":{}}` -- fetches successfully and produces two empty maps.

On a package not previously bound:

- `hasStorage=false`, `hadStorage=false` -> storage tier skipped
- `hasColocated=false`, `hadColocated=false` -> `reloadAllModels` skipped, so `recordManifestBinding` never runs
- `markManifestBound(uri)` -> status `bound`, `boundManifestUri` set

Result: `bound` / `manifestEntryCount: 0` / no `storageServeBindings`, on a package serving entirely live. On `main` this reported `unbound`, which was accidentally right.

The sharper sub-case is a **previously bound** package. There `hadColocated` is true, so `reloadAllModels({})` runs, `recordManifestBinding` sets `unbound` and nulls `boundManifestUri` -- and `markManifestBound` then overwrites both back to `bound` plus the URI. So the same input yields `bound` or `unbound` purely as a function of prior state, and in the previously-bound case `markManifestBound` is undoing a status `recordManifestBinding` deliberately computed.

This matters because it is the failure mode this PR fixes, pointed the other way. A caller reading the field to confirm the worker honoured a distributed manifest is told it was honoured when nothing routes; and per the reclaim-gate argument added to the docs here, `boundManifestUri` is then a non-null "live reference" to a manifest whose tables nothing serves from.

A guard matching the existing `recordManifestBinding` semantics:

```ts
pkg.markManifestBound(
   hasColocated || hasStorage ? manifestLocation : null,
);
```

with `markManifestBound` taking `string | null` and reporting `unbound` on null. Whether an empty manifest *should* report `bound` is a legitimate design call -- the ask is that it be a decided one rather than a function of whether the package happened to be bound before.

## Finding 2: the "fully in place" claim does not cover a mid-reload throw

The `manifestBindingStatus` description asserts that on a failed rebind "its colocated substitutions stay compiled in ... and `boundManifestUri`, `manifestEntryCount` and `storageServeBindings` all continue to describe it", and the PR body says the throw precedes the model swap.

That holds for the fetch failure, which is the case the new test exercises: patching to a nonexistent path throws in `fetchManifestEntriesWithTimeout` before either tier applies. It does not hold for the throw the reorder exists to protect against. In `reloadAllModels`, `this.models = nextModels` happens well before `recordManifestBinding(entries)`. A throw between them -- from `pushPreaggregateServeModels`, `discoverDashboards`, or the outer `validateRenderTags` path -- leaves models already swapped to the new manifest's compiled substitutions while `buildManifestEntries` / `manifestEntryCount` / `boundManifestUri` still describe the old one.

`pool.loadPackage` failing does precede the swap, so the trigger named in the PR body is genuinely safe. But the doc states the invariant unconditionally, and a caller building a reclaim gate on it is the stated audience. Suggest narrowing the claim to the fetch/timeout failure, or saying plainly that a throw after the model swap is out of scope.

## On the reorder itself

Traced rather than taken on the "otherwise immaterial" claim, and it holds. `bindStorageServeBindings` writes `storageServeBindings` then pushes to models; `reloadAllModels` re-pushes via `pushStorageServeBindingsToModels` after the swap. `pushPreaggregateServeModels` reads only `buildManifestEntries` (colocated), never the storage bindings, so recompile-first does not starve the preaggregate companion. `bindStorageServeBindings` also cannot throw as claimed -- `deriveServeBindings` plus a filter whose miss branch logs and returns false.

Hoisting both `had*` reads above either tier is genuinely required once the order flips. `hadStorage` read after `reloadAllModels` would still be correct today, but the hoist removes the dependence on that.

## Smaller notes

- The `markManifestBound` docstring is roughly fifteen lines of rationale on a two-line setter, and most of it describes `bindManifest`'s tier split rather than this method's contract. The reasoning is worth keeping, but it reads better at the call site, which already carries part of it.
- The three new tests each end with `patchPackage({manifestLocation: null})` + `getPackage(true)` for cleanup, but not in a `finally`. In a serial file an assertion failure mid-test leaks bound state into the next one, and the failed-rebind test deliberately leaves the package in `live_fallback`, so a failure there would cascade rather than fail alone.
- Calling out what is not covered -- the reordered failure path, and the unmapped package-level versus per-model Malloy failure boundary -- is the right call. Stating the trace boundary beats overclaiming it.

## Recommendation

Approve the core fix. Finding 1 is worth resolving before merge: it is a one-liner in the same function, and left as-is the field is self-inconsistent across two paths. Finding 2 is a doc-scope correction. Neither undermines the central change, which closes a real non-convergent rebind loop and a real fail-dangerous latch.

sagarswamirao (approved):

PR #1070, opened 2026-08-25
feat(server): return the model file's text on the compiled model

girishjeswani:
`GET …/models/{path}` now carries `sourceText`: the `.malloy` file verbatim, beside the compiled `sourceInfos` / `queries` / `sources`. A client that shows a model's code next to its compiled view (Credible's model explorer is the first) had no way to fetch the text: nothing on the API served it, and `PublisherStaticProxy`-style file routes cover `public/` assets only.

## Mechanism

- The text is read from the `Package` on `getPackage`'s **lock-free fast path** (`Package.getModelFileText`), the path `Model.getModel()` is already on. `environment.ts` requires callers reachable from that path to stay lock-free, and the package lock would buy nothing here: it protects multi-step disk work from a half-swapped tree, while one `readFile` against a swap sees the old file, the new file, or `ENOENT`, and only the last needs handling (below). The read itself is cheap: page-cached, and `loadViaWorker` already reads every `.malloy` on the main thread for `assertPersistNamesQuoted`.
- A read that fails logs a warning and **omits** the field rather than failing the request: the compiled model was never on disk to begin with, and the spec marks `sourceText` optional. A client reads it structurally and treats absence as "this server doesn't provide it".
- Named `sourceText`, not `source`: `sources` on the same object are the model's Malloy sources, and a public field one letter away with a different meaning would be a permanent trap for generated clients and agents reading the response.

## Decisions and tradeoffs

- **Additive and optional**, so older servers and clients stay compatible in both directions; no version bump of the API is needed.
- **Read per request, not stashed on `Model`.** The alternative is to keep the text on `Model` at load (the main thread already has the bytes and discards them), which would remove request-path IO *and* close a small skew: `Promise.all` pairs an in-memory compile (the cached `modelDef`) with a separate disk read, so an `updatePackage` swap landing between them can return the new tree's text beside views compiled from the old one, with nothing saying they disagree. Low harm for a read-only pane, and the fix is not one lock around both (that would put materializer hydration inside an exclusive per-package mutex). The cost of stashing is every loaded model's text resident under a `package_memory_governor` that already sheds queries. Left as a follow-up; revisit if the skew or the read shows up.
- **Not a query parameter.** The field is optional regardless, so every consumer shape-checks anyway; a request param would also have to be declared and plumbed through every hand-maintained downstream spec and proxy, which is what makes "the client lights up with no further change" possible. The one consumer that pays for the payload without wanting it is a per-model cache with an hourly TTL.
- **Not done:** a separate raw-file endpoint. One request for "the model and its text" is what the client needs, and a second endpoint would be a second contract to keep compatible.

## Testing

`packages/server`: `bun test src/controller/model.controller.spec.ts` 3/3; `bunx tsc --noEmit` and `bunx eslint` clean on the touched files.

The assertion is load-bearing: with the controller returning `compiled` alone, the first test fails

```
(fail) ModelController.getModel > returns the compiled model with the file's text alongside, read on the lock-free fast path
  Expected  - 4
  Received  + 0
-   "sourceText": "source: flights is duckdb.table(\"flights.parquet\")\n",
```

The degrade path is covered (`getModelFileText` rejecting with `EACCES` → the compiled model returns without `sourceText`), as is the pre-existing notebook refusal happening before any read; the fake environment has no `getModelFileText`, so a regression back onto the locked read fails the fixture.

Verified end to end against a local build of this branch serving the FAA sample: `GET /api/v0/environments/demo/packages/faa/models/flights.malloy` returns `sourceText` with 158 lines, and the Credible app's code view renders it.


packages/server/src/controller/model.controller.ts
```diff
@@ -63,7 +63,30 @@ export class ModelController {
...
             throw new ModelNotFoundError(`${modelPath} is a notebook`);
          }
-         return await model.getModel();
+         // The compiled view and the file's own text, together: the file is on
+         // disk beside the package, and a client showing code next to the model
+         // otherwise has no way to fetch it. Read through the environment, whose
+         // package lock keeps the read off a tree mid-install/delete swap; a
+         // read that still fails withholds only the text, never the compiled
+         // model the spec marks it optional beside.
+         const [compiled, sourceText] = await Promise.all([
```
Sha-Bang, line 74: the `readFile` itself is noise — page-cached, and `loadViaWorker` already reads every `.malloy` on the main thread for `assertPersistNamesQuoted` (`package.ts:672`). the cost is that `getModel` now acquires the package's **exclusive** mutex, the same one `installPackage` / `updatePackage` / `deletePackage` hold across a tree swap. so concurrent `getModel` calls on one package serialize against each other, and one landing during an install waits the install out instead of being served from memory.

that matters because `getPackage`'s fast path is lock-free on purpose, and its comment states the rule this crosses: *"callers that consume the returned Package on the fast path (notably the MCP query tools and `Model.getModel()`) must remain in-memory only."* this is the first thing to put lock acquisition on that read.

I don't think the lock will be impactful for a single `readFile`, though. it protects multi-step disk work — `compileSource`, `reloadAllModelsForPackage` — from a half-swapped tree. one read against a staging-then-swap sees the old inode, the new file, or `ENOENT`; two are valid text and the third already degrades to omit-with-a-warning, which is the path this PR built. so what if the read just skips the lock and stays on the fast path?

related, and it's why I'd not go the other way and widen the lock to cover both: `Promise.all` pairs an in-memory compile (`getStandardModel()` off the cached `modelDef`) with a separately-locked disk read, so an `updatePackage` swap landing between them returns text from the new tree beside views compiled from the old one, with nothing saying they disagree. low harm for a read-only pane — but the fix isn't one lock around both, that would put materializer hydration inside an exclusive per-package mutex. stashing the text on `Model` at load closes the skew *and* the request-path IO, since the main thread already has the bytes and discards them; the cost there is every loaded model's text resident under a `package_memory_governor` that already sheds queries, which is worth naming in the tradeoffs section either way.

on the option not taken: I'd stay away from making this a query param. the field is optional regardless, so every consumer shape-checks anyway; on our side the dataplane spec has to declare the response field no matter what, and a param means also declaring a request parameter in a drift-checked hand-maintained spec plus plumbing it through `EnvironmentService.getModel`, which costs the "lights up with no frontend change" property. the one consumer that pays for the payload without wanting it is the router's `DeclaredGivensCache`, and that's a 60-minute-TTL cache keyed per model — once an hour, not once a query.

Sha-Bang (commented):
looks good. the field name is correct — `sourceText` beside `sources` would have been a permanent trap one letter apart — and the degrade path is tested rather than asserted. read-only review, I didn't run the suite.

one thing I'd want changed before merge, and it's about the lock rather than the read. details inline at `model.controller.ts:72`.

packages/server/src/controller/model.controller.ts
girishjeswani: Agreed, and thanks for the trace: the lock protected nothing a single `readFile` needs, and it broke `getPackage`'s fast-path rule to do it. In 956f7653 the read is back on the package (`p.getModelFileText`), lock-free, keeping the catch → omit-with-warning; the spec's fake environment now has no `getModelFileText`, so a regression onto the locked read fails the fixture. The compile/read skew and the stash-on-`Model` alternative (with its memory-governor cost) are named in the tradeoffs section, as is why this stays a response field rather than a query param.

Sha-Bang (approved):

Issue #1069, opened 2026-08-25
Composite dashboards use a second chrome implementation that diverges from the # dashboard renderer


Sha-Bang:
Slack: https://credible-data.slack.com/archives/C085VT95H2T/p1787677123472499?thread_ts=1787612061.871039&cid=C085VT95H2T

Raised by @jswirhun while testing service#6077: "why do we have 2 different `# dashboard` implementations => one in malloy render, one in publisher sdk?" and, after a first pass claimed there was only one, "they render differently though".

He is right. Both dashboard forms delegate *leaf result* rendering to `@malloydata/render`, so there is one rendering library — but the dashboard chrome around those results is built twice, and the two produce visibly different pages.

## The two implementations

**Single-query form** (`# artifact` on a `query:`) imposes no publisher layout at all. `Dashboard.tsx:323-336` renders one `DashboardTile` with no grid, deliberately:

> Its `# dashboard {columns=N}` tag is the renderer's business, so no grid is imposed here: doing so would nest a grid in a grid.

Layout, cards, `# colspan`, `# break` and the responsive behaviour are all the renderer's (`malloy-render/src/component/dashboard/dashboard.tsx` + `dashboard.css`).

**Composite form** (`## artifact { tiles=[…] }`) builds its own grid in `Dashboard.tsx:344-350` and its own card chrome in `DashboardTile.tsx`, because each tile is a separate query execution and no single Malloy result spans them.

## How they diverge

| | single-query (renderer) | composite (SDK) |
|---|---|---|
| grid | `.dashboard-row-body`, flex or 12-col grid | MUI CSS grid, `repeat(dashboardColumns ?? 2, …)` |
| card | `.dashboard-item` — 8px radius, 20px padding, 1px ring + two drop shadows | `Paper elevation={0}` — host MUI radius, flat 1px `divider` border, no shadow (see the geometry table below) |
| tile title | `# label` on the nest | derived from the run expression (`tileTitle` / `humanizeSlug`) |
| per-tile width | `# colspan` | none, equal width only |
| row breaks | `# break` | none |
| responsive | container queries, collapse buckets at 400 / 600 / 900px | one `xs` → `md` breakpoint |
| height | virtualized, `# dashboard { table.max_height }` | `height ?? TILE_HEIGHT` (400) per tile, vs `WHOLE_PAGE_HEIGHT` (20000) for single-query |
| sticky row header | yes (`.dashboard-row-header`) | no |

[...]

PR #1072, opened 2026-08-26
test(app): pin that the instance theme's font reaches renderer-drawn text

mlennie:
Adds one Playwright test. No SDK, app or server source changes.

## Why

Nothing asserted that a host's `theme.font.family` actually affects the text the renderer draws. `buildTableCssVars.spec.ts:58` pins that `--malloy-render--font-family` is *emitted*, which is true regardless of how the cascade resolves, so emission and effect were never connected by a test.

That gap is currently hiding a trap, which is the real reason for this PR.

`@malloydata/render` declares `font-family: var(--malloy-render--font-family)` on `.malloy-render`, and then in the same stylesheet an `@supports (font-variation-settings: normal)` block re-declares `font-family: InterVariable, Inter, system-ui sans-serif`. That later rule is at equal specificity and would win, pinning every embedding host to Inter whatever it passes to `ServerProvider`.

It does not win, because the declaration is invalid CSS: `system-ui sans-serif` is an unquoted multi-word family name, and both `system-ui` and `sans-serif` are generic-family keywords, so the identifier sequence is invalid and the browser drops the whole declaration. Read back from `document.styleSheets`, that block computes to `font-feature-settings: "liga", "calt"` with no `font-family` surviving.

So host font theming currently works because of a missing comma. Adding the comma, which looks exactly like a typo worth tidying, would silently pin every host's dashboards to Inter, and nothing in this repo would fail. If anyone does touch that declaration upstream, the right change is to delete the `font-family` line and keep `font-feature-settings`, which is evidently what the block is for.

## What the test does

Serves an instance theme with a sentinel font family through `/api/v0/status`, opens the existing dashboards fixture's single-query `grid` dashboard (where the renderer draws the whole page rather than the SDK's tile chrome), and asserts the computed `font-family` on `.malloy-render` and on a `.column-cell.td` inside it follows the theme.

Three details that are deliberate:

- **Computed value, not the emitted variable.** That distinction is the whole point; a test at the variable level passes whichever way the cascade resolves.
- **`.column-cell.td`, not `.column-cell`.** Header cells carry `column-cell` too and come first in DOM order, so a bare selector pins the `th`. `.td` is the body text a reader actually sees, and matches this file's existing convention at :474 and :541.
- **The sentinel family need not exist on the machine.** `getComputedStyle` returns the specified font-family list rather than the face that resolved, so this checks the cascade and not the font inventory.

## Verification

Verified in both directions rather than only observed passing:

- Passes against unmodified `main`. It is a regression guard, so that is the correct result.
- Injecting the comma-corrected rule at runtime, which is what a tidy-up commit would produce, flips the computed font to `InterVariable, Inter, system-ui, sans-serif` and fails the assertion. So the guard discriminates on exactly the hazard it exists for.

Also measured, for the record, since it corrects something I got wrong on #1069: the operator's font reaches renderer tables, cells and chrome today. It is not confined to the Vega marks. Unthemed, the computed family is the SDK's `DEFAULT_FONT_FAMILY`.

Gate: lint, typecheck and prettier clean, server unit suite 3177 pass / 0 fail.

## Notes for review

Timeouts are sized against the 60s per-test budget on purpose: a 30s visibility wait plus a 15s poll fits, where the 60s wait I first wrote could never be spent and would have surfaced a cold-render failure as a bare test timeout naming no wait.

This may be worth landing alongside the Phase 3 assertion discussed in #1069 rather than separately, since it is the same shape of gap: something the theme is expected to control, with no test connecting the token to its effect.

Refs #1069

### DIFF
--- a/packages/app/tests/playwright/package-dashboards.spec.ts
+++ b/packages/app/tests/playwright/package-dashboards.spec.ts
@@ -943,4 +943,94 @@ test.describe("package-dashboards", () => {
       ).toBeVisible({ timeout: 30_000 });
       await expect(page.locator("iframe")).toHaveCount(0);
    });
+
+   // REGRESSION GUARD, not a fix. The operator's font already reaches the text
+   // the renderer draws, and this pins that it keeps doing so.
+   //
+   // It is guarding a genuinely fragile arrangement. The renderer declares
+   // `font-family: var(--malloy-render--font-family)` on .malloy-render and then,
+   // in the same stylesheet, an `@supports (font-variation-settings: normal)`
+   // block re-declares `font-family: InterVariable, Inter, system-ui sans-serif`.
+   // That later rule would win on source order and pin every host to Inter. It
+   // does not, because `system-ui sans-serif` is an unquoted multi-word family
+   // name, and both `system-ui` and `sans-serif` are generic-family keywords, so
+   // the identifier sequence is invalid and the browser drops the whole
+   // declaration. Verified against `document.styleSheets`: the block declares
+   // `font-feature-settings: 'liga' 1, 'calt' 1` in source and computes to
+   // `font-feature-settings: "liga", "calt"`, with no `font-family` surviving.
+   //
+   // So a well-meaning upstream fix to that missing comma would silently pin
+   // every embedding host's dashboards to Inter, and nothing else in this repo
+   // would notice. This test is what notices.
+   //
+   // Asserted on the computed value rather than the emitted variable: the
+   // variable was always emitted correctly, so a test at that level passes
+   // whether or not the value reaches any text. The sentinel family need not
+   // exist on the machine, since getComputedStyle returns the specified
+   // font-family list rather than the face that resolved.
+   test("the instance theme's font reaches renderer-drawn text", async ({
+      page,
+   }) => {
+      const SENTINEL = "PublisherThemeProbe";
+      await page.route("**/api/v0/status", async (route) => {
+         const res = await route.fetch();
+         const body = await res.json();
+         await route.fulfill({
+            response: res,
+            json: {
+               ...body,
+               theme: {
+                  ...(body.theme ?? {}),
+                  font: {
+                     ...(body.theme?.font ?? {}),
+                     family: `"${SENTINEL}", monospace`,
+                  },
+               },
+            },
+         });
+      });
+
+      // `grid` is the single-query form, so every card and cell on it is the
+      // renderer's own output rather than the SDK's tile chrome.
+      await openDashboard(page, "grid");
+      // The renderer builds its tables from divs, not <table>, which is why the
+      // rest of this suite locates cells by .column-cell.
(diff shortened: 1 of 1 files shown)


packages/app/tests/playwright/package-dashboards.spec.ts
```diff
@@ -943,4 +943,95 @@ test.describe("package-dashboards", () => {
...
+                     ...(body.theme?.font ?? {}),
+                     family: `"${SENTINEL}", monospace`,
+                  },
+               },
+            },
+         });
+      });
+
+      // `grid` is the single-query form, so every card and cell on it is the
+      // renderer's own output rather than the SDK's tile chrome.
```
Sha-Bang: this pins computed font on single-query `grid`, which is the right place for the InterVariable trap. leaf cells on a composite ride the same `.malloy-render` cascade, so they'd follow the sentinel too — but the composite Paper title is MUI `Typography` on app `SANS_FONT_FAMILY` Inter (`packages/app/src/theme`), which never reads `theme.font.family`. "instance theme's font reaches …" is true for renderer-drawn text and still false for SDK chrome titles, so a composite page can still ship Inter titles inside a themed app, and nothing here would catch it. worth a one-liner in the comment naming that split?

Sha-Bang (approved):
the InterVariable trap is real and this is the right pin. one inline on the chrome-title split, not a hold. approving.

packages/app/tests/playwright/package-dashboards.spec.ts
mlennie: Yes, and it is worth more than a one-liner because I checked it and you are exactly right, including the part that makes it non-obvious.

`820bdb1e` extends that comment. What I verified rather than took on faith: a composite tile's title is `<Typography variant="subtitle2">` inside MUI `Paper` at `DashboardTile.tsx:122`, and its `sx` sets only `fontWeight` and `color`, so the family falls through to the app's MUI theme, which is `fontFamily: SANS_FONT_FAMILY` at `packages/app/src/theme/index.tsx:76`, resolving to Inter and reading nothing from `theme.font.family`. So the split is real in both directions you named: a composite's leaf cells do ride the same `.malloy-render` cascade and would follow the sentinel, and its chrome titles would not.

That second half is what makes the guard's own title a little generous. "the instance theme's font reaches renderer-drawn text" is precisely true, and the comment now says what the words "renderer-drawn" are doing there, so the next person to read this test does not conclude the page as a whole is covered.

Not extending the test to cover it, deliberately, and it is worth saying why rather than leaving it implied. A guard that asserted a composite title carries the sentinel would fail today, and it would be right to fail, because that is a real gap rather than a test bug. Pinning current behaviour would say the opposite, that Inter titles inside a themed instance are intended. Either the SDK's chrome should read the instance theme, in which case the assertion belongs in whichever change does that, or it should not, in which case there is nothing to assert. Both are decisions above a test file, so this PR names the gap and stops.

Comment only, assertions untouched. Gate: typecheck and lint clean, and prettier clean on this file (the six standing warnings are the git-ignored generated clients).

PR #1073, opened 2026-08-26
fix(server): scope each dashboard grid-width spelling to the form that honors it

Sha-Bang:
Part of #1069 — the grid-width half.

A dashboard's grid width has two spellings, and the server read both on both forms even though each form honors only one. `dashboard_columns` on the artifact tag is what the composite SDK grid lays out from; `# dashboard { columns=N }` on the query is what `@malloydata/render` reads on a single query, and Publisher never feeds that one.

So on a single query, `# artifact { dashboard_columns=6 } dashboard { columns=3 }` put `dashboardColumns: 6` on the wire, linted clean, and laid out at 3. Worse, `dashboard_columns` *alone* left the renderer with no `# dashboard` tag at all, so the whole grid was lost and the result came out as one nested table — with nothing said at load.

## What it does

- `readArtifactTag` takes the form and gates only the `dashboard_columns` half of the `??`.
- A new load warning when `dashboard_columns` sits on a query-level artifact. It branches on whether a `# dashboard` tag is **present**, not on whether its `columns` is usable, because presence is what decides whether the renderer enters dashboard mode at all: with no tag you get one nested table, but with the tag and a bad `columns` you still get a dashboard in flex mode, so a nested-table claim there would be false. Up to three messages can land on one page (this one, the value check, the renderer's own `renderLogs`) and they must not disagree about what the page did.
- `warn`, never `error`. Every package spelled this way loads today and its page doesn't change.
- **`# dashboard { columns }` stays live on both forms.** A composite spelled that way is honored today, and dropping it would quietly move those dashboards to the SDK's default of 2. Only the inert direction went away; there's a regression test pinning the composite fallback.
- The `dashboard_columns` value check ("must be a positive integer") is now composite-only. On a query-level tag it sent the author to fix a number that was never going to be read.

## Why this isn't the one-spelling fix

This is the deliberately-different fork of the open question in #1069, not convergence — and the reason is in the renderer, not in taste. Publisher can't feed the renderer's grid on the single-query form at all (`Dashboard.tsx:323-337` imposes no layout, on purpose), and the renderer reads `dashboard` off the per-query tag. A model-level `## dashboard { columns=N }` on a composite would be live for Publisher and dead for the renderer — today's bug transposed — and wouldn't even be one spelling, being `#` on a query and `##` at model level.

What this buys: the author who mixes spellings gets told, and the wire field stops reporting a number no page used. What it doesn't buy: anything for the author who spells both correctly and still has to learn two names. That's grammar-home work upstream, so the syntax axis of #1069 stays open.

Keeping the composite fallback means the deferral is "we don't document or recommend the model-level render tag", not "the server doesn't read it."

One gap this deliberately does *not* close, since "the spelling that does nothing now warns" is the pitch: a model-level `## artifact` with no `tiles` falls through to the query loop and its tags go unlinted, so that position still warns about nothing.

## Compatibility

`ApiDashboard.dashboardColumns` narrows in one case: on a single-query dashboard it now reflects only `# dashboard { columns=N }`, so a host reading that field will see it absent where it used to carry an inert number. That's what the `RELEASE_NOTES.md` entry is for.

`api-doc.yaml` is deliberately untouched. Its wording — "from the `# dashboard {columns=N}` render tag, or `dashboard_columns=N` on a composite artifact tag" — is still true, and narrowing it to "on a single query" would document the composite fallback out of existence. No spec change also means no python-client regeneration.

## Tested

- `bun test src/service/dashboard.spec.ts` — 89/90, three of them new. The one failure, `tile normalization > stays linear`, is pre-existing and unrelated: it fails identically at base `f0f3b296`, and what blows the 5s per-test timeout is the deliberately quadratic `oracle()` in the test's *setup*, not the assertion under test. `normalizeTileExpression` is untouched here.
- `bun test tests/integration/dashboards/dashboards.integration.spec.ts` — 45/45. Needs `--timeout` above the 5s default on a slow machine, since the hook that boots the server exceeds it.
- `npm run test:skills` — 15/15. `npm run lint:server` and `npm run typecheck:server` clean (the latter after `npm run generate-api-types:server`, which writes the gitignored `src/api.ts`).
- Skills bundle regenerated and confirmed idempotent; `packages/skills/package.json` bumped 0.1.8 → 0.1.9 as `skills/README.md` requires.
- Checked against a running server rather than only in assertions: a purpose-built package carrying all three single-query spellings, read back off `ApiPackage.warnings`. All three branches produce the right message at severity `warn`, the packages load and serve, and each message agrees with what the renderer actually did — `Unknown render tag 'colspan'` for the no-tag case, `Ignored # colspan … only applies in columns mode` for the flex-mode case, nothing for the good one. Other direction too: `dashboards-test`'s `combined` keeps 4, `overview` keeps 6, `grid` keeps 12, and that package still reports zero warnings.


Sha-Bang:
Superseded by #1080, which takes the other fork of the open question this PR deliberately left open.

This one scoped each grid-width spelling to the form that honors it and kept both. #1080 removes the choice instead: `dashboard_columns` is deleted, reader included, and `# dashboard { columns=N }` is the one spelling on both forms. The `form` parameter added here was used at exactly one line, the gate that #1080 deletes, so there is nothing left of this diff to rebase.

What made the other fork viable in the meantime is per-tile layout. This PR's reasoning for keeping both was that Publisher could not feed the renderer's grid on the single-query form, so consolidating meant equal-width tiles and an inexpressible flagship layout. #1080 closes that from Publisher's own side: `readDashboardModelFacts` already walks each source's view fields, so the renderer's per-child dashboard tags come off the view a tile names and drive Publisher's grid directly, with no upstream dependency.

Closing. The warning this PR added is not lost, it is generalized: #1080 enumerates the artifact tag and names every property the reader does not read, `dashboard_columns` included, so the same author gets told the same thing plus the spelling to use.

PR #1074, opened 2026-08-26
feat(server): let a build session opt out of preserve_insertion_order

housejester:
## What

Adds `PUBLISHER_BUILD_PRESERVE_INSERTION_ORDER`, an opt-in override for DuckDB's `preserve_insertion_order` on the materialization build session. **Unset — the default — issues no statement at all**, so DuckDB's own default stands and no existing deployment changes behaviour.

## Why

A storage build is one pipeline: a passthrough `SELECT` the source warehouse executes, whose result is written straight into the destination table. DuckDB's execution is pull-based, so the writer paces the read and the pipeline is naturally bounded — the write-side buffers are bounded too (`write_buffer_row_group_memory_limit`, `write_buffer_row_group_count`).

Order preservation sits outside that. To hand rows to the sink in input order, DuckDB buffers them between a parallel scan and the writer, and that buffer grows with **how far out of order the pipeline runs** — not with any batch size, and not under the per-batch bound that makes the rest of the pipeline safe. On a large result it is the one part of an otherwise-bounded pipeline that can outgrow the process.

And a materialization has no row-order contract to keep. The built table is read back through the serve transform, which neither preserves nor promises row order, so on this path the guarantee costs memory and buys nothing.

## Why opt-in rather than just turning it off

It is a global session setting, and the blast radius of flipping it for every build is wider than the one build that needs it. Leaving it off by default means:

- no behaviour change on upgrade,
- the value is a deployment decision rather than a code one, so tuning it needs no release,
- and an operator can turn it on for the environment that needs it without reasoning about every other build.

Scoped to the build session for the same reason — that is the one place the tradeoff is unambiguous.

## Shape

- `config.ts` — `getBuildPreserveInsertionOrder(): boolean | undefined`, via the existing `parseBoolEnv`, alongside the `PUBLISHER_MEMORY_*` family. Returns `undefined` when unset; throws on a malformed value so a typo in a manifest is a loud startup failure rather than a silently ignored setting.
- `materialization_build_session.ts` — `applyBuildInsertionOrder(session)`, a sibling to the existing `pinSessionToUTC`, called on both build-session paths (both write, so both buffer). Returns without issuing anything when unset.

## Test

Four cases in `materialization_build_session.spec.ts`, against a recording stub so they assert the SQL actually issued rather than the parse alone:

- unset → **no statement issued** (the default is genuinely a no-op)
- `false` → `SET preserve_insertion_order = false`
- `true` → honoured explicitly, not treated as unset
- malformed → throws

Verified red/green rather than just green: with the body of `applyBuildInsertionOrder` stubbed out, 3 of the 4 fail (the unset case correctly still passes, since a no-op satisfies it). With it in place the file is **31 pass / 0 fail / 36 expect() calls**. `tsc --noEmit` and `eslint` clean.

## What this does not do

It does not bound the pipeline's memory on its own. `memory_limit` is still DuckDB's default — 80% of the detected limit — which for an embedded engine sharing a process with a resident host is an overcommit before any query runs. That is a separate change; this one removes the unbounded buffer, it does not add a ceiling.

### DIFF
--- a/packages/server/src/config.ts
+++ b/packages/server/src/config.ts
@@ -274,6 +274,28 @@ export const getMemoryGovernorConfig = (): MemoryGovernorConfig | null => {
 };
 
 /**
+ * Resolve the materialization build session's `preserve_insertion_order`
+ * override from `PUBLISHER_BUILD_PRESERVE_INSERTION_ORDER`.
+ *
+ * `undefined` when unset — the default — and the build then issues no such
+ * statement at all, so DuckDB's own default stands and this knob cannot change
+ * an existing deployment until someone sets it. Opting in is the whole
+ * interface.
+ *
+ * Scoped to the build session on purpose. It is a global session setting, and a
+ * build session is the one place where the tradeoff is unambiguous: the
+ * pipeline is a single passthrough read written straight into a destination
+ * table, and nothing downstream depends on the order the rows were written in.
+ * See `applyBuildInsertionOrder` for why the buffer it governs is not bounded
+ * by batch size.
+ *
+ * Throws at startup on a malformed value, so a typo in a manifest is a loud
+ * failure rather than a silently ignored setting.
+ */
+export const getBuildPreserveInsertionOrder = (): boolean | undefined =>
+   parseBoolEnv("PUBLISHER_BUILD_PRESERVE_INSERTION_ORDER");
+
+/**
  * Settings for the optional embedding provider behind semantic
  * `malloy_getContext` retrieval. See {@link getEmbeddingConfig}.
  */
--- a/packages/server/src/service/materialization_build_session.spec.ts
+++ b/packages/server/src/service/materialization_build_session.spec.ts
@@ -1,7 +1,7 @@
 // Copyright (c) Credible Data Inc.
 // SPDX-License-Identifier: MIT
 
-import { describe, expect, it } from "bun:test";
+import { afterEach, describe, expect, it } from "bun:test";
 import { DuckDBConnection } from "@malloydata/db-duckdb";
 import { mkdirSync, mkdtempSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
@@ -10,6 +10,7 @@ import { BadRequestError } from "../errors";
 import { storageDestinationRoot } from "./connection_config";
 import type { components } from "../api";
 import {
+   applyBuildInsertionOrder,
    assertStorageServeShapeCompiles,
    buildDownstreamIntoStorage,
    buildSourceIntoStorage,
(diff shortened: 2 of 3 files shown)


girishjeswani (approved):
Approving. The code is correct and correctly scoped, and I verified the safety property rather than taking it on trust: mutating the guard to `getBuildPreserveInsertionOrder() ?? false` produces exactly one failure, the unset case, with the diff showing the spurious statement. Unset really does issue nothing. Two documentation items I'd fix before merge, and one substantive note on the framing.

**The "no row-order contract" argument is right, but I think the reasoning inverts.** A declared `order_by` does get baked into the table at build and is not re-applied at serve: `PersistSource.getSQL()` compiles the inner query as a top-level query so Malloy emits ORDER BY into the persist SQL, and `SourceRefinement` in `materialization_serve_transform.ts` has no ordering category at all. Every `reduce` persist source even picks up a default ORDER BY via `calculateDefaultOrderBy()` with nothing written. So physical row order is the only place a declared ordering survives, which makes "the serve transform doesn't promise order, therefore order doesn't matter" the wrong way round.

The change is still safe, and I think for a better reason than the body gives. I ran the four real pipeline shapes against the pinned `@duckdb/node-api` 1.5.5, measuring `corr(rowid, id)` after resetting the flag so the read itself was in physical order:

| shape | preserve=true | preserve=false |
| --- | --- | --- |
| local `ORDER BY` then CTAS (colocated) | sorted, corr 1.0 | sorted, corr 1.0 |
| ordered single-stream table function then CTAS (`postgres_query`/`snowflake_query`) | sorted, corr 1.0 | sorted, corr 1.0 |
| `ORDER BY` in a derived table plus outer watermark WHERE (`deltaSelect`) | sorted, corr 1.0 | sorted, corr 1.0 |
| parallel multi-file parquet scan, no ORDER BY (`buildDownstreamIntoStorage`) | sorted, corr 1.0 | scrambled, corr 0.04 |

A sort operator or a single ordered producer is a pipeline breaker feeding one stream, so `preserve_insertion_order=false` has nothing to reorder. Order is only lost where the producer genuinely runs parallel and out of order, which is where the input order was arbitrary to begin with. That seems worth putting in the body, since it is the property that actually makes this safe.

**On "the guarantee costs memory and buys nothing":** insertion order does buy serve-side row-group pruning. On the chained shape I measured the `ts` zonemap average span widening from 31,495 to 215,393, roughly 6.8x, and the segments overlapping a narrow range filter going from 2 to 3. It scales with thread count, and since DuckLake prunes files and row groups by catalog stats the same cost applies there and is probably larger. Not an objection to the flag, but an operator reading "buys nothing" will turn it on without measuring read cost, so I'd state the tradeoff.

Related, and the thing I am most curious about: the win may be zero on the primary path. `buildSourceIntoStorage` reads through `postgres_query`/`snowflake_query`, and those table functions are single-stream producers, so there is no reorder buffer to bound. The win looks concentrated on the chained/parallel path, which is also where the pruning cost lands. Which path and source dialect showed the unbounded buffer, and what were the before/after numbers? #929 cited "8.0 to 0.70MB per build" and a figure here would settle whether this wants to be scoped to `buildDownstreamIntoStorage` only.

**Two doc fixes:**

1. `docs/configuration.md` is currently complete for every var `config.ts` reads. I diffed the full set of `PUBLISHER_*`/`EMBEDDING_*`/`PACKAGE_LOAD_*`/`EXTENSION_FETCH_*`/`PERSIST_STORAGE_*` names against the table and the difference is empty, so this would be the first gap. A release note too, since the recent feature commits all touch one or both.

2. The docstring at `config.ts:279-292` says it "throws at startup on a malformed value, so a typo in a manifest is a loud failure." The only call site is inside `applyBuildInsertionOrder`, so it is read per build: a typo boots the server fine and then fails every materialization mid-build, after the destination is attached. `EMBEDDING_INDEX_CONNECTION_SCHEMA` documents exactly this distinction the other way ("read when the tool is called, not at startup"). Either call the getter once in `server.ts` next to `getMemoryGovernorConfig()`, or reword.

**Two smaller ones.** `duckdb_settings()` on 1.5.5 reports `preserve_insertion_order` with `scope = GLOBAL`, and I confirmed a second connection on the same instance sees the change, so it is instance-scoped rather than connection-scoped. Nothing leaks today because `createIsolatedBuildSession` gives each build its own instance through the unique-`workingDirectory` share-key trick, but that function documents itself as a workaround for a `db-duckdb` pooling quirk, so it is worth saying "instance" and naming the dependency: if that isolation ever regressed, a GLOBAL set would reach the long-lived serve connection. And the `SET` is unguarded where `tagSnowflakeSession` warns and continues and the ducklake `SET` at `:820` is dialect-guarded. Failing loudly is defensible when someone has opted in, so just a one-line note on why this one differs would match the file.

Scoping and lifecycle both check out, for the record: both write paths get it, the two non-writing sessions correctly do not, and it lands after the destination ATTACH and `pinSessionToUTC` but before `federateSourceForPassthrough`, so the incremental DML on the same session is covered. Parsing reuses `parseBoolEnv` verbatim and keeps `boolean | undefined` with no defaulting, so unset and explicit false stay distinct.

Last one, unrelated to this PR but it surfaced while checking the claim: should the source's baked `order_by` be re-applied by the serve transform, or documented as deliberately dropped? It is the one refinement category with no comment, doc line, or spec case explaining its absence, where filters, joins, views and access modifiers all have one.

PR #1076, opened 2026-08-26
fix(server): storage serving on a case-folding warehouse, and surface the fallback reason

housejester:
## What

Two changes to the `storage=` serve path, one the reason the other was findable.

1. **`narrowSchemaToPublic` now matches a captured column to its author field case-insensitively**, emitting the author's spelling. This is the fix: on a warehouse that folds unquoted identifiers, storage serving never worked at all.
2. **The serve-shape fallback logs at `info` instead of `debug`**, matching the storage-hit line beside it.

## The bug

`narrowSchemaToPublic` intersects the captured `DESCRIBE` schema against the compiled source's field names:

```ts
return schema.filter((c) => publicNames.has(c.name));
```

Those two sides are in different identifier namespaces. `publicNames` holds the names the **author wrote**; `schema` holds what the **source warehouse reported** for the built table. Snowflake, Oracle, Redshift and Teradata all fold unquoted identifiers to upper case, so a source the author wrote as

```malloy
#@ persist storage="lake"
source: traffic_by_product is traffic -> {
  group_by: product_id
  aggregate: page_views is total_page_views
}
```

captures as `PRODUCT_ID`, `PAGE_VIEWS`. The intersection is **empty**.

An empty narrowed schema is not a partial loss. The caller drops such a binding entirely — deliberately, because an empty `type: X__shape is {}` is a parse error that would fail the whole shape model. So the source leaves the serve shape, and every query on it fails the per-query compile with `Reference to undefined object 'traffic_by_product'` and is served live.

Three things made this hard to see:

- It is **silent**. The query succeeds and the rows are correct; only the tier is lost. It presents as materialization having no effect, not as a fault.
- **Aggregate-only queries fall back too.** `-> { aggregate: c is count() }` names no column, so nothing about the *query* is case-sensitive — but the whole *source* is gone from the shape. Any fix that only reconciled field references would look right and still be wrong, so there is a test pinning exactly this.
- **No tier drop fires.** `compileServeShape` degrades full → drop views → drop views+joins → base-only, and logs a `warn` each time. None of that happens here: the shape compiles cleanly, it just has no sources in it. So the counters read normal and only `outcome="live_fallback"` moves, with no reason attached — which is change (2).

BigQuery and Postgres preserve lower case, which is why every existing test and every deployment on those warehouses routes correctly.

## The fix

Match on the case-folded name; emit the **author's** spelling, because that is the name a query references. The declaration still reads the right physical column: the serve shape resolves against DuckDB, which matches identifiers case-insensitively — so **no already-built table needs rebuilding**. This is a read-path change only.

Access filtering is unchanged, and this is the property most worth stating. Folding widens which *physical* column can match an author field; it never adds an author field. An `except:`-ed or access-restricted column has no public name to match in any case, so it stays out of the shape — pinned by a test that puts a physically-present `SECRET` next to a hidden `secret` and asserts the query is still refused rather than served from storage.

One edge case: two physical columns differing only in case fold onto one author field. Declaring that name twice is a duplicate that would fail the entire shape, so the first wins and the rest are dropped — the source keeps serving instead of the ambiguity costing storage serving for every source in the model.

## Why the log level moved

The reason for a fallback is assembled at the throw site and then discarded at `debug`. The storage **hit** already logs at `info`, and a shape **tier drop** logs at `warn`, so the per-query fallback was the outlier — and it is the half an operator needs:

```
storage serve-shape ineligible for this query; serving live
  {"error":"Reference to undefined object 'traffic_by_product'"}
```

That one line is the whole diagnosis above. At `debug` it is unreachable without raising the level on a live process and re-running the query, which is the one thing you cannot do for a fallback that already happened. Volume is bounded by the hit line's own bound: one per routed query, only on packages that declare `storage=`.

## Test

`bun test src/service/model_storage_serve.spec.ts src/service/model_storage_serve_joins.spec.ts src/service/materialization_serve_transform.spec.ts` — **76 pass / 0 fail / 103 expect() calls** (was 70 before the six added here).

Verified red/green, not just green. Reverting `narrowSchemaToPublic` to the exact-match one-liner and re-running fails exactly the six new tests and nothing else:

```
(fail) narrowSchemaToPublic > matches an UPPERCASE captured name and emits the author's spelling
(fail) narrowSchemaToPublic > still hides a non-public column when the captured name differs in case
(fail) narrowSchemaToPublic > emits a case-colliding author field once rather than failing the shape
(fail) storage= serve routing (end-to-end) > routes when the captured schema is in the warehouse's identifier case
(fail) storage= serve routing (end-to-end) > routes a query that references no field when the case differs
(fail) storage= serve routing (end-to-end) > does not serve an except:-hidden column from storage when the case differs
 65 pass  6 fail
```

The end-to-end cases go through the real `Model.getQueryResults` against a real in-memory DuckDB, where the original source and the storage table return different values — so the value observed proves which path ran rather than asserting on an internal. The physical column is created as `"TOTAL"` and the binding captures `TOTAL`, reproducing the fold.

### Validated against a live Snowflake account

The unit and end-to-end tests above reproduce the fold with a DuckDB column created as `"TOTAL"`, which exercises the same code path but not the same warehouse. So the whole chain was also run against a real Snowflake connection — build a source into a storage destination, then serve from it — with the fix and with `narrowSchemaToPublic` reverted to the exact-match one-liner:

```
buildSQL: SELECT 'alpha' AS who, 41 AS n UNION ALL SELECT 'beta', 1
captured schema = ["WHO","N"]          # BigQuery, same shape, gives ["who","n"]

  with the fix:     n.sum() -> servedFrom=storage  value=42     count() -> servedFrom=storage  value=2
  without the fix:  n.sum() -> servedFrom=null     value=0      count() -> servedFrom=null     value=1
                    "storage serve-shape ineligible for this query; serving live"
                      {"error":"Reference to undefined object 'X'"}
```

The live source and the built table return different values (0 vs 41, 1 row vs 2), so the number observed proves which path ran. Storage values only appear with the fix, and the fieldless `count()` is covered alongside the field-referencing query.

Worth noting for anyone reproducing this: the Snowflake extension is a wrapper over the ADBC Snowflake driver and `INSTALL snowflake FROM community` does not bring it, so a local run fails at query time with `ADBC Snowflake driver (libadbc_driver_snowflake.so) not found`. The `Dockerfile` installs it; a host run needs it fetched separately. That is plausibly part of why this path had no coverage.

`tsc --noEmit` clean (0 errors); `eslint` clean on all four touched files.


girishjeswani:
Nice find, and the diagnosis is the good part: the namespace split between author-written and warehouse-reported names is exactly right, and I confirmed `narrowSchemaToPublic` really is the only place the two sides meet, so the fix is in the right spot. I also checked the assumption it rests on: `CREATE TABLE t AS SELECT 60 AS "TOTAL"; SELECT "total" FROM t` returns 60, so DuckDB folds even quoted identifiers and emitting the author's spelling does read the right column with no rebuild. The comment at `:547` writing that dependency down is worth keeping.

One case I'd handle before this goes in. `publicNames.set(name.toLowerCase(), name)` at `:554` is last-write-wins, and Malloy allows two fields that fold together (a `TitleCase` dimension over a `snake_case` column):

```
source: X is duckdb.sql("SELECT 1 AS total, 2 AS other") extend {
  dimension: Total is other * 10       // live answer: 20
}
```

Through `getQueryResults` on head, `group_by: Total` comes back `servedFrom=storage` with `Total=1`. On main it's 20. Narrowing emits `Total` for physical `total`, the duplicate makes the tier drop, and `compileServeShape` returns base-only without a probe since base-only always compiles, so the wrong column serves silently.

Exact match first, then fold, and drop an ambiguous fold, fixes it in about two lines. I tried it: `Total=20`, `total=1`, and the upper-case fold still works. Only fallout is the captured-side collision test at spec:638 needing a new expectation, and that input looks unreachable anyway since DuckDB rejects `("AMOUNT" INT, "amount" VARCHAR)` and dedupes to `amount_1` on CTAS.

Worth noting this one only bites the case-preserving warehouses, so it's a narrow window on the way to fixing Snowflake.

housejester:
Fixed in a8fdc1a4 — you were right, and it was worse than a failed shape. Reproduced your case exactly before changing anything:

```
narrow -> [{"name":"Total","type":"BIGINT"}, {"name":"other","type":"BIGINT"}]
warn: shape failed to compile ... Cannot redefine 'Total'   (tier 0, then 1, then 2)
"group_by: Total" -> servedFrom=storage  Total=1        <- should be 20
"group_by: total" -> servedFrom=null
```

So the collision also cost the *physical* column its own storage serving, not just the computed one — `group_by: total` was falling back on top of `group_by: Total` returning the wrong number.

Took your fix. Exact hit first, then a fold, and an ambiguous fold is dropped rather than guessed — a query touching it goes live, where the author's own names still distinguish the two. After:

```
"group_by: Total" -> servedFrom=storage  Total=20
"group_by: total" -> servedFrom=storage  total=1
```

Both correct, both from storage, and no tier drop at all now.

Two notes on the details:

- The captured-side collision test did **not** need a new expectation. With exact-first, captured `AMOUNT` takes the sole folded candidate `amount`, then captured `amount` hits exactly, finds it already emitted, and drops — so the result is unchanged. Left it as is, and agreed the input looks unreachable anyway. Added the ambiguous-fold case as its own test instead.
- Added the end-to-end case as well as the unit one, asserting the **value** rather than that the shape compiles. The unit fix alone wouldn't have caught the symptom you found, because the shape compiling was never the property that mattered.

Also kept the `:547` note on the quoted-identifier dependency, and thanks for confirming it independently — that assumption is what makes this a read-path fix with no rebuild.

`bun test src`: 3178 pass / 3 skip / 0 fail. Reverting `narrowSchemaToPublic` to fold-first fails exactly the three new tests and nothing else.

girishjeswani (approved):
LGTM!

sagarswamirao:
Traced this at aac779fd. The diagnosis holds up and the exact-before-fold ordering is the right call -- I reproduced all five new unit cases against the extracted logic and checked the DuckDB premises against a real DuckDB 1.5.3, including the one the fix rests on: `CREATE TABLE t AS SELECT 60 AS "TOTAL"; SELECT total FROM t` returns 60, so emitting the author's spelling does read the right column with no rebuild.

One thing worth adding to the write-up, because it makes the fix look narrower than it is. I went looking for whether `projectToPublicColumns` already prevents the fold, since it wraps the build in `SELECT "total", "amount" FROM (...) AS __public` with quoted author-cased names. It does not, and the reason is worth stating: a quoted *reference* is not an alias, so it does not rename the output column. In DuckDB:

```
SELECT "total" FROM (SELECT 1 AS TOTAL) AS x;          -- label comes back TOTAL
SELECT "total" AS "total" FROM (SELECT 1 AS TOTAL) AS x; -- label comes back total
```

So the public projection passes a folded label straight through, and the premise survives. Might be worth a line in the commit message, since "the projection already quotes the names" is the first objection a reader will raise.

Four things I'd fix before merge:

**1. A comment that this PR makes false.** `packages/server/src/materialization_metrics.ts:454` still ends with "otherwise the fallback side is only a DEBUG log" -- describing exactly the behavior being changed here.

**2. The JSDoc at `materialization_serve_transform.ts:524` was spliced mid-sentence.** The new text lands inside the pre-existing one:

```
 * rather than guessed. See the body for why each of those three is load-bearing. Dropped columns stay physically
 * in the table but become unreachable through the source: a query that
```

That line runs to ~113 columns against the file's 80-col prettier width, and "Dropped columns stay physically in the table" now reads as a continuation of the load-bearing sentence. Needs a paragraph break before "Dropped columns" and a re-wrap.

**3. The `info` volume claim does not hold for one class that is guaranteed to hit it.** The comment says volume is bounded like the hit line beside it, but the routing gate tests the un-narrowed set:

```ts
const storageRoutingPossible =
   getPersistStorageMode() === "on" &&
   this.serveBindings.length > 0 &&      // model.ts:4494 -- before narrowing
   !!this.serveDestinationConfig;
```

and bindings are pushed package-wide to every loaded model (`package.ts:1405-1409`). So a model that defines none of the bound sources narrows to `[]`, gets dropped by the `.filter(b => b.schema.length > 0)`, and then logs this line on **every query, permanently, while never logging a hit** -- so the claimed symmetry with the hit line does not hold there. That is the sibling-model case your own test at `model_storage_serve.spec.ts:145-163` covers, so it is routine rather than hypothetical, and for a permanently-ineligible package it repeats an identical multi-line compiler error per query with no dedup. Since `recordStorageServeRouting("live_fallback")` already carries the rate, logging once per binding-set (the memo at `model.ts:3869-3871` invalidates on exactly the right boundary) would keep the diagnosis and drop the repetition.

**4. The security test covers the safe direction only.** Every access-modifier fixture uses a hidden name that shares no fold with any visible field (`ssn` next to `id`, `secret` next to `total`), so those tests pass under any roughly-correct implementation -- including one without the ambiguity guard. The direction this change actually opens is a hidden field's name folding onto a *visible* one. Concretely, for a public raw `other`, a public `dimension: Total is other * 10`, and an access-restricted raw `total`, the restricted field is excluded from both lookups, which leaves exactly one folded candidate and emits `Total` bound to captured `TOTAL`:

```
narrow([{TOTAL},{OTHER}], [total(private), other, Total(dimension)])
  -> [{name: "Total", ...}, {name: "other", ...}]
```

Whether that is reachable depends entirely on whether an access-restricted atomic column can reach the built table -- i.e. whether `intrinsicFields` excludes it. `publicColumnNames` never consults `accessModifier`, and I could not find a test pinning the exclusion anywhere; the closest thing is the fixture comment at `build_plan.spec.ts:166-168` asserting it as an assumption. If that assumption holds the case is unreachable and this is just a missing test; if it does not, a restricted column's values serve under a public dimension's name with `servedFrom: storage` and no warning. Could you add the folded-hidden-name test either way, and if the exclusion really is load-bearing, pin it with an assertion rather than a fixture comment?

Two smaller notes, neither blocking:

- The captured-side dedupe is order-dependent in its *type*: `[AMOUNT:BIGINT, amount:VARCHAR]` yields `BIGINT` and the reverse yields `VARCHAR`. Unreachable through a real build (DuckDB renames the collision to `amount_1` on CTAS, and rejects it outright in explicit DDL), so this is dead defensive code -- but the new test pins the lucky ordering, and a comment saying the input is unproducible would stop a future reader from trusting the order.
- The live-Snowflake evidence in the description uses unquoted aliases (`SELECT 'alpha' AS who, 41 AS n`), which is the shape of the harness at `storage_build_bigquery.integration.spec.ts:60-88` -- it calls `buildSourceIntoStorage` directly and so bypasses `projectToPublicColumns`. The conclusion still stands per the quoted-reference behavior above, but the repro as written is not the production path's SQL shape, so it would be stronger re-run through the persist-source path.

PR #1077, opened 2026-08-26
fix(server): bound every DuckDB session's memory instead of one per container

housejester:
## The bug

DuckDB sizes `memory_limit` from the container — roughly 80% — and it does that **independently per instance**. Publisher runs several instances in one process: the environment lookup funnel, a per-package sandbox, and a disposable session for every materialization build. None of them accounts for the resident runtime baseline, and none of them knows the others exist.

Measured in a 3 GiB container:

```
cgroup memory.max = 3 GiB
  instance 0: memory_limit = 2.3 GiB
  instance 1: memory_limit = 2.3 GiB
  instance 2: memory_limit = 2.3 GiB
  => 6.9 GiB of committed budget against 3 GiB of real memory
```

The kernel kills the process while **every session still believes it is comfortably inside its budget**. That is what makes this hard to see from the inside: no single session looks at fault, the JS heap is flat, and the growth presents as untracked native memory. A per-session memory metric cannot show it either, because per-session is exactly the wrong denominator.

## The fix

Two settings, both **no-ops when unset**, so nothing changes on upgrade:

- **`PUBLISHER_DUCKDB_MEMORY_LIMIT`** — a flat `memory_limit` (`1GB`, `512MB`) applied to every Publisher-owned session, passed to DuckDB verbatim. `off` restores DuckDB's default.
- **`PUBLISHER_DUCKDB_TEMP_DIRECTORY`** — where sessions spill.

**Flat, not a fraction.** This is the part worth not re-deriving: any fraction is still computed *per instance*, so N instances still commit N times it. Only an absolute per-session value bounds the sum — and the operator is the only party that knows both the container size and how many sessions to budget for. Set it to roughly (container memory − resident baseline) ÷ concurrent sessions.

Verified against the case that exposed the bug:

| `PUBLISHER_DUCKDB_MEMORY_LIMIT` | per instance | combined vs 3 GiB real |
| --- | --- | --- |
| unset | 2.3 GiB × 3 | **7066 MiB** — 2.4× oversubscribed |
| `512MB` | 488 MiB × 3 | **1465 MiB** — under |

## Where it is applied, and why there

Ahead of the autoinstall guard inside `applyExtensionSessionSettings`, which is the one funnel every Publisher-owned session already reaches. That guard returns early in the common case, so the limits sit before it — the resource bound must not inherit the extension policy's conditions. Putting it there means coverage does not depend on remembering a call site.

Plus an explicit call on each build session, for two reasons: a build reaches that funnel **only** for a DuckLake destination (a plain-DuckDB destination attaches a file and never passes through it), and a build session is the one session that owns a disposable working directory — so its spill is pointed there instead of at the configured default, unique per build and removed with it. The GC/drop session is bounded too, even though a `DROP` allocates almost nothing: the cost being guarded is the *instance*, which claims its share the moment it exists, whatever it goes on to run.

Bounding only the session running the largest job would fix nothing. The sum is what overcommits, so a partial rollout of this is worth zero.

## On spill, which is narrower than it looks

Setting a `memory_limit` does **not** by itself introduce spill on the `storage=` build path. That pipeline pushes its SQL to the source warehouse and streams the result into the destination — no join, no sort, no aggregation, so there is no spilling operator. Measured on a 60M-row build into a DuckLake destination in a 3 GiB container:

| `memory_limit` | outcome | peak RSS | bytes spilled |
| --- | --- | --- | --- |
| default (2.3 GiB) | ok, 20s | 408 MB | **0** |
| 488 MiB | ok, 18s | 398 MB | **0** |
| 122 MiB | fails: `Out of Memory Error: could not allocate block of size 76.5 MiB` | 353 MB | **0** |

Zero bytes at every limit, including the one tight enough to fail. An over-tight limit fails the *query* and leaves the process up, which is the intended trade against losing the pod and every package served from it. Local compute — and therefore spill — is the chained-build and serve paths, which is why `PUBLISHER_DUCKDB_TEMP_DIRECTORY` exists but is not a prerequisite for the limit.

## Test

`bun test src` — **3180 pass / 3 skip / 0 fail** across 151 files, so the funnel change is exercised by every existing session-opening test.

Nine new cases in `session_resource_limits.spec.ts`, against a recording stub so they assert the SQL actually issued rather than the parse: unset issues **nothing**; a flat value is passed verbatim; `off` reads as unset; a configured temp directory; a session-owned directory taking precedence over it; `temp_directory` set **before** `memory_limit` (a limit low enough to force spill must never be in effect while the directory is still DuckDB's default); idempotence when the funnel is reached twice for one connection; quote escaping; and a rejected value propagating rather than silently opening an unbounded session.

`tsc --noEmit` clean; `eslint` clean on all touched files.

## Note for review

Both settings default to unset, so this ships as a no-op and the oversubscription persists until an operator sets a value. That is deliberate for a public default — a flat limit that suits an 8 GiB worker would badly constrain a single-session deployment on a large host. Happy to bake a default in if reviewers would rather the safe behaviour be the default, but it is a real behaviour change for existing deployments and I did not want to make that call silently.


girishjeswani:
The diagnosis here is the valuable part, and I think it's right: per-instance independence is genuinely invisible from the inside, and a per-session metric can't show it because per-session is the wrong denominator. The funnel placement is a good call too, ahead of the autoinstall early-return, and it does catch more than I expected. I checked every DuckDB creation site and the env lookup, the per-package sandbox, both attach paths, the Azure and DuckLake connections, and even `testConnectionConfig` all resolve through it. Failure-mode change also checks out end to end: an OOM propagates as a plain error into the existing path, `reclaimStorageTablesFromFailedRun` drops stranded tables, and nothing anywhere keys on DuckDB error text, so an over-tight limit can't be misclassified as retryable. I reproduced the over-tight case independently (100MB limit, clean `Out of Memory Error`, zero spill files) so the "spill is narrower than it looks" framing holds.

Three things before this lands. All small and local.

**1. The build session never gets its own `workDir` as `temp_directory`.** `applySessionResourceLimits` early-returns on the `WeakSet`, and in both build functions the explicit `{ tempDirectory: workDir }` call sits after `attachDestinationReadWrite`. For a DuckLake destination that attach already runs through `attachDuckLakeWithMode` to `applyExtensionSessionSettings`, which calls `applySessionResourceLimits` with no `tempDirectory` and marks the WeakSet, so the later call is a no-op:

```
Expected to contain: "SET temp_directory = '/tmp/malloy-build-xyz'"
Received: [ "SET temp_directory = '/var/spill'", "SET memory_limit = '512MB'" ]
```

`ducklake` is the only type `DECLARABLE_STORAGE_DESTINATION_TYPES` admits, and the comment at `:69-72` says the `duckdb` branch isn't reachable from configuration today, so the case the new comment says needs the explicit call is the unreachable one and the reachable one is where it silently does nothing. That makes the docs line about a build "always" spilling inside its own disposable working directory false as merged, plus the matching release note: concurrent builds share the configured global directory, and spill from a crashed build isn't cleaned up by `dispose()` because it isn't under `workDir`. Moving both calls to immediately after `createIsolatedBuildSession`, before the attach, fixes it, and is the ordering `dropStorageTable` already uses at `:1218`.

**2. The shared gate session at `:1161` is unbounded.** `sharedGateSession ??= createIsolatedBuildSession("gate_shared").session` is process-wide and deliberately never disposed, and it reaches `assertServesInDuckDB` through a `FixedConnectionMap` that builds a `Runtime` directly, so no `lookupConnection`, no attach, neither hook. By your own argument for bounding the GC session ("the cost being guarded is the instance, which claims its share the moment it exists") this is the longest-lived instance in the process and arguably the first thing that should be bounded. One line after the `??=`.

**3. The metadata store at `StorageManager.ts:73` is unbounded.** `DuckDBInstance.create(this.dbPath)` is a real instance held open for the server's lifetime, sizing from the container like the others. Different class from Malloy's `DuckDBConnection`, which is presumably why the typed helper missed it. Confirmed the default empirically: a bare `DuckDBInstance.create(":memory:")` on a 17GB box reports `memory_limit = 12.7 GiB`. Not idle either, the embedding index runs real compute on it via `get_context_tool.ts:578`. `DuckDBInstance.create` takes a config map so it's two lines, same for the boot mirror at `schema.ts:382`. If you'd rather scope this PR to Malloy-owned sessions that's reasonable, but then the title, the docs row ("every Publisher-owned DuckDB session") and the release note want to stop saying "every".

**On flat-not-a-fraction.** Agree with the conclusion, but the argument may be defeating an alternative that doesn't exist: `SET memory_limit = '80%'` is rejected outright by the bundled v1.5.5 with `Parser Error: Unknown unit for memory: '%'`. The alternative worth addressing is Publisher computing bytes from the cgroup and dividing, and what defeats that is that Publisher can't know N when a session opens, and re-limiting live instances would shrink a cap under a running query. That reasoning is solid, it just isn't quite what `config.ts:239-242` says.

Related, on sizing guidance: the denominator isn't a builds number. It's 1 metadata store + 1 gate + 1 env funnel + one sandbox per loaded package times `PACKAGE_LOAD_WORKERS` (package loads run in `worker_threads`, same address space and cgroup) + one per in-flight build, and nothing bounds concurrent builds. "Divide by sessions you expect concurrently" doesn't give an operator a way to count that, so naming the contributors in the docs row would help. Also worth knowing DuckDB doesn't reserve `memory_limit` up front, so "claims its share the moment it exists" at `:1214-1216` is conservative rather than literal, and it's the framing that makes idle instances look like they belong in the denominator.

**Two smaller ones.** The new getters landed between `getMemoryGovernorConfig`'s docstring and its body, so the "Throws at startup on malformed input" block now sits above `getDuckDBMemoryLimit` and `getMemoryGovernorConfig` at `:287` is undocumented. And on that subject: a malformed value here is silent until the first session and then throws on every DuckDB path, while `/health` and `/health/readiness` never touch DuckDB, so the pod reports ready and serves nothing. `getExtensionFetchPolicy` one screen below resolves the same tradeoff the other way, with a docstring explaining exactly this reasoning, and `server.ts:280` validates it at boot. Either call these two getters there as well, or a shape check that throws with the var name would do it.

Separately, `SET temp_directory = '/nope/does/not/exist'` succeeds silently on v1.5.5 and only fails at first spill with an `IO Error` that names the directory but not the variable that set it. A boot-time `mkdirSync(dir, { recursive: true })` when the var is set would turn that into a startup failure and cover unwritable too.

**On coverage claims and tests.** Deleting every call site (the funnel plus all three in `materialization_build_session.ts`) still leaves 78 pass / 0 fail across the three relevant specs. Same pre-existing pattern as `pinSessionToUTC` so not a regression, but the body specifically says "coverage does not depend on remembering a call site" and nothing holds that line. One test that opens a real `createIsolatedBuildSession` under a set env var and reads back `current_setting('memory_limit')` would pin what the PR is actually selling, and would have caught item 1.

**#1074 interaction:** two trivial conflicts, both at the `pinSessionToUTC` call sites, resolution is keep-both. Semantically they're complementary, #1074 removes the unbounded reorder buffer and this makes it survivable, so landing both is right. Suggest #1074 first, and fixing item 1 moves these calls out of the conflicting hunks anyway.

sagarswamirao:
Reviewed at head `40b014cb`. The diagnosis and the design both hold up; one bug silently defeats a stated guarantee on both production build paths.

The central claim -- that a fraction cannot fix a per-instance calculation, so only a flat absolute value bounds the sum -- is correct, and it is the non-obvious part. Placing the limits ahead of the autoinstall early-return in `applyExtensionSessionSettings` is the right call: all four production call sites (`connection.ts:711`, `:1336`, `:2060`, `package.ts:2891`) do funnel through it.

Test claim reproduces exactly: `bun test src` gives 3180 pass / 3 skip / 0 fail across 151 files.

## Blocking: the build session's `workDir` is discarded for a DuckLake destination

`applySessionResourceLimits` latches a `WeakSet` on first call and returns early on every later one. On both real build paths the explicit call carrying `tempDirectory: workDir` runs *after* `attachDestinationReadWrite`:

- `materialization_build_session.ts:815` then `:840` (`buildSourceIntoStorage`)
- `materialization_build_session.ts:1009` then `:1031` (`buildDownstreamIntoStorage`)

For a DuckLake destination that attach chains `attachDuckLakeReadWrite` -> `attachDuckLakeWithMode` -> `applyExtensionSessionSettings` -> `applySessionResourceLimits(session)` with no `tempDirectory`. The WeakSet is set, so the later call carrying `workDir` is a no-op.

Against the real module:

```
await applySessionResourceLimits(conn);                                          // the attach funnel
await applySessionResourceLimits(conn, {tempDirectory: "/tmp/malloy-build-abc"}); // the build site

SQL ISSUED: ["SET temp_directory = '/var/spill'", "SET memory_limit = '512MB'"]
```

The build's disposable directory never appears.

The effect is inverted from the intent: it fires precisely on the DuckLake path, which is the reason the comment gives for the explicit call existing ("a build reaches that funnel only for a DuckLake destination"). The comment correctly identifies that the funnel is reached there, then draws the opposite conclusion about the outcome.

So on a DuckLake build the spill goes to the shared configured directory -- or DuckDB's `.tmp`, the node's shared disk -- instead of the build's own; it outlives the build rather than being removed by `dispose`; and concurrent builds can collide. Those are the three properties this change is sold on. It also makes the docs claim at `configuration.md:115` untrue as written, since a build does not "always" spill inside its own working directory.

The correct pattern is already in the diff: `dropStorageTable` (`:1218`) applies the limits before the attach. Only the two paths that actually move data have it backwards.

Fix: move both explicit calls above `attachDestinationReadWrite`, matching `dropStorageTable`. That also puts the memory bound in effect during the attach itself, which is strictly better. Letting an explicit `tempDirectory` override a prior latch would also work, but the reorder is simpler and makes all three sites consistent.

The nine new cases all call `applySessionResourceLimits` directly on a fresh stub, so none covers funnel-then-build-site ordering. A case asserting that `workDir` wins after the funnel has already run would have caught this.

## Non-blocking

- `off` is asymmetric: `PUBLISHER_DUCKDB_MEMORY_LIMIT` honors it, `PUBLISHER_DUCKDB_TEMP_DIRECTORY` does not (`config.ts:283`). Defensible, since a directory literally named `off` is legal, but it is undocumented and mildly surprising.
- The rationale is valuable, but roughly 90 lines of prose for ~25 lines of logic, largely duplicated across `config.ts`, `connection.ts`, `configuration.md` and `RELEASE_NOTES.md`, is four copies to keep true. Consider making the doc canonical and pointing the code comments at it.

## On shipping as a no-op

Defaulting to unset is the right call for a public default and I would keep it. The tradeoff is that the oversubscription this diagnoses persists for every existing deployment until an operator reads the release notes. A startup warning when the process holds N DuckDB instances with no limit configured would surface the condition to operators who would otherwise never know, without changing behaviour for anyone.

Good change overall -- the analysis behind it is the hard part and it is sound. Worth fixing the ordering before merge.

housejester:
Both of you found the same blocking bug and you were right — thanks, that one silently inverted the whole point. Fixed in `fc95d96f`, and I took essentially everything. Confirmed the ordering before changing anything: `buildSourceIntoStorage` attach at `:815` then limits at `:840`, `buildDownstreamIntoStorage` `:1009` then `:1031`, while `dropStorageTable` at `:1218` had it right. The two paths that move data were the two that had it backwards.

**Reordered all three to apply before the attach**, matching `dropStorageTable`. But the reorder alone leaves the guarantee dependent on call order, which is girish's "nothing holds that line" point applied to the fix itself — so a session-owned `tempDirectory` now also **overrides** one already applied from the global default. That makes it a property rather than an ordering convention, and it is directly testable. Verified end to end on a real DuckLake build against a local Postgres catalog + MinIO, with the global var deliberately set and the attach funnel having already run:

```
REALBUILD workDir=/tmp/malloy-build-xI3WdS | session temp_directory=/tmp/malloy-build-xI3WdS | memory_limit=488.2 MiB
```

Reverting the override fails exactly the new test and nothing else.

**Bounded the three instances I'd missed.** The gate session — you're right that by my own argument it should have been first, being the longest-lived one here and reached through a `FixedConnectionMap` so neither hook sees it. And the metadata store plus the boot mirror at `schema.ts:382`, via `DuckDBInstance.create`'s config map; I'd rather widen the change than narrow the claim, since "every" is the property that matters. Your 12.7 GiB measurement on a 17GB box is exactly the shape of the problem.

**Boot validation and `mkdirSync`**, wired next to `getExtensionFetchPolicy` in `server.ts`. The `/health` observation is the part that made this worth doing rather than nice-to-have — a pod reporting ready while every DuckDB path throws is the worst version of this. Silent-until-first-spill on a nonexistent temp dir is covered by the same call.

**The flat-vs-fraction argument was defeating a non-thing** — good catch, and I'd have kept repeating it. Rewritten around the alternative that does exist: Publisher computing bytes from the container and dividing, defeated because the divisor is unknown when a session opens and revising it would shrink a live cap under a running query.

**Sizing guidance** now names the contributors rather than saying "sessions you expect concurrently": metadata store + gate + env funnel + one sandbox per loaded package × `PACKAGE_LOAD_WORKERS` (`worker_threads`, same address space and cgroup) + one per in-flight build, with nothing bounding concurrent builds. Also noted that DuckDB does not reserve up front, and softened the "claims its share the moment it exists" comment to say it carries a cap and the sum of caps is what must fit — you're right that the original framing is what makes idle instances look like they belong in the denominator at full weight.

**Trimmed the prose.** Four copies was fair; `docs/configuration.md` is canonical now and the code comments point at it. Documented the `off` asymmetry in the same row (a directory named `off` is a legal path, hence no sentinel there).

**Startup warning added** when no limit is configured, per sagar's suggestion — it surfaces the condition without changing behaviour, which is the right answer to shipping this as a no-op. Keeping the default unset.

**On the test that would have caught it:** added the real-session case, opening an actual `createIsolatedBuildSession` and reading back `current_setting`. It asserts `488.2 MiB` rather than `512MB` precisely because DuckDB's own rounding is the evidence the value was parsed and not merely issued — a stub cannot tell those apart, and that is what let this ship with nine green tests.

`bun test src`: 3183 pass / 3 skip / 0 fail. `tsc` and `eslint` clean.

`#1074` is closed — the measurement went against it, so the conflict is moot. `preserve_insertion_order=false` turned out not to reduce build memory on the direct passthrough path at all (it trended higher), and a parallel-parquet control showed no memory delta either, so there was nothing left holding it up.

girishjeswani:
Re-reviewed at `fc95d96f`. All three coverage gaps are genuinely closed, and the ordering fix is better than what I suggested: making the guard a `WeakMap` keyed on the applied directory means the override recovers even under the old sequence, so ordering stops being load-bearing. I checked that by issuing the real statements rather than reading it:

```
tip order (limits(workDir) then the attach funnel):
  [ "SET temp_directory = '/tmp/malloy-build-xyz'", "SET memory_limit = '512MB'" ]

old order (funnel first, then limits(workDir)):
  [ "SET temp_directory = '/var/spill'", "SET memory_limit = '512MB'", "SET temp_directory = '/tmp/malloy-build-xyz'" ]
```

Both end on `workDir`. The metadata store is right too: I confirmed the signature against the pinned `@duckdb/node-api@1.5.5-r.4` and that `create(":memory:")` and `create(":memory:", {})` both report 12.7 GiB, so the empty-options path really is the old behaviour. `DuckDBConnection.ts:42` is now the only `DuckDBInstance.create` in the repo, which picks up the `schema.ts:382` boot mirror for free. Boot validation plus the effective-values log is more than I asked for and lands in the right place, next to the extension-fetch line.

Two things I'd fix, then this is good to go.

**1. The gate `??=` is no longer race-free.** `materialization_build_session.ts` checks at `:1163`, awaits at `:1173`, assigns at `:1176`. Two concurrent `assertStorageServeShapeCompiles` calls can both pass the check, both `mkdtempSync` and construct a session, and one loses the assignment: an orphaned connection and temp dir that nothing disposes, for the process lifetime. The round-one form was fully synchronous and couldn't interleave, so this came in with the fix. Worth noting `applySessionResourceLimits` is `async` and awaited even when both vars are unset, so the window is open on every deployment rather than only opted-in ones. Assigning `sharedGateSession = gate.session` before the await closes it.

**2. `off` doesn't silence the new warning.** `getDuckDBMemoryLimit()` maps `off` to `undefined`, so `server.ts:291` takes the unset branch and an operator who deliberately opted out with the documented sentinel still gets the five-line OOM warning on every boot. The warning itself is good and correctly stays quiet when a real value is set, but as written every existing deployment gets it every start with no way to acknowledge it short of setting a value, which is the shape of warning people learn to filter. Treating an explicit `off` as acknowledgement would fix it.

**On the shape check**, since I raised bare bytes last time and was wrong: DuckDB rejects `1073741824`, `1024`, `0` and `100` with `Unknown unit for memory: ''`, so the validator rejecting them is correct, and `1.5GB` / `1gb` / `1 GB` / `1GiB` all pass both sides. The one real hole is `1GBB`, which the regex accepts and DuckDB rejects. It is partly self-mitigating now that the metadata store takes the value as instance config, since `DuckDBInstance.create` throws and `environment_store.ts:819-822` calls `markNotReady()`, so the pod at least doesn't report ready. The message just names neither the value nor the variable. `.5GB` and `-1` being rejected are cosmetic given `off` exists.

**One gap against the docstring.** `config.ts:330-336` says creating the directory turns a nonexistent path "and an unwritable path" into a startup failure. I probed the cases: path-exists-as-file gives `EEXIST`, nested-under-a-file gives `ENOTDIR`, unwritable *parent* throws, all good. But an **existing** directory the process can't write to returns silently from `mkdirSync(dir, {recursive: true})`, and that's the common k8s case where a volume is mounted with the wrong ownership. An `accessSync(dir, W_OK)` would deliver what the docstring already promises.

**Two smaller ones.** The gate's spill goes to its own `mkdtemp` under `os.tmpdir()` rather than `PUBLISHER_DUCKDB_TEMP_DIRECTORY`, and is never removed. Low volume for gate compiles, but an operator who sets that variable specifically to keep DuckDB off the node's shared disk will still get spill there, and the temp-dir row's "every other session and instance uses this" doesn't hold for it. Also `duckDBInstanceResourceOptions` and its docstring landed between the `DuckDBConnection` class docstring and the class, so the class is undocumented and a 15-line rationale hangs off a private helper. Same shape as the `config.ts` one last round, one file over. Might be worth a glance at where new helpers land in these files.

**On the coverage claim.** The real-session test is a genuine improvement and catches something the SQL assertions can't, a value issued but not parsed. But it calls `applySessionResourceLimits` directly, so it doesn't hold the wiring: I deleted the funnel call, all four `materialization_build_session.ts` calls, and the metadata-store options, each independently, and got 125 pass / 0 fail every time. Reverting the override branch does fail exactly one test, so that half of your claim is accurate. I'd keep this non-blocking as before, but "hold the guarantee" in the commit title is stronger than what a test currently backs, so either soften it or add one wiring test.

Docs read accurately now. The "always spills inside its own disposable working directory" line is true as implemented, the enumeration picked up the gate and the metadata store, and the sizing guidance names the real contributors including the `PACKAGE_LOAD_WORKERS` multiplier and that nothing bounds concurrent builds. The reworked flat-not-a-fraction argument is the right one.

Also noting for anyone following the thread that #1074 is closed, so the conflict I flagged is moot. For the record the prediction held: `materialization_build_session.ts` auto-merges cleanly now that the limits calls moved above the attach, and the only collision left was an adjacent-insertion one in `config.ts`.

housejester:
Round two addressed in `e9ad86fa`. The gate race is mine and it's a good catch — I introduced it turning a synchronous `??=` into a check-await-assign, and you're right that because `applySessionResourceLimits` is `async` even when nothing is configured, the window was open on every deployment rather than only opted-in ones.

Rather than assign before the await, I made the assignment synchronous again and moved the bound to per-use:

```ts
sharedGateSession ??= createIsolatedBuildSession("gate_shared").session;
await applySessionResourceLimits(sharedGateSession);
```

The WeakMap makes that one lookup after the first compile, and there's no window at all. It also resolves your spill-location point in the same change: **the gate no longer takes a directory of its own.** It outlives every build, so a per-session directory would accumulate with nothing to remove it — and an operator who set `PUBLISHER_DUCKDB_TEMP_DIRECTORY` specifically to keep spill off the node's disk should get it there. So the temp-dir row's "every other session and instance uses this" is now true as written.

**`off` silences the warning.** You're right that a warning with no way to acknowledge it is the shape people learn to filter. Added `isDuckDBMemoryLimitDisabled()` so an explicit `off` takes the quiet branch and logs `off (explicitly disabled)` instead. Unset still warns.

**Shape check anchored to DuckDB's real units**, closing `1GBB`. Thanks for probing the boundary both ways — `1073741824`, `1024`, `0`, `100` rejected by DuckDB and now by the validator; `1.5GB` / `1gb` / `1 GB` / `1GiB` accepted by both.

**The `accessSync` gap was real and my own docstring promised it.** `mkdirSync(…, { recursive: true })` returning silently for an existing directory whatever its mode is exactly the k8s wrong-ownership case. Added the `W_OK` check, with a test that `chmod 0o500`s a real directory.

**Helper placement — same mistake, one file over.** Moved `duckDBInstanceResourceOptions` above the class docstring so `DuckDBConnection` keeps its own. Twice in two rounds is a pattern rather than an accident; I'll insert above a preceding docstring rather than adjacent to a declaration from here on.

## On the coverage claim

You've raised this twice and both times you were right, so I closed it rather than softening the wording. Two wiring tests, against real DuckDB instances reading their own settings back:

```
delete the funnel call        -> (fail) is reached through the extension-settings funnel        19 pass / 1 fail
delete the metadata options   -> (fail) is reached by the metadata store, ...                   19 pass / 1 fail
restored                                                                                        20 pass / 0 fail
```

Being straight about what is still uncovered: **the three build-session call sites have no test.** They need a live destination, so they're covered by review plus the manual DuckLake run from last round, not by CI. I've said so in the commit body rather than implying otherwise, and dropped "hold the guarantee" from the framing — you were right that it outran what a test backed.

`bun test src`: 3191 pass / 3 skip / 0 fail. `tsc` and `eslint` clean.

Also confirming your `#1074` note: closed, and the merge prediction held.

girishjeswani (approved):
Approving at `e9ad86fa`. Everything from the last two rounds is addressed, and the gate fix is better than what I suggested: keeping `??=` synchronous and applying the limit per use instead is the same guarantee with no race, and it costs one WeakMap hit after the first compile. The comment explaining why that line has to stay synchronous, including that the window was open on every deployment because the call is `async` even when nothing is configured, is the kind of note that stops someone reintroducing it. Pointing the gate at the configured directory rather than giving it one of its own is also the right resolution, and not the one I proposed.

Verified the rest directly rather than reading it: the regex now rejects `1GBB` and still accepts every spelling DuckDB takes, `accessSync(W_OK)` catches the pre-existing unwritable mount that `mkdirSync` returned silently on, `off` suppresses the boot warning and reports itself in the info line, and the class docstring in `DuckDBConnection.ts` is attached to the class again.

One honest note on the wiring, since two commit titles have now claimed it. I mutation-tested all six call sites against the spec:

| deleted | result |
| --- | --- |
| funnel, `connection.ts:213` | 19 pass / 1 fail |
| metadata store options | 19 pass / 1 fail |
| build site `:823` | 20 pass / 0 fail |
| downstream site `:1018` | 20 pass / 0 fail |
| gate site `:1183` | 20 pass / 0 fail |
| GC site `:1241` | 20 pass / 0 fail |

The two new tests do hold the two sites that were hardest to reason about, the indirect funnel and the different-class metadata store, so this is real progress rather than a token test. But the four `materialization_build_session.ts` sites are still deletable with nothing failing, and `session_resource_limits.spec.ts` is still the only spec that references any of these symbols. Not blocking, and I wouldn't hold the PR for it, but "hold the guarantee" and "test the wiring" are both stronger than what a test currently backs, so worth either softening the claim or picking up the build sites in a follow-up.

15/15 CI green, DCO on all three commits, no `Co-Authored-By`. Needs a rebase for `BEHIND` before it goes in.

PR #1078, opened 2026-08-26
docs: retire the dormant next pre-release channel

mlennie:
## What this is

James asked why `@malloy-publisher/server`'s npm `next` tag points at something old. It turns out the tag was never pinned and nobody configured it specially: `npm-sdk.yml` derives the dist-tag from the version string alone, so any hyphenated version publishes under `next` and everything else under `latest`. `next` last moved on 2026-05-21 when `0.0.198-dev6` was published. Fifty-one stable versions have shipped since then and not one prerelease, so the pointer simply stayed where the last prerelease left it. Docker Hub's `:next` is frozen at the same date and resolves to the same image by digest.

That made our own documented tag scheme wrong in a way that mattered. `docs/deployment.md` told readers that `*-dev` tags were frozen and that `:next` was the current pre-release channel, when `:next` resolved to one of those very `*-dev` builds. Anyone who followed the docs to `:next` got a three-month-old image and had no way to tell from the tag name.

## What changes

`docs/deployment.md` drops `:next` from the tag list and replaces the paragraph that misdescribed it. The two sibling references that repeat the same inline list, `docker-compose.example.yml` and `packages/server/README.docker.md`, are updated to match so the three files agree. `.github/workflows/CONTEXT.md` records the decision, the constraint behind it, and what a future prerelease will do.

Docs only. No TypeScript, no workflow logic.

## Why retire rather than automate

Wiring this into the workflows is the obvious idea and it is not available in its obvious form. npm OIDC trusted publishing authenticates `npm publish` and `npm stage publish` only; `npm dist-tag add` and `npm dist-tag rm` need a granular token, and this repo deliberately has no `NPM_TOKEN` (`.github/workflows/CONTEXT.md` says so and says it should not be added back). So CI can move `next` only by publishing a real prerelease under it. Re-pointing the tag at an existing stable release would mean reintroducing a token, which is a poor trade for a channel nothing reads.

Removal is self-healing. The routing in `npm-sdk.yml` and `docker-image.yml` is untouched, so the first genuine prerelease recreates both tags with the correct meaning and no change here.

## Please read before merging

The docs describe the intended end state and say the tag "has been retired". At the time of writing the removals have not run yet: all three packages still show `next: 0.0.198-dev6` and Docker Hub still lists `:next`. The two registry actions cannot run from CI and are being done by hand. Either merge after they land, or accept a short window where the docs lead the registries.

## How it was verified

The diff is four files with zero TypeScript, so the typecheck, lint, and test legs cannot be affected by it and were not run. Saying that explicitly rather than implying a full gate. What was run: the enumeration was checked with `git grep ':next'` so the fix is not partial, and the only surviving hits are my own prose and the two intentional lines in `docker-image.yml` that recreate the tag on a prerelease; the relative link from `packages/server/README.docker.md` still resolves; `docs/deployment.md` is now prettier-conformant and the other three files were already non-conformant on `main` and sit outside CI's prettier glob, which passes clean; the branch is zero behind `origin/main` and merges clean by exit code.

Every number quoted above was measured against the live registries immediately before committing, not carried over from earlier in the investigation.

## Follow-up

Nothing blocking. If anyone does want a live pre-release channel, the shape that works with OIDC is publishing a prerelease from `main` on merge or nightly under `--tag next`. That is a real publish per merge on a package already at 192 versions, so it is worth doing only if someone actually wants to consume it.

### DIFF
--- a/.github/workflows/CONTEXT.md
+++ b/.github/workflows/CONTEXT.md
@@ -129,6 +129,31 @@ Prereleases keep the hyphen convention: `npm-sdk.yml` routes a hyphenated versio
 to the `next` dist-tag and `docker-image.yml` withholds `latest`, so a prerelease
 never takes over either.
 
+**The `next` dist-tag was retired on 2026-08-26, on npm and on Docker Hub.** It
+had pointed at `0.0.198-dev6` since 2026-05-21 while 51 stable versions shipped
+past it, because a dist-tag only moves when a publish uses it and no prerelease
+had been cut since. So `@malloy-publisher/server@next` silently installed a
+three-month-old build, and `ms2data/malloy-publisher:next` served the matching
+image. Nothing in this repo, in the Credible service, or in the published docs
+installed from either. Three consequences, and the first one decides what
+automation is possible here at all:
+
+- **npm OIDC cannot move a dist-tag.** Trusted publishing authenticates
+  `npm publish` and `npm stage publish` only; `npm dist-tag add|rm` needs a
+  granular token, which this repo deliberately does not have (see the OIDC note
+  below). So CI can move `next` only by publishing a real prerelease under it.
+  Re-pointing `next` at an existing stable release is not available without
+  reintroducing `NPM_TOKEN`, which is not a trade worth making for a tag nothing
+  reads. The same follows for any other dist-tag surgery: it is a human org
+  member acting by hand, never a workflow.
+- **The routing was left in place, so this is self-healing.** `npm-sdk.yml`
+  still sends any hyphenated version to `next`, and `docker-image.yml` still
+  tags `:next` on a prerelease. The first prerelease after this recreates both
+  tags, correctly, with no change here.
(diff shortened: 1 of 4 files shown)


mlennie:
Update 2026-08-27: the npm half of the merge gate in the description has cleared. The `next` dist-tag is removed from `@malloy-publisher/server`, `@malloy-publisher/sdk` and `@malloy-publisher/app`; all three now return `latest` only. Nothing was unpublished, `0.0.198-dev6` is still installable by exact version on all three, and `latest` is untouched at 0.0.250.

Still outstanding: the `:next` tag on Docker Hub, which is the one the changed paragraph in `docs/deployment.md` is actually about. Please hold the merge until that is deleted, otherwise this page ships a claim that is not yet true. I will comment again when it lands.

Leaving the description as written, since it was accurate when the PR opened.

Sha-Bang (approved):

sagarswamirao (approved):

PR #1079, opened 2026-08-26
skills: add malloy-model-as-you-go

jswir:
## What this does

Adds a skill for the step after a data question is answered: writing down what the answer assumed, in the most durable place the session can write. Today an agent either builds a full model before answering anything, or answers ad hoc and formalizes later if someone remembers to. Both lose the reasoning behind each definition, because nobody wrote it down when the decision was actually made.

`malloy-analysis` answers the question. This skill runs right after, while the query is still in front of the agent, and decides what to codify and where.

## Where the assumption lands depends on what the session can write

The same judgment call ("returns don't count as revenue") lands differently depending on the tools available, and the skill says so explicitly instead of assuming everyone can edit the model:

| You can | Codify as |
|---|---|
| Edit the model files (local package, draft package) | a `measure:` / `dimension:` / `view:` with a `#(doc)` in the `.malloy`, after confirming binding decisions |
| Author notebooks or reports, not the model (a viewer) | `source: orders_q is orders extend { ... }` with the `#(doc)` in a notebook cell, plus a markdown cell stating the assumption |
| Only chat (claude.ai, Cowork) | the assumption stated in the answer, plus the Malloy snippet a modeler could paste |

Pick the highest rung the tools allow; never skip codifying because the model is read-only.

## What reviewers should look closely at

The skill asks the user to confirm a judgment call **before** it goes into the model, not after. While a decision lives in one query it's the analyst's own call. The moment it becomes a `measure:`, it's the definition everyone downstream inherits, and they never see the reasoning. So the skill stops and asks first, with both numbers shown, instead of picking one and mentioning it in the write-up. It only asks when the alternative would change the number, when the call is about to become permanent, or when a different analyst could reasonably pick the other option. Everything else gets a one-line note.

## Routing

`malloy-modeling` and the `malloy` index now send a first-message data question (and "build a model but not sure what metrics") to this skill. `malloy-analyze` keeps open-ended exploration with no intent to keep anything. Retiring `malloy-analyze` and folding its EDA loop into `malloy-analysis` is a separate change (proposed in ms2data/service#6177); this PR goes ahead of it.

## Shared-skill compliance

Shared skill, so it lives here first and gets vendored downstream. No `credible-` prefix; no platform-specific tools or answers (no `execute_query_draft`, no hosted draft/publish flow, no `#(index)`/`#(agent-hidden)`). All `skill:` references point at shared skills: `malloy-analysis`, `malloy-analysis-pitfalls`, `malloy-charts`, `malloy-dashboards`, `malloy-notebooks`, `malloy-model`, `malloy-lookml-review`, `malloy-analyze`. Tools are named bare, with the host note.

## Packaging

- `.claude/skills/malloy-model-as-you-go` symlink so Claude Code discovers it.
- `skills_bundle.json` regenerated: 57 entries, 32 skills plus 25 reference files.
- `packages/skills` bumped to `0.1.10`. `0.1.9` is already published on npm, so leaving the version alone would make a release skip this package and stay green while shipping none of these skills. The release guard catches that.

## Testing

`bun test src/mcp/skills` in `packages/server` (18 pass: bundle in sync, no em dashes) and `bun run --cwd packages/skills test` (15 pass) on this branch.

The earlier shape of this skill was run end to end against a real BigQuery dataset: three questions produced a five-field model, re-verified by running the codified measures back through a served package and matching the original numbers to the cent. **Not yet exercised on a fresh dataset:** the confirm-before-codify checkpoints, and the notebook and chat-only rungs of the ladder. Those were written from the four user situations (local developer, Credible viewer, Credible modeler, chat-only), not from a run. Worth a reviewer's eye.

## History

Earlier commits on this branch carry the previous name (`malloy-reproducible-analysis`) and shape; the last commit does the rename and the reshape.


kylenesbit (approved):
Approving — the skill is strong and the mechanics check out. Three shared-skill governance items to address, the first before merge if possible.

## What I verified

- The writable-surface ladder (model file → notebook `extend` → stated assumption + snippet) keyed off tools present, not guessed roles, is the right design, and confirm-before-codify with both numbers attached is the right control — "reporting it afterwards is not confirming it" is exactly the distinction.
- Symlink target correct; `skills_bundle.json` regenerates byte-identically from this tree (57 entries, new skill present); DCO signed; all checks green; bare tool names with the host note; and the explicit departures from `skill:malloy-model` are declared in **both** skills, which is how to disagree with a sibling skill without creating a contradiction.

## Items

1. **`skill:malloy-dashboards` breaks the shared-skill reference rule.** `skills/README.md` on this branch says a shared skill may only `skill:`-reference other shared skills, and classifies `malloy-dashboards` as Publisher-specific ("dashboards are a Publisher surface"). The new skill references it in "Save the view when a question is worth re-asking" (the PR body lists it among shared references, which contradicts the README). The README's own escape hatch fits: neutral prose — "views wanted side by side belong in a notebook, or in a dashboard surface if the host has one" — keeping only `skill:malloy-notebooks`. One-line fix; would be good before merge.

2. **`malloy-model` is a shared skill edited downstream-first.** The README rule is upstream-first (`ms2data/agent-skills`), then copy, or the next sync conflicts. If the agreed direction for the *new* skill is publisher-first, fine — but the hunk in the existing shared `malloy-model` is squarely what the rule exists for. Please mirror it (and the new skill, if it is to be shared) upstream at or before merge, and note it here. The new skill also needs adding to the README's shared-skills list.

3. **The narrowing doesn't reach where hosts actually route from.** `malloy-analyze` is narrowed to "open-ended exploration" in the two routers, but its own front-matter `description` still says "Explore data for insights and build views/dashboards/notebooks… or any analysis task" — and that description is what Claude Code matches for auto-invocation. A first-message data question will keep landing on `malloy-analyze` directly, bypassing the routers. Either narrow that description too (again upstream-first), or state in the README/PR that description-routing hosts won't see the new behavior until the `malloy-analyze` retirement lands.

## Notes, not blockers

- Branch is behind `main`; needs a branch update before merge (protection is strict; no conflicts).
- `skills-npm.yml`'s PR job packs and tests but doesn't compare versions against npm — moot here since `0.1.9` > published `0.1.7`, so no bump needed.
- On the honest flag that the confirm checkpoints and the notebook/chat rungs weren't exercised on a fresh dataset: given they're the heart of the skill, one real run on the middle rung would be worth doing before this ships broadly — not holding the PR for it.

jswir:
Thanks — all three addressed, plus the branch update.

**1. `skill:malloy-dashboards` reference.** Fixed with the README's own escape hatch: neutral prose for the dashboard surface, keeping only `skill:malloy-notebooks`.

> views wanted side by side belong in a notebook (`skill:malloy-notebooks`), or in a dashboard surface if your host has one

You were right that the PR body contradicted the README; the body was wrong, not the classification.

**2. Direction of the `malloy-model` edit.** This is publisher-first on purpose, and I should have said so here rather than leaving it to be inferred. The proposal in ms2data/service#6177 reverses the upstream-first rule: this repo becomes the source of truth for the shared set and `agent-skills` vendors from a pinned Publisher tag with a CI drift check, so the hand-carried `cp` goes away. malloydata/publisher#1088 lands the first half of that.

I've added `malloy-model-as-you-go` to the README's shared-skills list and recorded the reversal next to the upstream-first rule, including that this skill and the `malloy-model` hunk need mirroring by hand until the vendoring script exists. Happy to reorder if you'd rather this waited behind #1088.

**3. The narrowing not reaching description-routing hosts.** This was the most useful catch — the routers were doing nothing for auto-invocation. `malloy-analyze`'s own description now scopes it and names where a specific question goes instead:

> Open-ended exploration with no specific question to answer, and building views/dashboards/notebooks on an existing model. … Not for a specific data question - answering one is `malloy-analysis`, and writing down what the answer assumed is `malloy-model-as-you-go`.

**Notes.** Rebased onto current `main` (was 4 behind, no conflicts). Bundle regenerated: 57 entries, 32 skills. `packages/server` MCP-skills specs 18 pass; `packages/skills` 15 pass.

On the unexercised middle rung: agreed, and it stays flagged in the PR body. A real viewer-session run is the check I'd want before this ships broadly, and it's the one claim in the skill I haven't been able to back with evidence.

jswir:
Folded #1081 into this PR (its commit, cherry-picked unchanged) and re-signed the history.

**Why here:** #1081 was unreviewed and is 14 lines of the same kind of content this PR carries — pointing consumer skills at `given:` instead of the deprecated `#(filter)`, across `malloy-analysis-report`, `malloy-analyze`, `malloy-define` and `malloy-document`. It was also the thing blocking the next skills PR, which deletes `malloy-analyze`; folding it here dissolves that.

@kylenesbit this is new content since your approval, so it wants a fresh look — sorry for the moving target. It's one commit (`87597a24`) and independent of the three items you raised.

**Also fixed:** DCO went red after the rebase. The original remediation commit named a SHA that the rebase had rewritten, so it no longer covered anything. Signed every commit properly and dropped the remediation commit instead of stacking a second one; the tree is byte-identical, only the history changed.

Bundle regenerated (57 entries, 32 skills). `packages/server` MCP-skills specs 18 pass, `packages/skills` 15 pass.

PR #1080, opened 2026-08-26
feat(dashboards): one form, and per-tile layout that makes it sufficient

Sha-Bang:
Closes the syntax half of #1069. Supersedes #1073, which fixed the narrow grid-width bug and left the two forms in place.

There were two ways to build a dashboard, with two spellings between them. Now there's one.

## What this unlocks

**One page over different sources.** This is the headline, and it was the actual complaint behind
#1069. A dashboard can now show tiles that come from unrelated sources, because each tile runs as its
own query. Verified on a running server, four tiles over four separate DuckDB tables, all four
rendering together at `colspan=3`:

```malloy
## artifact { title="Company" tiles=["catalog -> size", "people -> size", "coverage -> size", "sales -> revenue"] } dashboard { columns=12 }
import { order_items, products, customers, regions } from '../storefront.malloy'

source: catalog  is products  extend { # colspan=3
                                       # label="Products"
                                       # big_value
                                       view: size is { aggregate: product_count } }
source: people   is customers extend { … }   // 1,000
source: coverage is regions   extend { … }   // 50
source: sales    is order_items extend { … } // $2,098,177.97
```

**A single query can never do this**, and that is a language fact rather than a preference: a nest's
pipeline starts from its own query's source, and there is no union construct. Measured against the
lockfile-pinned 0.0.432, with a same-source nest as a control so an API mistake could not read as a
language answer, a bare name, a path, an arrow and a full pipeline all fail undefined, and even a
`join_one` gives `Cannot use view from join`. So the form that spans sources is the form that has to
win.

**And the same source now presents consistently.** The second half of #1069 was that one view looked
different depending on which form consumed it. Before this, `tiles=` had no per-tile layout at all,
so a view that laid out correctly as a `# dashboard` nest lost its width, its break and its heading
the moment it became a tile. Now `# colspan`, `# break`, `# label`, `# subtitle` and `# borderless`
are read off the view either way. **One view, one appearance, whichever form names it** — which is
also what makes the consolidation honest rather than a downgrade: before the layout work, the
flagship storefront page simply was not expressible as tiles.

## The syntax, before and after

One running example throughout: the storefront overview page — a KPI strip, a monthly trend beside a
state map, then category and top-product charts. The repo carries it in both forms, so every block
below is real code (trimmed for length), not an invented grammar.

**Before — the query form.** One Malloy query; `@malloydata/render` lays out its nests. This is
`business_overview` in `examples/storefront/storefront.malloy`, unchanged by this PR:

```malloy
source: order_items is duckdb.table('data/order_items.parquet') extend {
  # dashboard
  view: business_overview is {
    nest:
      # big_value
      kpis is { aggregate: total_sales, total_margin, order_count }
    nest:
      sales_by_month
      sales_by_state
      by_category
      top_products
  }
}
```

This form could lay out: `# dashboard { columns=N }` on the view sets the grid, and `# colspan` /
`# break` / `# label` on a nest place it (this page takes the renderer's defaults). What it could not
do is span sources: every nest's pipeline starts from this one query's source, so the whole page is
`order_items` or nothing — the language fact the section above establishes.

**Before — the tiles form.** `## artifact { tiles=[…] }` at model level. Each string in `tiles=` is a
Malloy run expression, `source -> view` — "run the view `sales_by_month` on the source
`order_items`" — and each runs as its own query, which is exactly why this form *can* span sources.
The same page as tiles, before this PR:

```malloy
## artifact { title="Storefront overview" tiles=["order_items -> sales_by_month", "order_items -> sales_by_state", "order_items -> by_category", "order_items -> top_products"] dashboard_columns=2 }
```

Its genuine gap was layout: the grid width (spelled `dashboard_columns=N`, inside the braces) was the
only layout property. Every tile rendered at the same width, headed by its view's name. A page of
uniform cards was fine; this page was not expressible — no full-width KPI strip, no trend at 8 beside
a map at 4. So each form did half the job, which is why both survived: the query form laid out but
could not span sources, the tiles form spanned sources but could not lay out.

**After — one form: tiles, plus the query form's layout vocabulary.** This is the shipped
`examples/storefront/dashboards/overview.malloy`, trimmed (the full file has five tiles, and imports
that wire up the page's filter controls; `scoped_orders` is the model's orders source behind those
controls). The layout tags sit on the views, one tag per line, each applying to the view below it —
the `extend` exists to carry them, while the views themselves stay the model's reusable ones:

```malloy
## artifact { title="Storefront overview" tiles=["overview -> kpis", "overview -> revenue_trend", "overview -> revenue_by_state"] } dashboard { columns=12 }

source: overview is scoped_orders extend {
  # colspan=12
  # label="Key figures"
  # big_value
  view: kpis is { aggregate: total_sales, total_margin, order_count }

  # colspan=8
  # break
  # label="Revenue by month"
  view: revenue_trend is sales_by_month

  # colspan=4
  # label="Revenue by state"
  view: revenue_by_state is sales_by_state
}
```

The tiles here all read one source because this page only needs one; the cross-source shape is the
one at the top of this description.

### Every spelling that changed

| | before | after |
|---|---|---|
| Grid width, tiles form | `dashboard_columns=N`, inside the artifact braces | `# dashboard { columns=N }`, beside the artifact tag — already accepted as a fallback, now the only spelling |
| Grid width, query form | `# dashboard { columns=N }` | unchanged — the tiles form now matches it |
| Tile width | none; every tile the same width | `# colspan=N` on the view |
| Tile row break | none | `# break` on the view |
| Tile heading | the view's name | `# label="…"` on the view |
| Tile subtitle / chrome | none | `# subtitle="…"`, `# borderless` on the view |

The tags go on the view rather than inside the `tiles=[…]` string deliberately: a per-tile syntax
inside the string would have been a third grammar to learn, and `colspan`, `break`, `subtitle` and
`borderless` are `@malloydata/render`'s own per-child dashboard tags, read with the same validation —
which is what makes one view lay out identically whether a tile names it or a `# dashboard` nest
does, the consistency half of #1069.

`# artifact` on a `query:` still works and is still served. It's reframed for what it actually is — a
rendered Malloy query, the same thing a notebook cell shows — rather than a co-equal dashboard form.

## What else is in here

- **A lint that enumerates the artifact tag** against a per-form allowlist, so a property the reader doesn't read gets named instead of silently doing nothing. `dashboard_columns` is its main customer; it also catches a misspelling, a `tiles=` on a query-level tag, and `dashboard` nested *inside* the artifact braces rather than beside them — the likeliest new mistake, since the two are siblings on one line. Plus a `# colspan` on a dashboard that declares no width at all, and a model-level `## artifact` with no `tiles=` that a query-level `# artifact` in the same file silently rescues, which is what half a migration looks like.
- **Those tags need the render-log filter.** Not defensive — measured. The renderer resolves them only for the children of a `# dashboard` nest, and every tile is a standalone query, so without this each tile of every dashboard answered `Unknown render tag 'colspan'`. Filtered only on the **top-level result**, which is the shape a tile has, so a `# colspan` on an ordinary nest is still reported and the renderer's own `Invalid`/`Ignored` findings are untouched.
- **Tile chrome catches up.** The tiles card is now the card every dashboard draws, so it takes the instance theme's `tileTitle`, font and border instead of MUI's `text.secondary`, app font and `divider`, and gets the renderer's 20px body padding. Radius comes from the host's `shape.borderRadius` on both cards, pushed into the renderer through its own `--malloy-theme--dashboard-*` at raised specificity rather than over the top — no `!important`. That incidentally fixes a real bug: the existing `!important` on `.dashboard-item` was repainting `.dashboard-item-borderless`, drawing a card the author asked not to have.
- **First dashboard in the bundled examples.** `examples/storefront/dashboards/overview.malloy` is the same figures as the model's `business_overview` view, as tiles.

## Compatibility, and where Publisher diverges

This repo previously described the format as byte-compatible with Malloyyo (`docs/dashboards.md` on `main`: "kept byte-compatible on purpose… works unchanged in either"). That is no longer exactly true, in two ways, and the places that claimed it now say so rather than papering over it.

**Subtractive: Publisher no longer reads `dashboard_columns`.** The grid width is `# dashboard { columns=N }`, which the renderer already reads on a tagged `query:` — and which Publisher already accepted as a fallback before this PR — so one spelling now covers both forms. A Malloyyo model still spelling the old name is reported as a property Publisher does not read, rather than quietly laying out at the default width.

**Additive: Publisher reads per-tile layout off the view a tile names.** As documented, Malloyyo's composite has no per-tile layout, so a dashboard authored here with `# colspan` / `# break` / `# subtitle` / `# borderless` on its tile views renders in Malloyyo as it does today, without those widths. This is the larger of the two differences and the one worth a maintainer's attention.

**Two caveats a reviewer should weigh.** Both characterizations of Malloyyo's grammar come from `docs/malloyyo-dashboards-design.md` in this repo, not from `@malloyyo/mcp-engine`, which is a private package I cannot read — so if that grammar has moved, this section is stale. And [malloyyo#160](https://github.com/malloydata/malloyyo/issues/160) raised the grid-width change with those maintainers and currently has **no response**, so nothing here is agreed, only proposed. Neither is a fait accompli: if the shared grammar keeps `dashboard_columns`, Publisher re-adds the reader, and the enumeration lint is what makes either direction visible to an author.

`DashboardManifest.dashboardColumns` narrows to `# dashboard { columns=N }` only. `DashboardTile` gains `label`, `subtitle`, `colspan`, `break` and `borderless`, each spelled exactly as its tag. An earlier revision of this PR called the field `rowBreak`, on the assumption that a generated Python client could not carry a reserved word; that was wrong, and measured — `openapi-python-client` escapes the attribute to `break_` and keeps the wire key `break`, round-tripped both ways. So the five per-tile fields now match the five tags with no exception. Downstream artifacts spelling `dashboard_columns=2` (the viewer's default) don't visibly move, but their lint output changes. `RELEASE_NOTES.md` has the entry.

### Tested

- `dashboard.spec.ts` — 105 pass, 17 new: per-tile layout off both a view and a model-level query, a dropped colspan, the clamp warning, the enumeration lint's cases, the two render-log filter behaviours, the grid-width-written-inside-the-braces message, and the tiles-less model tag that a query-level tag rescues.
- Cross-platform CI runs the full suite green on ubuntu, macos and windows; `tsc --noEmit` and `bunx prettier --check './packages/**/*.{ts,tsx}'` are clean on every tracked file.
- `dashboards.integration.spec.ts` — 47/47, two new. The layout assertion is end to end on purpose: the reader walks a compiled `ModelDef`, and the unit tests hand it annotations directly, so only a real compile shows the tags land where it looks.
- **Playwright, 31/31**, two new. One is the claim the whole change rests on: `tiled` is `grid` re-authored as tiles, and the two pages have to come out with the same grid shape. Compared as proportions of each page's own grid, since the renderer draws inside its own padded container and Publisher draws straight into the page. The other pins the tile heading's computed font and colour against an injected instance theme — the hole #1072 called out in its own comment and couldn't close.
- Full browser suite: 110 pass, 2 fail. Both are `environment-connections` and **pre-existing** — reproduced identically on a clean `origin/main` worktree with its own `bun install` and build.
- **Opened in a browser, not only asserted.** The shipped dashboard renders a full-width KPI card row, a trend at 8 beside a map at 4, then a chart and a table at 6 each, with no failed request. That caught a real defect in it that no lint can: an unimported source left the Brand picker answering `400 Undefined source 'order_items'` with no options — the suggest query resolves *by name*, so it passes the lint and only fails when run.
- Against a running server: the package reloads with zero warnings, the manifest carries every tile's label/colspan/break and `dashboardColumns: 12`, and every tile runs with `renderLogs: null`. The filter entry was verified load-bearing by removing it and restarting — both tags come back as unknown-tag warnings.
- `npm run lint`, `npm run typecheck`, `npm run test:skills` (15/15), `bun test scripts/release-notes.spec.ts` (23/23) clean. Skills bundle regenerated and confirmed idempotent; `packages/skills/package.json` bumped 0.1.9 → 0.1.10.


jswir:
Read through this closely and it holds up. Approving. I went looking for the places this could be quietly wrong and mostly came back empty: `RENDERER_CHILD_TAGS` matches the renderer's shipped `DashboardChildConfig` exactly, the grid-parity Playwright test is a real comparison rather than a simulation, bad colspans are dropped and reported instead of validated-then-used, and the `.malloy-render.malloy-render` specificity actually beats the renderer's own rule. Good change.

Two things worth a follow-up commit:

**The new per-tile tags have read support but no lint.** `readTileLayout` (`dashboard.ts:648-662`) reads `label`/`subtitle`/`break`/`borderless` by literal key, and the only per-tile check in `lintDashboard` is the `colspan` block. So `# subtitel="Q3"` sets nothing and says nothing — which is the exact failure the artifact-tag enumeration in this same PR exists to end.

Same gap has a second half in the filter. `isDashboardModelPath` + `field === "root"` is a proxy for "this is a tile," not the thing itself; the single-query form lives in the same file and reports against `root` too, and its colspan lint loop iterates `manifest.tiles ?? []`, which is undefined there. So `# colspan` misplaced on a top-level `query:` is now silent where it used to be reported. Both call sites already have `queryName` in scope. Narrow edge, same root cause, probably one fix.

**The divergence record only names one of the two divergences.** `docs/malloyyo-dashboards-design.md` §Where Publisher diverges says "One property, deliberately" and covers `dashboard_columns`. `docs/dashboards.md`, `RELEASE_NOTES.md` and `choosing-a-surface.md` all repeat that framing, and two of them point at that section as the authority — but your own description calls the additive per-tile divergence "the larger of the two." A Malloyyo maintainer following the link to evaluate the break sees half of it. Worth adding there and letting the others keep pointing.

Minor: the commit carries a `Co-Authored-By: Claude` footer (sign-off is fine). "Supersedes #1073, which fixed the narrow grid-width bug" reads as though #1073 shipped — it was closed unmerged. And `choosing-a-surface.md:34` and `:94` still describe a dashboard as a tagged query / two parallel options.

On malloyyo#160 being unanswered: fine to land. The lint makes the break visible and either direction reversible, and waiting indefinitely on a silent issue isn't better.

jswir (approved):

jswir (approved):

PR #1081, opened 2026-08-27
skills: point consumer skills at given: instead of #(filter)

jswir:
`malloy-model` and `malloy-notebooks` already mark `#(filter)` legacy and document `given:` as the replacement. Four consumer-facing skills were never updated and still present `#(filter)` as the current mechanism:

- `malloy-analyze` — tells the agent to add a `#(filter)` to the source when the analysis needs a knob the model doesn't expose.
- `malloy-define` — instructs the agent to write `#(filter)` annotations into a model it is authoring. This is the one that actually produces new `#(filter)` in new models.
- `malloy-document` — its annotation table and its `#(filter)` section both describe it as current, and point at a `malloy-model` section heading that no longer exists under that name.
- `malloy-analysis-report` — same as `malloy-analyze`, for reports.

Each now points at `given:` and names the two exceptions `malloy-model` already documents — `required` and `implicit` — so the guidance doesn't overshoot into cases `given:` cannot cover yet.

Found while following `malloy-reproducible-analysis` on a real analysis: the model-authoring path recommended `#(filter)` with nothing marking it deprecated, and the stale cross-reference in `malloy-document` sent the reader to a heading that had already been renamed to *Legacy: Parameterizable Filters*.

Docs only, no code. Independent of #1079 — no overlapping files.


jswir:
Superseded by #1079, which carries this commit cherry-picked unchanged (`87597a24`).

Folding it there rather than merging separately: it was unreviewed, it is the same class of change as that PR, and it was the last thing blocking the skills PR that retires `malloy-analyze` — one of the four files this touches.

PR #1083, opened 2026-08-27
fix(server): rename a built table into the schema its persist name asked for

housejester:
## The problem

The staging-table swap that finishes a colocated `#@ persist` build names the
`RENAME TO` target by its **bare** name:

```sql
CREATE TABLE "MY_SCHEMA"."orders_v3_<staging>" AS (…)
DROP  TABLE IF EXISTS "MY_SCHEMA"."orders_v3"
ALTER TABLE "MY_SCHEMA"."orders_v3_<staging>" RENAME TO "orders_v3"   -- ← bare
```

That rests on the assumption stated in `bareTableName`'s docstring — that a
rename "names a table within its existing schema rather than re-stating the full
path". It holds on most engines, but not all: some resolve an unqualified target
against the **session's** current container, so a bare target *moves* the table.
On those, a source whose `#@ persist name=` points at any container other than
the connection's default is built in the right place and then moved to the wrong
one.

Two ways it surfaces, both observed against a real warehouse:

| Condition | Result |
| --- | --- |
| Nothing of that name in the default container | Build reports success. The manifest records `MY_SCHEMA.orders_v3`; the table is actually in the default container. The serve path then resolves a path holding no table. |
| Something already there | Build fails with ``Object '"orders_v3"' already exists`` — while `MY_SCHEMA` is empty, which reads as nonsense. |

The first is the worse one: a green build, a manifest that points at nothing, and
no error anywhere.

## Which dialects

Measured rather than assumed — one table per dialect, renaming `s1.t_staging` to
a bare `t` with the session pointed at `s2`:

| dialect | bare target | qualified target | affected |
| --- | --- | --- | --- |
| snowflake | **moves to s2** | accepted, correct | yes |
| mysql | **moves to s2** | accepted, correct | yes |
| postgres | stays in s1 | syntax error | no |
| duckdb | stays in s1 | parser error | no |
| bigquery | stays in s1 | `name "s1.t2" cannot contain dot` | no |
| trino | stays in s1 | (bare already correct) | no |

So this cannot be fixed by qualifying unconditionally — three of the six reject a
qualified target outright.

Note the two axes are independent: whether a **persist name** may carry a
container (which is what enabled BigQuery materialization) is a separate question
from what form the **rename target** must take. BigQuery needs the dataset in the
name *and* requires the rename target to be bare.

Databricks is not covered by that sweep and stays out of the set, keeping its
current behaviour.

## The fix

`quoteRenameTarget(tableName, dialect)` in `quoting.ts` qualifies the target on
the dialects that resolve a bare one against the session, and keeps it bare
elsewhere. `QUALIFIED_RENAME_TARGET_DIALECTS` holds `snowflake` and `mysql`, with
the measurement table recorded beside it so a future addition is a measurement
rather than a guess.

## Testing

- Six conformance cases in `quoting.spec.ts` covering both affected dialects
  (two-part and catalog-qualified), the bare-target dialects, and an
  already-unqualified name.
- Full server suite: 3208 unit + 387 integration, 0 failures.
- **Verified end to end against a real Snowflake warehouse** in the Docker image,
  driving an orchestrated colocated build whose persist name targets a
  non-default schema:
  - before: build green, table in the default schema, target schema empty
  - after: table in the schema the model named, default schema empty

Note for anyone who has already run colocated builds into a non-default container
on Snowflake or MySQL: existing tables are in the default container with manifests
pointing elsewhere, so those sources need a rebuild rather than just an upgrade.


girishjeswani (approved):
Approving. The diagnosis is right and gating the qualification rather than applying it blanket is the correct shape: `QUALIFIED_RENAME_TARGET_DIALECTS` holding exactly `{snowflake, mysql}` means Postgres keeps working, and an unknown dialect falls to the bare target, which is today's behaviour. I checked that rather than assuming it, running the whole `materialization_service.spec.ts` including the BigQuery container-path swap test, which still asserts the bare form. 146 pass.

I also confirmed the case I care about most, since we persist into a non-default Snowflake schema. With `#@ persist name=B.orders_v3` on a connection defaulting to `A` the swap now emits:

```sql
DROP  TABLE IF EXISTS "B"."orders_v3_<staging>"
CREATE TABLE          "B"."orders_v3_<staging>" AS (…)
DROP  TABLE IF EXISTS "B"."orders_v3"
ALTER TABLE           "B"."orders_v3_<staging>" RENAME TO "B"."orders_v3"
```

The invariant that makes this hold is worth naming: the rename target is now textually the same path `quoteTablePath` gave the CREATE, so the two can no longer resolve to different containers. That also explains the "already exists while MY_SCHEMA is empty" symptom, since the DROP at `:2465` cleared `B.orders_v3` while the bare target collided in `A`. And the swap is consistent end to end, the delta DML in `incremental_apply.ts` already interpolates the fully-qualified path, so this isn't a fix that relocates the failure downstream.

Two things worth picking up, neither blocking.

**The call site isn't covered.** I reverted just `materialization_service.ts:2439` back to `quoteIdentifier(bareTableName(physicalTableName), dialect)` and got 146 pass / 0 fail. `quoting.spec.ts` covers the helper well, both mutation directions get caught there, but every existing swap assertion uses either a 1-part name, where bare and qualified are identical and can't discriminate, or `standardsql`, which stays bare. Since the failure mode here is a silent green build writing to the wrong schema, I'd want the wiring pinned. This follows the pattern at `materialization_service.spec.ts:2843` and `callBuildOneSource` already takes a dialect:

```ts
it("qualifies the rename target on a session-relative dialect (Snowflake)", async () => {
   const runSQL = sinon.stub().resolves();
   const entry = await callBuildOneSource({ runSQL }, "MY_SCHEMA.orders_v1", "snowflake");

   const sql = runSQL.getCalls().map((c) => c.args[0] as string);
   expect(sql).toEqual([
      'DROP TABLE IF EXISTS "MY_SCHEMA"."orders_v1_abcdef123456"',
      'CREATE TABLE "MY_SCHEMA"."orders_v1_abcdef123456" AS (SELECT * FROM t)',
      'DROP TABLE IF EXISTS "MY_SCHEMA"."orders_v1"',
      'ALTER TABLE "MY_SCHEMA"."orders_v1_abcdef123456" RENAME TO "MY_SCHEMA"."orders_v1"',
   ]);
   expect(entry.physicalTableName).toBe("MY_SCHEMA.orders_v1");
});
```

I ran that verbatim: passes at head, fails under the reverted call site.

**I think the Trino row may be wrong, and Trino is still affected.** This is general knowledge rather than something I can show from the repo, so treat it as a question. Trino's `RenameTableTask` resolves the target through `MetadataUtil.createQualifiedObjectName(session, statement, statement.getTarget())`, which fills a missing schema from `session.getSchema()` and a missing catalog from `session.getCatalog()` — session-relative for a bare target, same as Snowflake. `TrinoConnection` carries both `catalog` and `schema`, so the precondition exists. My guess at why the sweep saw "stays in s1" is the connector rather than the engine: `BaseJdbcClient` throws `NOT_SUPPORTED, "Table rename across schemas is not supported"` while Hive and Iceberg allow it. But a refusal would surface as an error rather than as a no-op, so the observation is hard to explain either way. Which catalog was it run against, and was the session schema definitely pointed away from the table's schema?

The useful asymmetry: qualifying Trino is risk-free even if the measurement is right. The grammar accepts a qualified target, and qualifying with the source table's own schema satisfies both the same-catalog check in `RenameTableTask` and the same-schema check in the JDBC connectors. So leaving it bare risks a silent misplacement and adding it risks nothing.

**Two nits.** `bareTableName`'s docstring at `quoting.ts:5-8` still says the rename target "names a table within its existing schema rather than re-stating the full path", which is the exact sentence the PR body cites as the wrong premise, and the function is no longer the rename target directly. And `quoteRenameTarget` inherits `quoteTablePath`'s split-on-every-dot, so a literal dot inside an identifier is mis-segmented — pre-existing, consistent between the CREATE and the new ALTER, which is what matters, just noting the new call site rests on it.

Last thing, and it's the part I'd most like somewhere durable: the migration note in the description. Tables built before this lands are sitting in the connection's default container with manifests pointing elsewhere, so those sources need a rebuild. Since Credible consumes this as a worker, that's operational guidance a release note or the persist docs would carry better than a PR description. I'll go check our own Snowflake scratch-schema tables on the strength of it.

housejester:
Thanks — all four picked up, and the Trino question was the right one to ask.

**Call site is pinned now.** You were right that nothing discriminated it: every existing swap assertion uses a 1-part name or `standardsql`. Added your test plus the opposite direction (a container-qualified name on `standardsql` must still rename bare), and mutation-checked both — reverting the call site to `quoteIdentifier(bareTableName(...))` fails the Snowflake one.

**Trino: you're right about the risk, and my row was under-scoped rather than wrong.** I re-measured, and the duplicate row that bothered you was my own misreading of two `tail`'d listings — `s1` held both tables and `s2` was empty throughout. So the observation reproduces, but only against the `memory` connector, which ignores the target's container, and no real deployment runs on it. Your `NOT_SUPPORTED` hypothesis doesn't explain it either — I got `RENAME TABLE` success, not a refusal — so `memory` is simply permissive in a way that tells us nothing about Hive or Iceberg.

I also confirmed your asymmetry directly: Trino accepts a qualified same-schema target and performs it correctly. So `trino` is in the set, and the matrix row now says which connector it was measured against and why that doesn't generalize. Databricks stays out, unmeasured, and the comment says so.

**Both nits.** `bareTableName`'s docstring no longer claims to be the rename target — it now points at `quoteRenameTarget` for the dialect-specific part. On the dot-splitting: agreed it's pre-existing and that CREATE/ALTER consistency is what matters; left as is.

**Migration note.** Moved out of the PR description into a `## [Unreleased]` release-notes section marked ACTION REQUIRED, with the rebuild step and what's unaffected, and `name=`'s container form is now documented in `docs/materialization.md` where `name=` is introduced. `release-notes.mjs extract` parses it.

girishjeswani:
LGTM!

PR #1084, opened 2026-08-27
fix(server): report what a failed materialization actually failed on

housejester:
## The problem

A materialization that failed could report only this:

```
Maximum call stack size exceeded.
```

…with nothing in the server log at all. Three defects combine to produce that.

**1. The failure was redacted against a *live* connection.**

```ts
const reason = redactConnectionSecrets(errMessage(buildErr), connection);
```

`connection` there comes from `connections.get(...)` — a live `MalloyConnection`,
not its config. This is the only one of six redaction sites that does; the other
five pass an `ApiConnection` via `getApiConnection`/`destinationForCleanup`, and
the function's own doc says "from the given connection configs". So it walks
driver and pool internals that hold nothing worth redacting, and that walk is
unsound in both directions:

- An **enumerable accessor whose getter throws** — a pool or socket property that
  throws once its resource is gone, which is exactly the state a build failure
  runs in — makes `Object.entries` throw out of the walk. The throw escapes the
  per-source catch, so one source's failure becomes a whole-run failure: the
  sources that did materialize lose their manifest, and the failed-run reclaim
  drops the storage tables the run wrote.
- **`Map`/`Set` are not walked.** `Object.entries(new Map(…))` is `[]`, so a
  secret held in a pool's `Map` passes through in the clear — a shape a config
  never has.

**2. The redactor's walk had no cycle guard.**

`collectSensitiveValues` recursed with no `seen` set, while the equivalent walker
in `logger.ts` (`redactSensitive`) has always carried one. A back-reference in
that live graph recursed until the stack was exhausted, and the resulting
`RangeError` **replaced the build error being redacted** — which is where the
headline message came from.

**3. The background run's throw was never logged.**

`runInBackground`'s `.catch` recorded the message on the materialization record
and returned. A background run answers no request, so nothing surfaced it: no log
line, and — the half that actually locates a throw — no stack.

## The fix

- **Redact against the config**, not the live connection. This removes the
  unnecessary walk rather than making it survivable, and is the primary fix: with
  a config there is no cycle, no throwing accessor and no `Map` to miss.
- **Fallback that cannot leak.** `getApiConnection` throws for a connection the
  environment does not hold, so the fallback is `redactPgSecrets` —
  connection-free, already applied inside `redactConnectionSecrets`, and never the
  unredacted message.
- **Keep the cycle guard.** It still covers the other five sites and any future
  caller, and it matches `redactSensitive`.
- **Log the failure with its stack.** A cooperative abort logs at info instead,
  carrying the error it raced — otherwise an abort-then-fail race records only
  `"Cancelled"` and the real error is reported nowhere.
- **Redact the run-level message too**, in both the log and the record. A
  whole-run throw has no connection to redact against, but a connect failure while
  resolving the package's connections echoes a DSN, and `redactPgSecrets` needs
  none.
- One `seen` is shared across a call's connections, so the multi-connection sites
  stop walking a shared subgraph twice.

No change to a successful build: the redactor is only reached on the error path,
and the new logs only fire when a run settles abnormally.

## Testing

Every change is pinned by a test that fails when the change is reverted — checked
by mutation, not assumed:

| Mutation | Result |
| --- | --- |
| pass the live connection instead of the config | `redacts a failed source's reason before recording it` fails |
| drop the `redactPgSecrets` fallback | 5 fail, incl. 4 pre-existing tests whose stub environment throws |
| drop the run-level `redactPgSecrets` | `redacts a DSN from the run-level message…` fails |
| drop `error` from the cancel branch | `logs a cancellation as info…` fails |
| delete the `seen` guard | the cycle test fails |
| delete the `stack:` field | the stack test fails |

The redaction test was reworked to discriminate: the secret is now declared only
in the config, and the live connection deliberately carries none — as written
before, it carried the secret too, so it passed either way. `callExecute` takes an
optional connection-config map, defaulting to a throwing `getApiConnection`, which
is what exercises the fallback.

Full server suite: 3208 unit + 387 integration, 0 failures.

## Note on exposure

The logged message is the same value already persisted to the record on `main`,
and stack frames carry no credentials — so the log adds no new exposure, and the
run-level path is now strictly better than before. Cardinality is per run, not per
source.


girishjeswani (approved):
Approving. Both halves are real and both are pinned by tests that fail when mutated: deleting the `seen` guard fails the cycle test, deleting the `stack:` field fails the stack test, and deleting the whole new logging block fails both. The guard itself is correct — keyed on object identity, added before descending, threaded through both the array and property branches — and I checked it doesn't over-prune, since a DAG where the same secret-bearing object is reachable by three paths still redacts. Pruning skips re-walking, not collecting.

The second half is the part I'd underweighted reading the description. Because the outer catch rethrows, the new `logger.error` at `:3564` emits a stack for the variants this PR *doesn't* fix too, pointing straight at `collectSensitiveValues`. That makes a previously invisible class of failure diagnosable server-side, which is arguably more durable than the cycle fix.

One thing I'd like to understand, and it's the only item I'd act on before merge.

**The comment at `:346` says the callers pass a live connection, and that's true of 1 of 5 sites.** `:1961` passes `connections.get(...)`, a live `MalloyConnection`. But `:2179`, `:2771`, `:2839`, `:2926` and `:3426` all pass `ApiConnection` configs, via `getApiConnection`, `getStorageDestination` and `destinationForCleanup`, and the function's own doc comment says "from the given connection configs". So `:1961` reads as drift from its four siblings rather than an intended contract, and the fix makes an unnecessary walk survivable rather than removing it.

That matters because the live walk is unsound in both directions, and I confirmed both by probe against the patched function:

- **A throwing getter still eats the cause.** An enumerable accessor whose getter throws — a pool or socket property that throws once the underlying resource is gone, which is the state the redactor runs in — makes `Object.entries` throw out of `collectSensitiveValues`:

  ```
  GETTER threw: Error: socket closed   out: ""
  ```

  The throw escapes the per-source catch, so one source's failure becomes a whole-run failure: the sources that did materialize lose their manifest, and `reclaimStorageTablesFromFailedRun` at `:2050` drops the storage tables this run wrote. Larger blast radius than the bug being fixed.

- **`Map`/`Set` aren't walked, so a secret can pass through unredacted.** `Object.entries(new Map(...))` returns `[]`. Probe: `{pool: new Map([["cfg", {password: "sup3rsecret"}]])}` against `"boom sup3rsecret"` returns the secret in the clear. Pre-existing on main, but it bites specifically at this one site, because pools commonly hold state in Maps and a config doesn't.

So I'd switch `:1961` to the config like its siblings. `environment` is in scope at that point via `executeInstructedBuild`, so it's about a line — with the caveat that `getApiConnection` throws `ConnectionNotFoundError` when absent (`environment.ts:1028-1032`), so it needs a guard or you reintroduce the same bug class on the error path. Separately, wrapping the redaction call would help, as long as the fallback isn't the unredacted message: `redactPgSecrets(errMessage(buildErr))` is a pure string-to-string redactor needing no connection and is already applied inside `redactConnectionSecrets`, so it's a safe floor.

**Smaller things.** Depth still isn't bounded, and a 50k-deep acyclic chain reproduces the exact headline `RangeError` — unlikely on real driver graphs, which are shallow and cyclic, and `redactSensitive` in `logger.ts` has the same property, so the new guard is at least consistent with it. `:408` creates a fresh `WeakSet` per argument through the default parameter, so the two-connection sites at `:2771`, `:2839` and `:2926` walk a shared subgraph twice; hoisting one `seen` above the loop is free. The cancel branch at `:3560` logs only the id, so an abort-then-fail race reports a clean cancellation with nothing recording the real error — `error: message` at info would cover it. And the run-level error isn't redacted in either the log or the record: the comment at `:3556` is right that there's no connection to redact against, but `redactPgSecrets` needs none and would catch a DSN echoed by a connect failure from `pkg.getMalloyConnection`.

On the log path itself: it's unredacted, but it adds no new exposure, since `error` is the same string already persisted to the record on main and stack frames carry no credentials. Cardinality is per run rather than per source, so no noise concern.

I'd keep the two walkers separate, for what it's worth — they collect versus return a redacted copy, with different key sets — so the parallel `seen` idiom looks like the right amount of sharing rather than something to factor out.

housejester:
Thanks — this changed the shape of the PR, and for the better.

**The live-connection site: agreed, and it's now the primary fix.** I verified your read before acting on it: `:1961` is the only one of six sites passing a live `MalloyConnection`, the other five pass an `ApiConnection`, `environment` is in scope, and `getApiConnection` throws `ConnectionNotFoundError` at `environment.ts:1033`. So it reads as drift, exactly as you said, and removing the walk beats surviving it. The description now leads with the call site and demotes the cycle guard to what it actually is — cover for the other five sites and for depth.

Fallback is `redactPgSecrets`, as you suggested, never the unredacted message. Worth noting it turned out to be load-bearing beyond the new path: removing it fails five tests, four of them pre-existing ones whose stub environment throws.

**One correction to my own work, prompted by your standard.** My first pass at the redaction test didn't discriminate — the live connection object carried `password: secret` too, so it passed whether the config or the live object was walked. The secret is now declared only in the config and the live object deliberately carries none, so reverting the call site fails it. I also wrote a hostile-getter test, found it passed under mutation (my `try/catch` absorbs the throw regardless of which object is walked), and deleted it rather than keep a test that pins nothing.

**The three smaller ones, all taken:** one `seen` hoisted above the loop; the cancel branch carries `error: message` for the abort-then-fail race; the run-level message goes through `redactPgSecrets` in both the log and the record.

**Depth:** left alone, as you suggested — noting it matches `redactSensitive`, so the two stay consistent. Happy to bound both together in a follow-up if you'd rather.

Every change is mutation-checked; the table in the description lists which mutation kills which test. And agreed on keeping the two walkers separate.

girishjeswani:
LGTM!

PR #1085, opened 2026-08-27
fix(server): say why a server with no config is serving nothing

mlennie:
## What this fixes

The Docker image runs with `--server_root /publisher` and nothing mounted there, so no config resolves and `getPublisherConfig` returns an empty one. The server then reports `operationalState: serving` with `environments=0 packages=0 load_errors=0`, and nothing in the output at any level names the path it checked. Every one of those numbers is accurate, which is the problem: nothing failed because nothing was asked for, so a degraded deployment is indistinguishable from a healthy one. An operator who forgot the config mount sees a running server and an empty catalog, and `malloy_getContext` answers `{"results":[]}` with `isError: false`, which reads to an agent as "this server has no models" rather than as a fault.

This is the same shape as #1049 in a second place, and it carries that PR's conventions rather than inventing new ones: keep the boot, name the cause, and match the level to how the condition is actually reached.

## Reproduced before it was changed

Booted the published image `ms2data/malloy-publisher:0.0.241` with `--network none` and queried it from inside the container. `PUBLISHER_READY ... environments=0 packages=0 load_errors=0`, `/api/v0/status` reporting `serving` with `environments: []` and no `loadErrors` key, and `malloy_getContext` returning `{"results":[]}`. Not one log line mentioned the config.

**The bundled default config is not involved, despite being the obvious suspect.** `CMD` has passed `--server_root` since #732 precisely so the zero-arg bundled-default trigger does not fire, so the image never reads that file. Booting the bundled-default path offline for comparison gives `environments=1 packages=0 load_errors=3` with each failed clone named in `/api/v0/status`. That path already reports itself completely, thanks to #898. The hole is narrower: it is the branch where no config resolved at all.

## The change

`getUnresolvedPublisherConfigPath(serverRoot)` in `config.ts` returns the path that was looked for and not found, or null once a config resolved. It returns null for an explicit `--config` in either state, because `getPublisherConfig` already reports that case at `error` and two shapes of one complaint help nobody.

`environment_store.ts` logs the notice from `initialize()`'s success tail, immediately before the readiness line. `initialize` is private and called once from the constructor, so this is once per process by construction rather than by care, which is the per-read re-emission trap #1049 had to fix twice. It is gated on having zero environments, so a server that took its environments from the database stays quiet.

**Why `info` rather than `warn`, since #1049 established `warn` for a skipped config entry.** This is not a skipped entry. Nothing was skipped and nothing failed: the server was given no config and says so. Booting unconfigured is a supported way to run, because environments can be created over the API afterwards, and deployments do run it that way, so `warn` would fire on every one of those boots for the intended path. The sibling case in the same function, falling back to the bundled default, logs at `info` too.

**Nothing is added to `load_errors`.** Nothing failed to load, and `packages=` and `load_errors=` are disjoint counts of configured things. Inflating the count would make a legitimate boot read as degraded. The counts already were the machine-readable distinction; what was missing was the reason and, in CI, a reader.

## The smoke test could not have caught it

`docker_smoke_test` booted the image with no config and asserted only that `/api/v0/environments` returned a JSON array, which an empty catalog satisfies. It now checks both shapes.

The unmounted container must return an empty array and must say why it is empty, which pins the diagnostic against silent regression. A second container mounts a config plus the checkout's `examples/` over a bind mount and must load an environment with no load errors, serve a package, and return a non-empty context from `malloy_getContext`. Nothing is baked into the image, so the default boot is byte-for-byte what it was, and packages come from the checkout rather than GitHub so the leg does not depend on egress.

I verified the new `malloy_getContext` assertion **fails** against a live-but-empty server (0 results) and passes against the configured one (1 result), so it discriminates the case that shipped rather than merely passing today.

## Considered and not done

Shipping `examples/` inside the image so it boots with data. It is only 1.7 MB, so size is not the objection: it would change what every existing deployment does on upgrade, since an `examples` environment would appear where operators configure their own, and the Dockerfile already records a deliberate decision to keep sample loading off the startup path. The reporting gap was the defect; the config was not.

## How it was verified

Rebuilt the production image from this branch (`DUCKDB_VERSION=1.5.5`, the version CI derives from the lockfile) and booted it offline. The unmounted boot now prints the notice naming `/publisher/publisher.config.json` and its counts are unchanged. A configured boot gives `environments=1 packages=1 load_errors=0` with the notice correctly absent. Every new CI step was run against real containers, three of them verbatim out of the YAML and one with its port substituted so as not to bind 4000 while other work was using it.

Three mutations, each red without the change: removing the call takes the reporting test, breaking the helper's resolution check takes one parser and one reporting test, and defaulting `frozenConfig` to true takes the runtime-creation test.

That last test exists because the notice, its code comment and both docs all tell the reader that runtime creation still works on an unconfigured server. I verified that live first (`POST /api/v0/environments` returned 200 and the environment appeared in the list) and then pinned it, because nothing else in the suite did and all four artifacts would have gone quietly wrong together.

I also ran the new "says why it is empty" step verbatim against the **published, unfixed** image, after confirming that container had really booted. It exits 1 and dumps the container log. So both new assertions are proven to fail on the artifact that shipped rather than merely to pass on this branch.

Gate on the merged tree: typecheck, lint and prettier clean, unit 3209 pass / 3 skip / 0 fail across 153 files. The file count moved from 151 because the merge brought exactly two new spec files. Integration is 387 pass / 0 fail across 40 files, and that 40 is inflated: `test:integration` passes `tests` as a substring path filter, so once `publisher_data` is populated it also collects the bundled storefront example's 72 tests across 3 files. The server suite alone is 315 across 37 files. Everything passes either way.

One thing a reviewer should know rather than discover: the notice is `info`, so an operator running at `warn` or `error` will not see it. That is a deliberate request to see less, and the durable signal is unaffected, because the readiness line is written straight to stderr and never passes through the logger. The counts print at any log level. Publisher's logger defaults to `debug` and deployments that set the level set it to `info`, so the notice is visible in both.

## Docs

Both Docker docs are updated, since between them they are what teach a reader to treat `serving` as the finish line. `packages/server/README.docker.md` now says `serving` reports that the server is up rather than that it loaded anything and points at the `PUBLISHER_READY` counts, including what `load_errors=N` means and where the reasons live. `docs/deployment.md` describes the unmounted boot and quotes both lines in full.

## Base

Merged `main` at `c9aa9926` and re-ran the whole gate rather than assuming a merge moved nothing, which mattered here: main changed `api-doc.yaml`, so codegen ran first, and #1077 added a boot-time assertion at `server.ts:291` that sits upstream of this code path. `config.ts` is touched by both diffs, so I read main's change rather than trusting a clean merge: it is purely additive DuckDB memory helpers around line 274, and all five symbols this PR touches return zero hits in it. I re-booted the merged tree and confirmed the notice, the counts and runtime creation all still behave.

## Review note

The review on this branch was the author's own, not an independent lens. It found two things, both in my own prose: the deployment doc quoted my new log line truncated at the config path, dropping the two sentences that carry the remedies, and I had updated only one of the two Docker docs, leaving the more specific one still implying that `serving` means loaded. Both are fixed.


packages/server/src/service/environment_store_unconfigured.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+      infoSpy.mockRestore();
+      rmSync(serverRootPath, { recursive: true, force: true });
+   });
+
+   it("names the missing config once when nothing loaded", async () => {
+      const store = new EnvironmentStore(serverRootPath);
+      await store.finishedInitialization;
+
+      const notices = noticeLines();
+      expect(notices).toHaveLength(1);
```
Sha-Bang, line 105: **Blocking — these two tests fail on ubuntu and windows in CI.** `logUnconfiguredNotice` only runs on `initialize()`'s success tail. On those runners they finish in 1ms with `Expected length: 1, Received length: 0`, and the line immediately above is `PUBLISHER_INIT_FAILED error="attach failed: host=db port=5432 *** dbname=x"` — that's the redaction fixture in `environment_store_clone.spec.ts`, not a real attach.

That file `mock.module`s `StorageManager` with a process-wide `storageInitFails` flag, sets it true in the last tests, and does not clear it in `afterEach`. Cross-platform runs `test:unit --serial`; this file sorts after the clone spec, so it inherits the mock with init still failing. macOS is green, which reads as glob-order luck.

The "stays quiet when a config resolved" case prints the same `PUBLISHER_INIT_FAILED` and still passes, because it asserts zero notices — a failed boot and a successful empty config look the same.

The comment says no mocking so the **config** resolver stays real. Anchoring already mocks only `StorageManager` and keeps the real resolver. Same shape here, plus an assertion that init actually succeeded (no `PUBLISHER_INIT_FAILED`), would make "no notice" mean what it says.

I confirmed this by reading the spec, the clone mock, and the Ubuntu job log; I did not re-run the suite.

Sha-Bang (changes requested):
the notice-on-empty-boot shape is right — once, at info, gated on zero environments, not stuffed into `load_errors`. one blocker: the two boot-path tests fail on ubuntu and windows because they construct `EnvironmentStore` against the leaked `StorageManager` mock from `environment_store_clone.spec.ts` (`storageInitFails` left true; `PUBLISHER_INIT_FAILED` is that spec's redaction fixture). the "stays quiet" case fails the same way and still passes. mock storage the way anchoring does, and assert init actually succeeded. docker smoke is green.

Sha-Bang:
Closing — this landed as #1184, merged in `d460fada`.

Your change went in unmodified: `getUnresolvedPublisherConfigPath`, the notice logged once from `initialize()`'s success tail at `info`, the docs, and the smoke assertion that replaces `type == "array"` — which passed for an empty catalog and a full one alike, so an image serving nothing stayed green. You're credited as co-author and your `Signed-off-by` is preserved.

Two things changed on the way in.

The blocking finding from your review round was the spec's storage mock. `mock.module` is process-wide and every spec file shares one bun process, so this file inherited `environment_store_clone.spec.ts`'s `StorageManager`, whose `initialize` throws its redaction fixture — construction failed, `PUBLISHER_INIT_FAILED` fired, and the success tail carrying the notice never ran. The sharp part is that `stays quiet when a config resolved` asserts the notice is ABSENT, so it passed under the leak for the opposite of the reason it names. Storage is mocked the way `environment_store_anchoring.spec.ts` does it, and `assertInitialized` now pins the success line in every boot case so a notice-absent assertion can only mean something once the boot happened.

Review on #1184 then found a second one, in the gate itself. It read `this.environments.size > 0` — "did anything load" — where the sentence it prints claims "was anything asked for". An environment the database holds whose directory has gone is skipped at the `not found in config and files missing` branch with a logged error and a bare `return`, recording nothing in `failedEnvironments` either, so the notice fired and told the operator to create a config file that was never the problem, directly under the one line naming the real cause. Reachable exactly the way the notice's last sentence invites: boot unconfigured, create an environment over the API, restart onto a root where `publisher_data/<env>` is gone. The declared count is now carried out of `initialize()` — manifest entries plus database entries — rather than inferred from what survived.

The configured-container and MCP smoke legs did not ship. They were a "prove the built image can serve" test rather than a test of this change, and they're worth having on their own terms; the corrected version is preserved at `32c6c6ab` on that branch if anyone picks it up.

Your measurement of the Docker case — `--network none` against the published image, `PUBLISHER_READY ... environments=0 packages=0 load_errors=0` with nothing naming the config — is what made this straightforward to verify rather than argue about.

PR #1086, opened 2026-08-27
test: anchor every bun test path filter so a runtime copy cannot be collected

mlennie:
`bun test <arg>` does not take a directory. Its own `--help` says it runs "all test files with "foo" or "bar" in the file name", so the argument is a substring matched against the whole path. `packages/server`'s `test:integration` passed `tests`, and once anything boots a server the gitignored `publisher_data/` holds a copy of the storefront example, its own `tests/` directory included. Those three files were being counted as the server's.

Measured on one worktree, at `1580c4f8`, same command each time:

| state | collected |
|---|---|
| fresh tree, no `publisher_data/` yet | 315 tests / 37 files |
| same tree after a run populated it, argument unchanged | 387 tests / 40 files |
| same populated tree, anchored to `./tests` | 315 tests / 37 files |

315 + 72 = 387 and 37 + 3 = 40. The smaller number is the server's suite; the larger is the server's suite with 72 of an example's tests folded in. Those 72 already run as `test:examples`, from repo source, under a step that fails on a skip. Nothing is ever marked skipped here, so both totals look like clean sweeps and only the file count tells them apart. Re-measured after merging `main`, at `7833971b`: 315 tests / 37 files, 0 fail.

The wrong total is the smaller half of it. The copy is a snapshot, so it drifts from repo source, and a drifted copy fails correctly and reports the server's suite red for a reason outside the server. That was measured on 2026-08-24, when a vendored file was still stamped 0.0.427 against 0.0.432 in repo source and node_modules. I have not reproduced that failure here, only the count inflation; the copies in my worktree are current.

It is easy to assume this is local-only, and it is not. A fresh checkout has no `publisher_data/` and collects the right 37 files, which is why the first measurement above passes with the unfixed script. But the integration step in `cross-platform-tests.yml` retries up to three times, and attempt 1 boots a server. Attempts 2 and 3 measure something attempt 1 did not.

## What changed

The fix is a leading `./`, which makes the argument a path instead of a substring.

Every `bun test` in the repo used the unanchored form, so all of them are anchored here rather than only the one that fires today: the server's `test:unit` and `test:integration`, the root `test:hammer` and `test:scripts`, the CLI's `test` and `test:integration`, `packages/skills`' `test`, and `create-malloy-package`'s `test` and `test:e2e`. For each I ran the old and the new form and compared collected file counts; every one is identical, so this changes what could be collected, not what is. `packages/sdk`'s `bun test` has no filter to anchor and is left alone.

Grepping for the command form rather than for package.json found more that no manifest knows about. `connection-integration-tests.yml` invokes `bun test` directly with an unanchored file path in two steps, and that job runs with real BigQuery and Postgres credentials. Three comments and one README documented the bare form as a way to run "one directory", which is the misconception behind the bug; one of them described `test:unit` as `bun test src` and this change had just made that wrong. After those, the only bare `bun test` left in the repo is the SDK README's deliberately unfiltered one.

## The guard

`scripts/test-path-filters.spec.ts` parses every `bun test` invocation out of every workspace manifest and every workflow file, and asserts that each path filter collects its own path and does not reach into `publisher_data/`. Against a tree with the anchoring reverted it fails 12 times, one per filter plus a named case for the three storefront copies.

Three things it deliberately does not do. It does not walk the filesystem: `publisher_data/` is a runtime directory a fresh checkout does not have, so a walk would find nothing to collect and pass on every CI run while saying nothing, which is the shape the example-tests step in `cross-platform-tests.yml` already documents. It does not assert a file count, which would need raising by hand on every new test file. And it does not check that the argument starts with `./`, which would test the fix rather than the property; it reproduces bun's matching rule instead.

Because everything in it is a loop over what its parser found, a parser that quietly found nothing would report a clean sweep, so it asserts up front that it found the scripts it exists to check, named rather than counted.

One case is different from the rest: it shells out to bun in a scratch directory and checks that `./tests` collects one file where `tests` collects two. That case exists for Windows. The rest of the file is synthetic, and synthetic cannot tell "./x anchors on this platform" from "./x is not recognised here, so it fell back to a substring and matched anyway". The second would mean the anchoring silently stops applying on the one platform where the retrying integration step creates the copies it exists to exclude, with CI staying green. So `test:scripts` now runs on the cross-platform matrix rather than ubuntu alone.

## How this was verified

The mechanism was confirmed from `bun test --help` and reproduced in a throwaway fixture before the repo was touched. The anchoring rule was probed separately, since the interesting case is not documented: `src/tests` collects `nested/src/tests/b.test.ts` and `./src/tests` does not, so a slash alone does not anchor.

Anchoring can only narrow what is collected, and two of the changed lines are in a credentialed job, so the obvious question is whether a bad anchor could leave it testing nothing. It cannot: `bun test` on a path matching no file exits 1, not 0. That is what `--pass-with-no-tests` exists to override, and nothing here passes it.

Every assertion in the guard was mutation-tested: the code it covers was reverted and the assertion watched go red before being kept.

Full gate at `7833971b`, with file counts beside the test counts, since that is what this change is about: server integration 315 tests / 37 files, server unit 3206 / 152, `test:scripts` 74 / 3, sdk 356 / 22, skills 15 / 3, create-malloy-package 376 / 10, hammer 20 / 1, examples green. Typecheck clean, lint clean and it made no autofix edits, prettier clean apart from the six git-ignored generated files CI never sees.

An adversarial review of the first commit found six things, four of them defects in the guard itself, and the second commit fixes them. The sharpest was that the parser split scripts on `&&` alone, so a script joined with `;` or `||` disappeared from the parsed set entirely and no assertion fired for it while the run still reported a clean sweep. That is the guard exhibiting the bug it exists to catch.

## One thing left alone

`packages/server`'s test scripts pass `--max-workers=1`, and `cross-platform-tests.yml` appends `--serial`. Neither appears in bun 1.3.13's `--help`, which has `--parallel` instead, and passing `--max-workers=1` exits 0 with no effect I could observe. `bun test` runs serially without `--parallel`, so intent and behaviour coincide today, but nothing enforces it, and the comment justifying the unit/integration split relies on `mock.module` not leaking across files. That is a separate change with a real behavioural risk, so it is not in this PR.


scripts/test-path-filters.spec.ts
```diff
@@ -0,0 +1,443 @@
...
+          : `${target}/example.test.ts`;
+
+        it(`${filter} collects its own path`, () => {
+          expect(collects(filter, own)).toBe(true);
+        });
+
+        it(`${filter} does not reach into publisher_data`, () => {
+          const copy = `publisher_data/examples/storefront/${own}`;
+          expect(collects(filter, copy)).toBe(false);
+        });
```
Sha-Bang, line 429: `copy` is built from the filter, so a filter without `./` always shows up inside `copy`. "does not reach into publisher_data" is then the same as asking whether the filter starts with `./`, which the file says it will not do. "collects its own path" is true for both `tests` and `./tests`, so it does not catch the bug this PR is fixing. the temp-dir run above is the check that talks to bun; this loop isn't doing extra work.

scripts/test-path-filters.spec.ts
```diff
@@ -0,0 +1,443 @@
...
+    const labels = ALL.map(label);
+    expect(labels).toContain("packages/server test:integration");
+    expect(labels).toContain("packages/server test:unit");
+    expect(labels).toContain(". test:scripts");
+    // The workflow scan is a regex over plain lines, so it is the part most
+    // likely to quietly stop matching. Asserted as a floor of one rather than
+    // by file and line, which move whenever a step is edited.
+    expect(
+      ALL.filter((i) => i.pkg.startsWith(".github/")).length,
+    ).toBeGreaterThan(0);
```
Sha-Bang, line 386: this is supposed to prove the parser still sees every script this PR touched. it names three, plus "at least one workflow line". leave out cli, skills, create-malloy-package, hammer, or one of the two BigQuery workflow commands, and this still passes — the loop over filters just gets shorter. name the scripts from the commit, and both workflow lines, not a count of one.

Sha-Bang (changes requested):
the `./` in the package scripts is the fix. the only test that actually runs bun is the temp-dir one that checks 2 files vs 1. the rest of the new spec talks like it is keeping the filters honest, but it wouldn't notice if whole packages dropped out of the parse, and the publisher_data loop is just "does it start with ./?" written the long way. I'd want those two tighter before calling the guard the thing that keeps this from coming back.

Sha-Bang:
Closing — this landed as #1181, merged in `b6268903`.

The `./` fix is yours and went in unchanged, applied to every `bun test` in the repo rather than only the one firing at the time: the server's `test:unit` and `test:integration`, the root `test:hammer` and `test:scripts`, the CLI's two, `packages/skills`, `create-malloy-package`'s two, and both direct invocations in `connection-integration-tests.yml`. The cross-platform guard step and the doc-comment anchors went with it. You're credited as co-author and your `Signed-off-by` is preserved.

The guard spec is the one thing that changed, and it was the review's two findings: `collects()` reimplemented bun's matching rule and then asserted repo filters against that model, which made "does not reach into publisher_data" a restatement of the model's own definition rather than a fact about bun — and the non-vacuity check named three scripts, so a whole package dropping out of the parse would not have failed anything. Both trace to the same root cause, so the model is gone. Each filter now runs through the bun executing the suite, twice — once as written, once with the anchor stripped, so the unanchored leg proves the anchored one per filter — and the script and workflow roster is pinned by name in full.

Mutation-checked in three places: unanchoring `packages/server` `test:integration` goes red through real bun, dropping the workspace members from `packageDirs()` fails the roster, and breaking the workflow scan fails the workflow roster.

Two things your description turned up that outlived it: the measurement that a populated `publisher_data/` inflates the server's suite by 72 tests and 3 files, and the observation that `cross-platform-tests.yml` retries three times so attempts 2 and 3 measure what attempt 1 created. Both are recorded in the spec's header.

PR #1088, opened 2026-08-27
feat(skills): a manifest that decides what ships, and tests that hold every channel to it

jswir:
## The problem

Publisher ships skills four ways, and each one independently took "everything under `skills/` minus `credible-*`":

| Channel | How it picked skills |
|---|---|
| `.claude/skills/` symlinks | `ln -s` by hand |
| `@malloy-publisher/skills` | `copy-skills.ts` globbed the directory |
| MCP prompt bundle | `build_skills_bundle.ts` globbed the directory |
| `create-malloy-package` | copied whatever the pack produced |

So a skill added to the tree shipped everywhere by default, there was nowhere to say otherwise, and nothing noticed when the four disagreed. `docs/agent-skills/authoring-skills.md` documented a `manifests/` directory and a registration step that did not exist.

## What this does

`manifests/publisher-local.json` becomes the single answer to what ships. The pack and the bundle resolve it instead of globbing; the scaffolder inherits it, since it copies the packed `skills/` the pack has already filtered.

`groups` names optional subsets (`html-apps`, `lookml`, `materialization`) so a consumer wanting fewer skills gets a filter rather than a second directory. `supporting` is empty on purpose: agents discover a second skills directory poorly, and an SDK `Skill` tool cannot invoke from one at all.

**The bundle is byte-identical to `main`.** That is the intended outcome — the same 31 skills ship, now through something that can be reasoned about and tested.

## The tests are the point

`packages/skills/src/manifest.spec.ts`, 134 cases, ported from `ms2data/agent-skills` `tests/test_manifests.py` so both sides of the vendoring boundary enforce the same contract:

- the manifest names only skills that exist, **and** ships every skill in the tree — the second is what catches forgetting to register one
- frontmatter names its own directory; no skill declares its own `version:` (the pack stamps that)
- no body hardcodes a host install path (`.claude|.credible|.cursor/skills/`)
- relative `reference/…` paths resolve inside their own skill
- cross-skill references are **closed over the manifest**, and never carry a `skill:<name>/subpath`
- the `.claude/skills` symlinks match the manifest

**Confirmed the checks fire rather than passing for free:**

- dropping `malloy-charts` from the manifest → 3 failures (tree coverage, ref closure, symlinks)
- naming a skill that doesn't exist → 6 failures
- the reference scan sees **186 real `skill:` references across 57 files** hitting all 31 skills, so the closure check passes on evidence, not by matching nothing

## Testing

`packages/skills` 149 pass · `check-pack` clean (56 skill files in 61 tarball entries) · `packages/server` MCP suite 320 pass · `create-malloy-package` 376 pass · prettier clean on every changed file.

Note `skills/*.md` is deliberately **not** prettier-managed (main's own copies fail `--check`), while `docs/**.md` is; only the latter was formatted.

## Context

First of four PRs in the consolidation proposed in ms2data/service#6177 (private). This one stands alone: it changes no skill content and is a no-op for what ships. `skills/README.md` also records that this repo is the source of truth for the shared skills rather than a copy kept in step by hand.

**Deliberately not included:** adopting `agent-skills`' `skill-review` as repo tooling. It's Credible-authored content in a private repo, and copying it into a public one is a call for a maintainer, not something to slip into an infrastructure PR.


Sha-Bang (commented):
6 findings: 1 blocking, 3 should-fix, 2 nits.

- `manifests/publisher-local.json:51` — `analysis` and `modeling` are documented as standalone but neither is closed over its own `skill:` refs (2 and 9 dangling); the new closure test only checks the full manifest.
- `packages/skills/scripts/copy-skills.ts:49` — `check-pack.ts` and `index.spec.ts` still glob `skills/` filtered by `isExcluded` alone, so the first manifest omission breaks `prepublishOnly`, not just the pack.
- `packages/skills/scripts/copy-skills.ts:24` — the "deliberate non-ship" comment promises a state three guards redden.
- `packages/skills/src/manifest.spec.ts:209` — `isSymbolicLink()` is the first thing in the Windows-run skills suite to depend on committed symlinks surviving checkout.
- `docs/agent-skills/authoring-skills.md:30` — "analysis only" is used as the example for a second manifest, and for `groups`, in the same PR.
- `skills/README.md:14` — "asserts they agree" overstates what the spec checks for two of the four channels.

manifests/publisher-local.json
```diff
@@ -0,0 +1,92 @@
...
+      "malloy-html-data-apps",
+      "malloy-html-data-app-runtime",
+      "malloy-html-data-app-embedding"
+    ],
+    "lookml": ["malloy-lookml-review"],
+    "materialization": [
+      "malloy-materialization",
+      "malloy-materialization-tuning"
+    ],
+    "analysis": [
```
Sha-Bang, line 41: **Blocking** - `analysis` and `modeling` are documented as installable on their own ("An eval that measures one of those agents installs the matching group, not the whole set"), but neither group is closed over its own `skill:` references. Walking the markdown of each group's members: `analysis` leaks 2 — `malloy-getting-started` -> `skill:malloy-gotchas-modeling`, `malloy-analysis-report` -> `skill:malloy-model`; `modeling` leaks 9 — `malloy` alone points at `malloy-dashboards`, `malloy-getting-started`, all three `malloy-html-data-app-*`, `malloy-materialization-tuning` and `malloy-notebook-chat`, plus `malloy-modeling` -> `malloy-getting-started` and `malloy-publish` -> `malloy-html-data-apps`.

That is exactly the failure this PR's own "are closed over the manifest" test exists to stop — "Unknown skill at runtime for anyone who installs from it" — except it checks closure over the whole manifest only, so it cannot see a group. The exclusion groups have the mirror image: drop `html-apps` and `malloy` and `malloy-publish` both dangle.

Could the closure test run per group as well as over the full set, so a group advertised as standalone is proved to be? If the leaks are deliberate, it's the README's "installs the matching group" line that needs to change instead.

packages/skills/scripts/copy-skills.ts
```diff
@@ -14,17 +14,39 @@ import * as path from "node:path";
...
    filter: (from: string) => {
       const relative = path.relative(source, from);
-      return relative === "" || !isExcluded(relative);
+      if (relative === "") return true;
+      if (isExcluded(relative)) return false;
+      // Keep the manifest's skills and everything under them; drop the rest.
+      // path.sep rather than "/" so this holds on Windows, which the
+      // cross-platform job actually runs.
+      const [top] = relative.split(path.sep);
+      return shipping.has(top);
```
Sha-Bang, line 54: **Should-fix** - Two channels still derive their expectations from the source tree filtered by `isExcluded` alone, so the manifest isn't yet the single answer to "what ships":

- `scripts/check-pack.ts:73-74` — `sourceFiles().filter((file) => !isExcluded(file))` is what the tarball is required to contain.
- `src/index.spec.ts:25-26` — `skillNamesIn(sourceDir).filter((name) => !isExcluded(name))` is what `listSkills()` is required to return.

Both are green today only because `manifest.spec.ts` forces the manifest to equal the tree. The first manifest omission fails `check-pack` with "N of M skill files are missing from the tarball" — and it runs under `prepublishOnly`, so it breaks a publish while pointing at the copy rather than at the manifest. `exclusions.ts`'s own header names this hazard: "a filter that only the copy applies makes the audit demand a file that will never be there."

Worth passing the manifest set through both, so the audit forbids what the copy drops?

packages/skills/scripts/copy-skills.ts
```diff
@@ -14,17 +14,39 @@ import * as path from "node:path";
...
 import { isExcluded } from "./exclusions";
 import { locateFrontmatterClose } from "./frontmatter";
+import { manifestPath, manifestSkillNames } from "./manifest";
 
 const packageDir = path.dirname(path.dirname(fileURLToPath(import.meta.url)));
 const source = path.join(packageDir, "..", "..", "skills");
 const destination = path.join(packageDir, "skills");
 
+// What ships is the manifest, not whatever happens to be in skills/. A skill
+// added to the tree but not the manifest is a deliberate non-ship (work in
```
Sha-Bang: **Should-fix** - This comment promises a capability three guards forbid. `manifest.spec.ts:97` ("ships every skill in the tree") reddens the build on any skill on disk the manifest omits; `check-pack.ts` reddens the publish on the same state; and `skills/README.md` says the opposite in plain terms — "forgetting to [register] is a red build rather than a silent non-ship."

So a contributor who reads this and holds a work-in-progress skill back gets a red build, not a non-ship. Is exhaustiveness the intended contract? If so this comment probably wants to say the manifest is exhaustive by construction and earns its keep through `groups` and channel agreement; if a real non-ship is wanted, `manifest.spec.ts:97` and the two globs above are what stand in the way.

packages/skills/src/manifest.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+});
+
+describe("the .claude/skills symlinks", () => {
+   it("match the manifest", () => {
+      // This is how contributors get the skills in their own agent, and it
+      // drifted by hand before the manifest existed.
+      const dir = path.join(repoRoot, ".claude", "skills");
+      const linked = fs
+         .readdirSync(dir, { withFileTypes: true })
+         .filter((entry) => entry.isSymbolicLink())
```
Sha-Bang: **Should-fix** - `entry.isSymbolicLink()` makes this the first assertion in the skills suite that depends on committed symlinks surviving checkout, and `cross-platform-tests.yml` runs `bun run --cwd packages/skills test` on `windows-latest`. Git for Windows materializes a mode-120000 entry as a plain text file holding the target path unless `core.symlinks` is on, and in that state `linked` is `[]` and this fails with a 32-name diff that has nothing to do with the manifest. Nothing else in the Windows leg reads a committed symlink today, so the leg being green so far isn't evidence either way — worth confirming before this lands.

Would comparing directory names (entries under `.claude/skills` holding a `SKILL.md`) rather than the entry type give the same drift protection without the platform dependency?

docs/agent-skills/authoring-skills.md
```diff
@@ -21,14 +21,13 @@ skills/
...
-  modeling_and_analysis-ide.json
-  ...
+  publisher-local.json
 ```
 
 `skills/` is a flat collection of self-contained skills. A skill has no knowledge of which use
 case it belongs to. `manifests/` defines skill compositions for each use case and tells the
-host how to deploy them.
+host how to deploy them. This repository ships one manifest today; a deployment that wanted a
+narrower set (analysis only, say) would add a second rather than filter at the consumer.
```
Sha-Bang: **Nit** - This says a deployment wanting "analysis only, say" adds a second manifest rather than filtering at the consumer; `groups` fifty lines below, and `skills/README.md`, say `analysis` is the group a consumer installs to get exactly that. Same example, opposite advice — worth picking one.

skills/README.md
```diff
@@ -9,9 +9,19 @@ Task-specific guides for working with Malloy through this Publisher deployment.
 
 [`packages/skills`](../packages/skills) publishes this directory to npm, for consumers that need the files themselves without cloning. The MCP prompts carry the same tree: each `SKILL.md` body as one prompt, plus every `reference/*.md` as its own prompt named `<skill>/<file stem>`. It copies this tree in when it is packed, so adding a skill here needs no packaging step. It does need a version bump: `skills-npm.yml`'s PR check requires the version in [`packages/skills/package.json`](../packages/skills/package.json) to be ahead of what is on npm whenever a PR touches this directory, because a published version can never be replaced. A PR that adds or edits a skill without that bump goes red.
 
+## What ships: `manifests/publisher-local.json`
+
+[`manifests/publisher-local.json`](../manifests/publisher-local.json) decides which skills a deployment ships, and **every channel resolves it: none globs this directory.** The npm pack, the MCP prompt bundle, the `.claude/skills` symlinks, and the scaffolder all read the same list, and `packages/skills/src/manifest.spec.ts` asserts they agree with it and with each other.
```
Sha-Bang: **Nit** - "`manifest.spec.ts` asserts they agree with it and with each other" is a little stronger than what the file does (same sentence in `authoring-skills.md:111` and the `manifest.ts` header). The spec asserts the manifest against `skills/` and against the `.claude/skills` symlinks; the npm pack and the MCP bundle agree by construction because they filter on it, and the scaffolder only transitively, through the packed directory. "reads it" fits those two, keeping "asserts" for the pair actually checked.

packages/skills/scripts/copy-skills.ts
jswir: fixed

Sha-Bang (approved):

PR #1091, opened 2026-08-28
fix(sdk): send the dashboard's versionId, the way every other component does

Sha-Bang:
`Dashboard` was the only SDK component whose backing endpoints accept a `versionId` and that threw it away. `list-dashboards` and `get-dashboard` declare it in `api-doc.yaml` exactly as `list-notebooks` and `get-notebook` do, the generated client has always exposed it, and `Package`, `Model`, `Notebook` and `QueryResult` all thread it out of `parseResourceUri` already. A comment recorded the omission as intentional — *"the dashboards endpoint takes only env + package"* — which is simply false.

Four call sites: the manifest fetch, each tile's query, the package's dashboards listing, and `useSuggestOptions`, which had no version slot at all. **That last one is shared with `Notebook`**, so a notebook has been versioning its own fetch and its cell execution while its suggest dropdowns ran unversioned — this closes that too.

The **cache keys** are part of the fix, not a follow-up. A version that reaches the request but not the key is worse than no version at all: two versions of one dashboard collide on one entry and one serves the other's data, which looks correct. Each key now carries every value its request is built from, so the two can't drift.

`useGivensState`'s `documentKey` is in that set: without the version in it, a host swapping versions kept the values a reader had applied to the old one. That covers the reader's **edits** only — `initial` is `startingValues ∪ params`, so a host round-tripping givens through its own URL hands them back across the swap and they still apply. Deliberate: if the host's URL says `REGION=CA`, the host is asking for CA, and no key here should overrule it.

**Data apps stay unversioned** and that's deliberate: `list-data-apps` declares no `versionId` because `/data-apps` serves static files. The two exceptions used to be written down as one comment; the specs now pin them apart.

### One behaviour change, and one hole it opened

Point a **versioned** URI at Publisher and the page now reports an error where it used to render the latest — Publisher answers `501 Not Implemented` to a `versionId` on all eleven routes that declare one, and these calls no longer hide it. Same contract `Notebook` and `Model` have always had, and the reason these components are useful to a host that resolves versions at its own routing layer. `RELEASE_NOTES.md` has an `[Unreleased]` section for it.

A URI **without** a version sends nothing, which is every call site in this repo, the Console included. The only thing not byte-identical there is the cache keys: an unversioned dashboard's key gains an empty slot, the shape the package's other listings already have. In-memory, per page load, no request or render changes.

Sending a version on the dashboards listing also gave a host that resolves the parameter on some routes and refuses it on others a way to fail that one call — and the failure was invisible, because `dashboards` falls back to `[]`, the section hides when empty, and only the package query reaches the error page. A transport error and a package with no dashboards painted the same page. It now says so. The 404 stays swallowed: that one is an older Publisher with no route, which genuinely has nothing to list.

### Tested

Specs mount the real components against a shared `ServerProvider` stub and assert on the actual react-query cache, so they pin both what goes over the wire and what it's filed under:

- `Dashboard.spec.tsx` — manifest request + key; two versions get two entries; two spellings of one unversioned URI get one; end-to-end that the version survives the manifest fetch into the single-query tile, every composite tile, and the suggest query; and that applied control values don't cross a version switch.
- `DashboardTile.spec.tsx` — the prop alone lands in the body and the key.
- `Package.spec.tsx` — the dashboards listing is versioned, the data-apps listing still isn't, a failed listing says so, and a 404 still quietly shows nothing.
- `Notebook.spec.tsx` — the notebook actually forwards its version to the suggest queries. A bare trailing optional argument is the kind that goes missing with a green suite.
- `useSuggestOptions.dom.spec.ts` — request body + `useQueries` key.

The no-version cases assert the **whole** key rather than the absence of a version, which is what catches a slot that shouldn't have appeared.

`test/serverProvider.tsx` is the one place that stub lives, and that's load-bearing rather than tidiness: `mock.module` is process-global and can't be undone, so whatever shape a spec registers is what every later spec in the run inherits. `test/README.md` has the details.

```
bun test (packages/sdk)   384 pass, 0 fail
typecheck:sdk / :app      clean
lint:sdk                  clean
build:sdk                 clean, no test files in dist/
```

Each new test was checked against a reverted source rather than assumed load-bearing. The ones that pass both ways are exactly the ones that should: data apps unversioned, a notebook opened without a version, two unversioned spellings sharing a key, and the 404 listing.


housejester (approved):
Approving. This is the right shape: the version is read once from the URI and threaded to every request derived from it, and the cache keys ship in the same change rather than as a follow-up — which is correct, because a version that reaches the request but not its key is the failure that looks like success. Catching that `useSuggestOptions` is shared with `Notebook`, and that a notebook's dropdowns had therefore been running unversioned, is the part a narrower fix would have missed.

I verified the claims rather than reading them:

| claim | check | result |
| --- | --- | --- |
| the client exposes these params positionally as called | generated the client from `api-doc.yaml` | `getDashboard(env, pkg, dashboardName, versionId?)`, `listDashboards(env, pkg, versionId?)`, `listDataApps(env, pkg)`. All four call sites match, and `versionId` on the query path is a body field, which is how `DashboardTile` sends it |
| data apps genuinely declare none | spec + generated client | confirmed |
| additive — nothing in-repo sends a version | `grep -rn versionId packages/app/src` | zero hits |
| the reference server still 501s it | `server.ts` route map | 11 `setVersionIdError` guards, both dashboards routes among them (`server.ts:1680`, `server.ts:1703`) |
| tests, typecheck, lint | `bun test`, `typecheck:sdk`, `lint:sdk` in `packages/sdk` | 382 pass / 0 fail; both clean |
| the new tests are load-bearing | reverted the five source files to `main`, reran, restored | 16 fail. (Your body says 15 of 18; `main` moved under the branch, so the delta is expected) |

Four notes, none blocking.

**A newly-reachable error renders as a silently empty section.** `listDashboards` now sends `versionId`, so a host that rejects it (501) or errors (500) lands in the `throw e` branch at `Package.tsx:155`. But `dashboards` falls back to `[]` (`Package.tsx:171`), `isLoading` clears on `.isError` (`Package.tsx:226-228`), and only `pkgQuery.isError` renders an error page (`Package.tsx:230`) — so the rethrow and the deliberate 404-swallow produce identical UI: a package page with no Dashboards section. Against the reference server this can't bite, because `get-package` 501s a versioned URI first (`server.ts:1571`) and the page errors as a whole. Can you confirm there's no host where the package resource accepts a `versionId` but `list-dashboards` doesn't? If there is, "this package has no dashboards" is a fail-dangerous default for a transport error. Folding 501 into the swallowed statuses *and* surfacing `dashboardsQuery.isError` as a section-level notice would close it either way.

**The `documentKey` change closes less than the comment claims.** `useGivensState` uses `documentKey` only in `initialKey` (`useGivensState.ts:165-174`), which gates whether *edits* survive; `initial` is `startingValues ∪ params` (`useGivensState.ts:138-145`), and `Dashboard` passes `params: givens` (`Dashboard.tsx:191`). So for a host that round-trips givens through the URL — the Console, via `onGivensChange` — a reader's applied values still cross a version switch, carried in `params` and untouched by this change. The fix is right for a plainly-mounted `<Dashboard>`; it's the comment at `Dashboard.tsx:205-208` that reads as if the URL path were covered too. Worth narrowing so the next reader doesn't conclude more is closed than is.

**"the literal argument list handed to `getDashboard`"** — the key is `["dashboard", env, pkg, versionId, dashboard]` and the arguments are `(env, pkg, dashboard, versionId)`. Same values, different order. The commit message outlives the PR, and that phrasing invites the next reader to keep them positionally in step.

**Docs and release note.** `docs/dashboards.md:554` still shows `encodeResourceUri({ environmentName, packageName })` and doesn't mention the URI can now carry a version that will be sent — one line there would help a host author. And this does change behaviour for a host passing a versioned URI to a server that rejects the parameter: it used to silently render the latest, it now errors. `RELEASE_NOTES.md` asks for an `[Unreleased]` section in the PR that changes behaviour, and the two nearest sibling fixes each carried one; your call whether this clears that bar.

<sub>Reviewed with a lens pass — session "Intrepid Tamarin".</sub>

PR #1095, opened 2026-08-31
skills: teach the shared freshness rule, close the given: gap, and get diagnostics with no editor

Sha-Bang:
The `malloy-materialization` skill already names `freshness.window` + `fallback="live"` as the one control that bounds a gated source's stale **access decision**, and lists the two details that decide whether it actually binds. There is a third, and it is the one an author cannot see by reading their own source.

Reuse is keyed on the content-addressed `sourceEntityId`, which folds the connection and the SQL but **not** the source name. Two persist sources whose bodies compute the same SQL resolve to one table carrying one freshness policy — so a sibling's declaration decides part of yours.

Until now that was unstateable, because which sibling's policy reached the wire was decided by DB order. `ms2data/service#6064` makes the control plane min-combine the window across the binders of a shared address, so the rule is now well-defined and worth teaching:

> The tightest window any of them declares governs all of them — a sibling declaring nothing cannot loosen yours, and yours pulls that sibling's reads off the table once it lapses. If two sources need genuinely different windows, give them genuinely different SQL.

The Gotchas list already noted that reuse is keyed on `sourceEntityId` rather than `name=`, but only as a fact about *republishing unchanged logic*. It did not say that the same mechanism couples two **different** sources' freshness policies inside one package, which is the part that bites a gated source.

## Scope

One paragraph inside the existing freshness bullet ("Two details" → "Three details"). The surrounding "three controls that look like that bound" framing is deliberately untouched — a sibling is not a control, it is a condition on the first one.

`packages/skills/package.json` 0.1.11 → 0.1.12 and the regenerated `skills_bundle.json` ride along, per the skills-edit convention.

## Testing

`bun run packages/server/src/mcp/skills/build_skills_bundle.ts skills packages/server/src/mcp/skills/skills_bundle.json` — 57 entries (32 skills + 25 reference files), and the new text is present in the bundle. No em-dashes in the edited file.

---

## Also here: `malloy-notebook-chat` forbids `given:`, not just `#(filter)`

The skill told the agent:

> do NOT add `#(filter)` annotations: filters live on the source's model file and are inherited by this notebook automatically. Query-level `where:` filtering inside a cell is fine; declaring new filter UI is a model change, not a chat-time change.

The rule is right; the spelling went stale once `given:` became the recommended replacement for a presentation filter. An agent can satisfy the letter of that sentence by declaring a `given:` instead — the same model change under another name — and nothing else in the skill closes it.

It now names both on the same reasoning, and generalizes "new filter UI" to "a new runtime parameter" so the rule does not go stale again the next time the spelling moves. Query-level `where:` inside a cell stays explicitly allowed; that is the line the sentence exists to draw and it is unchanged.

It rides this PR's existing 0.1.12 bump rather than adding a second one, and the bundle was regenerated once over both edits.

Mirrored to [ms2data/agent-skills#105](https://github.com/ms2data/agent-skills/pull/105) alongside the materialization change. `diff` between the two copies of `malloy-notebook-chat/SKILL.md` reports only the license header Publisher adds — no other divergence, no em-dashes added.

Re-run after the second edit: bundle regenerates to the same 57 entries, and `bun test packages/server/src/mcp/skills/skills_bundle.spec.ts` — 12 pass, 0 fail.

---

## Also here: `malloy-debug` could not get diagnostics on a host with no editor

Its "Get Diagnostics" section offered three branches, one per host, and hardcoded `mcp__ide__getDiagnostics` and `ReadLints`. Both halves of that break the shared-skill authoring rules, which require bare tool names and forbid hardcoding a host's prefix.

The practical cost is worse than the rule violation. The three branches read as exhaustive, so an agent on a surface with neither tool matches the third and follows its instruction: *"Ask the user to open the file in VS Code with the Malloy extension and report the errors."* On a hosted chat surface there is no local checkout to open, so a compile error has no path to resolution and the agent hunts for two tools that are not on its surface.

Replaced with a host-neutral ladder that always terminates in something available:

> 1. **An editor-diagnostics tool**, if your host offers one.
> 2. **Otherwise, compile by running.** Run any query against the source with your query tool and read the error it returns. Every host that can run Malloy can do this, including chat surfaces with no editor.

The "ask the user to open an editor" advice stays, but conditioned on actually knowing they have one. `malloy-review` also builds on `mcp__ide__getDiagnostics`, but guards every call with an availability check and degrades to its LLM rubric pass, so it is correct as written and untouched.

It rides this PR's existing 0.1.12 bump rather than adding a second one, and the bundle was regenerated over all three edits.

Mirrored to [ms2data/agent-skills#105](https://github.com/ms2data/agent-skills/pull/105), where a second commit also names the host-specific answer for that repo's own surface — a Credible-only tool that the shared rules keep out of `malloy-debug`.

Re-run after this edit: bundle regenerates to the same 57 entries, and `bun test packages/server/src/mcp/skills/` — 18 pass, 0 fail.

---

## Also here: the layer rule and the sibling rule read as one rule

Review on this PR caught a real ambiguity in the section the first commit added. The gate paragraph now carries two rules that pull in opposite directions and sat one paragraph apart:

- window and fallback resolve **independently per layer**, so a package-level `stale_ok` defeats a window set on the source, and
- the **tightest window across siblings** sharing a content address governs all of them.

Read together, an author could reasonably conclude a sibling's `stale_ok` voids their bound the same way a package-level one does. It does not — the fold keeps whichever fallback bounds staleness. Each rule now says which of the two it is about, so the layer rule can no longer be carried over to siblings.

The section also promised more than the annotation can deliver. Both behaviors are properties of the **host** that assembles the manifest, not of `#@ persist`: a host that does not fold leaves it unspecified which sibling's policy reaches the wire, and one that folds at manifest-assembly time applies it when a version's manifest is next published rather than retroactively to manifests already distributed. So an author can declare the window correctly and not have it in force yet. That is now stated.

Mirrored to [ms2data/agent-skills#105](https://github.com/ms2data/agent-skills/pull/105) verbatim.

Re-run after this edit: bundle regenerates to the same 57 entries, and `bun test packages/server/src/mcp/skills/skills_bundle.spec.ts` — 12 pass, 0 fail. No em-dashes added; `diff` against the agent-skills copy still reports only the license header and the one deliberate `queryMetadata` rewording.


housejester:
Reviewed at `cfe2c37`. All three edits are the right call — the `malloy-debug` ladder in particular, since the old three-branch form read as exhaustive and its third branch told an agent on a surface with no checkout to ask the user to open one. Two notes on this diff and one on a claim it rests on.

**The verbatim-parity claim does not hold for `malloy-debug`, and this PR is the moment to close it.** I diffed both copies of all three files. `malloy-materialization` is exactly as described — the new paragraph is byte-identical and the only divergence is the deployment-agnostic `queryMetadata` wording. `malloy-notebook-chat` differs only by the license header. But `malloy-debug` carries three content divergences that predate this PR, with the downstream copy the stale side:

| This copy | The downstream copy |
|---|---|
| `` `unexpected 'from'` `` → "`from()` was removed from the language. Use the query directly: `source: x is q extend {...}`" | `` `from()` errors `` → "Verify the source query returns the expected columns" |
| `Query-based source "Can't find field"` | `` `from()` "Can't find field" `` |
| An `IO Error: No files found that match the pattern "data/x.csv"` row (DuckDB relative `workingDirectory`) | absent |

The first one has teeth, and it is the same failure shape this PR is fixing in that very file: an agent hitting a `from()` error is sent to verify the source query's columns, when the real answer is that `from()` no longer exists in the language. Since both copies of this file are already being edited, syncing the three rows costs nothing and is what the shared-skill contract is for.

**The new materialization paragraph teaches the window fold but leaves the fallback unstated, and the sentence above it now reads wrong for the shared-address case.** The paragraph's claims about the window are right — tightest governs, an absent declaration cannot loosen a sibling's, and a lapse pulls the sibling's reads off the table. But two lines earlier the bullet still says *"under `stale_ok` a stale artifact keeps being served, which voids the bound."* That remains true across **layers**, which do not combine. It is not true across **siblings** on a shared address wherever the host folds the policy — a fold that keeps the tightest window keeps the bounding fallback by the same logic, so a sibling's `stale_ok` cannot void your bound. A reader who has just learned that a sibling shares their window will assume the sibling's fallback reaches them too, and the truth is the reassuring direction. One clause fixes it:

> A sibling's `stale_ok` cannot void your bound either — the fold keeps whichever fallback bounds staleness.

**The rule depends on the host, and the paragraph states it unconditionally.** "The tightest window any of them declares governs all of them" is a property of the host that assembles the manifest, not of the annotation. On a host that has not adopted the fold, which policy reaches the wire is unspecified. It is also worth knowing that a host implementing the fold at manifest-assembly time will typically apply it when a version's manifest is next published, not retroactively to manifests already distributed — so an author can declare the sibling window correctly and still not have it in force yet. Both are worth a half-sentence, because a skill that asserts a guarantee the reader's deployment may not provide is the harder failure to debug: the author does everything the skill says and the bound is still absent.

Nothing here blocks. The `given:` generalization in `malloy-notebook-chat` is a clean fix — naming the mechanism ("a new runtime parameter") rather than the current spelling is what stops it going stale the next time the spelling moves.

housejester (approved):

PR #1097, opened 2026-09-01
feat(server): converge malloy_getContext's response shape on Credible's

jswir:
Stacked on #1028 (targets its branch). Previously the two surfaces shared no field name — Publisher a flat ranked `results[]` of entities, Credible a `sources[]` nesting its entities — so anything consuming both needed two parsers, and an eval harness written against one silently scored zero against the other.

Built as a serialization layer over the existing pipeline, so ranking, sibling grouping, alias collapse and the relevance floor are untouched and a shape bug stays separately revertible from a ranking bug.

Three judgment calls a reviewer should weigh:
- `entity_type` is a deliberate **superset** of Credible's `view | measure | dimension`, adding `join` and `query` — narrowing would delete the ability to retrieve a declared join, which exists so an agent stops concluding a model has none.
- Publisher-only fields ride along (`joins`, `entity_id`, `relationship`, `aliases`, `also_in`, response-level retrieval diagnostics).
- Credible fields Publisher cannot honestly fill are **omitted** rather than sent empty: `summary` needs an LLM summarizer, `prominence` needs usage telemetry, `values` needs a dimensional value index.

One behavioural change: a source's doc used to appear twice (full in the result row, truncated in the context block); it now appears once — full when the source itself matched, truncated when present only as a field's parent.

Testing: 10 new spec cases in `get_context_tool.spec.ts` / `server.protocol.spec.ts` pass; the 28 pre-existing failures on the base branch are unchanged (same 28 at base HEAD without this diff).


jswir:
Folded into #1028, which is now rebased onto `main` and carries the shape convergence as one of its twelve commits. Two things changed in the fold:

- The interface names and comments no longer name the hosted product; the response is described against the published retrieval shape it converges on.
- The shape now meets `main`'s drill-down behaviour from #956. A `sourceName` with no query returns that source as one card with every entity nested in it, rather than a card with nothing in it.

The stacked branch `jswir/get-context-credible-shape` is no longer needed.

PR #1098, opened 2026-09-01
test(hammer): fix the three red scenarios, and make a failed build distinguishable from a built one

housejester:
Hammer exits **1** on `main` with three failing scenarios that are not tagged
`known-red` — real failures by the harness's own contract. All three turn out to be
scenarios asserting contracts that have since moved, and each moved in a PR that
touched no scenario file. None is a regression, so none becomes `known-red`.

## The three

**`host-binding-honors-row-level-access`** (tagged `security`) — the security property
is intact. `Package.bindStorageServeBindings` still filters on a positive allowlist
keyed on `binding.sourceName` against `sourceEligibility.eligible`, computed over the
compiled sources independently of the build plan, so the forged binding is still
dropped. What moved is upstream of it: #1045 added `refusedSources` as a collection
separate from `sources`, because constructing a plan entry needs `getSQL()` and a
content address "which a free-parameter or given-referencing source cannot reliably
produce". The source therefore has no plan handle for a manifest to be keyed on,
which is what the scenario relied on. It now names the source `(unplanned)`, the way
a host working from its own records would.

**`failed-run-reclaims-its-tables`** — retitled, because the old title named a
mechanism rather than the guarantee. Since #1004 a part-way failure commits a
manifest recording both what built and why the rest did not, so the tables it wrote
are named and stay reachable to manifest-driven GC; reclaim is left for a run that
recorded nothing. The surviving table is correct, not stranded. The scenario now
asserts the guarantee ("leaves no table unreachable"), which survives either
mechanism, plus a total-failure leg proving a run that produces nothing still fails.

**`persisted-sql-select-colocated`** — the limitation this scenario pinned was fixed
upstream. Its own `## Note` asked whether the `query_source`/`sql_select` asymmetry
was deliberate; that became malloydata/malloy#3016, fixed by malloydata/malloy#3040
(`persistedTableFor`, called from both the FROM/join site and the `%{ }` segment
site), and adopted here by the routine `^0.0.432` bump in #1063. Confirmed by
version: `0.0.427` and `0.0.430` have no lookup, `0.0.432` does. The expectation
inverts, the `needs-attention` tag and the Note go, and the Note's publisher-side
suggestion (warn about a build nobody reads) is dropped — it would now warn about
something that works.

## Two harness changes, both needed to state the new contracts honestly

**`## Build (orchestrated)` could not tell a built source from a failed one.** It
asserted only `physicalTableName`, and since #1004 a failed source is mirrored into
`entries` carrying the name it was headed for — a table that does not exist. So the
step would have reported `failed-run-reclaims-its-tables`' failing source as built. A
per-source `(failed)` attribute now asserts against the manifest's `failures`, and an
unmarked source is asserted built **and** absent from `failures`, which closes the
same hole for the other twelve orchestrated scenarios.

**`## Build refusals`** asserts `BuildPlan.refusedSources`. With the storage tier's
refusals absent from `sources`, `host-binding-honors-row-level-access` and
`host-binding-of-unplanned-source` now reach the bind path by the same route, and
only a log line distinguished them. This asserts the difference where it actually
lives: the first pins examined-and-refused (`storage` / `given`), the second pins an
empty set — absent from both collections, because Malloy never admitted it as a
persist source. An empty table is the only way to state that side positively, which
is why this is a section rather than a column on `## Build targets`. That step was
also undocumented in `GRAMMAR.md`; its row is added alongside.

## Testing

Full suite, `bun hammer/run.ts`:

```
EXIT=0 — 77 PASS, 3 KRED, 0 FAIL, 0 MALFORMED, 0 FIXED
```

The 3 KRED are the pre-existing tagged debt (`extended-source-inherits-persist`,
`extended-source-inherits-persist-colocated`, `quoted-persist-name-colocated`),
unchanged by this PR. Re-run at `--workers 4` to check the low-core path: same
result, `EXIT=0`.

Before this branch, on the same checkout, those three scenarios fail with exactly the
signatures above.

## One thing worth a second opinion

While rewriting `failed-run-reclaims-its-tables` I could not construct a case that
reaches `reclaimStorageTablesFromFailedRun` at all. `builtThisRun` and `builtSources`
are appended together, so "every source failed" — the throw for a run producing
nothing — always implies there is nothing to reclaim. `isReclaimableStorageTable` is
`storageDestinationName && refresh === undefined`, so incremental entries are
retained by design. That leaves a throw escaping the per-source `try` after a source
has already built, and that `try` wraps `buildOneSource` entirely: a bad destination,
a failed CTAS and a missing schema are all captured as per-source failures.

What remains is `assertMaterializationEligible` on a later storage-targeted source —
reachable in production, where a host instructs by `sourceID` a source a model edit
has since made ineligible, but not expressible in Hammer, which resolves every
instructed source through the plan's `sourceEntityId`. So the reclaim is currently
pinned by unit tests only, and the guards it carries protect a path that almost never
runs. Recorded as the scenario's `## Note` rather than acted on here; it wants an
owner's call, not a test-only fix.


hammer/lib/scenario_md.ts
```diff
@@ -637,12 +658,43 @@ function parseMarkdown(text: string, fallbackId: string): ParsedMd {
                });
                break;
             }
+            // Tested before the `refused` prefix below: this asserts the plan's
+            // OTHER collection and runs nothing, where "## Build refused" runs a
+            // build and asserts it fails.
+            if (/^refusals\b/i.test(arg)) {
```
Sha-Bang, line 664: `scenario_md.spec.ts` pins Manifest `(unplanned)` with an accept case and a misspelling throw, for the reason at the top of that file: a typo that is simply not read is a check that never ran. `(failed)` has the same unknown-attr throw, and this `refusals` branch is deliberately parsed before `refused` because the two headers almost share a prefix. Neither has a spec.

A `## Build refused` written by someone who meant the plan collection would run a build and demand it fail — the comment here already says that. Worth the same two tests Manifest got?

hammer/lib/scenario_md.ts
```diff
@@ -163,6 +176,14 @@ type Step =
         givens?: Record<string, string>;
         exactColumns: boolean;
      }
+   | {
+        kind: "buildRefusals";
```
Sha-Bang, line 180: `buildRefusals` is a new assertion-bearing kind. Runtime `stepMustAssert` defaults to true for anything not in `SIDE_EFFECT_ONLY_STEPS`, so an empty Expect still has to contribute a check. The spec's pinned list of kinds that must assert does not include it, so the pin would not notice if it later got exempted. Add `buildRefusals` next to `buildTargets`?

Sha-Bang (approved):
The three red scenarios were asserting contracts that moved under them, and the restatements match the implementations: bind keyed on name against a positive eligible set, a part-way failure commits a manifest so the tables it wrote stay named, `sql_select` now reads its table. The harness hole is real — since #1004 a failed source is mirrored into `entries` under the name it was headed for, so a name-only check reads a failure as a success — and `(failed)` plus unmarked ⇒ absent from `failures` closes that in both directions.

Two nits inline, both one-test follow-ups. Approving.

On the owner’s call in `failed-run-reclaims-its-tables`: leave `reclaimStorageTablesFromFailedRun`. Hammer not reaching it does not mean production does not — a host instructing by `sourceID` after a model edit made the source ineligible is still a real path, and the still-referenced / orchestrated-only guards exist because of a real cross-environment drop. If we want it pinned, grow the step so it can instruct by `sourceID`; do not delete the guards because the harness cannot name them. Unit tests stay the backstop until then.

PR #1101, opened 2026-09-02
feat(skills): the model-evaluation loop -- eval-loop, eval-answer, eval-diagnose, eval-improve

jswir:
## What this is

The four skills that evaluate a semantic model, ported from ms2data/agent-skills#81 as shared skills, plus the scripts that make it a loop.

| Skill | Job |
|---|---|
| `eval-loop` | The procedure: baseline, noise band, diagnose, one edit, acceptance check, checkpoint. Conducts; does not restate the others. |
| `eval-answer` | Score one answer against a verified golden; score which of the entities the golden depends on retrieval delivered. |
| `eval-judge` | The judge itself — loaded as a skill, not pasted into a prompt. |
| `eval-diagnose` | Why a case failed and which artifact owns the fix: model, retrieval, agent skill, or dataset. |
| `eval-improve` | The smallest model edit for a diagnosed cluster, with probe receipts, never self-accepted. |

**Scripts** (`eval-loop/scripts`, `eval-answer/scripts`): `run_baseline.py` (a whole arm: verify goldens, blind answerer per case, contamination check, re-execute, judge, ledger), `verify_goldens.py`, `init_truth_package.py`, `serve.py`, `build_run_package.py` (a browsable run as a served Malloy package), `flip_table.py`, `check_judge.py`, `ledger.py`, `publisher_rest.py`, `agent_harness.py`.

**Stacked on #1088**: the manifest's `analysis` and `modeling` groups are what the loop installs for the agents it measures. Registered in `publisher-local.json` under a new `eval` group; symlinked in `.claude/skills`; indexed in `skills/malloy/SKILL.md`; a section in `skills/README.md`.

**Deliberately not here:** the engine-side evaluation of `get_context` itself. What a customer's run reports about retrieval is one number: of the entities each golden depends on, how many the agent's own `get_context` calls delivered.

**Supersedes #1032**, which ported a five-week-old snapshot without the scripts.

## Review notes — what changed after the initial port

The port was reviewed and then exercised end to end against a real model. Both passes changed it substantially; the commits are individually reviewable.

**Cut** (~350 lines): `run_all.py`, an undocumented overnight orchestrator whose three stated justifications were already covered by the scripts, and `mcp_client.py`, its only caller. The loop ships one mode: an agent conducts it.

**One definition per concept.** One `claude -p` runner (the answerer was the only agent missing the empty-reply retry). One REST path to a Publisher — the two copies had drifted, and the shallow one handed the judge raw cell envelopes for any `nest:` result. One definition of a flip — the run package counted a hedged arm as a non-pass and reported 2 flips where the gate's own instrument reported 1.

**"Gate" is now "acceptance check."** The ledger kind `gate` becomes `acceptance_check`; `ledger.py` reads the old kind, so existing run directories still validate.

**The judge is a skill, not a file a script pastes.** Measured on a case that had been oscillating: pasted gave `match/no_match/match/match`, loaded as a skill gives `no_match ×4`, with the reasoning citing the rule by name. Costs ~2.5x per verdict.

**Progressive disclosure.** Every SKILL.md is now under the 5k-token guidance (2154–3738), with eleven reference files loaded on demand. This surfaced a bug worth knowing about: `build_workspace` symlinked skills into an agent's scratch workspace, so `SKILL.md` loaded but nothing beside it could be read — bundled references were unreachable. Copied now.

**Four pins, taken from the server where the server knows.** `packageSha` (the server's `sourceContentSha`, which covers every model path where the old `modelSha` covered one file), `servedRevision` (documented and hardcoded to `None` until now), `datasetSha`, and the existing `targetVersion`. Runs name themselves `<set>-<phase>-<nn>`.

**CodeQL:** one real high-severity finding fixed — the run package's drawer built a Malloy literal escaping quotes but not backslashes, so a qid containing one closed the literal early and the remainder reached the parser as code.

**Tests:** `bun run test:skills-python` now runs the skills' Python suites in CI. They existed and nothing ran them.

## Testing

```
packages/server:  bun test src/mcp/skills        18 pass
packages/skills:  bun run test                  174 pass
skills' Python:   bun run test:skills-python    5 suites, 72 tests
```

Exercised end to end against `malloy-samples/ecommerce` on a local Publisher: two servers (model + truth), goldens each derived two independent ways, blind answerers, judge, flip table, run package. Then again against a deliberately incomplete model, through diagnose and improve: the improver added the missing measure, a fresh blind answerer discovered it, and the acceptance check correctly refused to accept on a reproducible untargeted regression.

**Known gaps.** The acceptance check has never run to an *accept* — it has only ever correctly refused. And a refusal scored against a golden that holds a value is settled by case authoring (`golden.kind: unanswerable` while the model lacks the concept) rather than by anything enforceable in the judge; four prompt edits failed to make it enforceable, and that is recorded in `eval-judge` rather than papered over.

A set anyone can run this on: credibledata/malloy-samples#23 (`evals/ecommerce`, with its truth package).


skills/malloy/SKILL.md
```diff
@@ -50,6 +50,15 @@ Every skill in this deployment, by what it is for. Start at a driver; it routes
 | `skill:malloy-notebook-chat` | The chat is bound to a notebook or saved report; answer from its cells |
 | `skill:malloy-phrase-detection` | Turning a plain-English question into search targets for the context tool |
 
+**Evaluating a model** (driven by `skill:eval-loop`)
```
Sha-Bang: **Should-fix** - The root router now routes to `skill:eval-loop`, `skill:eval-answer`, `skill:eval-diagnose` and `skill:eval-improve`, but `malloy` is in the `modeling` group and all four eval skills are in the new `eval` group. This PR's own `skills/README.md` says "`analysis` and `modeling` groups are what the loop installs for the agents it measures" — so installing `modeling` gives an agent a router pointing at four skills that are not there.

The `eval` group itself is closed over its own `skill:` references; it is only the cross-group edge that is open. This widens the same closure gap #1088 left (`modeling` already leaked nine refs) rather than introducing it, so it may be a deliberate deferral — but the router is the one file where a dangling ref is guaranteed to be read.

Either add the eval skills to the group the router ships in, or gate that table on the `eval` group being installed.

.github/workflows/build.yml
```diff
@@ -101,6 +101,15 @@ jobs:
           set -euo pipefail
           bun run test:scripts
           node scripts/release-notes.mjs extract >/dev/null
+      # The skills ship Python that no other suite touches: the ledger contract,
```
Sha-Bang: **Nit** - The comment says the step covers "the ledger contract, the retrieval scorer, the contamination check, the REST parse." Three of those have tests; the ledger does not — `ledger.py` is 352 lines plus a 324-line `reference/ledger-schema.md`, and no `*_test.py` in the tree imports or references it.

The step itself is fine (it discovers `*_test.py`, so a ledger test would be picked up automatically). It is the comment that overstates what is covered, which is the kind of claim that stops someone noticing the gap.

Sha-Bang (commented):
**1 should-fix, 1 nit.**

- `skills/malloy/SKILL.md:53` — the router ships in the `modeling` group but routes to four `eval`-group skills, so installing `modeling` leaves the table pointing at skills that are not there.
- `.github/workflows/build.yml:104` — the comment claims the step covers the ledger contract; `ledger.py` has no test.

Sha-Bang (approved):
Approving. Two findings posted inline — the router's cross-group refs are the one worth closing before this lands widely; the CI comment is editorial.

Sha-Bang (commented):
Second pass on the same commit — a reviewer that had timed out on the first round came back. **2 blocking, 5 should-fix**, all opened and confirmed against the code. My approval stands; these are for you to weigh before merge rather than a gate.

Both blocking findings are in the measurement path, which is the part of an eval harness where a bug is most expensive: it does not fail loudly, it just moves the numbers.

- `skills/eval-answer/scripts/score_retrieval.py:346` — `TypeError` on total retrieval failure. Reproduced against this branch; the scorer dies on exactly the run you most need the report for.
- `skills/eval-loop/scripts/run_baseline.py:599` — named-query submissions (`queryName`) are not captured, so `submitted` reads false, `final_query` is null, and the judge is handed "(none)" for an attempt that did query.

The should-fix set: two export-shaping mismatches in `build_run_package.py` (:382, :300), a `modelPath` / `targetModelPath` key drift in `check_judge.py:90` that silently checks the wrong model, a golden audit in `improve.py:132` that reports "not applicable" instead of running, and the `contaminated` string-vs-bool drift against the documented ledger contract.

That last one is worth reading alongside the `ledger.py` coverage note from the earlier round — the drift is precisely what a contract test would have caught.

skills/eval-answer/scripts/score_retrieval.py
```diff
@@ -0,0 +1,357 @@
...
+
+    s = summarise(rows)
+    print()
+    print(f"attempts {s['attempts']}, failures {s['failures']}, "
+          f"retrieval scored {s['retrieval_scored']}")
+    if s["mean_recall"] is not None:
+        print(f"complete retrieval {s['complete_retrievals']} of "
+              f"{s['retrieval_scored']} (every required entity returned); "
+              f"mean recall {100 * s['mean_recall']:.1f}%, "
+              f"mean precision {100 * s['mean_precision']:.1f}%")
```
Sha-Bang: **Blocking** - `100 * s['mean_precision']` crashes with a `TypeError` on a reachable run state, and it is the worst one: total retrieval failure.

`summarise` builds `scored` from rows where `recall is not None`, so `mean_recall` is a float whenever anything was scored — but `mean_precision` filters to `precision is not None` first, and `score_case:222` sets `precision = None` whenever `got_set` is empty. A run in which every scored attempt returned zero entities therefore passes the `mean_recall is not None` guard on :343 with `mean_precision` still `None`.

Reproduced against this branch:

```
retrieval_scored: 1
mean_recall: 0.0
mean_precision: None
guard `mean_recall is not None` passes: True
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

So the scorer reports fine on a healthy run and dies on the run you most need the report for. `pct()` is already defined four lines above and does exactly the right thing (`"  -  "` for `None`) — using it here, or guarding the two means separately, closes it.

Worth a regression test for the all-precision-`None` case; `score_retrieval_test.py` passes today because nothing exercises it.

skills/eval-loop/scripts/run_baseline.py
```diff
@@ -0,0 +1,1465 @@
...
+                    # execute_query.
+                    if (name.endswith("malloy_getContext")
+                            or name.endswith("__get_context")):
+                        n_get += 1
+                        pending[c["id"]] = {"tool": "get_context",
+                                            "targets": search_terms(c["input"])}
+                    elif (name.endswith("malloy_executeQuery")
+                            or name.endswith("__execute_query")):
+                        n_exec += 1
+                        q = c["input"].get("query")
```
Sha-Bang: **Blocking** - Only the ad-hoc `query` payload is captured. `execute_query` also accepts a named view or query — `queryName` (+ `sourceName`) — and its own tool description actively steers the answerer there: "Takes either ad-hoc Malloy in query, or a named view/query via queryName" (execute_query_tool.ts:87). So this is a shape the measured agent is being told to produce.

When it does, `n_exec` increments but `queries` stays empty, and the consequences land on the evidence the loop is built to collect:

- `:660` — `"submitted": bool(queries)` records **false** for an attempt that did submit.
- `:666` — `final_query` is `None`.
- `:886` — the judge is handed `"(none)"` for the queries it is meant to verify the answer against.

A run answered entirely through named views scores as if the agent never queried. That biases the measurement silently and in the direction that matters — it makes retrieval look worse than it is.

Capturing `queryName`/`sourceName` (and the `modelPath` alongside, which is dropped on the same path) would fix it.

One correction to how this was originally reported: the call *is* tracked in `pending` — that assignment is unconditional, outside the `if q:`. The tool_result is processed normally; it is only the query text and model path that are lost.

skills/eval-loop/scripts/build_run_package.py
```diff
@@ -0,0 +1,486 @@
...
+                required_rows.append({
+                    "attempt_key": ak, "run_id": run_id, "qid": qid,
+                    "entity_id": eid,
+                    "status": delivery(eid, set(got), tokens)})
+
+            roles: list[tuple[str, str]] = []
+            roles += [(eid, "required") for eid in sorted(required)]
+            roles += [(eid, "returned") for eid in sorted(got)]
+            if r["recall"] is not None:
+                roles += [(eid, "missing") for eid in sorted(required - got.keys())]
```
Sha-Bang: **Should-fix** - `required - got.keys()` is raw set subtraction, but six lines above (:376) the per-entity status for the same entities is computed with `delivery(eid, set(got), tokens)`, which resolves the `alias` and `in_docs` routes.

An entity delivered under an alias therefore appears as `status: alias` in one table and `role: missing` in the other, in the same export. That inflates the missing counts and the backlog built from them, and it is the exact failure `summarise`'s own comment warns about elsewhere: "the two tables on any report built from this disagree."

Deriving the roles from `delivery(...)` too keeps the one definition.

skills/eval-loop/scripts/build_run_package.py
```diff
@@ -0,0 +1,486 @@
...
+        events = read_jsonl(rd / "events.jsonl")
+        verdicts = {key(e): e for e in events if e.get("kind") == "score"}
+        tool_calls: dict[tuple, list[dict[str, Any]]] = {}
+        for e in events:
+            if e.get("kind") == "tool_call":
+                tool_calls.setdefault(key(e), []).append(e)
+
+        for e in events:
+            if e.get("kind") == "issue":
+                issues.append({"run_id": run_id, **e})
```
Sha-Bang: **Should-fix** - `issues.append({"run_id": run_id, **e})` copies the raw ledger event, but `issues.csv` is written with the columns `run_id, qid, component, owner, summary, evidence, status, lever, severity` (:451). The event shape does not match:

- the ledger carries `qids` (a list of affected cases, ledger-schema.md:260), not `qid` — so that column comes out blank;
- `status` is not on the issue event at all. It lives in separate `issue_status` events, and the schema says "Readers take the latest event for that `issue_id`" (:271) — so that column is blank too.

The CSV is mostly empty where it should carry the issue backlog. Exploding `qids` to one row per case and joining the latest `issue_status` by `issue_id` is what the schema asks for.

skills/eval-loop/scripts/check_judge.py
```diff
@@ -0,0 +1,177 @@
...
+    fx_path = a.fixtures or (a.set_dir / "judge-regressions.jsonl")
+    if not fx_path.exists():
+        print(f"no fixture file at {fx_path}, so nothing asserts this judge is "
+              f"correct rather than merely repeatable. Seed one from the cases "
+              f"an A/A pair disagreed on.")
+        return 0
+
+    cfg = json.loads((a.set_dir / "set.json").read_text())
+    a.package = a.package or cfg.get("package") or "ecommerce"
+    a.model_path = a.model_path or cfg.get("modelPath") or f"{a.package}.malloy"
```
Sha-Bang: **Should-fix** - This reads `cfg.get("modelPath")`, but the set config key is `targetModelPath` — that is what `run_baseline.py:1172` reads and what `reference/ledger-schema.md:67` documents. `modelPath` appears nowhere else in the tree.

So the lookup always misses and falls through to `f"{a.package}.malloy"`, silently checking a different model than the set configured. There is a second, smaller divergence in the same expression: `run_baseline` defaults to `"model.malloy"`, this defaults to `f"{package}.malloy"`.

A judge check pointed at the wrong model file is weak in a way that does not announce itself, which is why this is worth more than a rename.

skills/eval-improve/scripts/improve.py
```diff
@@ -0,0 +1,415 @@
...
+A reply that ends in prose is a lost result. The edit still happened and the
+receipts -- probes, edit tier, golden-suspect list -- are gone, so the
+acceptance check has nothing to weigh.
+"""
+
+
+def verify_goldens(a: argparse.Namespace, d: pathlib.Path,
+                   diff: str) -> dict[str, Any]:
+    """Run the set's golden verification against the edited model, if it has one."""
+    script = a.set_dir / "verify_goldens.py"
```
Sha-Bang, line 139: **Should-fix** - The golden audit probes `a.set_dir / "verify_goldens.py"`, but the only `verify_goldens.py` in the tree is the shared one this PR adds at `skills/eval-answer/scripts/verify_goldens.py`. Unless a set ships its own copy, the probe misses and the function returns `{"ran": False, "why": "set has no verify_goldens.py"}`.

That is the shape worth flagging: an acceptance check that reports *not applicable* rather than failing. The improve loop proceeds believing the goldens were audited when nothing ran.

Invoking the shared verifier directly would make the skip a real signal again.

skills/eval-loop/scripts/run_baseline.py
```diff
@@ -0,0 +1,1465 @@
...
+                      # as "package revision actually queried" and left
+                      # None until now, so nothing could tell an attempt
+                      # answered before a reload from one answered after.
+                      servedRevision=served_identity.get("servedRevision"),
+                      n_get_context=att["n_get_context"],
+                      n_execute=att["n_execute"],
+                      n_execute_errors=att["n_execute_errors"],
+                      host_tool_uses=att["host_tool_uses"],
+                      reported_calls=att["n_get_context"] + att["n_execute"],
+                      contaminated=("true" if att.get("breaches") else "false"),
```
Sha-Bang: **Should-fix** - `contaminated` is written as the strings `"true"` / `"false"` (also :1385), but the ledger contract documents it as `bool or "unknown"` (ledger-schema.md:162, :216).

Today this holds together only by accident: `diagnose.py:340` compensates with `e.get("contaminated") == "true"`. Any consumer written to the documented contract — testing the field's truthiness, as a bool field invites — reads the string `"false"` as **truthy** and treats every clean attempt as contaminated. Since contamination nulls the verdict, that silently voids a whole run.

Writing real booleans and accepting both on read is the compatible fix.

This is the same `ledger.py` contract flagged separately as having no test coverage; the two reinforce each other — the drift is exactly what a contract test would have caught.

skills/malloy/SKILL.md
jswir: Confirmed the dangling refs, took a third option. Fixed in 0968f84d.

**Measured on this head**, `skill:` refs pointing outside their own group:

| Group | Skills | Out-of-group refs |
|---|---|---|
| `analysis` | 11 | 0 |
| `modeling` | 31 | 5, all from `skills/malloy/SKILL.md` |
| `eval` | 5 | 0 |

So the exposure is `modeling` only. The router is not in `analysis`, so an answerer never sees these rows. And the pre-existing leak is one ref (`malloy-notebook-chat`), not nine.

I did not take either fix you offered. Adding the eval skills to `modeling` defeats the split: `modeling` is what the loop installs into the agents it measures, so shipping the evaluator's doctrine to the thing being evaluated is a contamination path this PR spends a whole skill guarding. Gating the table needs conditional rendering of a static markdown file, which the manifest cannot express.

Instead I stated the contract in the file. `manifest.spec.ts:126` already exempts `malloy` on the reasoning that "a catalogue row is not an instruction to go read something" - but nothing in the file said so, and the column header reads "Use when...". So the reader had no way to apply the rule the test relies on. It now opens with:

> This table is a catalogue of what exists, not of what is loaded. A host that installs one group takes that group's skills alone [...] A row naming a skill from a group you did not install says that the skill exists. It is not an instruction to load it.

That covers `malloy-notebook-chat` too.

**One thing worth flagging, from #1088 rather than this PR.** That exemption is justified partly on "10+ skills reference it in turn, so it cannot be dropped from a group either" (`manifest.spec.ts:126`, repeated in `skills/README.md`). That is not true on this head: exactly one skill references the index, `malloy-modeling/SKILL.md:129`, and it is in the same group. I checked `skill:malloy` excluding the `malloy-*` prefixes, plus prose forms and the file path.

The other half of the rationale survives, and is what I leaned on. But the "cannot be dropped from a group" half is resting on a count that is off by an order of magnitude, which means the index is cheaper to move or duplicate than #1088 assumed. I left both files alone rather than silently rewriting your rationale. Happy to correct the count here or in a follow-up, your call.

`bun test packages/skills/src/manifest.spec.ts packages/server/src/mcp/skills` is green (189 pass), bundle regenerated.

jswir: Correction to my reply above, and a better fix. 13df5965.

**I named the wrong agent.** I said `modeling` is what the loop installs into the agents it measures. It is not. Tracing the harness: the blind answerer, the measured agent, defaults to `publisher-local#analysis` (`run_baseline.py:549`, `agent_harness.py:111`); it is the improver that loads `modeling` (`improve.py:276`). So the golden-leak concern I raised does not apply, and my "contamination path" argument was aimed at the wrong place.

I also claimed folding `eval` into `modeling` would not buy closure, because `malloy-notebook-chat` would still dangle. That was true but not load-bearing, since that ref is trivially removable. Both of my reasons were weak.

**The real reason `eval` stays separate** is narrower and survives. `improve.py:195` spawns the improver with `skills=["eval-improve", *role_skills]`, one eval skill prepended by name, and the docstring at `improve.py:18` says that is deliberate. Folding `eval` into `modeling` would silently also hand that agent `eval-judge`, `eval-answer`, `eval-loop` and `eval-diagnose`. The agent writing the model edit would hold the rubric its edit is scored against, and `improve.py:24` states as a hard boundary that an improver never accepts its own edit, a boundary currently enforced by the absence of `eval-loop` from its workspace.

**What I did instead of the caveat.** You were right that the exemption is the problem, so I removed it rather than documenting around it. The index now names cross-group skills as plain names instead of `skill:` references, which is already what `malloy-getting-started` does for `malloy-gotchas-modeling` per `skills/README.md`. That covers the four eval rows and `malloy-notebook-chat`. Measured after:

| Group | Out-of-group refs |
|---|---|
| `analysis` | 0 |
| `modeling` | 0 |
| `eval` | 0 |

So `if (member === "malloy") continue;` is gone and the test now covers the index. A future dangling row fails the build, which is the outcome your comment was asking for. "A catalogue row is not an instruction" is now true in the syntax rather than only in the rationale.

**Two corrections to #1088's text**, both found while checking this. The exemption was justified partly on "10+ skills point at it in turn"; the real count is one (`malloy-modeling/SKILL.md:129`). And `skills/README.md` said the `analysis` and `modeling` groups are "what the loop installs for the agents it measures", which is the sentence that misled me above. Both fixed in this commit.

`bun test packages/skills/src/manifest.spec.ts packages/server/src/mcp/skills`: 177 pass. Bundle regenerated. The 7 failures in `index.spec.ts` are pre-existing on a clean tree (it needs the packed `packages/skills/skills/`).

skills/eval-answer/scripts/score_retrieval.py
jswir: **Confirmed, reproduced.** Not just read: I executed `summarise` against this branch with one scored attempt whose retrieval returned zero entities.

```
retrieval_scored: 1
mean_recall: 0.0
mean_precision: None
guard `mean_recall is not None` passes: True
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

Your mechanism is exactly right. `summarise:254` builds `scored` on `recall is not None`, so `mean_recall` is a float whenever anything scored, but `:269` filters `precision is not None` before taking the mean, and `score_case:222` sets `precision = None` when `got_set` is empty. All-empty retrieval gives `mean([])` -> `None`, past a guard that only tested recall.

Taking the fix you propose: `pct()` on both, which already renders `None` as `"  -  "`. Plus the regression test for the all-precision-`None` case, since `score_retrieval_test.py` is green today precisely because nothing constructs it.

skills/eval-loop/scripts/run_baseline.py
jswir: **Confirmed.** `:598` is `q = c["input"].get("query")` and everything downstream sits inside `if q:`, so the `queryName`/`sourceName` shape is dropped whole, and `modelPath` with it on the same path.

I checked the steer you cite. `execute_query_tool.ts:87` reads verbatim: "Takes either ad-hoc Malloy in query, or a named view/query via queryName (with sourceName for a view)." So the tool description does actively point the answerer at the shape the harness cannot see.

Confirmed the three consequences at the lines you name: `:660` `"submitted": bool(queries)`, `:666` `final_query`, `:886` the judge's `"(none)"`. The direction matters and you named it: a named-view answer scores as if the agent never queried, which biases retrieval *downward*. An eval that under-reports its own subject is the worst failure shape here.

Thanks for the correction on `pending` too. Confirmed: `pending[c["id"]] = {...}` at `:610` is outside the `if q:`, so the call is tracked and only the query text and model path are lost. Fixing by capturing `queryName`/`sourceName`/`modelPath` on that branch.

skills/eval-loop/scripts/run_baseline.py
jswir: **Confirmed, and the truthiness trap is real.** `run_baseline.py:1354` and `:1385` both write `"true"` / `"false"`; `ledger-schema.md:162` and `:216` both document `bool or "unknown"`.

`bool("false")` is `True` in Python, so a consumer written to the documented contract reads every clean attempt as contaminated. And since contamination nulls the verdict, the whole run voids. Silent, and in the direction that destroys the data rather than merely biasing it.

Confirmed `diagnose.py:340` is the only thing holding it together, with `e.get("contaminated") == "true"`. That is the accident you describe: the one consumer that happens to test string equality rather than truth.

Taking your compatible fix, write real booleans and accept both on read. And you are right that this pairs with the `ledger.py` coverage note. Same contract, and a contract test over the writer would have caught this exact drift. I am treating the two as one piece of work.

skills/eval-loop/scripts/build_run_package.py
jswir: **Confirmed.** `:376` computes `status` with `delivery(eid, set(got), tokens)`, and `delivery` (`score_retrieval.py:145`) resolves four routes: `exact`, `alias`, `in_docs`, `missing`. `:382` then computes the role with `required - got.keys()`, plain set subtraction, which knows only two.

So an entity delivered under an alias is `status: alias` in `required.csv` and `role: missing` in `entities.csv`, from the same run, in the same export. Same for `in_docs`.

You are right that it inflates the missing counts and anything built on them. And the comment you quote is in this tree at `summarise:265`, warning about exactly this failure for a different pair of numbers, which makes it a fair standard to hold this to. Deriving both from `delivery(...)` so there is one definition.

skills/eval-loop/scripts/build_run_package.py
jswir: **Confirmed, both columns.** `:300` copies the raw event, and the writer at `:451` asks for `run_id, qid, component, owner, summary, evidence, status, lever, severity`.

Checked against the schema:

- `ledger-schema.md:260` is `| qids | list | Affected cases. |`. The event has no `qid`, so that column is blank.
- `issue_status` is a separate event kind (`:269`), and `:272` says "Readers take the latest event for that `issue_id`." There is no `status` on the `issue` event at all, so that column is blank too.

Which leaves `issues.csv` missing the case linkage and the open/closed state, i.e. most of what makes it a backlog. Taking your fix: explode `qids` to one row per case, and join the latest `issue_status` by `issue_id`.

skills/eval-loop/scripts/check_judge.py
jswir: **Confirmed.** `check_judge.py:90` reads `cfg.get("modelPath")` out of `set.json`, but the documented key is `targetModelPath` (`ledger-schema.md:67`, "Model path within the package"), and that is what `run_baseline.py:1172` reads. So the lookup always misses and silently checks a different model than the set configured.

Confirmed the second divergence too: `run_baseline` falls back to `"model.malloy"`, this falls back to `f"{package}.malloy"`. Two defaults for one concept.

One small correction to the report, which does not change the finding. `modelPath` is not absent from the tree; it exists as a *per-case* key (`run_baseline.py:831`, `:851`) and as a `run` event field (`ledger.py:117`). What is true, and is the actual bug, is that it is not a `set.json` key. That makes the typo a bit more dangerous than a bare one, since the name is real elsewhere and reads as correct.

Agreed on why it is worth more than a rename: a judge fixture check pointed at the wrong model still produces verdicts, so nothing announces the miss.

skills/eval-improve/scripts/improve.py
jswir: **Confirmed.** `improve.py:132` probes `a.set_dir / "verify_goldens.py"`, and `find . -name verify_goldens.py` across the tree returns exactly one path: `skills/eval-answer/scripts/verify_goldens.py`, the shared one this PR adds. No set ships a copy, so the probe misses and the function returns `{"ran": False, "why": "set has no verify_goldens.py"}` every time.

The shape is the part worth flagging, and you put it well: a check that reports *not applicable* rather than failing. `{"ran": False}` reads as "nothing to do here", so the improve loop proceeds believing the goldens survived the edit when nothing looked. That is worse than a hard error, because it is indistinguishable from a legitimate skip.

It also lands on the one thing this skill most needs to be true. `improve.py:379` notes an improver editing its own answer key removes the only independent check on the edit, and this is the check that was supposed to notice. Fixing by invoking the shared verifier directly, so a skip becomes a real signal again.

.github/workflows/build.yml
jswir: **Confirmed, and the comment is the part that is wrong.** `build.yml:104` claims the step covers "the ledger contract, the retrieval scorer, the contamination check, the REST parse." The tree has six `*_test.py`:

```
eval-answer/scripts/check_contamination_test.py
eval-answer/scripts/init_truth_package_test.py
eval-answer/scripts/mcp_payload_test.py
eval-answer/scripts/publisher_rest_test.py
eval-answer/scripts/score_retrieval_test.py
eval-loop/scripts/platform_target_test.py
```

Three of the four claims land. The ledger does not: no test imports `ledger.py`. Grepping for it hits only two *test names* that contain the word ("test_ledger_read_is_contamination", "test_the_cli_reads_a_ledger_and_attributes_each_attempt"), neither of which touches the module.

Agreed the step itself is fine, and that the comment is the problem: a claim of coverage is what stops the next person looking for it. This is not really a nit given where it landed. The `contaminated` string-vs-bool drift you filed separately is a writer-side violation of that same unverified contract, so the false claim of coverage sits directly upstream of a real bug. Fixing both together: correct the comment, and add the contract test that makes it true.

skills/eval-answer/scripts/score_retrieval.py
jswir: **Fixed** in df2156ec. Each mean is now guarded on its own, and precision renders `n/a` rather than reaching `100 * None`.

Two regression tests in `score_retrieval_test.py`, both of which fail against the old code: one pinning that `summarise` returns a float recall with a `None` precision for this shape, and one running the CLI end to end on a single zero-entity attempt and asserting exit 0 with `mean precision n/a` in stdout.

skills/eval-loop/scripts/run_baseline.py
jswir: **Fixed** in df2156ec. New `named_query()` renders the named form as the Malloy it stands for: `run: source -> view` for a view on a source, `run: name` for a model-level query. The capture is now `input.get("query") or named_query(input)`.

Rendering rather than storing it structurally was deliberate. The harness stores query TEXT, and `final_query` has to stay re-executable and readable by the judge, so one shape for every consumer means `submitted`, `final_query` and the judge prompt all follow with no further changes.

`modelPath` comes along for free, since it was dropped inside the same `if q:` you identified.

skills/eval-loop/scripts/run_baseline.py
jswir: **Fixed** in df2156ec, and the fix closes the class rather than the instance.

Both writers now pass real bools. `ledger.is_contaminated()` is the single reader, accepting bool and the legacy strings; `diagnose.py:340` calls it instead of `== "true"`. `"unknown"` reads as contaminated rather than clean, since no host log means contamination could not be ruled out.

Then the part that makes it stick: `_check` now type-checks the field, so `ledger.event()` raises on the strings. I ran the original expression through it verbatim:

```
CAUGHT: attempt: legacy shape, contaminated='false'; expected bool or "unknown"
```

Compatibility went the way you suggested. `validate_run` re-runs `_check` over stored events and turns problems into errors, so a hard check would have failed every historical run. The legacy shape is grandfathered to a warning on read, matching how unknown fields are already handled, and rejected only on write, where the drift starts.

skills/eval-loop/scripts/build_run_package.py
jswir: **Fixed** in df2156ec. The `missing` role is now `delivery(eid, set(got), tokens) == "missing"` over `sorted(required)`, the same call the `status` column uses six lines above. One definition, so an alias or in_docs delivery can no longer be `alias` in one table and `missing` in the other.

skills/eval-loop/scripts/build_run_package.py
jswir: **Fixed** in df2156ec. `qids` is exploded to one row per case, and `status` comes from a `latest_status` map built off the `issue_status` events, latest wins per `issue_id`, as the schema specifies.

One case your comment did not raise and I had to decide: an issue with no `qids` at all. It now emits a single row with a null `qid` rather than vanishing from the CSV, since disappearing from a backlog is the worse failure.

skills/eval-loop/scripts/check_judge.py
jswir: **Fixed** in df2156ec. Reads `targetModelPath`, and the default is `"model.malloy"` to match `run_baseline.py:1172` rather than `f"{package}.malloy"`. The comment records why the wrong name was dangerous specifically because `modelPath` is a real key elsewhere.

skills/eval-improve/scripts/improve.py
jswir: **Fixed** in df2156ec. Falls back to the shared `eval-answer/scripts/verify_goldens.py`, resolved relative to `__file__`. A set shipping its own copy still takes precedence, so the override you would expect still works, and `{"ran": False}` now means the verifier is genuinely absent instead of meaning nothing looked. The stale docstring saying "if it has one" went with it.

.github/workflows/build.yml
jswir: **Fixed** in df2156ec, and not treated as editorial.

`ledger_test.py` is new: 15 tests over the event contract, covering the `contaminated` type in both directions, the grandfathering of stored runs, missing-required still being an error, duplicate scores, unknown fields and kinds, and `run_config` identity. The comment now says what it actually covers.

You were right that the two findings reinforce each other, and it turned out to be load-bearing: writing the test meant type-checking the field, which is what makes the string drift impossible to reintroduce. The uncovered contract and the bug in it were one piece of work.

Full suite: 7 files, 98 tests, green. The step discovered the new file without editing the workflow, as its comment promised.

jswir:
Thanks for running it end to end. Follow-up is #1114; the mirror is ms2data/agent-skills#81 and the set changes are credibledata/malloy-samples#23.

**First, the thing that cost you time.** Every line number in your review matches the `ms2data/agent-skills` copy of the scripts, not this repo's. The malloy-samples runbook sent you there, and that mirror was behind by 393 lines in `run_baseline.py` alone. Three of your findings were already fixed here: named-view capture, `contaminated` as a bool with write-time type validation, and `servedRevision` from the server. I have made Publisher the runbook's target and said why, so the next person does not hit it.

**Named views.** Fixed on main before your run: `named_query()` renders `sourceName` + `queryName` as `run: <source> -> <view>`, which re-executes and reads like any other query, so every consumer sees one shape. I replayed your saved transcripts through it. Of your five null-query cases, `ecom_yoy` now records `run: order_items -> sales_summary_yoy` and `ecom_cumulative_4m_crossing_2022` records its view. The other three (`ecom_2022_sales_bookings`, `ecom_cart_abandonment`, `ecom_roas`) contain **zero** `execute_query` calls, so `submitted: false` is right for them. Three answerers answered a numeric question from `get_context` alone, which is a finding about the agent rather than the harness.

**Your question, and it is the better half of that point.** The gate stays on answer text, and `submitted` is not what decides. `not_submitted` now means neither prose nor a query, so the three above get judged instead of being labelled unscorable while receiving a verdict anyway. Against a golden that holds a value, an answer containing none of it is `no_match` however well it reasons. Gating on `submitted` would have dropped a confident refusal on an answerable case out of the pass rate, which is the one thing the four answerable-sounds-unanswerable cases exist to measure. `ledger-schema.md` now says this.

**`mustNotUse`** was an omission, not a decision. `check_must_not_use.py` vetoes an identifier or a full field path found in the final query, forcing `no_match` and keeping the judge's verdict beside it as `judge_verdict` so a veto is auditable. It vetoes only what text can prove: prose entries and a path's bare leaf go to the judge, because a veto that fires on a correct answer is worse than one that misses. On your set that is 11 mechanically checkable entries of 41.

**Which retriever answered** is now recorded per call from `get_context`'s own `retrieval` field, summarised as `retrievalMode`, and `flip_table.py` exits 2 on a mismatched pair. Your run would have printed `unreported`, which is the honest answer to what you had to infer.

**The noise band: yes please, and after #1114 lands.** `flip_table.py --calibration` writes the block into the set's `CALIBRATION.md`, which now exists and specifies what it holds. Two reasons to wait rather than measure now: #1114 changes what gets re-executed and adds the retrieval gate, and the block records the pins a band is a property of, so one measured on the old harness could not be matched to any later run anyway.

**Judge family: keeping Sonnet, dropping the rule.** The design record and the code disagreed and whoever ran it decided by accident. A larger judge on every case of every arm did not justify itself against the control we actually have, which is the human-settled fixture file rather than the model pairing, and every run records `judgeModel` beside `answererModel` so the pairing is visible. The design record now says so with the date. Revisit if a fixture set fails in a way that tracks the pairing rather than the rubric.

**Fixtures.** `check_judge.py` now names the unpinned ones, skips a stale-pinned one rather than failing it (the judge did not regress, the key moved), and warns for each of five decision classes nothing covers: `required_disclosure`, `refusal_correct`, `refusal_wrong`, `gold_status`, `near_match_boundary`. On the deliberate break: no, it has not been done. `--skills-root` makes it one flag against a copied tree, and `reference/checking-the-judge.md` has the drill. I did not pin your 11: the whole set landed in one commit and they were settled 2026-08-31 against judge v3 with several goldens revised afterwards, so stamping today's revision would claim a check nobody performed.

**Two of the nits are not defects, and one of them nearly cost us a bug.**

`contaminated = 'true'` in `eval_run.malloy` is deliberate. `build_run_package` writes a bool as lowercase `true`/`false`; DuckDB sniffs a populated column as BOOLEAN and a header-only one as VARCHAR. Verified against the Malloy compiler and DuckDB: the string form compiles and evaluates correctly under both, and a bare `where: contaminated` fails as `non-boolean-filter` on the VARCHAR case, which is the empty-run path `write_csv`'s docstring exists to protect. So "the template has already adapted to the bug" is the wrong mechanism, and the fix would have broken it. It is now a comment. The real bug was next door: `app.js` tested `=== 'true'` alone, so the contamination chip stopped rendering the moment the column sniffed boolean.

On the doctrine: `eval-loop/SKILL.md` forbids a *sequencing* script, and its own line 13 says the scripts run the loop. `run_all.py` exists only in the mirror and is exactly what it forbids, so it is deleted there rather than documented here. The wording now says "no script decides what to do next" plainly.

**The rest.** `final_query` prefers the query the answer declared, then the last one the server answered, and records which as `final_query_source`, so a trailing probe is visible rather than silently re-executed. `reExecution` counts attempted/ok/failed/noQuery beside the one-bool `predictionsReExecuted`. `modelGitSha` comes from `--model-repo` and is null without it. `host_tool_uses` counts every tool use, with the MCP subset as `mcp_tool_uses`, so the under-report rule stops being arithmetically true, and `validate_run` applies it. A `near_match` stable across both arms reaches diagnose via `--verdicts`; a one-armed one still does not, deliberately.

Two things your run surfaced that nobody had reported. A run reporting no token usage, which every platform run does, failed the whole run package compile on an uncast `sum()`. And `platform_target_test.py` asserted the presence of a Publisher-only skill, so it passed here and failed in the very checkout you were running from.

Issue #1001, opened 2026-08-13
storage= serve bindings rebind by source name, so a same-named source in another model can be served from the wrong table


housejester:
A `storage=` serve binding identifies its source by NAME (`ServeBinding.sourceName`, "the Malloy
source name to rebind"), and `Package.pushStorageServeBindingsToModels` pushes every binding to
every model in the package. Nothing checks that the model's source of that name is the source that
was actually materialized.

Source names are not unique in a package — the build plan is keyed by `sourceID` for exactly that
reason — so two models may each declare `daily`. If one is persisted and the other is an unrelated
source that happens to share column names, queries against the unrelated one are answered from the
persisted table.

## The mechanism is already demonstrated in the suite

`model_storage_serve.spec.ts` is built on it. Its model's own source is
`source: X is duckdb.sql("SELECT 0 AS total")`; the binding names `X` and points at `mz_real`
(`total = 60`); the first test asserts **60**. The two are semantically unrelated — the harness
exists to prove routing works, and it simultaneously shows that rebinding is name-only, with no
check of provenance.

The one thing standing between an unrelated same-named source and the table is
`narrowSchemaToPublic`, which **intersects** the captured schema with the bound source's public
fields rather than validating that they describe the same relation. Disjoint columns ⇒ empty shape ⇒
live fallback (fine). **Partial overlap ⇒ the overlapping columns are served from the wrong table.**
So this is not fail-safe; it is guarded only by shape overlap.

## Scope

Pre-existing, and in mechanism rather than merely in kind: the seam has always been name-keyed and
package-wide. What #1000 changed is *which* names get bound — it binds every source sharing a
materialized table's content address, so alias names (an `extend` of a persisted source, which the
author never annotated) are now eligible to be rebound where previously only the builder's name was.
The surface grew; the mechanism did not.

Worth stating plainly because #1000's alias-grouping rules cannot reach this: they are computed from
`buildPlan.sources`, which contains persist sources only, so a shadowed non-persist source is
structurally outside their input.

## The fix that does NOT work

Scoping a binding to the model that declared its source. This breaks imports: a persisted source is
legitimately queried through an **importing** model (`cross-model-dag`,
`imported-persist-from-flagless-model`), so being served through a model that did not declare you is
a tested, correct case. Path scoping would regress it.

## Proposed fix — match on `sourceID`

Malloy already maintains the identity that separates the two cases. `sourceID` is
[...]

PR #1103, opened 2026-09-02
feat(preaggregation): build and serve a rollup from a storage destination

housejester:
A `#@ preaggregate` line may now name a storage destination. The rollup is built into that
destination and served from it; the query is unchanged, still naming the base source and
still knowing no rollup exists.

```malloy
source: orders is orders_pg.table('public.orders') extend {
  measure:
    #@ preaggregate grain="category" storage=lake
    total is amount.sum()
}
```

`storage=` on a `#@ preaggregate` line previously parsed, passed validation, and did
nothing: the reader took only `grain` and `namespace`, nothing rejected the unknown key,
and the rollup was built alongside its base. Nothing documented the key, so no advertised
path reached it — which is why support and the refusals for what remains unsupported are
one change.

## Authoring

**The destination is written on the line, never inherited.** A `storage=` base lends its
rollups nothing, as before. Inheriting reads as obviously right — a rollup of X belongs
where X's rows live — and cannot work, exhaustively, because inheritance would require
`#@ persist` on the base:

- **A query-shaped base** builds a stored table of its own, and rollups are served by
  rebinding the base's name to a composite of them. That table's binding already claims
  the name, so its rollups could never take it: each would be built, refreshed, and never
  read.
- **A base that is not query-shaped** — a table extended with measures, the ordinary case
  — parses the annotation, and the build refuses the whole run for a `#@ persist` source
  dropped from the plan.

A `name=` namespace still inherits, and the asymmetry is not about inheritance: colocated
rollups compose members Publisher names itself, with the base under an import alias, so
nothing there is keyed on the author's source name.

**Refused, each with the measure and the fix:**

- `namespace=` with `storage=` on one line. Placement inside a destination is derived, not
  authored: a freshly provisioned catalog has no schema to create the table in.
- Two measures at one grain naming different destinations. One grain is one table.

**A hidden field WARNS rather than refusing**, at publish and at load alike, and appears in
the package's `warnings`. `#@ preaggregate` on a measure the source does not publicly
expose, or a grain naming a hidden dimension, builds nothing: a rollup stores its grain and
each measure's partial and is served under the base's name with none of the source's field
visibility applying, so the planner refuses to plan one at all. It warns rather than
refusing for two reasons, neither being that it matters less. Such a package published and
loaded before the rule existed, and an upgrade must not break either. And the refusal was
unfollowable — an annotation inherited onto a source that then hides the measure raises it,
while that source produces no plan and exposes nothing, so the advice to expose the field
or drop the annotation asked an author to give up either the hiding or the rollup over a
model that was already safe.
- `#@ preaggregate` on a measure the source does not publicly expose, or a grain naming a
  hidden dimension. A rollup stores its grain and each measure's partial, and the stored
  table is served under the base's name with none of the source's field visibility
  applying to it, so pre-aggregating a hidden field would publish it.

Warning there is affordable because the refusal is not the enforcement: the planner refuses
to plan a hidden field at all, so no table is built and nothing can be served whether or
not anyone reads the message. For every other violation the refusal IS the enforcement, so
those stay fatal at both gates.

A rollup's generated name is a single unqualified identifier. It is also the physical
table name — the build self-assigns it, there being no `name=` — so a dot in it would read
as schema qualification downstream.

## Serving

Rollups are offered through a composite of themselves on the destination's connection,
re-exposed under the base's author name. There is no base member — the base lives on the
source warehouse and a composite cannot span connections — so a query no rollup covers
does not compile against it and is answered live, with the answer it always had.

**Two grains on one base naming different destinations is not refused at publish — it is
dropped at serve.** Across grains a `namespace=` legitimately splits across two schemas, so
nothing at publish has grounds to refuse the same for `storage=`: two grains are two
tables. But a base's rollups are offered through ONE composite and every member of a
composite must live on one connection, so such a base serves from its rollups not at all.
Its queries are answered from the base.

A rollup's virtual source is shaped from its plan rather than from the author's model,
which has no source of that name. Its stored columns are used as captured: not a
relaxation of the serve shape's narrowing rule but a consequence of it, since a rollup's
stored columns are its whole surface and there is no wider table to hide part of.

Where several rollups cover one query the **coarsest** is used. This changes colocated
routing too, deliberately: members were ordered by generated name, so with `grain="b"` and
`grain="a, b"` a query grouping by `b` alone read the `a, b` table. Ordering only the lake
path would leave two mechanisms to drift apart, and the docs' existing non-promise about
ordering is what makes this a fix rather than a breaking change. Grain dimensions are
counted, not measured, so it is a proxy for size; `rowCount` would be accurate and is
deliberately unused, because member order would then depend on the bound manifest and the
build and serve legs would synthesize different text.

One consequence, documented: a rollup is offered whether or not it has been built, so
adding a coarse grain to a package that already has a built finer rollup costs
acceleration until the new one builds. Answers are unaffected and it lasts one build.

## A `storage=` source is never built into its own warehouse

Independent of pre-aggregation, and it applies to any `#@ persist storage=` source.

Every guard on the instructed build path tested what the CALLER asked for
(`instruction.destination`), and none tested what the source DECLARED. A host that
resolved no destination — the tier off, the organization not enabled for one, the field
simply omitted — left the source to be CTAS'd into its own warehouse.

Declaring `storage=` is a statement about where the data may be written and is not consent
to that, and the read-only credential a server normally holds for a source warehouse would
fail the write anyway, on an error naming neither the destination nor the annotation.
Refused now, independent of *why* no destination was available, since none of those
reasons makes the warehouse the right answer.

The refusal is scoped to that source rather than the run. A host that resolved no
destination for one source has partly resolved its list — the org not enabled, the field
omitted — so failing the run would take down colocated siblings that have no destination
to resolve and nothing to do with the refusal.

## Two properties the tests exist for

**Every failure mode in this feature is silent.** A rollup that cannot be used falls back
to the base and returns the right answer — correct answers, no acceleration, no error
anywhere. That is what makes the tier safe and also what makes every defect in it
invisible, so the coverage is heavier than the surface area suggests and the scenarios
assert placement and staleness rather than values.

**A bad rollup binding must not cost the package its storage serving.** The serve shape is
built once per package, so a rollup that cannot be satisfied would otherwise take
unrelated stored sources with it. Two mechanisms hold that line: a base whose name is
already claimed has its rollups dropped rather than emitted, and the shape's escalation
ladder has a final tier that drops a pre-aggregation group whole, restoring base-only as
the guaranteed floor it is documented as. Both are pinned, the second by asserting the
un-droppable shape fails to compile first, so the test cannot pass vacuously.

## A design note for anyone extending this

`ServeBinding.sourceName` means two different things depending on `origin`: an author's
source name, or a generated rollup name that appears in no model file. Every consumer that
resolves a binding back into the author's model — by name, by field list, or by author
intent, since a declared `freshnessFallback` is author intent — needs an explicit decision
about rollups. Every consumer that reads the binding as an opaque
`(virtualHandle, tablePath, destinationName)` triple does not.

`deriveServeBindings` is the single place `origin` enters the system, which makes it the
one place a future field on `ServeBinding` needs that decision.

## Observability

The routing counter gains an `origin` attribute so rollup hits are separable from ordinary
storage hits. It makes the total splittable, not smaller: a consumer summing by `outcome`
alone starts counting rollup hits and must group by `origin` to get what it had.
`servedFrom` is unchanged — a lake-served rollup reports `storage`, which is the tier it
came from — and `Package.storageServeBindings` entries carry `origin` with the same
"may name no source in any model file" caveat `ManifestEntry.sourceName` already had.

## Known gaps, deliberately not here

- **A rollup that is BOUND and never HIT is not surfaced.** The `origin` attribute is the
  numerator; the denominator needs its own design.
- **A rollup that can never be served is surfaced only in a worker log.** Both cases where
  a base's rollups are dropped at serve — the base already claiming its own name, and two
  grains split across destinations — are reported by one `info` line and nothing else: not
  on the manifest, the package status, `storageServeBindings`, or the routing counter. The
  table is still built and refreshed. Unlike the denominator above this is statically
  decidable at publish from the build plan, with no traffic needed, so it wants a different
  fix rather than the same one.
- **No product surface shows that anything is served from a storage destination**, for
  rollups or for plain `#@ persist storage=`. The column stops before the API.
- **Serve bindings are keyed by source name** (#1001). A generated rollup name is far less
  collision-prone than an author name, but the keying is unchanged.

## Testing

- `bun run test:unit` — 3401 pass, 0 fail. `bun run test:integration` — 410 pass, 0 fail.
- `bun hammer/run.ts` — 86 pass, 1 known-red (`quoted-persist-name-colocated`, unrelated).
- Six hammer scenarios: build-and-serve from the store (with the uncovered-query
  fall-through in the same scenario, so the covered query is stale and the uncovered one
  fresh after the same mutation — either alone would pass while the other was broken),
  mode-off, stale-falls-back, hidden-field-warns-and-loads, a rollup's namespace
  round-trip, and `storage=` never falling back to the warehouse.
- Two spikes kept as permanent tests: whether pre-aggregation works over virtual sources at
  all, including the documented promise that a coarser truncation of a stored time
  dimension routes to the same rollup; and that the compiler's artifact id equals ours for
  every source in every target, which is the manifest key, the serve binding's handle, and
  the key an incremental refresh's ledger is stored under.


packages/server/src/service/preaggregation_synthesis.ts
```diff
@@ -341,12 +480,26 @@ function emitRollup(plan: RollupPlan): string {
...
+   // (`selfAssignTableName`), identically for both tiers, and a rollup's source
+   // name carries the grain digest — so it is unique per grain without a `name=`
+   // to make it so.
+   //
+   // Otherwise the colocated rule, unchanged: named only when a namespace was
+   // named or inherited, since the build self-assigns from the source name, which
+   // is what every dialect but BigQuery accepts, and inventing one would be a
+   // guess.
+   const persist = plan.storage
+      ? `#@ persist storage=${plan.storage}`
```
kylenesbit: The destination is spliced unquoted here, unlike the quoted `name="…"` branch just below, and nothing upstream restricts a destination name to a bare identifier (`readPreaggregateAnnotation` only checks non-empty and not `duckdb`).

I checked this against the installed `@malloydata/malloy-tag` parser: `#@ persist storage=my-lake` parses as `storage="my"` plus a bare `lake` tag with **no parse log**, so the rollup silently targets a destination named `my`. `storage=lake.prod` fails to parse (`Expected an identifier`), `deriveAnnotationFields` sees no storage key, and the persist declaration is lost, which also means the new `declared && !instruction.destination` guard can never fire for it.

Either quote it (`storage="${plan.storage}"`) or refuse non-identifier destinations in `readPreaggregateAnnotation`.

packages/server/src/service/model.ts
```diff
@@ -4015,13 +4144,41 @@ export class Model {
...
+      // — which mattered because a group that will not compile takes the whole
+      // package's storage serving with it. Base-only was documented as the
+      // guaranteed floor, and a group breached it: a composite whose members
+      // disagree about a grain column's captured type fails at every tier,
+      // including the one that carries no refinements at all.
+      const tiers: Array<{
+         keep: ReadonlySet<string>;
+         groups: RollupShapeGroup[];
+      }> = [
+         ...keepKinds.map((keep) => ({ keep, groups: rollupGroups })),
```
kylenesbit: Rollup groups are carried unchanged through all four refinement-thinning tiers and only dropped in the appended fifth tier, where every ordinary refinement is already stripped. So a single uncompilable group (e.g. the case the comment below names, members disagreeing on a grain column's captured type) degrades every authored `storage=` source in the package to base-only and drops every other base's rollups too. The doomed group also gets compiled four times before that.

Pre-PR a package like this served at tier 0. Suggest adding an "all refinements, no groups" tier before the thinning ladder, or better, probing each group's compilability independently in `rollupServeBindings` so one bad group is dropped on its own rather than taking the shape down with it.

packages/server/src/service/preaggregation_serve_bindings.ts
```diff
@@ -0,0 +1,205 @@
...
+         baseSourceName,
+         members: entries
+            // Coarsest first, the same order synthesis emits compose() members in.
+            .sort((a, b) => compareRollupBreadth(a.plan, b.plan))
+            .map(({ plan, binding }) => ({
+               ...binding,
+               // Derived from the plan, never from the author model: see this
+               // module's header for what each of the two lookups does to a
+               // rollup.
+               refinements: rollupServeRefinements(plan),
```
kylenesbit, line 226: A manifest entry is matched to the current plan by `rollupSourceName` (base + grain, measures excluded) and its captured schema is used as-is. If the plan's measure set grew since the build, e.g. the author adds `#@ preaggregate grain="category"` to `revenue` where a `category` rollup with only `total` is already built, the member here declares `revenue is revenue__partial.sum()` over a schema with no `revenue__partial`. Nothing compares `binding.virtualHandle` to the plan, and Package's `allowed` filter checks only the name.

Combined with the tier ladder in `compileServeShape`, that one stale entry fails every tier and drops all groups for the package until the next build, with only warn lines as signal.

Suggest comparing the binding's `virtualHandle` to the plan's expected artifact id, or checking that every `partialName` and grain column in the plan exists in `binding.schema` and skipping the member otherwise.

packages/server/src/service/model.ts
```diff
@@ -3626,7 +3674,23 @@ export class Model {
...
+      // compiled for the previous one.
+      this.serveShapeCache = undefined;
       if (!this.modelDef || this.modelType !== "model") return;
+      // Planned directly rather than taken from the compile below, so a companion
+      // that fails to compile still leaves the storage tier able to serve these
+      // rollups: planning is pure, the compile is not.
+      const { planModelPreaggregation } = await import(
+         "./preaggregation_synthesis"
+      );
+      this.preaggregateRollupPlans = planModelPreaggregation(
```
kylenesbit, line 3710: Race: `serveShapeCache` is cleared above, then the dynamic import is awaited, then `preaggregateRollupPlans` is assigned here without clearing the cache again. `Package.reloadAllModels` publishes `this.models = nextModels` before awaiting `pushPreaggregateServeModels`, and each fresh Model starts with plans `[]`, so a query against model N+1 while model N's build is awaiting sees zero plans, compiles a group-less shape, and writes it to the cache after its own await. The cache key covers bindings only, so every later query reuses the group-less shape and storage rollups stop serving until the next bind/reload.

Fix is either re-clearing the cache after this assignment, or stamping a plans version into the cache key and discarding a write whose stamp no longer matches.

Separately, the dynamic import looks unnecessary: model.ts already value-imports `preaggregation_serve_bindings`, which imports `preaggregation_synthesis`, so there is no cycle to break.

packages/server/src/service/build_plan.ts
```diff
@@ -1054,13 +1054,21 @@ export function deriveBuildPlan(
...
+            // one an authored `#@ persist storage=` passes. Neither subsumes the
+            // other, so a storage rollup runs both.
             try {
                assertColocatedPersistNotAuthorizeGated(
                   source,
                   source.name,
                   "preaggregate",
                   options?.sourceGateOutcomes?.[sourceID],
                );
+               if (declaresStorage) assertMaterializationEligible(source);
```
kylenesbit: The "deliberately no `continue`" rule below was written for the authorize gate, where a refused source can still coexist in `sources` for colocated serving. It does not hold for the storage gate added here: a rollup refused by `assertMaterializationEligible` is recorded in `refusedSources` but still lands in `BuildPlan.sources` with `annotationFields.storage` set.

Example: `#@ preaggregate grain="region" storage=lake` on a base like hammer 65's `scoped is base extend { where: region ~ $REGION }` with a defaulted given. It passes the colocated authorize gate, is refused here with reason `given`, and because the default lets `getSQL` succeed it is written into `wireSources` carrying `storage`. `deriveSelfInstructions` then calls `assertMaterializationEligible` with no try/catch and throws, aborting the publisher's whole auto-run on every tick; the orchestrated path 422s the entire run for a host that instructs it. api-doc.yaml says refused sources carry no entry in `sources`, and `Package.storageWarnings` reads `storage` off the entry with no `refusedSources` check, so in write-only mode it claims a rollup is built that never will be.

Suggest `continue` when the storage gate refuses, and keep the fall-through only for the authorize gate.

kylenesbit (approved):
Approving. Five findings below worth fixing, ideally before merge: the unquoted `storage=` emit and the cache re-clear are one-liners; the build_plan `continue` is small; the group-isolation and stale-manifest-matching ones need a little more thought but can land as a follow-up.

PR #1104, opened 2026-09-02
fix(server): build each persist source once per run, not once per importing model

housejester:
## The problem

`executeInstructedBuild` builds a persist source once for every graph that reaches it. A source imported by *N* models is therefore built *N* times in a single run, each time as a full reseed.

Two pieces combine, each correct on its own.

**1. The graph list holds one entry per model.** `compilePackageBuildPlan` accumulates every model's plan:

```ts
for (const modelPath of pkg.getModelPaths()) {
   const buildPlan = malloyModel.getBuildPlan();
   allGraphs.push(...buildPlan.graphs);        // wholesale, per model
   for (const [sourceID, source] of Object.entries(buildPlan.sources)) {
      allSources[sourceID] = source;           // keyed, so idempotent
   }
}
```

`allSources` is keyed and collapses a source seen twice; `allGraphs` is a list and does not. A source declared in one model appears in the build plan of every model that *imports* it — which the preaggregate branch a few lines up already documents: "a persist source declared there also appears in this plan — but under the SAME sourceID, since a sourceID embeds the model that DECLARES a source rather than the one importing it."

**2. The traversal dedups per graph, and is called per graph.**

```ts
export function* iterGraphSources(graph, sources) {
   const seen = new Set<string>();     // per call, i.e. per graph
   ...
}
```

Its doc is accurate as far as it goes — it deduplicates shared (diamond) dependencies "by sourceID so each is yielded once" — but only within one graph. The build walk needs that guarantee across the whole list:

```ts
for (const graph of graphs) {
   for (const persistSource of iterGraphSources(graph, sources)) {
```

`deriveSelfInstructions` walks the same list with its `seen` declared *outside* the graph loop, so the auto-run path has always been correct. Only orchestrated builds take the traversal without the guard, and every existing test of this loop constructs a single graph — so the cross-graph case was untested.

### What it costs

Measured on a package with three persist sources reachable across 21 models, in a run with no failures at all:

| # | Source | Duration | |
|---|---|---|---|
| 1 | A | 2m 54s | first build |
| 2 | B | 12m 36s | first build |
| 3 | A | 2m 52s | rebuild |
| 4 | B | 14m 45s | rebuild |
| 5 | C | 13s | first build |
| 6 | A | 2m 41s | rebuild |
| 7 | B | 10m 04s | rebuild |
| 8 | A | 3m 50s | rebuild |
| 9 | B | 10m 36s | rebuild |
| 10 | C | 15s | rebuild |
| 11 | A | 3m 58s | rebuild |

Eleven builds for three sources. All three had been built by visit 5, 33 minutes in; the run was still rebuilding them 32 minutes later — 15m 43s of necessary build time against 64m 44s spent, so **76% of the run was redundant**. Where a source takes ~24 minutes rather than three, four builds of it consume a two-hour run on their own, and an orchestrated caller that bounds a run never sees it finish.

Two further consequences worth naming, because neither looks like this bug from the outside:

- A run can build **every** instructed source successfully and still never finalize, because the traversal has not returned. The manifest is not published, so the caller sees no progress on a build that is complete and correct.
- With `strictUpstreams`, a source whose upstream failed on one visit reports "not found in manifest" — which reads as a dependency-ordering problem rather than a repeat visit.

## The fix

One build per declared source per run, in the build walk:

```ts
+ const seenSources = new Set<string>();      // hoisted above the graph loop

  for (const graph of graphs) {
     for (const persistSource of iterGraphSources(graph, sources)) {
        if (!instruction) continue;
+       const buildKey = `${graph.connectionName}\u0000${persistSource.sourceID}`;
+       if (seenSources.has(buildKey)) continue;
+       seenSources.add(buildKey);
```

**Keyed on `sourceID`, not `sourceEntityId`.** `deriveSelfInstructions` keys its guard on the content address, which is right for a path holding no caller instructions. It would not be right here: two *distinct* sources can share one content address — identical SQL on one connection — and each carries its own instruction naming its own physical table. Collapsing on the address would leave one instructed table unbuilt and reported as neither built nor failed. `sourceID` is the declaring model plus the source name, so a revisit through another graph is always genuinely the same declaration.

**And on the graph's connection, because that is part of what a build is.** `buildOneSource` executes on the connection resolved from `graph.connectionName`, not from `persistSource.connectionName`, and `getBuildPlan` groups only *root* nodes by connection (`collectSources` recurses into `dependsOn` regardless) — so a nested source can be yielded under a root of a different connection. Whether such a chain is reachable, and whether the graph's connection is the right one to build it on, are pre-existing questions; keying on the pair means this guard does not silently decide them by removing a build that happens today.

**Placed before the eligibility asserts and the content hash.** Those are per-source, so a repeat visit re-derives an answer the first visit already acted on. Skipping early is what makes a duplicate visit cost nothing rather than merely build nothing.

## Testing

`materialization_service.spec.ts` — **156 pass**. `build_plan.spec.ts`, `incremental_build.spec.ts` and `materialization_service.spec.ts` together — **208 pass, 0 fail**.

The new test builds one source reachable from two graphs and asserts a single `CREATE TABLE`. Against the unguarded traversal:

```
2730 |       expect(creates.filter((s) => s.includes("shared_v1")).length).toBe(1);
                                                                           ^
error: expect(received).toBe(expected)

Expected: 1
Received: 2

(fail) executeInstructedBuild > builds a source shared by two graphs exactly once
```

Revert-proved: with the added lines disabled, that is the only failure of the suite.

Both halves of the key are pinned by tests that discriminate, each failing alone under its own mutation:

| Mutation | Fails |
|---|---|
| key on `sourceEntityId` instead of `sourceID` | `builds two sources that share one content address, not just the first` |
| drop `graph.connectionName` from the key | `does not collapse a source across two connections` |

It also asserts the second graph's own work survives the dedup — the consumer still builds, still after the upstream it reads, and both appear in `entries` — so the guard cannot silently swallow work that a graph legitimately contributes.

**Not exercised:** no test builds the plan from real model files on disk, so the "one graph per importing model" shape is inferred from the observed visit sequence rather than asserted against `compilePackageBuildPlan` directly.

Also corrected: the `BuildPlan.graphs` description said "one per connection", where a package's plan holds one per connection *per model* — the fact this bug turns on.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -5498,7 +5498,11 @@ components:
       properties:
         graphs:
           type: array
-          description: Dependency-ordered build graphs, one per connection.
+          description: >-
+            Dependency-ordered build graphs. One per connection per model, not one
+            per connection: a model contributes a graph for each connection its ROOT
+            persist sources use, and a package's plan concatenates every model's, so
+            the same source appears in the graph of every model that imports it.
           items:
             $ref: "#/components/schemas/BuildGraph"
         sources:
--- a/packages/server/src/service/materialization_service.spec.ts
+++ b/packages/server/src/service/materialization_service.spec.ts
@@ -2664,6 +2664,84 @@ describe("executeInstructedBuild", () => {
       expect(rootIdx).toBeGreaterThan(midIdx);
    });
 
+   it("builds a source shared by two graphs exactly once", async () => {
+      // compilePackageBuildPlan pushes every model's graphs into one allGraphs
+      // list, so a persist source declared in one model and imported by another
+      // appears in BOTH models' graphs. iterGraphSources dedups per graph, so a
+      // shared source is yielded once per graph and built once per graph -- each
+      // a full reseed CTAS.
+      const runSQL = sinon.stub().resolves();
+      const connection = { runSQL } as unknown as MalloyConnection;
+      const shared = fakeSource({
+         name: "shared",
+         sourceEntityId: "bsharedaaaaaaaaa",
+      });
+      const consumer = fakeSource({
+         name: "consumer",
+         sourceEntityId: "bconsumerbbbbbbb",
+      });
+      const compiled = {
+         graphs: [
+            {
+               connectionName: "duckdb",
+               nodes: [[{ sourceID: "shared", dependsOn: [] }]],
+            },
+            {
+               connectionName: "duckdb",
+               nodes: [
+                  [
+                     {
+                        sourceID: "consumer",
+                        dependsOn: [{ sourceID: "shared", dependsOn: [] }],
+                     },
+                  ],
+               ],
+            },
+         ],
+         sources: { shared, consumer },
+         connectionDigests: { duckdb: "dig" },
+         connections: new Map([["duckdb", connection]]),
+      };
+
+      const { entries } = await callExecute(
+         compiled,
+         [
+            {
+               sourceEntityId: "bsharedaaaaaaaaa",
+               materializedTableId: "mt-s",
+               physicalTableName: "shared_v1",
+               realization: "COPY",
+            },
+            {
+               sourceEntityId: "bconsumerbbbbbbb",
+               materializedTableId: "mt-c",
+               physicalTableName: "consumer_v1",
+               realization: "COPY",
+            },
+         ],
+         {},
+      );
+
+      const creates = runSQL
+         .getCalls()
+         .map((c) => c.args[0] as string)
+         .filter((s) => s.startsWith("CREATE TABLE"));
+      // The shared source builds ONCE, not once per graph that reaches it.
+      expect(creates.filter((s) => s.includes("shared_v1")).length).toBe(1);
+      // And the dedup does not swallow the second graph's own work: the
+      // consumer still builds, and still after the upstream it reads.
+      expect(creates.filter((s) => s.includes("consumer_v1")).length).toBe(1);
+      expect(creates.findIndex((s) => s.includes("consumer_v1"))).toBeGreaterThan(
+         creates.findIndex((s) => s.includes("shared_v1")),
+      );
+      // Both are reported built, so an orchestrated caller settles neither as
+      // missing.
+      expect(Object.keys(entries).sort()).toEqual([
+         "bconsumerbbbbbbb",
+         "bsharedaaaaaaaaa",
+      ]);
+   });
+
    it("seeds a downstream build with the QUOTED upstream reference (case-folding dialect)", async () => {
       // A carried/seeded upstream (built in a prior run or unit, e.g. via
       // referenceManifest) must reach the downstream's SQL generation as the
--- a/packages/server/src/service/materialization_service.spec.ts
+++ b/packages/server/src/service/materialization_service.spec.ts
@@ -2742,6 +2742,114 @@ describe("executeInstructedBuild", () => {
       ]);
    });
 
+   it("builds two sources that share one content address, not just the first", async () => {
+      // Why the guard keys on sourceID and not sourceEntityId. Two DISTINCT sources can
+      // canonicalize to the same content address (identical SQL on one connection) while
+      // holding separate instructions that name separate physical tables. Keying on the
+      // address would build one and leave the other instructed table unbuilt — reported
+      // neither in entries nor in failures, which an orchestrated caller cannot settle.
+      const runSQL = sinon.stub().resolves();
+      const connection = { runSQL } as unknown as MalloyConnection;
+      const twinA = fakeSource({
+         name: "twin_a",
+         sourceEntityId: "btwinaaaaaaaaaaa",
+      });
+      const twinB = fakeSource({
+         name: "twin_b",
+         sourceEntityId: "btwinaaaaaaaaaaa",
+      });
+      const compiled = compiledWith(
+         { twin_a: twinA, twin_b: twinB },
+         [["twin_a"], ["twin_b"]],
+         new Map([["duckdb", connection]]),
+      );
+
+      const { entries } = await callExecute(
+         compiled,
+         [
+            {
+               sourceEntityId: "btwinaaaaaaaaaaa",
+               materializedTableId: "mt-a",
+               physicalTableName: "twin_a_v1",
+               realization: "COPY",
+               sourceID: "twin_a",
+            },
+            {
+               sourceEntityId: "btwinaaaaaaaaaaa",
+               materializedTableId: "mt-b",
+               physicalTableName: "twin_b_v1",
+               realization: "COPY",
+               sourceID: "twin_b",
+            },
+         ],
+         {},
+      );
+
+      const creates = runSQL
+         .getCalls()
+         .map((c) => c.args[0] as string)
+         .filter((s) => s.startsWith("CREATE TABLE"));
+      expect(creates.filter((s) => s.includes("twin_a_v1")).length).toBe(1);
+      expect(creates.filter((s) => s.includes("twin_b_v1")).length).toBe(1);
+      // Both were instructed, so both must be accounted for. They collide on one manifest
+      // key by construction — that collision is pre-existing and separately detected; what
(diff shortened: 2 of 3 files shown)


girishjeswani (approved):
Approving. I verified the invariant the fix rests on from the compiler rather than taking it on trust: `mkSourceID(name, url)` uses `DefineSource`'s own `this.location.url` (`source_def_utils.js:20-22`, `define-source.js:65`), and an import re-registers the same sourceID rather than minting a new one (`import-statement.js:188-202`). So a revisit through another graph really is the same declaration, and the declaring-model claim holds.

I also checked the axes where a collapse would be silently wrong, and none of them can share a sourceID with a different build: `preaggregatePlans` is keyed by sourceID (`build_plan.ts:704`), so a rollup cannot carry a different grain; destination, physical name and incremental mode all come from the resolved instruction, and `bySourceID` was already a `Map.set` (`materialization_service.ts:1733-1739`), so two instructions sharing a sourceID collapsed to the last one before this PR too. The dedup cannot lose a build that was reachable before it.

Mutation-checked both directions. Removing the two guard lines fails exactly one test (`Expected: 1 / Received: 2`). Coarsening the key to `sourceEntityId` leaves all 206 green, but with a probe of two distinct sources sharing one content address it fails with an instructed table never built at all, which is the failure mode your comment predicts. Ordering is safe: `iterGraphSources` is post-order DFS and the guard is a `Set` plus `continue` with no re-sort, and I probed the diamond where the shared source is a root in graph 1 and a dependency in graph 2 — still built once, still before its downstream.

Worth crediting a second thing the fix repairs that the body doesn't claim: `builtThisRun` and `retainedThisRun` are arrays pushed per visit (`:1965`, `:1971`), so on `main` a shared storage source was pushed N times and the failure-path reclaim iterated duplicates. `entries` was already keyed, so the committed manifest count was never inflated.

Three notes, none blocking.

**A cross-connection shared source now builds on the wrong connection, deterministically.** `getBuildPlan` groups a graph by the root's connection (`core.js:1208-1224`), but a nested `dependsOn` source is yielded under that graph regardless of its own `connectionName`, and `buildOneSource` receives the graph's connection rather than `persistSource.connectionName` (`:1784-1789`, `:1893`). I demonstrated it: a source whose `connectionName` is `duckdb`, reached as a dependency of a postgres root, gets its CTAS executed on postgres. On `main` it built on postgres *and* on duckdb, so the correct build also happened; with the dedup only the postgres one runs.

This is pre-existing and not caused by the fix, but the dedup removes the accidental second build that was masking it. I could not tell whether a multi-connection persist chain is actually reachable (eligibility or `manifestExcludingStorage` may block it upstream), so rather than assert impact: is it reachable? If so the guard probably wants the connection in its key, or `buildOneSource` should take `connections.get(persistSource.connectionName)`.

**The key choice has no test.** The comment spends 14 lines arguing `sourceID` over `sourceEntityId`, and mutation 2 shows nothing currently locks that in. Two distinct sources sharing one content address, each with its own instruction, is the test — I wrote it as a probe and it discriminates cleanly.

**Placement.** The guard sits after instruction resolution, both eligibility asserts, and `computeSourceEntityId` (`:1920`). I assume that's deliberate so a duplicate visit still hits the 422 gate, but the comment doesn't say so, and the next reader may try to hoist it to skip the redundant `getSQL()` and hash work per duplicate visit. One clause either way would settle it.

Unrelated and pre-existing: `api-doc.yaml:5309` and `api.ts:1827` describe the graphs as "one per connection", where `compilePackageBuildPlan` emits one per connection *per model* — the fact this bug turns on. Cheap to correct while the invariant is fresh.

DCO signed on both commits, 15/15 CI green.

housejester:
All three addressed, thanks — and the cross-connection one was worth chasing.

The mechanism is confirmed at the compiler: `getBuildPlan` builds `graphsByConnection` from **root nodes only**, while `collectSources` recurses into `dependsOn` regardless of connection. So a nested source really can be yielded under a root of a different connection.

On reachability, I could not construct one either, and I suspect it is not reachable in a working model: the CTAS executes on a single connection and would reference a table that is not there, and both `connectionDigests` and `resolvePackageConnections` are built from graph-root connection names only, so such a source would not even resolve a digest. But that is a belief, and not a basis for removing a build that happens today.

So the key is now `(graph.connectionName, sourceID)`. That reads as defensive, but I think it is simply more accurate: `buildOneSource` executes on the connection resolved from `graph.connectionName`, so `(source, connection)` is what identifies a build as the code stands. The common path is unchanged, and the guard no longer decides a question that is not its to decide. Whether the graph's connection is the *right* one to build a nested source on is still open, and still pre-existing.

Both halves of the key are pinned now, each failing alone under its own mutation:

| Mutation | Fails |
|---|---|
| key on `sourceEntityId` instead of `sourceID` | `builds two sources that share one content address, not just the first` |
| drop `graph.connectionName` from the key | `does not collapse a source across two connections` |

On placement, I hoisted rather than documented. Your guess was that it sat late so a duplicate visit still hits the 422 gate — but that gains nothing, since an ineligible source throws on the first visit regardless. It now sits directly after instruction resolution, so a duplicate visit also skips the eligibility asserts and the `getSQL()` + hash, which is the cost this change exists to remove.

`BuildPlan.graphs` description corrected and the types regenerated.

And thanks for the `builtThisRun` / `retainedThisRun` catch — I had not noticed the failure-path reclaim was iterating duplicates.

156 pass in `materialization_service.spec.ts`; 208 across it plus `build_plan.spec.ts` and `incremental_build.spec.ts`.

girishjeswani:
Second round looks good — all three items from my earlier pass are closed, and one of them was my mistake.

**The key choice is now genuinely pinned.** I re-ran the mutation that survived last time and both halves of the composite key hold: keying on `computeSourceEntityId` instead of `sourceID` kills only `builds two sources that share one content address, not just the first`, and dropping `graph.connectionName` from the key kills only `does not collapse a source across two connections`. With the guard disabled entirely, only `builds a source shared by two graphs exactly once` fails. Three mutations, three distinct single-test kills, no overlap.

**On the connection keying** — worth stating plainly for the record, since the commit subject is easy to misread: this doesn't fix the wrong-connection build, it restores the compensating correct one that my suggested `sourceID`-only key had deleted. I verified the underlying behaviour is identical on `main`, on the PR, and with the guard fully disabled:

```
single graph, duckdb-declared source as a dependsOn of a postgres root:
  duckdb CTAS:   []
  postgres CTAS: ["CREATE TABLE \"shared_v1_...\" AS (SELECT 1)"]   (unchanged in all three)

two-graph shape:
  PR as-is:          duckdb=1, postgres=1   <- correct build preserved
  sourceID-only key: duckdb=0, postgres=1   <- correct build deleted
```

So the guard's scope is right: it removes redundancy without adjudicating a question that isn't its to answer. And your reachability argument is stronger than the way you put it — `connectionDigests` is keyed by graph connection names (`build_plan.ts:935-955`) while `computeSourceEntityId` looks the digest up by `source.connectionName` (`build_plan.ts:594`), so a nested cross-connection source resolves `undefined` as its digest. That's an unstable content address, not just a build on the wrong connection, which means the shape would break loudly at addressing before the connection mismatch ever mattered. If you want an issue for the real question (which connection *should* build a nested source), that digest-lookup mismatch is the sharper framing.

**My `api.ts` point was wrong** — all three of those files are gitignored codegen regenerated from `api-doc.yaml`, so there was never a divergence to create. I was reading a stale local build artifact. The new spec wording is accurate to `compilePackageBuildPlan`.

**And you were right to override my placement guess.** Hoisting the guard above the eligibility asserts is correct: an ineligible source throws on visit 1 regardless, so nothing is lost and visits 2..N now skip `getSQL()` and hashing, which is the cost the PR exists to remove.

Also re-verified the things the new key could plausibly have broken: a source imported by 5 same-connection models still builds exactly once (all five graphs share a key, confirmed rather than assumed), and the diamond shape keeps post-order with 4 entries. Using NUL as the separator is the right call — I probed the crafted-collision shape a printable separator would fuse and both sources stay distinct.

208 pass at the tip, 18/18 CI, DCO on all three commits. Nothing outstanding from me.

PR #1105, opened 2026-09-03
fix: measure a notebook table after it settles, not before

jswir:
## The bug

A notebook table painted its cell at the 700px cap however few rows it had. A two-row table got 700px of cell for 84px of table.

## Why

A cell renders its result into a box capped at `CELL_MAX_HEIGHT`, measures the result, and shrinks the box to fit. The measurement ran **once**, from the renderer's `onReady` — and for a table `onReady` is not a paint signal. The renderer fires it as soon as it knows it can paint, which for a table is immediately (a chart waits for a non-zero parent size, a table does not), so the one measurement landed before the table's virtualized grid had laid out.

Measured in Chromium against `examples/storefront/storefront.malloynb`: a four-row table reported `scrollHeight` **10028** at that moment and **308** once settled. The cell latched onto the 10028, clamped it to 700, and never looked again.

## The fix

A table now keeps a `ResizeObserver` and reports each settled height, not only the first.

That **terminates**, rather than merely converging in practice: `.malloy-table.root` is `height: fit-content; max-height: 100%` in the renderer's own CSS, and the wrappers between it and the container add no padding (measured), so shrinking the container to the table's content height is the exact point at which the cap stops binding — the next measurement reports the same number. A table taller than the cap reports its full `scrollHeight`, the container clamps to the cap, and nothing resizes.

A root that **fills** its container stays on the one-shot path deliberately. A chart draws itself 4px-per-side inset from the box it is given, so feeding its height back as the new container height would ratchet the container down by that inset every pass and never converge.

Which root is which comes from the renderer's own typed API — `viz.getMetadata().getRootField().renderAs() === "table"` — not from sniffing its CSS class names. `getMetadata()` is already used here for the drill affordance; `renderAs()` is declared on `FieldBase` and inherited by `RootField`.

Also: the `MutationObserver` fallback now disconnects only once a measurement has actually landed. Disconnecting unconditionally gave up on a stage whose renderer output was still empty at the first settle, leaving nothing to measure it later. That cannot recreate the loop the unconditional disconnect was added to guard in #531 — *that* observer watched the container with `attributes: true`, so reporting a height re-triggered it through the container's own style; this one watches `stage`, a descendant, which that write is outside of.

## Measured, before → after

Same notebook, same build:

| cell | before | after |
| --- | --- | --- |
| 10-row table | 700 | 308 |
| 2-row table | 700 | 84 |
| chart | 692 | 692 |
| `# dashboard` | 700 | 700 |

The two-row cell is added to `storefront.malloynb` in this PR, so the second row is reproducible and the smallest table reaches CI. It had been living only in a local `publisher_data` copy, which is gitignored.

No console errors, no `ResizeObserver` loop warnings.

## Is a `# size=` chart affected too?

No, and it was worth checking, because a `# size=` other than `fill` puts a **chart** on the same "fixed" sizing strategy a table gets, so it signals ready just as early. But a tag-sized chart takes its dimensions from a synchronous lookup, so its first measurement is already correct. Measured against the real renderer, cell box vs rendered height, zero dead space in every case:

```
size=xs 112/112   sm 168/168   md 252/252   lg 336/336
size=xl 448/448   2xl 560/560  spark 28/28
viz=bar viz.size=lg 336/336    viz=line viz.size=2xl 560/560
bar_chart size=lg 336/336
```

Fill charts in the same run measured 692/684 — the renderer's own inset, which is why they stay on the one-shot path. What sets a table apart is not its sizing strategy but its virtualized grid, which lays out after the ready signal. So the predicate is `renderAs() === "table"` on purpose, not as an approximation of something broader.

## One assumption, named rather than defended

`scrollHeight` is the content height, so the measurement assumes the table's box adds no chrome of its own. Neither source exists today: `.malloy-table.root` declares no border, and a horizontal scrollbar costs no layout where scrollbars overlay. A wide table clipped by a few pixels on a platform that draws classic scrollbars would be this assumption breaking, and the fix is `+ (offsetHeight - clientHeight)`.

A correction term was written and then removed: it was a no-op in every reachable case and could not be tested here, because headless Chromium would not give up overlay scrollbars. There is a comment at the measurement site naming the symptom and the one-line fix, which is more useful than an untestable branch.

## Tests

New Playwright case in `package-notebooks.spec.ts`, asserted against the real renderer in a real browser and on the table-to-box ratio rather than a pixel count, so a row-height or font change doesn't break it. It runs over every table on the page, and pins the count at 3 first — notebook cells render progressively and the poll succeeds the moment nothing it can see has dead space, so without the count a run with one table painted would pass. Three is every untagged view in `storefront.malloynb` (`top_products`, `top_customers`, and the two-row cell this PR adds); the dashboard's nested tables are excluded by both `.root` and the direct-child step in the selector.

Confirmed it isn't passing for free: reverting only the SDK change makes it fail with `{box: 700, table: 308}` and `{box: 700, table: 84}`, past the count assertion. Note it names two tables, not three — one of the three shows no dead space even with the bug, which is why pinning the count matters.

Full app Playwright suite: **111 passed, 2 failed**. Both failures are in `environment-connections.spec.ts` and need the stub `bigquery` connection the CI workflow injects into `publisher.config.json` before the suite (`app-playwright.yml:86`); they fail the same way on a clean tree locally. `test:sdk` 384 pass, `test:examples` 72 pass. Lint and typecheck clean on both packages. I did not run `test:server` — it binds port 4000 and nothing here touches the server.

## Not in this PR

The sidebar change that was originally bundled here now lives on its own branch, since it is an unrelated product judgment that deserves to be judged separately.

### DIFF
--- a/packages/app/tests/playwright/package-notebooks.spec.ts
+++ b/packages/app/tests/playwright/package-notebooks.spec.ts
@@ -93,7 +93,16 @@ test.describe("package-notebooks", () => {
       // row-height or font change, and over every table on the page rather
       // than the first, because the first one to settle is not deterministic.
       const tables = page.locator(".malloy-render > .malloy-table.root");
-      await expect(tables.first()).toBeVisible();
+      // Pin the count before measuring anything. Notebook cells render
+      // progressively, and the poll below succeeds the moment nothing it can
+      // see has dead space, so without this a run where only one table has
+      // painted is a pass - and that is the run least likely to reproduce the
+      // measurement race this test guards. Three is every untagged view in
+      // storefront.malloynb: top_products, top_customers, and the two-row
+      // top_products cell. Every other view carries a render tag, and the
+      // dashboard's nested tables match neither .root nor the direct-child
+      // step. Adding an untagged cell to that notebook means updating this.
+      await expect(tables).toHaveCount(3);
       await expect
          .poll(
             () =>
--- a/packages/app/tests/playwright/package-notebooks.spec.ts
+++ b/packages/app/tests/playwright/package-notebooks.spec.ts
@@ -79,4 +79,45 @@ test.describe("package-notebooks", () => {
       );
       await expect(page.getByText(/does not exist/i)).toHaveCount(0);
    });
+
+   test("a table cell is as tall as its table, not as tall as the cell cap", async ({
+      page,
+   }) => {
+      await page.goto(
+         `/${DEFAULT_ENV}/${PACKAGES.storefront}/storefront.malloynb`,
+      );
+
+      // Every notebook table, measured against the box the cell gives it. The
+      // renderer signals ready before a table's virtualized grid has settled,
+      // so the cell used to latch onto the height it read at that moment --
+      // scrollHeight 10028 for a four-row table -- clamp it to the 700px cap
+      // and paint ~390px of blank space under the rows. Asserted on the ratio
+      // rather than on a pixel count so it survives a row-height or font
+      // change, and over every table on the page rather than the first,
+      // because the first one to settle is not deterministic.
+      const tables = page.locator(".malloy-render > .malloy-table.root");
+      await expect(tables.first()).toBeVisible();
+      await expect
+         .poll(
+            () =>
+               tables.evaluateAll((nodes) =>
+                  nodes
+                     .map((node) => {
+                        const table = node as HTMLElement;
+                        const box = table.parentElement as HTMLElement;
+                        return {
+                           table: table.offsetHeight,
+                           box: box.offsetHeight,
+                        };
+                     })
+                     // Only the ones with dead space left under the table, so a
+                     // failure names the measurements rather than saying false.
+                     .filter(
+                        ({ table, box }) => table === 0 || box - table > 4,
+                     ),
+               ),
+            { timeout: 30_000 },
+         )
+         .toEqual([]);
+   });
 });
--- a/packages/app/tests/playwright/package-notebooks.spec.ts
+++ b/packages/app/tests/playwright/package-notebooks.spec.ts
@@ -87,14 +87,11 @@ test.describe("package-notebooks", () => {
          `/${DEFAULT_ENV}/${PACKAGES.storefront}/storefront.malloynb`,
       );
 
-      // Every notebook table, measured against the box the cell gives it. The
-      // renderer signals ready before a table's virtualized grid has settled,
-      // so the cell used to latch onto the height it read at that moment --
-      // scrollHeight 10028 for a four-row table -- clamp it to the 700px cap
-      // and paint ~390px of blank space under the rows. Asserted on the ratio
-      // rather than on a pixel count so it survives a row-height or font
-      // change, and over every table on the page rather than the first,
-      // because the first one to settle is not deterministic.
+      // Every notebook table, measured against the box the cell gives it: a
+      // cell must end up as tall as its table, not as tall as the 700px cap.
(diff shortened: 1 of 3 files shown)


Sha-Bang (commented):
**1 should-fix.**

- `packages/app/tests/playwright/package-notebooks.spec.ts:95` — the test asserts over however many tables have rendered, not over the three the page has, so it can pass on exactly the run that would reproduce the race it guards.

The sizing change itself looks right: the fixed-point argument holds in both the under- and over-cap cases, the chart path stays one-shot, and moving the MutationObserver disconnect does not recreate the #531 loop (that one watched the container with `attributes: true`; this watches `stage`, a descendant).

packages/app/tests/playwright/package-notebooks.spec.ts
```diff
@@ -79,4 +79,42 @@ test.describe("package-notebooks", () => {
...
+      await page.goto(
+         `/${DEFAULT_ENV}/${PACKAGES.storefront}/storefront.malloynb`,
+      );
+
+      // Every notebook table, measured against the box the cell gives it: a
+      // cell must end up as tall as its table, not as tall as the 700px cap.
+      // Asserted on the ratio rather than on a pixel count so it survives a
+      // row-height or font change, and over every table on the page rather
+      // than the first, because the first one to settle is not deterministic.
+      const tables = page.locator(".malloy-render > .malloy-table.root");
```
Sha-Bang, line 95: **Should-fix** - The comment says this asserts "over every table on the page rather than the first, because the first one to settle is not deterministic" — but nothing pins how many tables `evaluateAll` runs over. `expect(tables.first()).toBeVisible()` only proves there is at least one, and the poll succeeds the moment the filtered list is empty. Notebook cells render progressively, so a pass with one of the three tables present is a pass.

That matters here because the bug being fixed is a measurement race: the run where only one table has rendered is exactly the run least likely to reproduce it.

The count is deterministic at 3 — every view in `storefront.malloynb` carries a render tag (`# bar_chart` x5, `# line_chart`, `# shape_map`, `# dashboard`) except `top_products`, `top_customers`, and the `top_products + { limit: 2 }` cell this PR adds; the `.malloy-render > .malloy-table.root` selector correctly excludes the dashboard's nested tables. So `await expect(tables).toHaveCount(3)` before the poll makes the assertion mean what the comment says.

Sha-Bang (approved):
Approving. The sizing fix itself checks out — fixed-point argument holds in both the under- and over-cap cases, and the MutationObserver change doesn't recreate #531. One should-fix on the test's table count, posted inline.

packages/app/tests/playwright/package-notebooks.spec.ts
jswir: Fixed in 3492073f. `await expect(tables).toHaveCount(3)` now replaces the `toBeVisible()` on the first table, with a comment recording where 3 comes from so the next person adding an untagged cell knows to update it.

Confirmed your count independently: `top_products`, `top_customers`, and the two-row `top_products` cell are the only views in `storefront.malloynb` without a render tag, and `toHaveCount(3)` passing is itself the proof that the dashboard's nested tables don't match the selector.

Your point about which run reproduces the race is sharper than I'd realized. With the SDK fix reverted, the failure names only **two** entries: `{box: 700, table: 308}` and `{box: 700, table: 84}`, not three. So one of the three tables shows no dead space even with the bug present, and the old assertion could have gone green on a run where that was the only one painted.

Verified both directions on this branch: 3/3 repeats pass with the fix, and reverting only `RenderedResult.tsx` still fails, now getting past the count assertion first.

PR #1106, opened 2026-09-03
fix(materialization): reject unsafe characters in persist table names

sagarswamirao:
## What

Extend the publish-time persist-name validator (`assertPersistNamesQuoted`) to require a `#@ persist name=` value to be a plain identifier path -- dot-separated `[A-Za-z0-9_-]` segments -- in addition to the existing quoted-value check.

## Why

The author-controlled `#@ persist name=` value is inlined into the `CREATE OR REPLACE TABLE <path>` / `DROP TABLE IF EXISTS <path>` DDL the materialization builder emits on the storage path. The validator previously required only that the value be quoted, so a quoted value whose content carried an embedded quote or backtick could close the identifier and append arbitrary statements.

## How

- Content check: the name must match `^[A-Za-z0-9_-]+(\.[A-Za-z0-9_-]+)*$`. Hyphens and leading digits stay allowed because a dialect container path can be a hyphenated project id (e.g. `my-proj.mydataset.engaged_events`); a value containing a quote, backtick, semicolon, or space is rejected with an actionable `ModelCompilationError` (424) at model load.
- Every `name=` occurrence on a line is validated, not just the first, so a repeated key cannot shadow an unsafe value past the check regardless of how the tag parser resolves duplicates.
- The `.malloy`-only scan is annotated at the call site: `#@ persist` is a model-file directive and does not appear in `.malloynb` notebooks.

## Tests

`packages/server/src/service/quoting_injection.spec.ts` (new): rejects embedded-quote / backtick / semicolon / whitespace / duplicate-key payloads; accepts simple, dotted, hyphenated, leading-digit, and three-part identifier paths. Existing `persist_annotation_validation.spec.ts` continues to pass unchanged.

## Compatibility

Behavior change: a package whose persist name contains a quote, backtick, semicolon, or space now fails its next load with an actionable 424 rather than publishing. A previously deployed package using such a name must rename it to a plain identifier path before its next reload.


housejester (approved):
The vulnerability is real and the fix is carefully built — the standalone-key lookbehind, validating every `name=` occurrence rather than the one the tag parser keeps, and mirroring the physical-name grammar are all the right calls. Two non-blocking findings, and one compatibility question I went and answered rather than leaving as a caveat.

## The vector is genuine, and worth naming where it actually lives

Reachability rests on `quoteManifestTablePath`:

```ts
return isQuotedIdentifierPath(tableName) ? tableName : quoteTablePath(tableName, dialect);
```

where `isQuotedIdentifierPath` is `tableName.includes('"') || tableName.includes("`")`. A value that merely *contains* a quote takes the pass-through branch and reaches the DDL unescaped — being a *well-formed* quoted identifier is never checked. The safety argument sits directly above it:

> a self-assigned `#@ persist name=` value is the author's own canonical SQL (they own quoting it for the dialect)

That assumption is what this PR retires, and retiring it is correct: `#@ persist name=` is author-controlled input on a server that loads packages it did not write.

Worth recording that `assertPersistNamesQuoted` was not previously a security check — it exists because an unquoted name is silently dropped from the build plan, so the source "publishes but never materializes". This changes what the function is for; the doc rewrite handles that well.

## Compatibility: checked, and it is clean

A new load-time 424 is a behaviour change for already-published packages. Rather than leave that as a note, I evaluated a real-world corpus against the three regexes from this diff, transcribed verbatim: **438 `#@ persist name=` annotations across 47 packages, 152 distinct values.**

**Newly rejected: 0. Already failing the existing unquoted rule: 0.**

The shape census is the more useful half, because it says which parts of `SAFE_NAME_PATH` are load-bearing:

- **307 dotted paths** — allowing `.`-separated segments is doing real work, not hypothetical.
- **0 hyphens, 0 leading digits** — both permitted on the hyphenated-project-id reasoning. That reasoning is sound, but it is currently headroom rather than a compatibility requirement, which is worth knowing if the grammar is ever tightened again.

So nothing in that corpus carries a quote, backtick, semicolon or space in a persist name, and this cannot fail a package that loads today.

## Finding 1 — this closes the perimeter, not the sink

`quoting.ts` is untouched, so `quoteManifestTablePath` still treats "contains a quote" as "is canonical SQL". This validates *one* producer; the doc comment names two, and argues the other is safe because those names "never contain a quote" — an assumption about a producer maintained elsewhere.

The PR is deliberate about this, mirroring the physical-name grammar so the two agree. But that makes DDL safety a property of two independently-maintained validators staying in step, rather than of the function that emits the SQL.

Cheap follow-up, need not be here: have `isQuotedIdentifierPath` require a well-formed quoted identifier path — balanced quotes, no embedded unescaped quote — instead of the mere presence of a quote character. Then the sink is safe whatever reaches it, and this validator becomes defence in depth rather than the only thing standing between an author and the DDL.

## Finding 2 — the `.malloy`-only scope is correct, but held by accident

The scope is sound, and I checked it rather than taking it on faith: `compilePackageBuildPlan` skips non-`.malloy` paths, so a `#@ persist` in a notebook never reaches the build plan or the DDL. Validator scope and build-plan scope agree.

The reason they agree is the part worth pinning down. That skip exists because:

> notebooks: `getModel()` parses a model file as a flat model and throws on the notebook's `>>>` cell delimiter

— a parser limitation, not a security decision. The new comment here anticipates the change ("If that ever changes, widen this guard"), which is the right instinct, but it lives on the validator. Whoever lifts the notebook restriction will be editing the build-plan compile loop and will not see it. A reciprocal one-line note at that `continue` would close the loop from the side that would actually break it.

## Minor

The validator reads raw text while the sink consumes the tag-parsed value, which is a parser differential. I checked which way it falls: `name="a\"b"` matches non-greedily to the first quote, yields `a\`, fails `SAFE_NAME_PATH`, and is rejected — so it fails closed. Worth a sentence confirming the tag parser performs no unescaping, so that stays a property rather than a coincidence.

housejester:
One thing my review missed: this wants a `RELEASE_NOTES.md` entry, and `RELEASE_NOTES.md` asks for it in the PR that changes the behaviour rather than at release time.

The case for one is that a `#@ persist name=` value which publishes and loads today can be refused at load tomorrow, with a 424 and no other signal. That is a change an author has to *do* something about — republish with a conforming name — and nothing else will surface it: the auto-generated "What's Changed" list gives a reader `fix(materialization): reject unsafe characters in persist table names`, from which they cannot tell whether their own package is affected or what the accepted grammar now is. The actionable half is the character set itself (dot-separated `[A-Za-z0-9_-]` segments), since that is what lets an author check a name without reading the diff.

On the level, I'd flag rather than assert, and my own compatibility finding argues *against* the heavier reading: 438 annotations across 47 packages, 152 distinct values, 0 newly rejected. The versioning policy's bar for a breaking change is a package that loaded yesterday and does not today, and no package in that corpus is affected — so patch looks defensible to me and `(BREAKING)` may be overstating it. The policy anticipates exactly this shape ("a section can be titled `(BREAKING)` while its bullets say *this cannot affect an existing package*") and says the call is the maintainer's, not something to infer from the diff. Worth putting the census in the note either way: "no package in a 47-package corpus is affected" is the sentence that tells an operator whether to care.

Two mechanics that bite quietly:

- The heading needs a title after the marker — `## [Unreleased] — <what changed>`, with an em dash, colon or hyphen. A bare `## [Unreleased]` fails CI on the PR that writes it, because the version becomes the release's title and the stripped marker leaves nothing to print.
- #1108, stacked on this branch, needs its own section for the generic 500/502 bodies. Those are two different behaviours, so two sections is right — amending in place is for a follow-up that changes the *same* behaviour. Both ship in the same release as separate entries.

For what it's worth, `main` is clean right now: no pending `[Unreleased]` sections and no unmerged `release-notes-stamp-*` branch, so a section added here goes out with the next release and nothing else rides along with it.

<sub>Reviewed with an assist from Dapper Ibex.</sub>

Sha-Bang (approved):

PR #1107, opened 2026-09-03
fix(app): move Visualization theme to the bottom of the sidebar

jswir:
## What

Moves **Visualization theme** out of a headed `SETTINGS` section under the environment list and into the pinned bottom group with the docs links, dropping the heading.

## Why

Visualization theme is the only setting there is — `/settings` is a redirect to `/settings/theme` — so a headed section directly beneath the environment list gave one operator preference the same visual weight as the data the sidebar exists to navigate. The bottom group is where the other persistent, non-navigational links already live.

## Notes

Split out of #1105, where it was originally bundled with an unrelated SDK measurement fix. It is a product judgment about sidebar weighting rather than a bug fix, so it is easier to agree or disagree with on its own.

`Typography` is still imported and used elsewhere in the file, so nothing is left dead.

## Tests

Full app Playwright suite run on the combined branch before the split: 111 passed, 2 failed — both in `environment-connections.spec.ts`, which needs the stub `bigquery` connection CI injects before the suite (`app-playwright.yml:86`); they fail the same way on a clean tree locally. Lint and typecheck clean.

### DIFF
--- a/packages/app/src/components/layout/Sidebar/Sidebar.tsx
+++ b/packages/app/src/components/layout/Sidebar/Sidebar.tsx
@@ -65,8 +65,8 @@ export default function Sidebar({
          <Box sx={{ flex: 1, overflowY: "auto", overflowX: "hidden", py: 1 }}>
             <PrimaryNav isCollapsed={isCollapsed} />
             <EnvironmentsSection isCollapsed={isCollapsed} />
-            <SettingsSection isCollapsed={isCollapsed} />
          </Box>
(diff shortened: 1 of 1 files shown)


Sha-Bang (approved):
Approving — clean, no findings. Verified the premise (App.tsx:92 redirects to /settings/theme), no dead imports left behind, and the missing flexShrink matches DocsFooter's existing convention rather than diverging from it.

PR #1108, opened 2026-09-03
fix(errors): return a generic body for internal 500/502 responses

sagarswamirao:
## What

`internalErrorToHttpError` returns a generic body for internal-failure responses (500 fallback and 502 `ConnectionError`) instead of echoing `error.message`, and logs the real error server-side.

## Why

An internal error's message can carry a filesystem path, an SQL fragment, an upstream host, or a driver-error oracle. Returning it to the client is an information-disclosure surface.

## How

- 500 returns `Internal server error.`, 502 returns `Upstream connection error.`; both `logger.error(...)` the real error so the detail is available for debugging.
- Every 4xx client-error branch is unchanged -- 400/403/404/422/424/409/413/503/504/501 still return their actionable message, which the caller needs to correct the request.

## Tests

The two existing tests in `errors.spec.ts` that pinned the old verbatim 500/502 message are updated to assert the generic body; `errors_internal_leak.spec.ts` (new) pins that internal detail is not echoed and that a 4xx message is preserved.

## Compatibility

A 500/502 response body now carries a generic message instead of the internal error text; the status code is unchanged. A consumer that displayed the raw 5xx message text will show the generic text.

## Note

The permissive `frame-ancestors` default that was part of the original finding is intentionally not included here: a served data-app page is embedded cross-origin by the consuming deployment, and tightening the default without a paired frame-ancestors override would break that embed. That change ships with the deployment coordination, separately.


housejester (commented):
Direction is right, and deferring the `frame-ancestors` half with a stated reason is the correct call — a served data-app embed is exactly the thing that would break, and pairing it with the deployment change is better than shipping a default nobody can use. The new spec file is also the right shape: it pins the negative *and* the positive (a 4xx message survives), which is what stops a future "genericize everything" pass from quietly going too far.

One finding below inverts the PR's own premise, so I'd like that resolved before merge. The rest are scope and follow-through questions.

## The detail isn't actually logged anywhere

`logger.error("...", { error })` serializes the `Error` to `{}` under **both** formats `logger.ts` configures. `message` and `stack` are non-enumerable own properties, so `format.json()` and `format.simple()` both emit an empty object, and `format.errors({ stack: true })` doesn't fire because it inspects the info object / splat rather than a nested property.

The PR's own test run shows it:

```
$ bun test packages/server/src/errors.spec.ts packages/server/src/errors_internal_leak.spec.ts
error: Upstream connection error {"error":{}}
error: Unhandled internal error {"error":{}}
 16 pass  0 fail
```

So at `errors.ts:52` and `errors.ts:76` the response is generic **and** the log is empty — the message and stack exist nowhere. That's a net loss against the status quo, because pre-PR the message reached the logs anyway via `loggerMiddleware`'s response-body logging (`logMetadata.response`, on by default). Today a 500 is diagnosable from the response; after this, from nothing.

Fix is the splat form — `logger.error("Unhandled internal error", error)` — which I verified carries message + stack in both telemetry and non-telemetry modes, or an explicit `{ error: { name, message, stack } }`.

Worth saying that the *placement* is right and I wouldn't move it: 8 of the 53 `internalErrorToHttpError` call sites in `server.ts` (618, 1026, 1584, 1993, 2012, 2029, 2053, 2075) log nothing themselves, so the mapper is the only thing guaranteeing coverage. Only the serialization is wrong. The same `{ error }` shape is used at ~9 other sites in `server.ts`, which is presumably why it reads as correct — pre-existing and not this PR's to fix, but it means those log lines are empty too.

## Guard the log line once you're logging the real thing

Two guards worth folding into the same helper, because `error.message` here is unbounded and caller-influenced:

**Size.** A `MalloyError` embeds the whole compile error text, and `ConnectionError((error as Error).message)` on the `sqlQuery` path echoes a driver error that can contain the caller's entire SQL statement — bounded only by the request body limit. Add `stack` and it's multi-KB per failure. The Console transport caps nothing, and a JSON log sink with a max line size will drop the biggest lines, which are precisely the ones worth having.

**Control characters.** In the default format (`colorize + simple`, whenever `OTEL_EXPORTER_OTLP_ENDPOINT` is unset) output is newline-delimited plain text, so a message carrying `\n` — and the caller's SQL can — can forge log entries. `format.json()` escapes them, so this is specific to the non-telemetry path.

Neither is a reason to hold the change; both are cheap in the one helper the fix above needs anyway.

## The finding isn't closed at 5xx

`ServiceUnavailableError` → 503 still returns `error.message` verbatim, and two of its producers interpolate a raw internal error:

- `service/package.ts:569` and `:2036` — `Package-load worker pool unavailable: ${realError.message}`
- `service/environment.ts:818` — `Package compile worker unavailable: ${...}`

That's a worker-crash message, i.e. exactly the class that carries an absolute path or a stack fragment, on the most frequently exercised internal failure in the server. Its message also isn't caller-actionable — "the pool broke" is nothing the caller can fix — so it reads like the same case as the 500 fallback rather than a deliberate exception.

Related: the new comment at `errors.ts:11-13` says "Every other status this mapper produces is a client error (4xx)", and the PR body lists "400/403/404/422/424/409/413/503/504/501" under "Every 4xx client-error branch". 501, 503 and 504 are 5xx. Not cosmetic — that sentence is what the next person adding a 5xx branch will read, and it tells them the genericize-or-not question doesn't arise.

## Is the 502 change intended for driver-rejected SQL?

`ConnectionError` isn't only "upstream unreachable" — it's also the wrapper for the warehouse rejecting the caller's own statement:

- `connection.controller.ts:427` — `fetchSelectSchema`, behind `…/connections/{conn}/sqlSource`
- `connection.controller.ts:683` and `:700` — `sqlQuery` execution
- `connection.controller.ts:790` — temporary table

So `SQL compilation error: object DB.SCHEMA.FOO does not exist` now comes back as `Upstream connection error.` on endpoints an unattended agent is documented to use. That's the most actionable error message in the product and the caller is the only one who can act on it.

If the intent is the hosted case, one option is splitting the class — network/auth/host failure genericized at 502, driver-rejected-statement as 422 with the driver message. If that's more than you want here, narrowing this PR to the 500 fallback and leaving 502 verbatim would also be coherent. Either is fine by me; I'd just rather it were a decision than a side effect.

## Three other surfaces still echo the same detail

Same process, same posture, different door:

- **MCP** has its own mapper: `mcp/error_messages.ts:33` concatenates `${baseMessage}: ${errorMessage}`, reached for anything unrecognized *and* for `ConnectionError` (`mcp/handler_utils.ts:162`).
- **watch-mode** takes only `{ status }` from the mapper and builds its own body with the raw message (`controller/watch-mode.controller.ts:205, 230`) — and now also emits a spurious `logger.error` as a side effect of asking what status an error maps to.
- **Connection test** returns 200 with `errorMessage: "Connection test failed: <raw>"` (`connection.controller.ts:825`).

I'd argue the MCP one is defensible on purpose — the consumer is a local agent and the detail is the product — and the connection test is the caller's own config failing back to them. But nothing says so, and the mapper's comment reads as a closed invariant. A line naming these as deliberately out of scope would settle it; watch-mode looks like a straight miss rather than a choice.

## Smaller

- **Log volume / severity.** The 502 branch fires `logger.error` on every driver rejection, so a caller can drive unbounded error-level volume — and skew error-rate dashboards — with a loop of bad SQL. `warn` for `ConnectionError` and `error` for the 500 fallback would fit the two cases better. (Also worth knowing that ~45 of the 53 call sites already log, so 500/502 now log twice; harmless until the fix above makes both lines verbose.)
- **No correlation handle.** A user reporting "I got `Internal server error.`" has nothing to join to a log line. `loggerMiddleware` already extracts a W3C `traceId` — echoing that (or a request id) into the generic body is the difference between hardened and hardened-and-supportable.

## Two things for the write-up rather than the code

**A release note.** This is a client-visible change in behaviour that no check will surface: `message` keeps its name, keeps its type, stays required in the `Error` schema, and changes its basis to a constant. A strict generated client doesn't break and CI stays green, so the auto-generated PR list is all a reader gets — which is the "changed meaning of an existing signal" case `RELEASE_NOTES.md` asks for a `## [Unreleased]` section on, written in the PR that changes the behaviour. Reads as a patch to me rather than breaking, since nothing that loaded yesterday fails to load, but that's the maintainer's call. Worth noting the note's natural mitigation sentence — "the detail is logged server-side instead" — isn't true until the first finding is fixed, which is the strongest argument for treating that one as blocking.

**A public/private split at the raise point.** The real fix for the tension running through all of the above is for an error to carry both an internal `message` and a caller-safe public message, decided where it's raised rather than inferred from its class at the boundary. Nothing to do here, but one thing makes it cheaper than it looks when someone gets to it: the `Error` schema already declares a `details` field (`api-doc.yaml:5222`) that nothing ever populates. `message` generic plus `details` carrying safe specifics needs no schema change and no client regeneration, which beats minting a new field on a response shape that generates strict validating clients.

---

**Before merge, in order:** the `{ error }` serialization, since the fix's whole premise is that the detail moves to the log; a decision on whether 502 was meant to cover driver-rejected SQL; and either genericizing 503's raw worker message or narrowing the comment that says no such branch exists.

<sub>Reviewed with an assist from Dapper Ibex.</sub>

sagarswamirao:
Thanks for the review -- the serialization finding was the important one and you were right that it inverted the PR's premise. Walking your points in the order you raised them.

**The detail wasn't logged.** Fixed. The fields are copied out explicitly (`name`, `message`, `stack` as plain strings) rather than passed as `{ error }`, which serialized to `{"error":{}}` under both configured formats. I kept the placement at the mapper for the reason you gave -- a number of call sites log nothing themselves, so it is the only guaranteed coverage.

**Guards on the log line.** Both added, in the one helper. A 2000-char cap per field, and control characters replaced. One correction to my own first pass: the range now also covers U+0085 and U+2028/U+2029. `JSON.stringify` does not escape the line and paragraph separators, so they reached the rendered line verbatim under *both* formats, not only the non-telemetry one.

**Was 502 intended for driver-rejected SQL?** Yes, and it is now stated as a decision at the branch rather than left implicit, including the part that is uncomfortable: `ConnectionError` is one class covering both a statement the warehouse rejected and a host that is unreachable, so the same text that names the caller's own typo names an internal hostname when the failure is ours. Splitting the class is the right end state, and a table path that names nothing already took that route in 0.2.6 via `TableNotFoundError` (404/400).

**The 4xx-only comment, and 503.** You were right that the sentence was wrong -- 501, 503 and 504 are 5xx. Corrected, and while rewriting it I audited all ten 5xx throw sites and found three that do interpolate a raw error (the worker-pool and compile-worker throws behind 503). That is now named in the comment rather than papered over.

Generalizing those at the mapper turns out not to be possible: the load pool serializes an error across the worker boundary as a plain `Error`, re-wrapping only compile errors, and `BadRequestError` does not set `name` -- so an unusable-manifest error and a worker crash arrive indistinguishable. Blanking both suppresses a message the caller needs to fix their own config (there is a test that catches exactly this). The fix belongs at those throw sites, where the two are still separable, and I would rather do it there than guess at the boundary.

**The other surfaces.** The MCP mapper is fixed, scoped to the driver-wrapping class only -- an operational failure keeps its message, since blanking those returns callers to the generic text `classifyToolError` exists to avoid. watch-mode was a straight miss and is fixed at both sites; it now takes the mapper's message rather than only its status, and there is a test pinning that shape. The connection test keeps the driver's text, with a comment saying why it is the payload rather than a leak: the config arrived in that same request, so the host, port and user in any message are the caller's own values.

**Log volume.** Split. `logInternalFailure` takes a level, defaulting to `error`; the 502 branch and its MCP counterpart pass `warn`, so a caller looping bad SQL cannot fill the error log or move an error-rate dashboard meant to track the server's own faults. Tests assert the split both ways.

**No correlation handle.** Punting this one, and I think it is the right call rather than the convenient one. No error response in this server carries a handle today -- the `details` field the `Error` schema declares is populated nowhere, and the MCP JSON-RPC path answers with a bare `Internal server error` too -- so adding one only here makes this the single exception instead of the new convention. It is worth doing server-wide, and your `details` observation makes it cheap (no schema change, no client regeneration); the one thing it needs beyond echoing the existing traceId is an id that exists for callers who send no `traceparent`. I have left a note at the constants saying so, so the next person finds the reasoning rather than the gap.

**Release note.** Added, as an `[Unreleased]` section, and its mitigation sentence is now true. The persist-name section shipped in 0.2.5 with the parent PR.

housejester (approved):
Approving. The premise-inverting bug is gone, and I verified that rather than reading the summary:

```
$ bun test packages/server/src/errors.spec.ts packages/server/src/errors_internal_leak.spec.ts
warn: Upstream connection error connect ECONNREFUSED 10.0.0.1:5432 {"name":"Error","stack":"Error: connect ECONNREFUSED..."}
 28 pass  0 fail  62 expect() calls
```

Calling `logInternalFailure` directly, message, name and stack all land, and a message carrying `\n` plus NEL, U+2028 and U+2029 comes back as plain spaces. `callerSafe` is better than the split-the-class suggestion I made: it puts the distinction where it is known, and a new throw site that forgets about it leaks nothing rather than leaking by default. The tests pin both directions, both log levels, and the watch-mode body shape, and each of those assertions would have failed before the fix -- which is the property that makes them worth having.

Two things where you were right and I was not:

**The control-character range.** I said the injection risk was specific to the non-telemetry format because `format.json()` escapes newlines. That is wrong for exactly the characters at issue: `JSON.stringify` leaves NEL, U+2028 and U+2029 verbatim and escapes none of them, so JSON is not a backstop for those the way it is for `\n`. Your range and the reasoning in the docstring are right, and mine was under-specified.

**The correlation handle.** Taking the punt. "No response in this server carries one, so adding it only here makes this the exception rather than the convention" is a better argument than the one I made for adding it, and parking the reasoning at the constants means the next person finds a decision instead of a gap.

Five things from this pass. None is an exposure, and none blocks the approval -- but the first two are inside the helper this PR just added and they degrade the exact claim the release note now makes, so I would do them here.

## The metadata `message` key collides with winston's reserved one

`logger[level](summary, { name, message, stack })` does not emit `message` as a field. Winston fuses it into `info.message`, which is visible in the test output above and in JSON mode:

```
{"level":"error","message":"Unhandled internal error connect ECONNREFUSED 10.0.0.1:5432","name":"Error","stack":"..."}
```

So the summary is no longer a stable string: it is unique per failure, which means it cannot be a grouping key or an alert condition, and there is no queryable error-message field. Nesting under an `error` key restores both, and it is not the original bug returning, because these are plain strings rather than an `Error` instance. `errors.ts:88-92`.

## The stack loses every frame exactly when it is most wanted

`sanitize` is applied to `stack` and truncates at 2000 chars, but `error.stack` *begins* with `Error: <message>`. With a 3223-char message -- a wrapped driver error echoing the caller's SQL, the case `MAX_LOGGED_DETAIL_CHARS`' own docstring cites -- I measured `stack.slice(0, 2000)` containing **zero** ` at ` frames. Both fields then hold the same truncated message and the frames are gone, on the one branch (an unrecognized 500) where the frames are the whole point.

Capping the message and slicing the stack *past* the message prefix fixes it. Neither guard is pinned by a test today, which is presumably how it slipped: one assertion on a long-message error catches it, and would also keep the U+2028 range from being quietly narrowed later. `errors.ts:83-87`.

## Sibling paths classify the same driver text differently

`driverErrorToPublisherError` (`connection.controller.ts:197`) routes a table-not-found to `TableNotFoundError` -> 404 with the message, and it is used at `:229`, `:373` and `:518` -- all schema and source introspection. The three runSQL / temporary-table sites (`:785`, `:802`, `:892`) construct `new ConnectionError((error as Error).message)` directly and bypass it.

So `.../sqlSource` answers "table X does not exist" with a 404 carrying the driver's text, while `.../sqlQuery` flattens the identical string to `Upstream connection error.` That is where most of what the 502 comment calls the uncomfortable half of the trade actually lands, and the classifier that would recover it already exists.

I would rather name the outcome than the mechanism here, because routing those three through it is not free: it changes the status for a class of failures on a query endpoint, and a 404 from a POST whose endpoint does exist is its own oddity. The 0.2.6 precedent makes either choice defensible -- it is worth being a decision.

## The 503 deferral holds, but its stated reason is one line from evaporating

The comment says the pool "serializes an error across a worker boundary as a plain `Error`, so an unusable-manifest error and a worker crash arrive indistinguishable". The wire shape does carry `name` (`protocol.ts:270`) and `deserializeError` restores it (`package_load_pool.ts:885`). What actually makes the two indistinguishable is that `BadRequestError` never sets `this.name`, where `AccessDeniedError`, `NotQueryableError`, `PayloadTooLargeError` and `MaterializationEligibilityError` all do.

So the sentence reads as a property of the boundary when it is an omission in one class, and the comment will outlive the omission. Worth rewording. I have no argument with deferring the fix itself, though I would rather it were tracked as owed work than recorded only in a comment.

## The MCP fix's own argument covers the case it leaves out

The docstring's reason for withholding there is that doing it on one transport only "would leave the same text retrievable over the other". That is exactly right, and equally true of the unrecognized-error path it deliberately keeps: an `ENOSPC ... watch '/var/lib/publisher/environments/acme'` is generalized over HTTP and concatenated over `/mcp` -- the string your own fixture uses. The trusted-local-agent rationale would justify keeping it, but it would have justified not fixing `ConnectionError` there either. Worth saying which principle is the operative one, since the docstring currently supports both.

<sub>Reviewed with an assist from Dapper Ibex.</sub>

sagarswamirao:
Thanks -- and for running it rather than reading the summary. All five checked out; I verified each independently rather than taking them on faith, and you were right on all five. Four are fixed here. The third is a decision I would rather put to a maintainer than make unilaterally.

**The `message` key collision.** Fixed, and it reproduced exactly as you described. A live probe:

```
top-level: {"level":"error","message":"Unhandled internal error connect ECONNREFUSED 10.0.0.1:5432","name":"Error",...}
nested:    {"error":{"message":"connect ECONNREFUSED 10.0.0.1:5432","name":"Error",...},"level":"error","message":"Unhandled internal error"}
```

The summary was unique per failure, so it could not be a grouping key or an alert condition, and no field held just the error text. Now nested under `error`. Your note that this is not the original bug returning is the reason it is safe: those were non-enumerable properties of an `Error` instance, these are plain strings.

**The stack losing every frame.** Fixed, and your numbers replicate. A 3223-char message gives a 4-frame stack; `slice(0, 2000)` retained **zero**, and both fields then held the same truncated message. The message prefix is now dropped before capping, so the budget buys frames -- 0 before, 5 after. You were also right that neither guard was pinned, which is how it shipped: both are now, along with the control-character range, and each fails when its production change is reverted.

**The 503 deferral's stated reason.** You are right and my comment was wrong. `protocol.ts` carries `name`, `deserializeError` restores it, and what actually collapses the two is that `BadRequestError` never sets `this.name` while `AccessDeniedError`, `NotQueryableError`, `PayloadTooLargeError` and `MaterializationEligibilityError` all do. Reworded to name the omission rather than blame the boundary, so the comment does not outlive the thing it describes. Agreed it should be tracked as owed work rather than living only in a comment.

**The MCP docstring supporting both principles.** Also right. Withholding-on-one-transport would equally have justified generalizing the unrecognized branch I keep, so the docstring proved too much. It now says the line is drawn by class -- `ConnectionError` is the one class whose message is always a driver's text -- and that keeping the rest rests on the endpoint being localhost and unauthenticated by design, per `docs/security-posture.md`, not on the text being safe. If it ever fronts a remote caller, that branch generalizes with it.

**The sibling classification.** This is the one I am leaving, because you framed it as a decision and I think that is right. `driverErrorToPublisherError` exists and the three runSQL / temporary-table sites bypass it, so `sqlSource` answers a missing table with a 404 carrying the driver's text while `sqlQuery` flattens the identical string. That is genuinely where the uncomfortable half of the trade lands.

Punting it, deliberately rather than by omission. Routing those three through the classifier changes the response status for a class of failures on a query endpoint, and as you say a 404 from a POST whose endpoint does exist is its own oddity -- that is a product call about what a query endpoint promises, not a cleanup, and it does not belong in a security fix that is otherwise done. Nothing here is an exposure either way: the flattened string is the safe direction, so punting costs diagnosability rather than adding risk.

Worth filing as its own change, with the 0.2.6 precedent as the argument for whichever direction is picked. Happy to take it once someone owns the contract question.

PR #1109, opened 2026-09-03
fix(server): gate compile and sqlSource behind the query-concurrency cap

sagarswamirao:
## What

Puts `/compile` and `/sqlSource` behind `queryConcurrency()`, the admission
middleware the query routes already use, on every surface that reaches the same
work.

## Why

`/query`, `/sqlQuery` and `/sqlTemporaryTable` pass through `queryConcurrency()`,
which caps concurrent work at `PUBLISHER_MAX_CONCURRENT_QUERIES` so a flood cannot
saturate the pod. `/compile` and `/sqlSource` reach the same controllers and the
same warehouse -- compile resolves a source's schema with a `DESCRIBE` against the
connection, sqlSource runs a live introspection -- but took no slot, so a flood of
either bypassed the cap its siblings enforce.

The memory governor does not cover this. It refuses only once RSS has already
crossed the high-water mark, so a burst that is admitted while memory looks healthy
is exactly the case it cannot prevent, and compile's cost is CPU on the shared event
loop rather than resident memory.

## How

Seven registrations, eight lines:

- The compile route and both sqlSource routes in `server.ts`.
- Their legacy `/projects/...` twins in `server-old.ts`. Leaving the legacy surface
  ungated would relocate the bypass rather than close it.
- The MCP `malloy_compile` tool, which takes a slot directly with
  `tryAcquireQuerySlot("mcp:compile")` rather than through Express middleware. Same
  compiler, one surface over.
- `PUT /environments/:env/packages/:pkg/models/*?`, the dashboard save. It arrived
  from main after this branch opened; it compiles the submitted text and rewrites
  the package under its lock across a full reload, which is more of the work this
  cap bounds than one `/compile` is.

## Testing

Three behavioural cases stand up a real Express app with the wiring `server.ts`
registers, cap at 1, and assert a second request gets a 503 while the first holds
the slot.

Those build their own app, so they pin the middleware rather than the
registrations: reverting the production gates left them green. Eight further
assertions read the registrations themselves, one targeted match per route literal
plus the MCP tool, in the shape `data_apps_route_parity.spec.ts` uses -- no window
extraction and no comment stripping, which is what made the source-scan spec this
replaced a maintenance liability. Each was verified by reverting its gate and
watching that assertion fail.

One trap the coverage had to handle: the dashboard-save path is also registered as
a GET that only reads and is correctly ungated, so a matcher on the path alone pins
the wrong route. That entry matches the verb and the path as a pair.

## Behavior

Under load, `/compile`, `/sqlSource` and the dashboard save can now return 503, the
same admission behaviour the query routes already have. The cap is
`PUBLISHER_MAX_CONCURRENT_QUERIES`, default 32, `0` disables it. A deployment that
finds it tight once authoring traffic counts against it raises the cap.

## Deliberately not here

Two things came up in review and are recorded rather than changed.

**Compile and sqlSource cannot be cancelled**, so `runWithQueryTimeout` does not
wrap them the way it wraps every other slot holder. The constraint is upstream:
`abortSignal` exists on `RunSQLOptions`, the query-execution path, and the compiler
API exposes no equivalent. Both were uncancellable before this change and still
are. Wrapping them anyway would answer the caller with a 504 and release the slot
while the compile carried on, leaving the pod holding more work than it had slots
outstanding. Cancellable compilation is upstream work and a separate task.

**A slot is taken before the per-package lock**, so K requests against one package
hold K slots with K-1 parked, charged pod-wide because the pool is module-scoped.
This predates the change -- the query routes already took a slot ahead of the same
lock, and `withPackageLock` reached `compileSource` with the dashboard builder.
Admitting compile widens who can trigger it rather than introducing it; closing it
means acquiring inside the lock or a pool per package.


Sha-Bang (commented):
The eight lines of production change are right, and the reasoning holds: compile resolves source schemas against the connection and sqlSource runs a live introspection, so both drive the work the cap exists to bound, and gating the `server-old.ts` twins in the same PR closes the bypass rather than moving it. No argument with any of that.

Four things, three of which I'd want before merge.

**The test asserts source text, and this repo already decided that isn't enough for this exact middleware.** `authorize_bypass_wiring.integration.spec.ts` opens by saying a source assertion "does not send a request" and then builds a real Express app — with `queryConcurrency()` in the chain — to make the claim behavioural. `concurrency_gate.spec.ts` is 128 lines of regex over `server.ts` to establish something that harness already stands up in about 20. I confirmed the scan does catch a plain removal (I deleted the gate from the package-scoped sqlSource route and the suite went red), so it isn't worthless — but it proves a string is present, not that a second request gets a 503 while the first holds a slot, and the window-extraction heuristic is a maintenance liability the behavioural version doesn't have. Could this reuse the `authorize_bypass_wiring` shape: cap of 1, hold one compile open, assert the next gets 503?

**`docs/configuration.md` still describes the old scope.** Line 108 reads "Per-pod cap on simultaneous in-flight *queries* (HTTP + MCP share the same slot pool)", and line 258 labels the rejection counter the same way. After this PR, compile and sqlSource draw from that pool too, so an operator sizing the cap from that row sizes it for query traffic and gets 503s on an authoring surface they didn't know was in scope. Both lines want a sentence.

**MCP compile stays ungated, and it's the same argument the PR makes about `server-old.ts`.** The spec header states the contract as "every route that drives compile or sqlSource work must include `queryConcurrency()`". `mcp/tools/execute_query_tool.ts:188` takes a slot via `tryAcquireQuerySlot("mcp:executeQuery")`; `mcp/tools/compile_tool.ts` and `search_database_schema_tool.ts` take none. So a flood of `malloy_compile` over MCP still bypasses the cap the HTTP twin now enforces — which is the unbounded-bypass reasoning in the legacy-route comment, one surface over. Either gate them or say in the PR why MCP is out of scope.

Non-blocking, worth a thought: compile and sqlSource now share one 32-slot pool with warehouse queries. Compile is cheap and bursty — an agent authoring loop or a notebook compiling on edit can hold slots that a real query then can't get, and the query is the one that matters. Is a shared pool the intent, or would compile want its own smaller cap?

Two small ones: `Finding F-11` in the spec header refers to an audit that isn't in this repo, so a reader in six months can't resolve it — the paragraph underneath says everything the label was doing. And the PR body is empty; the header comment is doing the work a description should.

Ran against `b52d35b8`: `bun test packages/server/src/concurrency_gate.spec.ts` — 6 pass. The removal experiment above is the evidence for the first finding.

packages/server/src/concurrency_gate.spec.ts
```diff
@@ -0,0 +1,128 @@
+// Copyright (c) Credible Data Inc.
+// SPDX-License-Identifier: MIT
+
+import { describe, expect, it } from "bun:test";
+import { readFileSync } from "fs";
+import { resolve } from "path";
+
+/**
```
Sha-Bang: `Finding F-11` points at an audit that doesn't live in this repo, so nobody reading this file later can resolve the label. The two sentences under it already say exactly what was wrong and what the contract is — drop just the identifier and keep the prose.

Same note applies to `Finding F-9` in #1110.

packages/server/src/concurrency_gate.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+const routeGatePresence = (source: string, pathSuffix: string): boolean[] => {
+   const needle = pathSuffix + BACKTICK;
+   const results: boolean[] = [];
+   let from = 0;
+   for (;;) {
+      const literalEndSuffix = source.indexOf(needle, from);
+      if (literalEndSuffix === -1) break;
+      const windowStart = literalEndSuffix + needle.length;
+      from = windowStart;
```
Sha-Bang: This is the part I'd most like to see go. `routeGatePresence` hand-rolls a parser: find the closing backtick of a route literal, guess where the handler starts with `/async\b|\((?:_?req|_)\b/`, strip comments inside that window but deliberately not outside it because route wildcards contain `/*`. Every one of those is a real constraint and the comment explains each honestly — which is the tell that the approach is fighting the problem rather than fitting it.

What it buys is proof that the characters `queryConcurrency()` sit between two other strings in `server.ts`. What it can't tell you is whether a request actually gets admission-controlled, so a refactor that keeps the token and breaks the gate — the middleware registered but short-circuited, `getMaxConcurrentQueries()` read from the wrong place, the handler responding before the chain runs — stays green.

I did check it isn't inert: removing `queryConcurrency()` from the package-scoped sqlSource route in `server.ts` turns the modern-surface sqlSource case red, so it catches the plain regression it was written for.

But `authorize_bypass_wiring.integration.spec.ts` in this same directory already made this call for this same middleware — its header says a source assertion "does not send a request" and then stands up a real Express app with `queryConcurrency()` in the chain and only the controller stubbed. That harness is ~20 lines and it's already imported the gate. What if this became: `PUBLISHER_MAX_CONCURRENT_QUERIES=1`, hold a compile open, assert the second request comes back 503? That tests the thing the PR is actually claiming, on both surfaces, without the parser.

Sha-Bang (commented):
Two inline, detail on the points in the verdict above.

sagarswamirao:
All four addressed, and the behavioural-test point was the right call -- it found a real bug the source scan could not have.

**The test asserts source text.** Replaced. `concurrency_gate.integration.spec.ts` uses the `authorize_bypass_wiring` shape you pointed at: cap of 1, hold one request open on a deferred promise, assert the second comes back 503 while the first still returns 200, so it proves the held slot was released rather than only that something refused. There is a cap-of-0 case asserting both requests pass, so the 503 cannot come from the harness. It installs the same `internalErrorToHttpError` handler `server.ts` registers -- without that the app returns Express's default 500, which is exactly the trap the source scan could not see. The 128-line regex file is deleted.

Worth recording what that swap caught: the gate rejects via `next(error)`, not by throwing, so the 503 only exists if the error handler is registered after the routes. The scan would have stayed green with the gate wired and the status wrong.

**`docs/configuration.md` described the old scope.** Fixed. The row now says the pool covers queries plus the two authoring surfaces that reach the database, names `compile` and `sqlSource`, and tells an operator to size for authoring traffic as well as query traffic. The rejection-counter row is updated too.

**MCP compile stays ungated.** Gated. `compile_tool.ts` takes a slot via `tryAcquireQuerySlot("mcp:compile")` with the release in a `finally`, and the refusal routes through `classifyToolError`'s back-pressure branch, so an agent gets a retry suggestion rather than advice to check its Malloy syntax. Same reasoning as the legacy routes: leaving one surface open just moves the bypass.

**`Finding F-11`.** Gone -- the label lived in the deleted file, and nothing else carries it.

On the shared pool: kept shared, and the tradeoff is now written down rather than implied. The release note says a deployment sitting near its cap may start seeing 503s on compile and sqlSource it did not see before, and the config row says to size for authoring traffic. A separate smaller cap for compile is the natural follow-up if authoring traffic turns out to be burstier than query traffic in practice; I would rather set that from a real signal than guess a split now.

One thing this branch does not close, and I do not think it should here: the schema-introspection routes (`/schemas`, `/schemas/{schema}/tables`, both scopes) and the MCP `search_database_schema` tool are still ungated, and they do the same live-warehouse work `sqlSource` does -- `listTables` resolves a real connection and issues an introspection. They are byte-identical before and after this PR and carry no gate on `main` either, so it is pre-existing rather than a regression, and outside the compile/sqlSource scope. But the "leaving one surface ungated relocates the bypass" argument applies to them verbatim, so it is worth its own change.

packages/server/src/concurrency_gate.spec.ts
sagarswamirao: Dropped, and the prose it labelled went with the file -- the source-scan spec is deleted, so the label has no remaining home. The behavioural replacement states the contract in its header without an identifier that only resolves outside this repo.

Same fix applied on the SSH host-key branch: that `Finding F-9` label lived in `proxy_hostkey.spec.ts`, which is also deleted.

packages/server/src/concurrency_gate.spec.ts
sagarswamirao: Agreed, and it is gone. `concurrency_gate.integration.spec.ts` uses the `authorize_bypass_wiring` shape you pointed at: cap of 1, hold one request open on a deferred promise, assert the second returns 503 while the first still returns 200 -- so it proves the held slot was released, not just that something refused. A cap-of-0 case asserts both pass, so the 503 cannot come from the harness.

Your read on what the scan could not see was exactly right, and it caught a real bug in the swap: the gate rejects via `next(error)` rather than throwing, so the 503 only exists if the error handler is registered after the routes. The behavioural app installs the same `internalErrorToHttpError` handler `server.ts` uses -- without it the response is Express's default 500. The scan would have stayed green with the gate correctly wired and the status wrong.

housejester:
Reviewed this fresh, downstream of @Sha-Bang's review rather than re-running it. Their four points were the right four and all four are addressed — the docs-scope fix and the MCP `compile_model` gate are both correct, and `classifyToolError` already homes `ServiceUnavailableError` to `BACK_PRESSURE_SUGGESTIONS` (`handler_utils.ts:82`), so an MCP refusal tells the agent to retry instead of sending it hunting for a typo. `RELEASE_NOTES.md` is additions-only against `main`, one heading, no duplicate from the `f12` merge. All 14 checks green.

The eight lines of production change are individually right. Three things below, plus a scoping ask.

**The shape of the change.** `PUBLISHER_MAX_CONCURRENT_QUERIES` stops being a *query* cap and becomes a *warehouse-touching-work* cap — but the two new members have different lifetime properties than the incumbents. Every pre-existing slot holder is wrapped in `runWithQueryTimeout` and is lock-free (a query fast-paths the package cache without the mutex). Neither new holder is either of those things. That asymmetry is where all three findings live.

---

### 1. The behavioural spec doesn't pin the registration, and the scan that did was deleted

`concurrency_gate.integration.spec.ts:107` builds its own Express app with `queryConcurrency()` hardcoded into the route registration. That pins the middleware's behaviour — which `query_concurrency.spec.ts` and the `authorize_bypass_wiring` harness already do — not that the compile/sqlSource routes are registered with it. `0c0c8a7c` deleted `concurrency_gate.spec.ts`, which you'd confirmed catches a plain removal. Net: the PR removes the only assertion that fails if the fix is reverted.

Verified by reverting it — deleted `queryConcurrency(),` from the modern compile route at `server.ts:1958` and re-ran the new spec:

```
$ bun test src/concurrency_gate.integration.spec.ts
 3 pass  0 fail   Ran 3 tests across 1 file. [674.00ms]
```

(restored afterwards; working tree clean.)

Worth noting the precedent the spec header cites keeps **both** halves — `authorize_bypass_wiring.integration.spec.ts:193-204` is a `readFileSync` source assertion over `server-old.ts` sitting alongside the behavioural harness, precisely because the behavioural half can't see the registration. So "superseded by the behavioural one" isn't quite what that file does. The thing that made the old scan a maintenance liability was `routeGatePresence`'s hand-rolled parser, not the idea — a registration assertion can come back without it.

### 2. The slot is acquired before a serializing mutex, so queued compiles burn the pool

This is the concrete mechanism under @Sha-Bang's non-blocking "would compile want its own smaller cap?", and I think it promotes that question to blocking.

`queryConcurrency()` runs as middleware, then `CompileController.compile` (`compile.controller.ts:15`) calls `Environment.compileSource`, whose entire body is inside `withPackageLock` (`environment.ts:578`) — an `async-mutex` `runExclusive`, i.e. FIFO queueing. Compiles against one package can only run one at a time by design. So K concurrent compiles on a single package hold K slots with K−1 of them parked on the mutex doing zero work.

At the default cap of 32, ~32 concurrent compiles against one package — the "agent loop or a notebook that compiles on every edit" traffic the release note itself names — consume the whole pod's pool and 503 every `/query`, `/sqlQuery`, `/sqlTemporaryTable` and `mcp:executeQuery` **pod-wide, across every package and every environment** (the pool is module-scoped, `query_concurrency.ts:20`), while the pod does one compile's worth of work. It compounds with anything else holding that mutex: `installPackage`'s rename swap, `deletePackage`, a package reload. Before this PR those queued compiles consumed no query slots at all.

Second-order: `publisher_query_active_slots` counts parked waiters, so `active / max` reads 100% utilization on an essentially idle pod — on the dashboard the `docs/configuration.md` change points operators at.

Recommending the outcome rather than a mechanism: mutex-queued compiles shouldn't count against the pool. Acquiring after the lock, a separate compile pool, or something else — they trade differently and that's your call.

### 3. Both new holders are unbounded, and MCP compile has no backstop at all

`runWithQueryTimeout` wraps `query.controller.ts:94`, `connection.controller.ts:766/791/868`, `model.controller.ts:154` and `execute_query_tool.ts:229` — every pre-existing slot holder. It wraps neither `CompileController.compile` nor `getConnectionSqlSource` (`connection.controller.ts:492`, whose `fetchSelectSchema` takes no `AbortSignal`). So a `DESCRIBE` against a warehouse that's resuming, or a compile whose schema fetch stalls, holds its slot for as long as the driver takes.

HTTP has a coarse backstop: `mainServer.timeout = 600000` (`server.ts:2285`) destroys the socket at 10 minutes and `res.on("close")` releases. That's 2× the query bound and it's accounting-only — the driver call keeps running. **MCP compile has neither.** The release is a `finally` around the awaited promise (`compile_tool.ts:240`), not tied to a socket, so a hung compile holds its slot until the promise settles, full stop. `mcp:executeQuery` avoids this because its slot wraps `runWithQueryTimeout`.

This also makes an in-code invariant false: `query_concurrency.ts:115` documents "Query timeout (Step 5) prevents one query from monopolising a slot indefinitely" as a property of everything in the pool. Either bound both paths, or narrow that docstring to say which holders it covers.

### Scoping ask: say what's deliberately out

The two admission controls now cover inverted halves of the authoring surface:

| | concurrency cap | memory governor |
|---|---|---|
| query / sqlQuery / sqlTemporaryTable / notebook / `mcp:executeQuery` | yes | yes |
| `compile`, `sqlSource` | yes (this PR) | **no** |
| `reload_package` / `?reload=true` | **no** | yes |
| REST `/connections/.../schemas`, `/tables` (both surfaces) | **no** | **no** |
| MCP `search_database_schema` | **no** | yes |

Two consequences. Under memory back-pressure `compile` and `sqlSource` are still not shed — `assertCanAdmitQuery` is called by every other query controller but neither of these (`connection.controller.ts:655/842` have it, `:492` doesn't). And `?reload=true` recompiles the whole package — strictly more compile work than one `/compile` — while taking no slot; `listSchemas`/`listTables` (`connection.controller.ts:439/463`) are live introspection by this PR's own definition and take neither control.

Not asking for any of that here. But the PR's reasoning — "leaving one surface ungated relocates the bypass rather than closing it" — lands on them too, so one sentence in the release note naming what is deliberately still open would stop the next reader taking the note as "done".

### Minor

`active` (`query_concurrency.ts:20`) is module-global and the new spec never resets it, though `resetActiveQueryCountForTesting()` exists for exactly this. Any spec sharing the process that leaks a slot makes the cap-1 cases fail on `first === 200` rather than on the thing they test. A `beforeEach` reset is free insurance.

### One stale claim in the description

**How** still says "A source-scan spec pins that every compile/sqlSource registration on both surfaces carries the gate…". That file is deleted and isn't among the PR's six. The description also doesn't mention the `compile_tool.ts` gate, which is in the diff. Worth fixing mainly because the claim it makes is exactly the guarantee finding 1 says is missing — as written it would talk a reader out of noticing.

---

Three to resolve before merge, in order: restore an assertion that fails when `queryConcurrency()` leaves a registration; decide the shared-pool question with the mutex convoy in view rather than on "compile is cheap"; bound the MCP compile slot hold, or say why a driver stall holding a slot indefinitely is acceptable.

*(Review by "Deft Gary" — an agent-assisted pass. Claims above were checked against the source at `156ded8c`; the revert experiment is the evidence for finding 1.)*

sagarswamirao:
Synced with main and worked through the four findings. All three blocking ones are
addressed; the fourth is answered rather than changed.

**The source-scan spec is gone.** `concurrency_gate.spec.ts` is deleted and replaced by
`concurrency_gate.integration.spec.ts`, built on the `authorize_bypass_wiring` shape you
pointed at: a real Express app, cap of 1, one request holding the slot, the next asserted
to get a 503. The run shows it doing the thing rather than matching a string:

```
warn: Rejecting query: 1/1 slots in use (PUBLISHER_MAX_CONCURRENT_QUERIES).
  {"route":"/api/v0/environments/:environmentName/packages/:packageName/models/*?/compile"}
warn: Rejecting query: 1/1 slots in use (PUBLISHER_MAX_CONCURRENT_QUERIES).
  {"route":"/api/v0/environments/:environmentName/connections/:connectionName/sqlSource"}
```

3 pass, 0 fail. The window-extraction heuristic and its maintenance cost go with the old
file.

**`docs/configuration.md` now states the new scope.** The `PUBLISHER_MAX_CONCURRENT_QUERIES`
row reads "Per-pod cap on simultaneous in-flight work" and names both authoring surfaces
and why each reaches the database, with the sizing consequence spelled out.

Your second doc pointer turned up one more line than the two you named: the metric table
documented `publisher_query_concurrency_rejections_total` as labeled by an HTTP route "or
`mcp:executeQuery`", which the MCP gating below makes incomplete. It now names both MCP
tools.

**MCP compile is gated.** `compile_tool.ts` takes `tryAcquireQuerySlot("mcp:compile")`, so
the contract in the spec header holds across both surfaces rather than stopping at HTTP.
You were right that leaving it was the same bypass reasoning one surface over.

**On the shared pool** -- not changed, and I think the shared pool is right for now. The
cap exists to bound what a pod is concurrently doing against its connections, and a
compile that resolves schemas is doing that whether or not it is cheaper than a query.
Splitting it into two pools means two numbers to size and a way for the sum to exceed what
the pod can carry, which is the failure the single cap prevents. If authoring traffic does
starve queries in practice the metric now distinguishes them by route, so that would be
visible before it needs a design.

The `Finding F-11` label is gone from the spec header, and the PR body is written.

Gates on the merged tree: `build`, `typecheck` (five packages) and `lint` all exit 0.

housejester:
Re-reviewed at `ea4bebc2`. All 15 checks green, and `bun run test:unit` in `packages/server` is 3819 pass / 0 fail across 173 files here. The eight lines of production change are unchanged and still right.

Two things are new since the last pass, and both come from the main sync rather than from the change itself.

### 1. The merge deleted a release note's body and left its heading

`RELEASE_NOTES.md:402` now reads `## [0.2.7] — 500 and 502 responses no longer echo the internal error` followed immediately by `## [0.2.6]`. The 36 lines under it are gone — the whole rationale for withholding internal error text from 5xx and from `/mcp`, and the two things that note is explicit about *not* generalising.

It is the merge, not an edit:

```
$ for c in 156ded8c fb1d7238 567ff7a7 ea4bebc2; do
    git show $c:RELEASE_NOTES.md | grep -c 'A 500 or a 502 returned'; done
1  1  0  0     # present pre-merge and on the merged main point; gone at 567ff7a7
```

Nothing else in the merge is damaged. Against `fb1d7238` — the commit the merge actually took, not `origin/main` — the branch has 40 deletions total: 36 are this, and the other 4 are the two rewritten `docs/configuration.md` rows and one replaced comment in `server-old.ts`.

Recommend restoring the section from `fb1d7238`. Worth flagging for the next sync too, since `mergeStateStatus` is `BEHIND` and main is four commits ahead: a heading-level check passes this cleanly (the `^## ` set is still additions-only, exactly one new heading) because the heading survived and only the body was lost. `git diff <the-merged-main-commit>..HEAD --stat` and reading every deletion is what catches it.

### 2. The merge also added a new ungated compile path, inside this PR's own contract

`PUT /api/v0/environments/:env/packages/:pkg/models/*?` (`server.ts:1683`) carries no `queryConcurrency()`. It runs `putDashboardSource`, which compiles the submitted text (`dashboard.controller.ts:198`) and then calls `writeModelFileTransactional` (`environment.ts:1528`), holding the package lock across the write, **a full package reload**, and the rollback-and-reload on failure.

That is strictly more compile work than one `/compile`, and it takes no slot. It also sharpens the shared-pool question below from the other side: a single dashboard save can now convoy up to 32 slot-holding compiles behind the package mutex while holding zero slots itself.

---

### The four findings from the last pass are unchanged at `ea4bebc2`

Re-verified rather than assumed. Flagging it in case that review was missed rather than declined — the 2026-09-17 comment reads as a reply to the 09-04 thread.

**No assertion fails when the fix is reverted.** `concurrency_gate.integration.spec.ts:95,132` registers the middleware into an app it builds itself, so it pins the middleware, not the routes. Reverting *both* production gates — `queryConcurrency(),` from `server.ts:1972` and `tryAcquireQuerySlot("mcp:compile")` from `compile_tool.ts:173` — leaves it at `3 pass 0 fail` (restored afterwards; tree clean). Nothing else covers it: only three specs in `packages/server/src` mention `queryConcurrency`, and all three build their own app.

Recommending the outcome, not a mechanism: something has to go red when a route loses its gate. It does not need the `routeGatePresence` parser that made the old spec a liability — `authorize_bypass_wiring.integration.spec.ts:193-204` keeps a `readFileSync` half beside its behavioural half for this exact reason, and `data_apps_route_parity.spec.ts` pins a route from source with one targeted regex on the route literal.

**The slot is acquired before a serializing mutex.** `CompileController.compile` → `Environment.compileSource`, whose body is `return this.withPackageLock(...)` (`environment.ts:600`) → `runExclusive` (`environment.ts:1470`). K concurrent compiles on one package hold K slots with K−1 parked doing nothing; at the default cap that 503s every `/query`, `/sqlQuery`, `/sqlTemporaryTable` and `mcp:executeQuery` pod-wide across every package and environment, since the pool is module-scoped (`query_concurrency.ts:21`). Second-order: `publisher_query_active_slots` counts parked waiters, so `active / max` reads 100% on an essentially idle pod — on the dashboard `docs/configuration.md` now points operators at.

**Both new holders are unbounded, and MCP compile has no backstop.** `runWithQueryTimeout` wraps every pre-existing slot holder (`query.controller.ts:94`, `connection.controller.ts:772/797/874`, `model.controller.ts:154`, `execute_query_tool.ts:229`). It wraps neither `CompileController.compile` nor `getConnectionSqlSource` (`connection.controller.ts:498`, whose `fetchSelectSchema` takes no `AbortSignal`). HTTP has the coarse `mainServer.timeout = 600000` with `res.on("close")` releasing — 2× the query bound and accounting-only. `mcpServer.timeout` exists too (`server.ts:2471`) but is not wired to the slot: the release is a `finally` on the awaited promise (`compile_tool.ts:239`), so a socket teardown releases nothing. Either bound both paths, or narrow `query_concurrency.ts:115`, which documents "Query timeout (Step 5) prevents one query from monopolising a slot indefinitely" as a property of everything in the pool.

**Say what is deliberately out.** The new row above joins `?reload=true` (governor yes, cap no), REST `/connections/.../schemas` and `/tables` on both surfaces (`server.ts:1258-1359`, neither), MCP `search_database_schema` (governor at `search_database_schema_tool.ts:396`, no slot), and `getConnectionSqlSource` missing `assertCanAdmitQuery` while its siblings at `connection.controller.ts:661/848` have it. Not asking for any of it here — one sentence in the release note naming what is still open, so the next reader does not take the note as "done".

### Minor

`active` (`query_concurrency.ts:21`) is module-global and the spec never resets it, though `resetActiveQueryCountForTesting()` exists at `:103`. The suite runs 173 files in one process and passes today, so it is latent — but a spec that leaks a slot makes the cap-1 cases fail on `first === 200` rather than on the thing they test.

Bounding the blast radius down: no SDK or Console code path calls `/compile` or `/sqlSource`, so the new 503 reaches agents over MCP and REST clients only, not the UI. Worth a line in the release note if it is being edited anyway.

### One stale claim in the description

**How** still says "A source-scan spec pins that every compile/sqlSource registration on both surfaces carries the gate…". That file is deleted and is not among the PR's six. The description also still does not mention the `compile_tool.ts` gate, the docs rows, or the release note. It matters mainly because the claim it makes is exactly the guarantee the first finding says is missing.

---

**Three to resolve before merge:** restore the `0.2.7` release-note body the merge deleted; restore an assertion that fails when a route loses its gate; decide the shared-pool question with the convoy in view, now that an ungated dashboard save holds the same mutex through a full package reload.

*(Review by "Snazzy Heron" — an agent-assisted pass. Claims checked against the source at `ea4bebc2`; the revert experiment is the evidence for the registration finding.)*

sagarswamirao:
Synced with main and worked the four. Two are fixed, two are answered rather than
changed, and I think the reasoning matters more than the outcome on both.

**The release-note body is restored.** You were right that the merge dropped it, and
right about the method: a heading-level check passes that cleanly because the heading
survived. I now diff against the commit the merge actually took and read every
deletion. The branch is additions-only against it.

**A route that loses its gate now fails a test.** The behavioural cases build their
own app, so they pinned the middleware rather than the registrations -- your revert
experiment was correct and my earlier reply claiming otherwise was wrong. There are
now eight assertions reading the registrations themselves, one targeted match per
route literal across both surfaces plus the MCP tool, in the shape
`data_apps_route_parity.spec.ts` uses. No window extraction, no comment stripping.
Verified by reverting each gate in turn: each fails and names what lost it.

One trap worth recording, since it is the same class that made the old source-scan
spec a liability: the dashboard-save path is also registered as a GET that reads and
is correctly ungated, so a matcher on the path alone pins the wrong route. That entry
matches the verb and the path as a pair.

**The dashboard save is gated.** `PUT /environments/:env/packages/:pkg/models/*?`
compiles the submitted text and writes it under the package lock across a full
reload, which is more of the work this cap bounds than one `/compile` is. It arrived
from main after this branch opened.

## The two I did not change

**The unbounded holders are pre-existing and not part of this finding.** Compile and
sqlSource were uncancellable before this change and still are; admitting them to the
pool does not alter that. The constraint is upstream: `abortSignal` exists on
`RunSQLOptions`, the query-execution path, which is why `runWithQueryTimeout` works
for `/query` and `execute_query`, and the compiler API exposes no equivalent, so
`Environment.compileSource` and `fetchSelectSchema` have nothing to forward.

Wrapping them here anyway would answer the caller with a 504 and release the slot
while the compile carried on regardless, leaving the pod holding more work than it
had slots outstanding -- worse than an honest unbounded hold. Cancellable compilation
is upstream work and belongs in its own task, separate from this finding.

**The slot-before-lock ordering predates this change.** The query routes already took
a slot ahead of the same package lock; `withPackageLock` came to `compileSource` with
the dashboard builder two days ago. Admitting compile widens who can trigger the
convoy rather than introducing it, and closing it properly means acquiring inside the
lock or a pool per package -- the same memory arithmetic that kept the per-tenant
caps out of the bypass PR.

Neither is the end of the world in the meantime: an operator who finds the cap too
tight once authoring traffic counts against it raises
`PUBLISHER_MAX_CONCURRENT_QUERIES` to 64 or 128, which is a better position than
leaving these routes ungated. The release note now says so.

Gates on the merged tree: `build`, `typecheck` (five packages) and `lint` all exit 0.

housejester:
Re-reviewed at `fb7a8034`. Three of the four are closed and I verified each one; the two you answered rather than changed are below, one of which I accept.

`bun run test:unit` here is 3889 pass / 0 fail across 178 files.

### Closed

**The registration pin does what it says.** I reverted each of the seven gates individually — the three `server.ts` routes, the three `server-old.ts` twins, and the MCP slot. Every one turns the suite red and names the route or the tool that lost it. That is the guarantee the change is making, and it is now guarded.

**The dashboard save is gated**, with the entry anchored on verb and path as a pair. The GET-on-the-same-path trap was a good catch and worth the comment you left on it.

**Release notes restored, and the second sync is clean too** — 0 deletions against `bc703c9e`, heading set additions-only.

**The description now matches the diff.**

### Two things to fix

**Windows CI is red on the new test.** `Cross Platform Tests (windows-latest)` fails one case: ``gates app.put(${API_PREFIX}/…/models/*?` in server.ts``. The verb anchor at `concurrency_gate.integration.spec.ts:320` embeds a literal `\n`:

```js
source.indexOf(`${verb}\n   \`${literal}`)
```

There is no `.gitattributes`, so a Windows checkout is CRLF and that anchor never matches. It returns `-1`, `indexOf(literal, -1)` restarts at 0, finds the **GET** registration of the same path — correctly ungated — and fails. Reproduced locally by converting `server.ts` to CRLF: same single case, same name.

Normalizing on read (`readFileSync(...).replace(/\r\n/g, "\n")`) fixes it and protects the other six entries from the same trap later. Separately, `toBeGreaterThan(-1)` cannot catch this: `start` lands on a real index, so the failure reads as "the PUT registers no `queryConcurrency()`" while the assertion is actually looking at the GET. Asserting the *anchor* was found is the sharper guard.

**The scan can be satisfied by a comment.** Delete the gate from the compile route and put this in its place:

```js
// queryConcurrency() is deliberately omitted here: see #1234.
```

`11 pass / 0 fail`. The window runs from the route literal to `async (req, res)` with no comment stripping, so any comment naming the token satisfies it — and that is a plausible shape, since someone removing a gate would naturally document why, naming the thing they removed.

This one is mine. I recommended "one targeted regex on the route literal, in the shape `data_apps_route_parity.spec.ts` uses" and said the parser was the problem rather than the idea. Comment-stripping was one of the things listed against the old spec, so reading that as "no comment stripping" was a fair reading of what I wrote — it was my recommendation that was under-specified, not your implementation of it. The spec header and the description both now say "no window extraction and no comment stripping, so it cannot be satisfied by a comment", and the second half does not follow from the first. `authorize_bypass_wiring.integration.spec.ts:199-201` already does the one-line strip.

### On the two you did not change

**The timeout argument is right, and the residual is not the timeout.** I checked: `abortSignal?: AbortSignal` is on `RunSQLOptions`, and `run_sql_options.d.ts` is the only file in the malloy dist that declares it, so the compiler genuinely has nothing to forward. Answering with a 504 while the compile carried on would leave the pod holding more work than it has slots outstanding — agreed, and agreed it is upstream work.

Two things survive that argument, both cheap. `query_concurrency.ts:115` still documents "Query timeout (Step 5) prevents one query from monopolising a slot indefinitely" as a property of everything in the pool, which is now false for three of its members; your reply above is the right content for that docstring. And MCP compile differs from HTTP in kind rather than degree: HTTP's `res.on("close")` returns the slot when the socket is destroyed at ten minutes, while `compile_tool.ts:239` releases in a `finally` on the promise, so nothing ever returns it. Uncancellable before and after is true, but the slot hold is new, and on that surface there is no path that ends it.

**The ordering point rests on two claims I could not confirm.**

> the query routes already took a slot ahead of the same package lock

Not on the hot path. `query.controller.ts:79` calls `getPackage(pkg, false)`, and `environment.ts:1793` returns the cached package *before* `withPackageLock` — the comment at `:1785-1792` calls that the fast path and says it must stay in-memory only. A query reaches the mutex only on a cold load or a reload. Compile takes it unconditionally, every time.

> `withPackageLock` came to `compileSource` with the dashboard builder two days ago

```
$ git log -L 600,601:packages/server/src/service/environment.ts
433e5f5b 2026-05-19 fix(server): stage-and-swap package installs + extend per-package
                    mutex to all disk reads (#756)
```

So the narrower version is the one that holds: the mutex queueing predates this change, but queued compiles *consuming query slots* does not. That is a new coupling rather than a pre-existing one. I am not asking you to change it — you have weighed it and it is your call — but it is worth the record saying which half is new.

One note on the interim, since the release note now recommends it: raising `PUBLISHER_MAX_CONCURRENT_QUERIES` to 64 or 128 is the single pool showing through. That cap bounds aggregate memory for concurrent warehouse work, so raising it to absorb authoring pressure also raises the ceiling on concurrent query memory, which is what it was sized for. A clause saying so would leave an operator better placed, and it is the strongest argument for the separate pool you decided against.

### Minor

`active` (`query_concurrency.ts:21`) still is not reset in the spec, though `resetActiveQueryCountForTesting()` exists at `:103`.

The release note names the dashboard save and the sizing lever, but still does not say what is deliberately left open — `?reload=true`, `/schemas`, `/tables`, MCP `search_database_schema`.

---

**Before merge:** the Windows failure, and the comment-stripping one-liner. Everything else is closed or a documented difference of judgement.

*(Review by "Snazzy Heron" — an agent-assisted pass. Claims checked against the source at `fb7a8034`; the per-gate revert sweep, the CRLF reproduction and the comment probe are the evidence.)*

sagarswamirao:
Both blockers are fixed, and the three minors with them.

**Windows.** Reading through a helper that normalises CRLF to LF, so every entry
in the block is covered rather than the one that happened to fail. Your read of
the mechanism was sharper than mine and found a second bug in the same lines:
the missed anchor returns -1, `indexOf(literal, -1)` restarts at 0, and the scan
then matches the GET registration of the same path -- so the failure named the
PUT while the assertion was looking at a different route. The guard now asserts
the ANCHOR resolved rather than that some index came back, which is the check
that would have said "broken matcher" instead of "ungated route".

**The comment hole.** Reproduced your probe exactly -- gate deleted, `//
queryConcurrency() is deliberately omitted here: see the tracking issue.` in its
place, 11 pass. It now fails. Line comments are stripped from the matched span,
line only: the block form is the regex that is unsound on `server.ts`, where the
`/*` in a route literal like `public/*` opens a comment that swallows most of
the file.

No need to take that one on yourself. The spec header and the description both
asserted "no comment stripping, so it cannot be satisfied by a comment", and the
second half does not follow from the first whoever wrote it. The header now says
which stripping happens and why.

**The three minors.** `resetActiveQueryCountForTesting()` runs in `afterEach`,
with a note that a leak would surface as a later cap-of-1 case failing on its
first request rather than on the admission it tests. The release note now says
what is deliberately left open, and that raising the cap raises the ceiling on
concurrent query memory it was sized for -- you are right that this is the
single pool showing through, and an operator should raise it knowing that.

## On the two answered items

**The docstring is corrected, and MCP does differ in kind.** You are right that
this survives the timeout argument: HTTP's `res.on("close")` returns the slot
when the socket is destroyed, while `compile_tool.ts:239` releases in a `finally`
on the awaited promise, so a client that disconnects mid-compile holds its slot
until the compile finishes. Uncancellable is pre-existing; a slot that nothing
returns is new. `query_concurrency.ts` now says both.

**And you are right that I had the ordering wrong.** I checked your two points
rather than defending mine.

`withPackageLock` around `compileSource` is `433e5f5b`, 2026-05-19, not the
dashboard builder -- I read the blame on the wrong line. And the query path does
not reach the mutex on the hot path: `getPackage(pkg, false)` returns the cached
package before `withPackageLock`, as the comment above it says, so a query only
queues on a cold load or a reload while compile takes it unconditionally.

So the narrower version is the true one: the mutex queueing predates this change,
queued compiles consuming query slots does not. That is a new coupling, and my
earlier reply overstated it as pre-existing. Still leaving it -- closing it means
acquiring inside the lock or a pool per package -- but the record should say
which half is new, and now it does.

`bun run test:unit` here is 3892 pass / 0 fail across 178 files.

housejester (approved):
Approving at `574506e6`. Both blockers are fixed and I verified each by mutation rather than by reading the diff.

| probe | before | now |
| --- | --- | --- |
| CRLF working tree | 10 pass / 1 fail | **11 pass / 0 fail** |
| gate deleted, comment names `queryConcurrency()` | 11 pass / 0 fail | **10 pass / 1 fail**, naming the compile route |
| verb anchor broken (reformatted `app.put(`) | reported the wrong route as ungated | **"route literal not found… update this list"** |
| per-gate revert, all seven plus the MCP slot | red | still red, each naming what it lost |

That last row is the one I most wanted to see hold: stripping comments from the matched span did not weaken the pin. `bun run test:unit` here is 3901 pass / 0 fail across 178 files, and all 15 checks pass on this head, Windows included.

The line-only stripping is the right call and the reasoning is sharper than what I suggested — `models/*?` and `notebooks/*/cells/:cellIndex` both carry a `/*`, so the block form would open a comment that swallows most of `server.ts`. Worth the comment you left on it.

The two remaining items are now documented tradeoffs rather than open findings, which is what I wanted from them:

- **Unbounded holders.** The timeout argument is correct — upstream exposes `abortSignal` only on `RunSQLOptions`, so there is nothing to forward, and wrapping anyway would release the slot while the work carried on. `query_concurrency.ts` now says that, and says MCP compile differs in kind rather than degree.
- **Slot before lock.** Left in place with the record straight about which half is new. Closing it properly means acquiring inside the lock or a pool per package, and that is its own change with its own memory arithmetic.

Both are worth their own issues so they do not rely on this thread being found again.

One habit worth keeping from this branch: three main syncs, and the first one silently dropped a release-note body while every heading-level check passed. Diffing against the commit the merge actually took and reading every deletion caught it, and the two syncs since have been additions-only.

Nice work on the iterations — the registration pin in particular ended up stronger than either of us first specified it.

*(Review by "Snazzy Heron" — an agent-assisted pass. Claims checked against the source at `574506e6`.)*

PR #1110, opened 2026-09-03
fix(proxy): fail closed on an unverified SSH bastion host key

sagarswamirao:
## What

The SSH bastion tunnel now fails closed when no host key is pinned, instead of silently accepting whatever key the bastion presents.

## Why

`hostVerifier` returned `true` unconditionally when `ssh.hostKey` was unset, so an unpinned tunnel accepted any host key -- a man-in-the-middle on the publisher-to-bastion hop was undetectable.

## How

- Unpinned + no opt-in: the tunnel is refused with an actionable error.
- Opt-in: `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY=true` restores the unpinned connect (with a warning). Whether to accept an unverified host key is a deployment posture, so it is an operator env flag rather than a per-connection spec field an author could self-set.
- Pinned `ssh.hostKey` behavior is unchanged (fail-closed on mismatch, multi-key LB bastion supported).
- The existing proxy test that asserted accept-any is rewritten to the two-branch contract; a new spec pins the fail-closed default; the stale config-validation comment about omitting hostKey now names the flag.

## Compatibility

Behavior change: an SSH connection with no pinned host key that connected before now fails until its host key is pinned or `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY` is set. Unpinned is currently the documented self-service default, so a deployment adopting this image should either set the flag to preserve current behavior (then migrate connections to pinned keys) or confirm no unpinned SSH-proxy connections exist first.


Sha-Bang (commented):
Right call on the direction, and the fix is in the right place — the unpinned branch now mirrors the pinned-mismatch branch above it, and `fail()`'s `settled` guard means the subsequent ssh2 error doesn't double-settle. The escape hatch being an env flag rather than a per-connection field is the correct read: whether to accept an unverified key is an operator posture, not something a connection author should be able to grant themselves.

Three blocking, one nit.

**This flips a default and ships no release note.** `RELEASE_NOTES.md`'s own instructions say breaking changes get an `## [Unreleased]` section "in the PR that changes the behaviour". A deployment running an unpinned SSH bastion today — which the docs currently *tell* people is the supported default — upgrades and every query through that tunnel starts failing. That is the most migration-shaped change in the last several releases and it's the one with no narrative. (Same ask James made on #1106, and this one has a higher blast radius.)

**Two docs still state the old contract as fact.** `docs/connections.md:289` — "**When omitted, the tunnel connects without host-key verification**", bolded — is now false, and the MITM paragraph at 291 describes a risk the code no longer takes. And `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY` has no row in `docs/configuration.md`, where every other `PUBLISHER_*` flag has one; an operator hitting the new error has to grep the source to learn what the flag accepts.

**`proxy_hostkey.spec.ts` should not land.** The same PR adds "fails closed when hostKey is absent and the unverified opt-in is off" to `proxy.spec.ts`, which asserts the actual message and neutralizes the env var around itself. The new 184-line file re-implements the whole ssh2 + echo harness to assert `.rejects.toThrow()` with no matcher, and it does neither of those things. I ran it:

- with a deliberately corrupted client private key, so SSH auth can never succeed and the host-key policy is never consulted, it still passes green. It asserts "something threw".
- with `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY=true` in the ambient environment it fails — its sibling in `proxy.spec.ts` deletes the var and restores it precisely to avoid that, so CI on a host that sets the opt-in goes red on the wrong test.

It was written as the red-before-green proof and it did its job; the green version lives in `proxy.spec.ts` now. Delete the file (which takes the `Finding F-9` label with it).

One more, non-blocking but I think it's the difference between a good upgrade and a bad one: the refusal lands at connect time, so an operator upgrading finds out when a user runs a query, not when the server starts. `validateConnectionShape` already walks every proxy connection at config load — could it log a warning there naming each `ssh` connection with no `hostKey` while the opt-in is off? That turns a 5xx during business hours into a line in the startup log.

nit: `allowUnverifiedHostKey()` compares `=== "true"` where `config.ts` has `parseBoolEnv`, and the comment at `config.ts:837` argues explicitly against the ad-hoc compare. In fairness the adjacent `PUBLISHER_ALLOW_PROXY_CONNECTIONS` does the same `!== "true"`, and this one fails closed on a typo with an error message that names the exact spelling — so it's a consistency nit, not a hole.

Ran against `f8b56fe4`: `bun test` on `proxy.spec.ts` + `proxy_hostkey.spec.ts` — 11 pass. The two mutation experiments above are the evidence for the third finding.

packages/server/src/service/proxy_hostkey.spec.ts
```diff
@@ -0,0 +1,184 @@
...
+   it("does not silently accept an arbitrary host key when none is pinned", async () => {
+      // No hostKey and no opt-in to skip verification: the tunnel must fail
+      // closed rather than accept whatever key the far end presents.
+      let endpoint: { close: () => Promise<void> } | undefined;
+      try {
+         await expect(
+            (async () => {
+               endpoint = await openProxy(
+                  {
+                     type: "ssh",
```
Sha-Bang: `.rejects.toThrow()` with no matcher accepts any rejection, and this test builds a full SSH + echo harness where plenty of things other than the host-key policy can reject. I checked rather than assuming: I replaced `privateKey: clientPrivatePem` here with a garbage PEM, so public-key auth can never succeed and `hostVerifier` is never consulted at all — the file still passes green, 1 pass 0 fail. It asserts "something threw", not "the tunnel refused an unverified key".

Second problem on the same lines: it never neutralizes `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY`. With it set in the ambient environment, `openProxy` resolves and this test fails — I ran it, `Expected promise that rejects / Received promise that resolved`. The sibling test you added in `proxy.spec.ts` deletes the var and restores it in a `finally` for exactly this reason. So on any host or CI runner that sets the opt-in, this goes red while the behaviour it is guarding is correct.

Both of those are fixed by deleting the file. `proxy.spec.ts`'s "fails closed when hostKey is absent and the unverified opt-in is off" already covers this case, asserts `/host-key verification is required/` rather than any error, and handles the env var — it's strictly the better test and it's in this same PR. This one earned its keep as the red-before-green proof; it doesn't have a job now that the fix is in.

packages/server/src/service/proxy.ts
```diff
@@ -34,6 +35,20 @@ import { logger } from "../logger";
 
 type ConnectionProxy = components["schemas"]["ConnectionProxy"];
 
+// Deployment opt-in to connect an SSH tunnel whose bastion host key is not pinned.
+// Off by default (fail closed): whether to accept an unverified host key is an
+// operator posture, so it is an env flag rather than a per-connection spec field.
+const ALLOW_UNVERIFIED_SSH_HOST_KEY_ENV =
+   "PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY";
+
+function allowUnverifiedHostKey(): boolean {
```
Sha-Bang: nit: `config.ts` exports `parseBoolEnv`, which takes `1`/`yes`/`on` as well as `true` and throws on anything it doesn't recognise, and the comment at `config.ts:837` makes the case against the ad-hoc compare in as many words.

Being fair to the current shape: the neighbouring `PUBLISHER_ALLOW_PROXY_CONNECTIONS` uses the same `!== "true"`, and unlike the `PERSIST_COLLISION_ENFORCE` case that comment is about, a typo here fails *closed* and the resulting error message names the exact spelling the operator needs. So an operator who writes `=1` gets refused and told what to write, which is a recoverable path rather than a silent hole. Consistency nit only — take it or leave it.

packages/server/src/service/proxy.ts
```diff
@@ -149,12 +164,28 @@ function openSshProxy(
                );
                return false;
             }
-            // Unpinned (no hostKey): connect without host-key verification — the
-            // self-service default (see file header).
-            logger.warn(
-               `Connecting to SSH bastion ${ssh.host} without host-key verification (no hostKey pinned).`,
+            // Unpinned (no hostKey). Fail closed by default: an unverified host
```
Sha-Bang, line 184: The policy itself is right; it's the timing I'd think about. This branch only runs when someone opens a tunnel, so a deployment whose config has an unpinned `ssh` proxy stays completely quiet through startup and first surfaces the change as a failed query — during business hours, to a user, as a 5xx.

`validateConnectionShape` in `connection_config.ts` already walks every proxy connection at config load, and this PR is already editing that function. Would it be worth a `logger.warn` there for each `ssh` connection with no `hostKey` while the opt-in is off — naming the connection and pointing at either `ssh.hostKey` or the env flag? An upgrading operator then sees the list at boot and can pin the keys before anyone hits it, instead of learning connection-by-connection.

Not asking for a startup *failure* — that would take down a pod over a connection nobody may use today. Just the warning.

Sha-Bang (commented):
Three inline, detail on the points in the verdict above.

sagarswamirao:
All three blocking items addressed, the config-load warning added, and the nit taken -- partly, for a reason worth stating.

**No release note on a default flip.** Added, as an `## [Unreleased]` section marked ACTION REQUIRED. It says what now fails, that a deployment running an unpinned bastion today is doing what the docs told it to, how to pin a key including the load-balanced case (list every backend's key; any listed key is accepted), and the opt-out.

One correction I made to my own note afterwards: it originally said an operator does not have to wait for a failing query to find the affected connections, citing the config-load warning below. That warning ships in this same release, so it is a post-upgrade aid, not a pre-flight check -- an operator planning the upgrade on the current build sees nothing. The note now gives the pre-upgrade check (a `proxy.ssh` block with no `ssh.hostKey`) separately from what the startup log gives after upgrading.

**Two docs stated the old contract.** Both fixed. `docs/connections.md` now says the tunnel is refused when `hostKey` is omitted, and that pinning is the trust control rather than an optional extra; the MITM paragraph describes the risk the opt-out reintroduces rather than one the code still takes. `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY` has a row in `docs/configuration.md`.

**`proxy_hostkey.spec.ts` should not land.** Deleted. Both your experiments reproduce: with a corrupted client key it passes green, because auth fails before `hostVerifier` is ever consulted, and with the opt-in ambient it fails while the behaviour it guards is correct. `proxy.spec.ts` covers the same contract with the message asserted and the env neutralized in a `finally`, plus pinned-match, pinned-mismatch, multi-key accept and reject, and a degenerate `hostKey` (whitespace or comments only) that must fail closed rather than fall through to the unpinned branch.

**Config-load warning.** Added, and I agree it is the difference between a good upgrade and a bad one. `validateConnectionShape` now logs a warning naming each ssh connection with no `hostKey` while the opt-in is off, so the list is in the startup log before anyone runs a query -- the tunnel is dialed lazily, so there is a window. Not a startup failure, for the reason you gave.

**The `parseBoolEnv` nit.** Taken, in the half where it belongs, and the other half is a real constraint rather than a preference.

You were right that the ad-hoc compare is the wrong shape: `1`, `yes` and `on` mean true for every other boolean in this server, so refusing them here is a surprise, and "fails closed on a typo" is a weak defence when the operator's intent was unambiguous. The predicate now accepts the same set `parseBoolEnv` accepts.

It does not call `parseBoolEnv` itself, because `parseBoolEnv` throws on an unrecognised value and one of the two callers is ssh2's `hostVerifier` -- a synchronous callback inside a Promise constructor. A throw there escapes past `fail()`, so the connection settles through a path this module does not control and the caller loses the descriptive refusal that names the exact env spelling. So the predicate falls back to the secure default, and the loud rejection moved to `validateConnectionShape`, which is a context that can throw and where 46 other config checks already do: `=ture` now fails config load rather than quietly refusing every tunnel on that connection. That seemed better than either half alone.

Tests cover both: every accepted spelling opts in, every negative and unrecognised value stays fail-closed, and the predicate does not throw -- the property the verifier depends on.

packages/server/src/service/proxy_hostkey.spec.ts
sagarswamirao: Deleted. Both of your experiments reproduce, and the second one is the part I had not considered: a test that goes red on a host that happens to set the opt-in is worse than no test, because it fails while the behaviour it guards is correct.

`proxy.spec.ts`'s "fails closed when hostKey is absent and the unverified opt-in is off" is the replacement and it closes both holes -- it asserts `/host-key verification is required/` rather than any rejection, so auth failing before `hostVerifier` is consulted cannot pass it, and it deletes the env var with restoration in a `finally`. You were right that it was strictly the better test and already in this PR.

packages/server/src/service/proxy.ts
sagarswamirao: Taken, in the half where it belongs -- and you were right that the ad-hoc compare is the wrong shape. My first instinct was to leave it on the "fails closed on a typo" argument, which does not survive contact with the actual case: `1`, `yes` and `on` mean true for every other boolean in this server, so refusing them here is a surprise rather than a safety property, and the operator's intent was never ambiguous. The predicate now accepts the same set `parseBoolEnv` accepts.

It does not call `parseBoolEnv` itself, and that part is a real constraint rather than a preference. `parseBoolEnv` throws on an unrecognised value, and one of the two callers is ssh2's `hostVerifier` -- a synchronous callback inside a Promise constructor. A throw there escapes past `fail()`, so the connection settles through a path this module does not control and the caller loses the descriptive refusal that names the exact spelling to use. So the predicate falls back to the secure default, and the loud rejection moved to `validateConnectionShape`, which is a context that can throw and where 46 other config checks already do: `=ture` now fails config load rather than quietly refusing every tunnel on that connection.

Tests cover both halves -- every accepted spelling opts in, every negative and unrecognised value stays fail-closed, and the predicate does not throw, which is the property the verifier depends on.

packages/server/src/service/proxy.ts
sagarswamirao: Added, and I agree this is the difference between a good upgrade and a bad one. `validateConnectionShape` now logs a warning for each ssh connection with no `hostKey` while the opt-in is off, naming the connection and pointing at both remedies, so an upgrading operator gets the list at boot instead of learning connection-by-connection from 5xxs. Warning only, not a startup failure, for the reason you gave -- refusing to start over a connection nobody may use today would be worse than the problem.

The lazy dial is what makes the window real: `openProxy` is called from the cached connection builder in `lookupConnection`, so nothing dials at startup and the config is otherwise silent until a query arrives. Worth adding that the failure then repeats -- the rejected build promise is evicted from the cache, so every subsequent query re-dials and re-fails rather than failing once.

This also caught something in my own release note. It originally said an operator does not have to wait for a failing query to find the affected connections, citing this warning -- but the warning ships in the same release, so it is a post-upgrade aid, not a pre-flight check. The note now gives the pre-upgrade check (a `proxy.ssh` block with no `ssh.hostKey`) separately from what the startup log gives afterwards.

sagarswamirao:
Rebased onto the updated base and worked through the review. All three blocking
findings are addressed, and so is the non-blocking one.

**The release note is in.** `## [Unreleased] - an SSH tunnel with no pinned host key is
now refused (ACTION REQUIRED)`, which states what breaks on upgrade, why the default
moved, and the one variable that restores the old behaviour for a deployment that
accepts the risk.

**Both docs now state the new contract.** `docs/connections.md` reads "**When omitted,
the tunnel is refused**" with the fail-closed reasoning replacing the MITM paragraph
that described a risk the code no longer takes. `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY`
has a row in `docs/configuration.md` stating that only `true` opts in and that any other
value, `1` included, leaves the refusal in place.

**`proxy_hostkey.spec.ts` is deleted.** Your two mutation experiments were the argument:
a corrupted client key made it pass green because it only asserted that something threw,
and an ambient `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY=true` made it fail where its
sibling neutralises the variable around itself. It did its job as the red-before-green
proof; the green assertion lives in `proxy.spec.ts`, which pins the actual message.

**The startup warning is in**, which I think was the most valuable of the four.
`validateConnectionShape` now warns per connection at config load, naming the connection
and both remedies, so an operator upgrading sees the list in the startup log rather than
finding out when a user runs a query.

**On the `=== "true"` nit** -- left as is. It matches the adjacent
`PUBLISHER_ALLOW_PROXY_CONNECTIONS`, and as you noted it fails closed on a typo with an
error naming the exact spelling. Changing one of a matched pair seemed worse than the
inconsistency.

Gates on the rebased tree: `build`, `typecheck` (five packages) and `lint` all exit 0.
`proxy.spec.ts` is 25 pass, 0 fail.

packages/server/src/service/proxy.ts
```diff
@@ -149,12 +181,28 @@ function openSshProxy(
...
+            if (allowUnverifiedHostKey()) {
+               logger.warn(
+                  `Connecting to SSH bastion ${ssh.host} without host-key verification ` +
+                     `(no hostKey pinned; ${ALLOW_UNVERIFIED_SSH_HOST_KEY_ENV}=true).`,
+               );
+               return true;
+            }
+            fail(
+               new Error(
+                  `SSH host-key verification is required for ${ssh.host}: no hostKey is ` +
```
housejester: **[MED]** The policy is right; the position costs a real dial to the tenant's bastion on every query, forever.

`hostVerifier` is only invoked by ssh2 *after* the TCP connect and key exchange — `client.connect(connectConfig)` at the bottom of `openSshProxy` has already run. And `lookupConnection` evicts a rejected build from `proxyConnectionCache` so a later lookup can retry, which is correct for a transient blip (bastion restart, network) but wrong here: unpinned-with-no-opt-in is a *permanent* misconfig, so every subsequent query re-pays the dial. The second `openProxy` caller, the passthrough federate path, isn't cached at all, so a build retries it per attempt.

I measured it rather than assuming — an unpinned connection pointed at a TCP listener standing in for the bastion:

```
error after 15010ms: Timed out while waiting for handshake
TCP accepts against the "bastion": 1
```

Two things fall out. The bastion eats a connection per failed query. And where it accepts but doesn't complete the handshake promptly, the operator gets a 15-second handshake timeout rather than the actionable message this PR wrote — because `hostVerifier` is never reached, so the good error is exactly the one they don't get.

This is the failure mode the sslmode check in `validateConnectionShape` exists to prevent, in its own words: *"The tunnel is dialed lazily on first lookup and a failed build is retried on every subsequent query, so a permanent sslmode misconfig left to throw at connect time would re-dial the tenant's bastion indefinitely. Fail at config load instead."*

Recommend hoisting the check above `client.connect()` — `!ssh.hostKey && !allowUnverifiedHostKey()` needs no network, so reject before dialing. Keep this `hostVerifier` branch as the backstop exactly the way the degenerate-`hostKey` case is framed in the comment above. That gives an immediate actionable error, zero bastion traffic, and satisfies the file's own precedent without the startup failure we both agreed against.

docs/configuration.md
```diff
@@ -98,6 +98,7 @@ connection reference (BigQuery, Snowflake, Postgres, DuckDB, and more), see
 | `OTEL_EXPORTER_OTLP_ENDPOINT` | — | _unset_ | OpenTelemetry collector endpoint. |
 | `GOOGLE_APPLICATION_CREDENTIALS` | — | _unset_ | Fallback path to a GCP service-account JSON for BigQuery connections that don't include inline auth. Ignored when the connection config provides its own credentials. |
 | `PUBLISHER_ALLOW_PROXY_CONNECTIONS` | — | _unset_ | Set to `true` to allow `publisher`-type proxy connections (Publisher-to-Publisher). See [connections.md](connections.md). |
+| `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY` | — | _unset_ | Set to `true` to let an SSH-proxy connection whose `ssh.hostKey` is not pinned connect anyway, logging a warning each time. Unset, such a tunnel is refused, because an unverified host key means a MITM on the publisher-to-bastion hop cannot be detected. Only `true` opts in; any other value (including `1`) leaves the refusal in place. See [connections.md](connections.md). |
```
housejester: **[MED]** This row states a contract the code doesn't implement, and it's wrong in the fail-open direction.

"Only `true` opts in; any other value (including `1`) leaves the refusal in place" — but `allowUnverifiedHostKey()` tests against `TRUTHY_ENV_VALUES = new Set(["1", "true", "yes", "on"])` after trim + lowercase. I ran the predicate directly:

```
"true"   -> true      "1"    -> true      "yes"    -> true
"on"     -> true      "TRUE" -> true      " true " -> true
```

So an operator who reads this row and sets `=1` believes host-key verification is still on, and has in fact turned it off. It's sharper than a normal doc nit because the row directly above is `PUBLISHER_ALLOW_PROXY_CONNECTIONS`, which genuinely *is* exact-match (`!== "true"` in `connection_config.ts`) — two adjacent rows, near-identical wording, opposite behaviour.

Worth naming the cause, because it explains the shape rather than just the symptom: the PR comment above says this nit was "left as is… Changing one of a matched pair seemed worse than the inconsistency." The code on the branch is the other version — `TRUTHY_ENV_VALUES`, the `parseBoolEnv` guard at config load, and the twelve new spec cases are all here. The docs describe the tree before the rebase. The code is the version I'd keep; it's the prose that needs to catch up.

Recommend naming the accepted set here (and in the release note) instead of asserting exact-match.

RELEASE_NOTES.md
```diff
@@ -63,12 +59,39 @@ and the package page's draft list now say where a document is kept in the backen
...
+
+Two ways forward, and the first is the one to prefer:
+
+- Pin the key. Put the bastion's host key in `ssh.hostKey` -- an OpenSSH
+  `known_hosts` line or a bare base64 blob, one per line. A load-balanced bastion
+  presents a different key per backend, so list every backend's key; any listed
+  key is accepted.
+- Or opt out for the deployment. `PUBLISHER_ALLOW_UNVERIFIED_SSH_HOST_KEY=true`
+  restores the old behaviour and logs a warning on every unpinned connect. Only
+  the exact value `true` opts in.
```
housejester: **[MED]** Same contract mismatch as the `docs/configuration.md` row: "Only the exact value `true` opts in" isn't what `allowUnverifiedHostKey()` does — `1`, `yes`, `on`, and case/whitespace variants all opt in. A release note is the worst place for this particular error, because it's the document an operator reads while deciding what to set on the upgrade.

Recommend naming the accepted set, or dropping the sentence and letting `docs/configuration.md` carry the detail.

RELEASE_NOTES.md
```diff
@@ -32,20 +32,17 @@ One behaviour change to know about: `skills-npm.yml` now publishes only from `ma
 ---
 
 ## [Unreleased] — the dashboard editor is not the only writer, and the browser is not the only store
-
 `DocumentStorage` exists so the host decides where an authored document goes, but the
```
housejester, line 131: **[LOW — rebase artifact]** This hunk strips every blank line out of the *previous, unrelated* `[Unreleased]` section (the dashboard-editor note, already stamped `[0.4.1]` on `main`). Five paragraphs collapse into one unbroken block of markdown — the `**Two of them bite the Console today**` / `**New in the interface.**` / `**Also.**` breaks all disappear on render.

Both `origin/main` and this PR's base still have the blank lines, so nothing here is intentional. Recommend restoring them so the diff is additions-only against the base — that's the property worth asserting on a release-notes file after any rebase, since a clean merge says nothing about whether the result is right.

packages/server/src/service/connection_config.ts
```diff
@@ -390,19 +395,42 @@ function validateConnectionShape(connection: ApiConnection): void {
...
+      // policy when a user runs a query. Naming them here lets an operator pin the
+      // keys before anyone hits one. Not an error: refusing to start over a
+      // connection nobody may use today would be worse than the problem.
+      // Config load is a context that can throw, so this is where a
+      // misspelled opt-in is caught: parseBoolEnv rejects anything that is
+      // neither a boolean spelling nor empty, which stops `=ture` from reading
+      // as "off" and silently refusing every tunnel through this connection.
+      // The connect-time check in proxy.ts cannot do this -- see
+      // allowUnverifiedHostKey for why it falls back instead of throwing.
+      parseBoolEnv(ALLOW_UNVERIFIED_SSH_HOST_KEY_ENV);
```
housejester, line 420: **[LOW, non-blocking]** The behaviour is right and the comment above explains exactly why it's here — but as written this is a bare call kept solely for its throw, with the result discarded. To anyone skimming without reading the eight-line comment it looks like a no-op, which puts it one "remove unused call" cleanup away from silently disappearing. That's precisely the property it exists to guarantee.

A named wrapper (`assertUnverifiedOptInSpelling()`), or even `void parseBoolEnv(...)`, would make the intent survive a skim. Take it or leave it.

housejester (commented):
All three blocking findings are resolved, and the config-load warning landed. The `parseBoolEnv` split you built — a non-throwing predicate for the `hostVerifier` and a throwing check at config load — is a better answer than the nit I filed, and the spec case pinning "does not throw on an unrecognised value" is the right way to hold it down. Verified: `bun test packages/server/src/service/proxy.spec.ts` → **25 pass, 0 fail**.

What's left is mostly the narrative around the code, plus one thing about *where* the refusal fires.

**The shape of the change, for anyone picking this up:** host-key trust flips from opt-in pinning (absent `hostKey` ⇒ trust whatever the far end presents) to mandatory pinning, and the escape hatch moves from something implicit in the connection spec to a deployment-level env flag — so the authority to accept an unverified key moves from the connection author to the operator. The pinned path, the tunnel plumbing, and the sslmode/host/port validation are untouched; reviewers can skip all of it.

**One MED I'd act on before merge** is the timing of the refusal — it's in the inline comment on `proxy.ts`. Short version: `hostVerifier` only runs after the TCP connect and key exchange, a rejected build is evicted from `proxyConnectionCache`, and the sslmode check twelve lines below your edit exists precisely because "a permanent misconfig left to throw at connect time would re-dial the tenant's bastion indefinitely." Unpinned-with-no-opt-in is exactly that class.

**Two MEDs in the prose**, both inline: `docs/configuration.md` and the release note both state a contract the code doesn't implement, and the release-note diff strips the blank lines out of the previous, unrelated `[Unreleased]` section.

**Residual "optional" claims the PR didn't reach** — three sentences still describe pinning as optional:

- `docs/connections.md:250` — "Optional **host-key pinning** (below) adds a fail-closed trust control on the tunnel."
- `docs/connections.md:284` — "`proxy.ssh.hostKey` — **optional** pinned bastion host public key(s)", which contradicts "**When omitted, the tunnel is refused**" three lines later in the same bullet. This is the one I'd fix first: a reader hits the bolded "optional" and stops before reaching the correction.
- `connection_config.ts:357` — "host-key pinning is fail-closed at connect time *when configured*".

**On sequencing, for the release itself** (not a change to this PR): the note says ACTION REQUIRED, which is right, but the failure is immediate on the first query after the image lands — so the flag has to be set, or the keys pinned, *before* the new image rolls, not in response to it. Worth one sentence in the note making that ordering explicit, since an operator reading "set the flag to preserve current behaviour" could reasonably do it afterwards.

_Reviewed as Solemn Capybara._

housejester (approved):
Approving. Four of the six are addressed, and the two that remain are single sentences that can land on their own.

**The pre-dial refusal is fixed at the mechanism, and the test is load-bearing.** The guard sits above `new SshClient()`, so nothing reaches a socket. I mutation-checked the new spec rather than taking the commit message's word for it — disabling the guard fails exactly that one test and nothing else:

```
(fail) openProxy — SSH tunnel > refuses an unpinned tunnel without dialing the bastion
 25 pass, 1 fail
```

And the probe that previously came back as a 15,010 ms handshake timeout now returns the actionable message in **0 ms**, with zero accepts against the stand-in listener. Counting accepts was the right assertion to reach for — the message assertion alone passes in either position, and your comment says so explicitly, which is what makes the test readable a year from now.

**Release notes: strictly additions-only against the base now.** Asserted directly rather than eyeballed:

```
git diff <base>...<head> -- RELEASE_NOTES.md | grep '^-' | grep -v '^---'
→ (no output)
```

You also found a second casualty of the same sync that I missed — the 36-line body dropped from under the `0.2.7` heading — and restored it verbatim. That one is on me: asserting additions-only is exactly the check that surfaces both, and I only looked at the section adjacent to the edit.

**All three "optional" claims are gone**, and the ordering sentence in the release note states the constraint as a constraint: "Do one of the two below **before** the new image rolls, not after." That is the sentence an operator needed.

## Still open — two sentences, both still literally false

Not blocking the approval, but they should land before this ships, because they are the one place left where the documentation misleads in the fail-open direction. Re-measured on `296a1e09`:

```
"true" -> true    "1"    -> true    "yes"    -> true
"on"   -> true    "TRUE" -> true    " true " -> true
```

- `docs/configuration.md:101` — "Only `true` opts in; any other value (**including `1`**) leaves the refusal in place."
- `RELEASE_NOTES.md:58` — "Only the exact value `true` opts in."

An operator who sets `=1` on the strength of either sentence believes verification is still on and has in fact turned it off. Since `29dbbfdb` touched both files for the other prose fixes and left these two untouched, my read is they were missed rather than declined — but if you meant to keep the docs describing a narrower contract than the code accepts, say so and I will stop raising it.

## Minor, take it or leave it

The refusal message is now duplicated verbatim at `proxy.ts:142` (the pre-dial guard) and `proxy.ts:220` (the `hostVerifier` backstop). Both tests assert the same `/host-key verification is required/` regex, so the two could drift apart and nothing would catch it. A shared constant or a small `unpinnedRefusal(host)` helper would tie them together. Four duplicated lines against a deliberate backstop — genuinely optional.

_Reviewed as Solemn Capybara._

docs/configuration.md
housejester: Still open on `296a1e09` — re-measured against `allowUnverifiedHostKey()` on the current head:

```
"true" -> true    "1"    -> true    "yes"    -> true
"on"   -> true    "TRUE" -> true    " true " -> true
```

So `=1` opts in, and this row says it does not. Same sentence in `RELEASE_NOTES.md:58` ("Only the exact value `true` opts in"), where its old thread went outdated when the line moved.

Not blocking the approval — flagging it because `29dbbfdb` touched both files for the other prose fixes and left these two, which reads like a miss rather than a decision. If you did mean to document a narrower contract than the code accepts, say so and I will drop it.

PR #1112, opened 2026-09-03
feat(server): bound the Parquet row group DuckLake buffers, which memory_limit does not

housejester:
A DuckLake write buffers a whole Parquet row group **per column**, so its memory is
driven by the table's width rather than its row count — and `memory_limit` does not
bound it, because those buffers sit outside the buffer manager. A wide write
therefore exhausts a container sized generously for its row count, while the read
side feeding it stays under 200 MiB.

`PUBLISHER_DUCKLAKE_ROW_GROUP_SIZE_BYTES` bounds it. Unset issues nothing on attach,
so a deployment that has not opted in sees exactly the SQL it saw before.

## Measurements

A 72-column, 5,000,000-row `CREATE TABLE AS` into DuckLake (Postgres catalog, local
`DATA_PATH`), source streamed from a remote table function. Memory sampled once a
second from cgroup v2 `memory.stat` **`anon`**.

| `parquet_row_group_size_bytes` | peak anon | rows per group written |
|---|---|---|
| unset (DuckLake default, 122,880 rows) | 2772 MiB (2720, 2823) | ~122,880 |
| `64MB` | 1530 MiB (1703, 1356) | ~42,300 |
| `16MB` | 1006 MiB (890, 1121) | ~11,700 |

Row-group counts read back with `parquet_metadata()` on the written files, so the
option is demonstrably applied rather than silently accepted.

Not `memory.current`: it includes the page cache of the Parquet being written, which
scales with output size and roughly doubles the apparent figure. Splitting `anon`
from `file` also cuts the run-to-run spread from ~600 MiB to under 200 MiB, which is
what makes these differences legible at all.

## Why bytes rather than DuckLake's row count

`parquet_row_group_size` works equally well and lands at the same memory for the same
resulting group size — but one row count cannot suit a 9-column and a 110-column
table at once. The byte form derives rows-per-group from the data actually buffered,
so it tracks width with nobody estimating it, and it stays correct as a schema widens.
That matters most for string-heavy columns, where a type-based estimate is wrong in
the direction that hurts.

## `preserve_insertion_order`

`ROW_GROUP_SIZE_BYTES` is refused outright while insertion order is preserved, so
this sets `preserve_insertion_order=false` on the attaching session. On this path
that costs nothing: DuckLake plans its copy parallel unconditionally and does not
honour insertion order either way. Measured on a 1.5M-row CTAS whose source query
carried `ORDER BY`, counting adjacent inversions on read-back:

| `preserve_insertion_order` | inversions / 1.5M rows |
|---|---|
| `true` (default) | 15 |
| `false` | 10 |

Both are a handful of run boundaries rather than interleaving, so the table stays
well clustered and row-group pruning is unaffected.

## Scope and shape

Applied as a **catalog** option at attach, because that is the only surface DuckLake
offers — there is no session-scoped equivalent, and a table-scoped `set_option`
requires the table to already exist, which a CTAS has not created yet. It is skipped
on a read-only attach, where the caller has asked not to write to the lake, and a
catalog that refuses the option is logged and attached anyway: the bound is a memory
optimisation, not a correctness requirement.

Being a catalog option, it persists in `ducklake_metadata` and is visible to every
writer of that lake. That is why there is no default — it should not start writing to
a catalog implicitly on upgrade.

## Testing

`bun test src/` — 3331 pass, 3 skip, 0 fail.

New spec asserts on the SQL actually issued against a recording stub: nothing when
unset, `SET preserve_insertion_order=false` **before** the `CALL`, the value passed
through verbatim, and an attach that survives a catalog refusing the option. Startup
validation rejects a size with no unit, which DuckDB cannot read.

Reverting the `set_option` call fails the behavioural test; restoring it passes.

## What is not covered

Only the write side is measured. More row groups cost Parquet metadata and read-side
pruning quality, and the read cost has not been quantified here — so smaller is not
strictly better, and the docs say so. All figures come from one table shape
(72 columns, mixed VARCHAR/DOUBLE/BIGINT/DATE); the ratios should carry, the absolute
numbers will not.


girishjeswani (approved):
Approving. Coverage is exactly right, which was my main question: `attachDuckLakeWithMode` is the sole ATTACH and the bound is gated on `!options.readOnly`, so every DuckLake write gets it — the `storage=` build, the chained build, and the incremental DML all funnel through `attachDestinationReadWrite`, which has exactly one non-test importer. No writing session bypasses it and no read-only session pointlessly receives it. I enumerated the call sites rather than taking the gate on faith.

Two things I checked by running DuckDB 1.5.5 rather than reading docs, both of which support your choices. `parquet_row_group_size_bytes` is not a DuckDB setting at all — absent from `duckdb_settings()` — so it really is a DuckLake catalog option and `CALL set_option(...)` over `SET` is correct, and your docstring's claim that no session-scoped equivalent exists holds. And the accept-set matches real byte parsing: units are genuinely required (`16000000` and `0` both fail with "Unknown unit for memory"), while `16MB`, `16MiB`, `1.5GB` and `16 MB` all parse. Boot validation is wired through `assertDuckDBResourceConfig` at `server.ts:296`, which is the right call for the same reason as #1077 — `/health` never touches DuckDB, so a typo would otherwise leave a pod reporting ready and serving nothing.

Also worth recording: no memoization on the new function, so it re-reads env every call and the `WeakMap`/`WeakSet` guards from #1077 cannot skip it. And the `SET preserve_insertion_order=false` is correctly ordered before the `set_option` CALL, with a test pinning that ordering.

**The `preserve_insertion_order` comment says the opposite of what is true.** `connection.ts:137-139` reads "Set on the session, not the catalog, so it binds this connection's writes and nothing else." I verified empirically: `preserve_insertion_order` has `scope: GLOBAL`, and with two connections on one instance, setting it on A flips B from true to false. `SET LOCAL` is not implemented in DuckDB, so there is no connection-scoped alternative.

The containment is real, but it comes from somewhere else: `createIsolatedBuildSession`'s per-build instance plus the read-only gate keeping this off the serve connection. That invariant is load-bearing and currently undocumented at this call site, while the comment actively tells the reader the write is contained by mechanism. Given `createIsolatedBuildSession`'s own comment describes the per-instance trick as a `db-duckdb` pooling workaround, someone simplifying it away would be reading a comment that says it is safe. Something like "GLOBAL in DuckDB (there is no `SET LOCAL`), so this binds the whole instance — contained because every read-write attach runs on a per-build instance from `createIsolatedBuildSession`, and the read-only gate keeps it off the serve connection" would both fix it and document what the gate is protecting.

**The docstring-orphaning has recurred, third time.** `connection.ts:107-112` — the new function landed between `sessionLimitsApplied`'s docstring and its declaration, so `sessionLimitsApplied` is now at line 153 with no docstring and there are two adjacent JSDoc blocks where the first describes nothing. Same shape as the two earlier instances. Moving the new function and its docstring below `sessionLimitsApplied`'s declaration fixes it. At three occurrences it might be worth whatever guard catches it — a lint rule, or just a habit of adding new config helpers at the end of the block.

**One behaviour worth a clause in the docs row.** The option is written to `ducklake_metadata`, so it is catalog-persistent rather than session-scoped. The docs row does say it "outlives the connection and is seen by every writer," which is honest — but the consequence is that **unsetting the variable does not restore the default**, and "Unset issues nothing" can be read as implying it does. Related open question: what is the intended way to revert a lake to DuckLake's default once a build has written the option?

**And the wiring is still untested, same as #1077.** Deleting the only call site at `connection.ts:855-857` — so the bound is issued nowhere, on any path — leaves the new spec at 4 pass and the full 2,493-test server suite exactly as green as baseline. Breaking the unset branch *is* caught, so the function is well covered in isolation; it is only the integration that nothing holds. There is a cheap fix in the existing house style: `connection_federation.spec.ts:274` already has an `attachDuckLakeReadWrite` describe block with a recording connection, so one case asserting the `set_option` CALL appears with the var set, plus a read-only counterpart asserting it does not, would pin both the wiring and the gate that currently provides the only containment for a GLOBAL setting.

Two smaller ones. The validator drops `TB|TIB` where `PUBLISHER_DUCKDB_MEMORY_LIMIT` accepts them, so `1TiB` is valid for one sibling byte var and a boot failure for the other — defensible if it is deliberately catching a unit slip, but then worth saying so, since the memory-limit regex carries exactly that kind of note. And there is no `RELEASE_NOTES.md` entry where #1077 had one; CI is green without it, so convention rather than gate.

The tests are the right shape, for the record — asserting issued SQL is the observable behaviour for this function and matches `incremental_storage.spec.ts`'s treatment of `pinSessionToUTC`, so no change-detector concern. The measurement reasoning also holds: a row group buffered per column predicts memory scaling with width and flat in row count, the byte-valued form is what makes that width-tracking (a row count cannot serve both a 9-column and a 110-column table), and the docs row is honest that only the write side was measured.

DCO signed, 15/15 CI green.

PR #1116, opened 2026-09-05
feat(server): bound the Parquet file DuckLake holds while writing to object storage

housejester:
`PUBLISHER_DUCKLAKE_TARGET_FILE_SIZE_BYTES` caps how large a Parquet file DuckLake writes
before rotating to the next one. Unset issues nothing at all on attach, so an existing
deployment is unchanged.

### Why

Writing to object storage, DuckDB copies each multipart part into a buffer it allocates
itself and holds it until the file completes. A file's bytes therefore stay resident however
they are grouped inside it, and peak memory tracks the **file** size. `memory_limit` does not
bound it — the allocation is tagged for the extension and the limit is not enforced against
it — and the row group bound added in 0.2.3 does not reach it, because that bounds the
per-column buffer *within* a file.

The same write to a local path streams and does not do this, so a deployment that
materializes to `s3://` or `gs://` pays a cost its local-disk testing will not show.

### Measured

72 columns, 20,000,000 rows, DuckLake CTAS to GCS, cgroup `memory.stat` `anon`. All six cells
in one batch — this figure moves with link speed, and DuckLake's `set_option` persists in the
catalog between runs, so cells measured apart are not comparable:

| `TARGET_FILE_SIZE_BYTES` | files | peak anon | write | full scan | catalog rows |
|---|---|---|---|---|---|
| unset — DuckLake's own ~512MB | 6 | 650 MiB | 45s | — | 432 |
| `1024MB` | 3 | 892 MiB | 36s | 13.0s | 216 |
| `512MB` | 6 | 550 MiB | 45s | 13.0s | 432 |
| `256MB` | 12 | 373 MiB | 50s | 14.5s | 864 |
| `128MB` | 23 | 262 MiB | 59s | 15.8s | 1656 |
| `64MB` | 45 | 255 MiB | 69s | 21.6s | 3240 |

DuckLake already rotates files, so this moves where it rotates rather than introducing
rotation — hence 650 → 373, about **1.74×**, at `256MB`.

The underlying effect is larger than that ratio suggests. A plain single-file `COPY` of the
same data measures 2979 MiB against 149 MiB writing to local disk, with S3 and GCS agreeing
within 0.1% (2976 vs 2979). DuckLake never writes that single file, but the contrast is what
identifies the mechanism, and it shows the behaviour is not specific to either cloud.

The two bounds are independent terms — on one 5M-row write, row group alone −18%, file size
alone −34%, both −65% — which is why this is a sibling of `applyDuckLakeRowGroupBound` rather
than folded into it.

### Choosing a value

**Largest that clears the ceiling, not smallest.** Memory is the only axis where smaller wins,
and it stops improving below ~`128MB`; write throughput, scan time and catalog rows all get
worse. Catalog rows are one per file per column, in a database every writer of that lake
shares.

Scan time is the least binding of the three where sources carry a required equality filter: a
selective lookup runs in 17–63 ms at every size tested, because file statistics prune to one
file and row groups prune within it.

### Relationship to the upstream fix

DuckDB reworked its object-storage upload in
[duckdb-httpfs#389](https://github.com/duckdb/duckdb-httpfs/pull/389) to stream from buffers
the engine already owns. Measured back to back on one network, the same single-file write
costs 545 MiB there against 1814 MiB on the pinned version. That work targets DuckDB 2.0, so
this option is the lever until it ships in a release we can take. It does not retire the row
group bound, which remains worth 16–21% on both engines.

### Tests

`ducklake_target_file_size.spec.ts` — nothing issued when unset; the CALL is catalog-scoped
rather than a session `SET`; a catalog that refuses the option does not fail the attach; and
the two bounds are independent in both directions. Boot-time validation rejects a unitless
byte count. `test:unit` green (3336 pass / 3 skip / 0 fail), typecheck and lint clean.


girishjeswani (approved):
Approving. I verified the mechanism against a live DuckLake catalog on DuckDB v1.5.5 rather than reasoning about it, and the claims hold up.

`target_file_size` is not a DuckDB setting — absent from `duckdb_settings()` — and `CALL lake.set_option(...)` then `FROM lake.options()` shows it with scope GLOBAL, value normalized to `256000000`, written to `ducklake_metadata` and surviving a fresh process. So catalog-persistent exactly like the row-group option, and "unset issues nothing" is accurate because unset simply leaves no row.

Two release-note claims I checked specifically, both correct. `target_file_size` genuinely does **not** require `preserve_insertion_order=false` — a 200k-row write succeeds with insertion order on and the option set, whereas `parquet_row_group_size_bytes` fails loudly with `Binder Error: ROW_GROUP_SIZE_BYTES does not work while preserving insertion order`. And ordering between the two options does not matter; I set them both ways and both land. Coverage and gating are right too: `!options.readOnly` at `connection.ts:890-893`, and across the four attach paths every write gets it and no read-only path does, with the plain DuckDB file destination correctly untouched since it is not a DuckLake.

The two mechanisms are also genuinely different terms rather than double-counting: the row group is a width-driven per-column buffer within a file, this is a length-driven cost of what the httpfs uploader holds resident, and both sit outside the buffer manager which is why `memory_limit` bounds neither. The local-disk control at 149 MiB against 2979 is the right experiment to isolate the upload path, and the duckdb-httpfs#389 citation is a good account of why this is temporary.

Two things I would fix before merge.

**`docs/configuration.md` has no row for the new variable.** I diffed every `PUBLISHER_*` symbol `config.ts` reads against the table at the PR head: the table carries both siblings at `:114-116` and stops there. #1077's review established that table as complete for every var `config.ts` reads and #1112 added its row, so this breaks a live invariant. A row mirroring the row-group one would want the counterintuitive parts an operator most needs from a config table: catalog option, skipped on read-only attach, persists in `ducklake_metadata`, validated at startup, and the "pick the largest that clears your ceiling" guidance.

**The `TB|TIB` gap got inherited rather than fixed.** `config.ts:434` is byte-identical to #1112's regex at `:424`, so the family now has two accept-sets rather than three, which is better than I feared — but I probed the live option and DuckDB accepts `1TB` (→ `1000000000000`) and `1TIB` (→ `1099511627776`) for this value, while refusing `1PB` and a bare `268435456` with "expected: KB, MB, GB, TB for 1000^i units or KiB, MiB, GiB, TiB for 1024^i units". So DuckDB's real accept-set here is exactly the memory-limit regex's, and the boot regex currently fails the pod on a value the engine would have taken — with a message reading "expected a size like 256MB", which sounds like the operator wrote something malformed rather than the validator being narrower than the engine. Adding `|TB|TIB` fixes it; hoisting one shared constant used by all three vars would end the question permanently. Minor related: the regex admits `0MB`, which DuckDB accepts and stores as `0`, and a zero file size is not a sensible request.

**One answer worth putting in the notes.** Since the option is catalog-persistent, unsetting `PUBLISHER_DUCKLAKE_TARGET_FILE_SIZE_BYTES` does not revert it — the last value written stays in `ducklake_metadata` for every writer of that lake. The notes do say it persists and is seen by every writer, which is prominent and correct, but they also advise tuning to the largest value that clears your ceiling, so operators will set this and then want to back it out. One sentence saying the revert is `CALL <lake>.set_option('target_file_size', '<old>')` against the catalog rather than removing the env var would close it.

**On the row-group figure, which spans this PR and #6309.** Row-group-only measures −18% on object storage here, against #1112's ~64% headline on a write whose destination the notes do not state. Separately I confirmed on the service side that all five deployed cells write to object storage unconditionally — both control planes hardcode a cloud bucket and there is no local `DATA_PATH` anywhere in `infrastructure/`. So the fleet-wide `32MB` from #6309 appears to be delivering roughly a fifth of what its own comment claims, in the only regime that actually runs. That is not an argument against this PR — the file bound is the larger term and clearly the thing that fixes the object-storage case — but it does mean #6309's justification is materially overstated where it is deployed, and `32MB` may deserve revisiting now that the dominant term is bounded. What destination was #1112's sweep measured against?

**And the wiring gap is a third occurrence.** Deleting the only call site at `connection.ts:892` leaves the new spec at 5 pass and the full server suite at 3334 pass / 2 fail, identical to baseline — the feature entirely disabled with 3,339 tests as green as before. Breaking the unset branch *is* caught, and well: two tests fail, including the independence one. So the function is covered in isolation and only the integration is not. I also checked whether the fix I suggested on #1112 landed — it did not; `connection_federation.spec.ts:274-297` is unchanged and asserts only the `ATTACH OR REPLACE` string. Setting both env vars in that block and asserting the recorded SQL contains both `CALL lake.set_option(...)` lines, with a read-only counterpart asserting neither appears, would catch this and retroactively cover #1112, in about a dozen lines.

Two smaller ones while you are in this file. The `sessionLimitsApplied` docstring stranded by #1112 at `connection.ts:108-113` is now separated from its declaration at `:188` by two functions rather than one — this PR did not create a new seam, but it widened that one. And the comment at `connection.ts:138-140` still says `preserve_insertion_order` is "Set on the session, not the catalog, so it binds this connection's writes and nothing else"; I re-confirmed on v1.5.5 that its scope is GLOBAL, so it binds the whole instance and containment comes from `createIsolatedBuildSession`'s per-build instance plus the read-only gate. Your new comment at `:157` points at that function as the reference for the same catalog-option mechanics, so it quietly inherits the wrong claim.

The spec is the right shape, for the record — asserting issued SQL is the observable behaviour here and matches the row-group spec beside it. The "scopes the option to the catalog, not the session" test is a real invariant that would catch a refactor from `CALL` to `SET` silently changing persistence semantics, and the independence test with its comment explaining why one direction is insufficient is genuinely good. Release notes verified claim by claim rather than just present. DCO signed, 15/15 CI green.

housejester:
Thanks — the live-catalog verification is worth more than the review itself. All four taken, plus both comment defects. Pushed in `fd451d53`.

**`docs/configuration.md`.** Added, mirroring the row-group entry: catalog option, skipped on read-only, persists in `ducklake_metadata`, validated at startup, and the largest-that-clears-your-ceiling guidance.

**`TB`/`TIB`.** Hoisted rather than patched. All three size settings now share one `DUCKDB_BYTE_SIZE` constant, so the family has a single accept-set instead of two — which fixes `PUBLISHER_DUCKLAKE_ROW_GROUP_SIZE_BYTES` retroactively as well. Your `0MB` note is in there too, as a leading-zero guard. Covered both ways: `1TB`/`1TIB`/`1tb`/`1.5GB`/`0.5GiB` accepted, `1PB`/`268435456`/`1GBB`/`0MB` rejected, and narrowing the constant back fails three of them.

**Revert.** In the notes and the docs row: unsetting the variable does not revert the lake, and the way back is `CALL <lake>.set_option('target_file_size', '<value>')` against the catalog.

**The wiring gap — the most useful thing in the review.** Reproduced exactly: deleting the call site left the new spec at 5 pass. `connection_federation.spec.ts` now asserts a read-write attach issues both `CALL ... set_option(...)` lines. Deleting *either* call site fails it, so #1112 is covered retroactively.

Two things about that fix you should know. **My first attempt at it was itself vacuous** — the block threw at registration on a missing `afterEach` import, never ran, and the suite reported all-green. Only the mutation check caught it, which is your own finding recurring one level down inside the fix for it.

And **I did not implement the read-only counterpart.** That attach is module-private and reachable only through the connection factory, and I wasn't willing to export it purely for a test. The gate is one `if (!options.readOnly)` that both calls sit inside, so it's covered by inspection rather than assertion. If you think exporting the pair symmetrically is warranted, say so and I'll do it — I'd rather flag the hole than quietly drop that half.

**Both comments.** Confirmed `preserve_insertion_order` is `GLOBAL` on v1.5.5 myself. The claim that it "binds this connection's writes and nothing else" is corrected to say what actually contains it — a per-build instance from `createIsolatedBuildSession` plus the read-only gate — and the new function's reference now names what is *not* shared, so it no longer inherits it. The stranded `sessionLimitsApplied` docstring is back against its declaration.

---

**On the row-group figure — you're right, and I can answer the question directly: #1112's sweep was against local disk.** Every cell was `DATA_PATH '/tmp/dl_b32_1/'`, `/tmp/dl_bytes16/` and so on. Not one object-storage destination in it.

So your reconciliation is the correct one, and combined with your finding that every deployed cell writes to a cloud bucket unconditionally, the `−51%` standing in `terragrunt.hcl:112-118` describes a regime that never runs. The deployed value is delivering something closer to `−18%`.

Worth being clear that `32MB` was still the right call on what was known then: the measurements available said the row group dominated, and `32MB` over `16MB` was chosen specifically to protect read-side pruning. What has changed is that the dominant term is now bounded elsewhere — so the trade has moved, and as you say a *larger* row group may now be the better choice. That is new information rather than a wrong decision, and it wants its own measurement on an object-storage sink rather than a correction here. I'll file it separately along with fixing that comment.

Gate after the changes: 3352 tests / 0 fail, typecheck and lint clean.

PR #1126, opened 2026-09-07
feat(givens): graft a #(partition) row filter at read time, off filter-typed givens

Sha-Bang:
Adds `#(partition) <column> = $GIVEN`: a source-level annotation whose predicate publisher grafts as a row filter at read time, and skips under the authorize bypass so an index build scans every slice in one pass.

It reuses `#(authorize)`'s machinery rather than paralleling it — same graft, same landing proof, same inheritance rule — and closes the three fail-closed protections that stop applying once the predicate leaves the source body.

## Why the predicate moves out of the source body

The Credible dimensional index is built one slice per partition value. Today a partitioned source carries its own `where: dim ~ $GIVEN`, which forces two things: the given must be `filter<T>` (index time binds it to `""`, and only a filter expression reads `""` as "match everything"), and every scan has to neutralize a predicate it does not want.

With the predicate grafted at read time there is nothing to neutralize at index time, so the given can be an ordinary `string` or integer `number`.

```malloy
given:
  TENANT :: string

#(partition) tenant = $TENANT
source: orders_scoped is orders extend { ... }
```

## Mechanism

`resolveEntryPointPartitions` answers which `(column, given)` pairs apply to an entry point, using the **same own-wins-else-ancestor rule** `#(authorize)` uses — the `.inherits` extend chain, the `sourceRegistry` link, and a `query_source`'s own base. It deliberately does not delegate to `ancestorGateExprs` — that function's `["false"]` sentinel is specific to a boolean access gate — but it matches that function's posture: an unreadable chain denies, because losing a partition axis serves every slice rather than merely producing a worse index. Only a chain read to its end with nothing found returns `[]`.

Resolved pairs are merged into the same `rowLevel` list authorize gates use, so `authorizeAndBindRunnable`'s existing graft, proof and deny path covers both. That is what makes the filter reach a **named query invoked by `queryName` alone** and a **notebook cell** — neither of which the legacy `#(filter)` text-injection path can see, because it resolves the source by run-target name.

Joined sources are not traced, matching the documented Q16 posture: a `#(partition)` is a statement about the entry point's own source.

## The three refusals, and why none of them is inherited

Moving the predicate out of the source body removes the given reference that three existing guards keyed on. Each needed its own check.

| Guard | Why it stops firing | Consequence if left alone |
|---|---|---|
| Storage materialization (`referencesGiven`) | The source no longer references a given | A frozen artifact holds every tenant's rows |
| Colocated persist | Checks only the authorize condition, by design | The storage refusal is sidestepped by dropping `storage=` |
| `routingBlockedByRowLevelGate` | Keys on `hasAnyAuthorizeNote()` | A storage-routed query has its givens **suppressed**, so the predicate cannot bind at all |

That third one is the sharp edge: the serve shape is built from given-free sources and carries no `given:`, which the code notes is *"safe only because the storage eligibility gate already refused any given-referencing source."* Remove that premise and the veto has to take its place.

`hasAnyPartitionNote` shares its struct sweep with `hasAnyAuthorizeNote` so the two cannot drift on which structs get walked, and returns `true` on unreadable IR so the veto engages rather than being skipped.

A composite source is refused at publish whichever way it reaches a `#(partition)` marker, and the two ways fail very differently. A marker on the composite *itself* denies every read — the graft lands on the composite's own struct, the landing proof fails, and the author gets an opaque error. A marker on a *member* was silent, and worse: measured before the refusal existed, wrapping a partitioned source in `compose(...)` served **every** partition. `run: combo` returned all three seed rows under `TENANT: 'acme'` while `run: marked` correctly returned two, and forcing the marked branch with a member-only dimension changed nothing. That is a cross-tenant read reachable by adding one `compose(...)`, so both are refused.

Relatedly, the marker walk now denies when it cannot read the derivation chain it would have to follow — an exhausted `inherits` cap or an unresolvable registry link. It previously answered "no partition", which serves unfiltered; its authorize twin `ancestorGateExprs` already returns `["false"]` for both shapes. The load-time sweep swallows that one cause on purpose: the read path is where the guarantee is enforced, and refusing a whole model over one unreadable IR link would take down its unmarked sources too. That sweep also covers non-composite sources now, so a malformed body or a duplicate given fails at load with its own named cause instead of denying every read opaquely.

A marker declared one line too low — on a dimension, a view, or inside an inline `compose(...)` — was the quietest shape of all. Resolution reads the source's own notes, so it found nothing, grafted nothing and refused nothing: the source read as partitioned in its own text and served every partition. A targeted resolver cannot validate itself, so the check is an unrestricted IR walk that refuses when it finds a marker resolution did not. It skips the root's own annotations (that is what resolution reads) and skips joined sources, keeping the Q16 posture rather than refusing a source merely for joining a partitioned one.

## Testing

`src/service` **2647 pass / 0 fail** locally, `tsc --noEmit` clean, 15 new integration tests plus 2 eligibility tests. The repo's watch-mode E2E specs do not complete on this machine (they bind a server and time out); they are untouched by this diff, so CI is the arbiter for the rest of the suite.

The graft is proven to land for a plain source read, a named query invoked by `queryName` alone, an ad-hoc `run: <source> + {...}`, and a notebook cell — and to be **absent** under `x-publisher-bypass-authorize`, which is what the index build depends on.

It is keyed on the authorize bypass, never on `bypassFilters`: that flag is a public query param and body field, so keying a confidentiality control on it would be a one-parameter cross-tenant read. A request sending `bypassFilters: true` still gets only its slice, and there is a test for it.

Each guard was reverted and the failure read, rather than trusted because the suite was green — the old guards still pass while covering none of these cases, so a green suite proves nothing here:

| Reverted | Failure |
|---|---|
| Storage eligibility | `Expected constructor: MaterializationEligibilityError` / `Received function did not throw` |
| Colocated persist | same shape |
| Routing veto | `Expected: true` / `Received: false` |
| The graft itself | `Expected: 2` / `Received: 3` (a row count, so the test checks slicing rather than the presence of a filter) |
| Composite refusal (own marker) | `Expected constructor: PartitionAnnotationError` / `Received value: undefined` |
| Composite refusal (member marker) | same shape — and the pre-fix measurement above is what it prevents |
| Misplaced-marker refusal | `Expected constructor: PartitionAnnotationError` / `Received value: undefined` |

## Notes for review

- No partition-specific metric labels; this reuses `recordRowLevelGateDecision` / `recordEligibilityRefused`.
- An unbound partition given surfaces Malloy's own "no value and no default" error naming the given, rather than authorize's opaque 403. A partition given is an ordinary model given, so the name-hiding posture authorize maintains for cross-import-hop givens did not seem owed here — worth a second opinion.
- Only `=` is accepted. Negation, comparisons, `in` and compound booleans are each refused with their own cause, because the index is keyed by a single positive slice value.


jswir:
Read through this carefully and ran the interesting parts. The design holds up: everywhere the marker **is** found, this fails closed four different ways (no graft scope, no graft target, given off the model surface, unreadable ancestry), and the revert table is the right way to have proven each guard. The `bypassFilters` test in particular is the one I went looking for and was glad to find.

None of the below is a blocker. But there are two shapes where the graft silently doesn't attach, and I think at least the first is worth closing here.

## Two placements where no filter attaches

Both compile clean, publish clean, and serve every partition. Ran both at `26cc88f2` against real DuckDB using this PR's own harness (seed: 2 `acme` rows, 1 `globex`).

**1. Annotation one line too low**

```malloy
source: X is duckdb.table('tenant_rows') extend {
   #(partition) tenant = $TENANT   // lands on the measure, not the source
   measure: n is count()
}
```

`run: X -> { aggregate: n is count() }` with **no `TENANT` bound at all** returns 3 rows. That's what makes it conclusive rather than just suspicious — if the filter had attached, Malloy refuses a given with no value and no default, which is exactly what your test at `partition_graft.integration.spec.ts:314` asserts for a comparable shape. `##(partition)` before any `source:` line behaves the same way.

**2. Inline `compose(...)` under a query-source**

```malloy
#(partition) tenant = $TENANT
source: marked is duckdb.table('tenant_rows') extend { measure: n is count() }
source: openm is duckdb.table('tenant_rows') extend { measure: n is count() }

source: qs is compose(marked, openm) -> { group_by: tenant, val }
```

`TENANT: 'acme'`, `TENANT: 'globex'`, and no given bound all return the same 3 rows and generate byte-identical SQL — no `WHERE`, no reference to the given anywhere. Naming the composite (`source: combo is compose(marked, openm)`) is correctly refused, so the difference is only whether it gets its own top-level `modelDef.contents` entry.

## The question I'd actually put to you

On #2 specifically: you already decided this bug was worth fixing in this PR. The refusal you added carries its own measurement —

> `MEASURED: querying the composite reads EVERY partition of that member — the graft lands on the composite's own struct, not the resolved member branch, so no filter reaches the executed query.`

The inline form is that same case with the composite unnamed. So rather than argue severity: if the named form cleared the bar for a publish-time refusal here, does the inline one, or is there a reason it doesn't that's worth writing down? `assertPartitionAnnotationsValid`'s doc reasons about where the graft would *land* for this shape, which is right, but the marker is never *discovered*, so it never gets that far.

## Why the fix looks cheap

`walkForPartition` (`materialization_eligibility.ts:469-510`) is an unrestricted deep walk and already finds the marker in both shapes above. So this can be a reconciliation rather than new machinery: refuse at load when a partition-routed note exists anywhere no entry point can graft it.

That's also the one piece of `#(authorize)`'s machinery this doesn't inherit — `source_extraction.ts:281-291` on main spells out the same hazard for the same reason:

> "every `#(authorize)` annotation this walk found attached to a `dimension:`/`measure:`/`view:` field (author-written one line too low) ... All are positions nothing here, or anywhere else, ever reads for enforcement; see `assertNoMisplacedAuthorizeAnnotations`'s doc for why that **fails OPEN** and has to be refused at load rather than silently ignored."

Worth noting `model.ts:1153-1156` justifies the narrower reach with "a field can never carry one" — that's the claim shape #1 falsifies, and this PR's own storage guard is written to catch exactly that case.

## Small, cheap, would do here

**The fail-open rationale is inverted.** The paragraph above `resolveEntryPointPartitions` (`gate_classification.ts:994-996`) says it "returns `[]` ... in every case `ancestorGateExprs` would deny," on the grounds that losing a partition axis is "only a worse index." Twelve lines down, at `:1045` and `:1049`, both of those cases `throw`. 26cc88f2 fixed the code and the two inline comments and left the JSDoc — and the PR description, which still carries the fail-open reasoning. The code is right; it's the explanation that's backwards, and it's the paragraph the next person reads to decide which way this fails.

## Follow-ups, not for here

- `#(partition)` has no docs anywhere in `docs/` or `skills/`. `docs/materialization.md:56` already walks through the identical three refusals for `#(authorize)`, and `docs/givens.md:24` is where someone would look. Also not marked provisional.
- The graft calls `liftGateCondition` directly (`gate_classification.ts:1289`), skipping `gateShapeCache` (`:522`, `:702`). That's a live probe compile per query that an authorize gate on the same source doesn't pay; the cache key is `(cacheScope, graftTarget, filterText)`, all of which partition has.
- `materialization_metrics.ts:245-246` still enumerates the old five reasons — the TS union and `api-doc.yaml` both got `partition`, this third hand-synced copy didn't.
- `ancestry_unresolvable` is the only rejection cause with no test, and it's the fail-closed path 26cc88f2 added. The synthetic-IR harness in `partition_resolution.spec.ts` already builds the shape.
- No test reaches a partitioned source through `import`. That hop was a blocking bug in #1045 on this same graft path; the notebook test declares its source inline.
- The veto's `hasAnyPartitionNote()` arm (`model.ts:4871`) isn't exercised end-to-end — the test calls `queryEntryPointHasRowLevelGate` through a cast, past the short-circuit.
- `partition_annotation.ts:15-17` still says "(in a follow-up change) the row-filter graft"; that's `c2d11285`, in this PR.

On the ancestor walk being re-implemented rather than shared with `ancestorGateExprs`: the stated reason is the `["false"]` sentinel, and 26cc88f2 made the terminal behavior match anyway — but you've been further into these IR shapes than I have, so if they diverge more than they look from outside, say so and I'll drop it.

jswir (approved):

PR #1129, opened 2026-09-08
fix(connections): answer a bad table reference with 404 or 400, not 502

Sha-Bang:
A table path that names nothing, or that the dialect cannot parse at all, is the caller's mistake. Both were answered with **502**, so browsing for a table that turned out not to exist read as a server fault.

That has a downstream cost beyond tidiness: a 5xx is counted against a server-error budget. On Credible's deployment it paged on-call **twice in one afternoon** — 616 errors in one 30-minute window, then 415 in another — for what was, both times, a modeler typing a table name into an editor. A compiler re-resolves an unresolved `table()` reference on every recompile, so one bad reference in an open file becomes hundreds of requests.

## What changes

| Condition | Before | After |
|---|---|---|
| `Not found: Table\|Dataset …` | 502 | **404** + `reason: "TABLE_NOT_FOUND"` |
| `Improper table path …` | 502 | **400** |
| Publisher-authored not-found (Azure, DuckDB, DuckLake) | 502 | **404** |
| Anything else (unreachable, credentials, quota) | 502 | **502**, unchanged |

**404 was already the declared contract** — 502 appears in no spec for these routes — so that half is conformance, not a break. 400 is newly declared alongside it.

## Why match on message text

BigQuery's driver signals failure by **returning** `error.message` instead of throwing, discarding the structured `code: 404` the Google client gave it:

```js
// @malloydata/db-bigquery — fetchTableSchema
catch (error) { return error.message; }
```

The text is the only surviving signal, so two of BigQuery's own API messages are matched. The second one, `Improper table path`, is not an edge case: BigQuery needs `dataset.table`, so **every prefix typed before the first dot** has one segment and lands here. It was 42% of the errors in the second incident.

Anything unrecognized stays `ConnectionError`/502, so a real outage is still loud.

**Deliberately not generalized to other dialects.** Postgres reports a missing table as the generic `Unable to read schema.`, indistinguishable from any other failure; Snowflake's `DESCRIBE TABLE` answers *"does not exist or not authorized"*, conflating absence with denial **by design**. Guessing on either would be worse than leaving them 502. The publisher's own wrappers need no matching and are converted directly.

## The trap this closes

`fetchTable`'s catch rewrapped *everything* as `ConnectionError`. Classifying without touching it would have silently undone the fix — and would have missed every dialect that throws rather than returning. It now preserves an already-classified error and classifies a thrown message the same way.

A not-found is logged at `warn` rather than `error`, so a path being typed cannot fill the error log. (Route handlers still log at error; this removes one of two lines per miss.)

## `reason` on the response

`TableNotFoundError` carries `reason: "TABLE_NOT_FOUND"`. A caller that retries a 404 by re-resolving which server hosts a resource — as Credible's router does, invalidating a shared cache entry and re-asking its control plane — needs to tell that case apart from a table that is simply absent, and status alone cannot say which. Optional and additive; callers that ignore it are unaffected.

## Testing

`bun test packages/server/src` — 3546 pass, 7 pre-existing failures unchanged against a clean baseline (`build_targets_address_equality`, `service/package` xlsx; unrelated to this change).

Six new tests, and I verified they fail for the right reason: reverting the catch to its blanket rewrap fails **5 of 6**, while the "unrecognized failure stays 502" floor test correctly stays green.

- BigQuery's returned not-found string → `TableNotFoundError`
- `Improper table path` → `InvalidArgumentError`
- a driver that **threw** its not-found → `TableNotFoundError` (the non-BigQuery path)
- an empty schema result → `TableNotFoundError`
- a `TableNotFoundError` thrown by a wrapper survives the catch — asserting the **class**, since a status assertion would not prove the rethrow branch ran
- `ECONNREFUSED` → still `ConnectionError`
- a BigQuery **auth** failure → still `ConnectionError`, so the match cannot swallow a credentials problem

## Follow-up, not in this PR

`db-publisher` implements only `fetchTableSchema` and never overrides `fetchSchemaForTables`, and the base implementation awaits each iteration — so a cold 40-table model is **40 serialized HTTP round trips**. Measured p95 on the affected route was 181 ms for a miss but **3.22 s for a hit**, so batching (or even parallelizing) is the real latency lever. Separate change.

The cleanest fix of all is upstream of this repo: `@malloydata/db-bigquery` should preserve the structured error instead of returning prose, which would make this classification unnecessary.

### DIFF
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -731,6 +731,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Table"
+        "400":
+          $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "404":
@@ -1089,6 +1091,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/Table"
+        "400":
+          $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "404":
@@ -5347,12 +5351,29 @@ components:
       type: object
       description: Standard error response format
       properties:
+        code:
+          type: integer
+          description: HTTP status code, repeated in the body for convenience
         message:
           type: string
           description: Human-readable error message describing what went wrong
         details:
           type: string
           description: Additional error details or context
+        reason:
+          type: string
+          description: |
+            Machine-readable discriminator, for callers that must branch on
+            which error they got rather than on prose. Absent unless a caller is
+            expected to act on it, so treat absence as "no special handling".
+
+            TABLE_NOT_FOUND distinguishes a table that is not in the database
+            from the other conditions that also answer 404 on these routes --
+            an environment or connection this server does not host. The former
+            says nothing about whether the caller reached the right server; the
+            latter is a routing problem worth retrying elsewhere.
+          enum:
+            - TABLE_NOT_FOUND
       required:
         - message
 
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -781,6 +781,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SqlSource"
+        "400":
+          $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "404":
@@ -1147,6 +1149,8 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/SqlSource"
+        "400":
+          $ref: "#/components/responses/BadRequest"
         "401":
           $ref: "#/components/responses/Unauthorized"
         "404":
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -5376,8 +5376,13 @@ components:
             an environment or connection this server does not host. The former
             says nothing about whether the caller reached the right server; the
             latter is a routing problem worth retrying elsewhere.
-          enum:
-            - TABLE_NOT_FOUND
+
+            Deliberately an open string rather than an enum. Generators render
+            an enum closed, so adding a second value here would break every
+            client generated before it -- on the one field whose purpose is to
+            grow. Callers must treat an unrecognized value as absent.
+
+            Known values: TABLE_NOT_FOUND.
       required:
         - message
 
--- a/packages/server/src/controller/connection.controller.spec.ts
+++ b/packages/server/src/controller/connection.controller.spec.ts
@@ -952,6 +952,25 @@ describe("ConnectionController getTable not-found mapping", () => {
       );
    });
 
+   it("classifies DuckDB's rejected catalog errors, which never resolve falsy", async () => {
+      // DuckDBCommon.fetchTableSchema returns a structDef or throws -- it never
+      // resolves empty -- so the falsy check above cannot see a DuckDB miss and
+      // the sandbox, Azure and DuckLake connections all arrive here instead.
+      // Strings verbatim from the installed driver.
+      for (const message of [
+         "Catalog Error: Table with name no_such_table does not exist!",
+         'Catalog Error: Table with name "s.t" does not exist because schema "s" does not exist.',
+         'Binder Error: Catalog "a" does not exist!',
+      ]) {
+         const { controller } = buildTableController(
+            sinon.stub().rejects(new Error(message)),
+         );
+         await expect(getTable(controller)).rejects.toBeInstanceOf(
+            TableNotFoundError,
+         );
+      }
+   });
+
    it("still reports an unrecognized driver failure as a 502-class fault", async () => {
       // The floor: anything the classifier does not recognize stays a server
       // error. Guessing wider would hide real outages behind a 404.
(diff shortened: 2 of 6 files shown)


housejester (commented):
The change is right, and the reasoning quality is high — the conservative fallback, the explicit refusal to generalize to Postgres/Snowflake, and the extension-vs-table carve-out all look like the correct calls to me.

Credit where it's due first: the testing discipline here is the standard I'd want. "Reverting the catch fails 5 of 6, and the floor test correctly stays green" is proof the assertions would have failed before the fix, rather than a coverage claim. And I checked the spec claim rather than taking it — `"502"` appears **nowhere** in `api-doc.yaml`, so "404 was already the declared contract" is if anything understated.

## Mental-model delta

The thing I'd want a future reader to absorb in thirty seconds: the HTTP status for a schema-introspection failure stops being a constant (`502` for everything) and becomes *derived from the driver's error prose*, with 502 demoted to the fallback for anything unrecognized.

|---|---|---|
| Where the status is decided | one `catch`, one answer: `ConnectionError` → 502 | a message classifier (`driverErrorToPublisherError`) picks 404 / 400 / 502 |
| Load-bearing invariant | *none* — the catch was total | **the regex set is the contract**; unmatched prose falls back to 502 |
| Failure direction on a classifier miss | n/a | fail-safe: a miss stays 502, so an outage stays loud |
| What a 404 on these routes tells a caller | "wrong environment / wrong connection" only | that **plus** "table genuinely absent" — disambiguated only by the new `reason` |
| Error-body contract | `{code, message, details?}` | `+ reason?: "TABLE_NOT_FOUND"` (optional, additive) |

Four things a future dev has to hold in their head:

1. Adding a dialect means adding a regex, and forgetting to means that dialect silently keeps returning 502 — safe, but invisible.
2. A driver version bump that rewords a message silently reverts that case to 502. Nothing detects this.
3. Postgres and Snowflake are *deliberately* excluded, because their not-found prose is ambiguous with auth denial.
4. `reason` is the only machine-readable way to tell "absent table" from "wrong server", and both are 404.

Worth naming the new coupling explicitly, since it's the part that isn't visible in the diff: these status codes can now drive a caller's **retry policy**, not just its error display. That's exactly what `reason` exists to disambiguate, and it's why I think the field earns its place rather than being speculative generality.

What this deliberately does *not* change: any 2xx path, any other route, any other dialect's classification, or the 502 fallback itself. Reviewers can skip those regions.

## Worth acting on

**A misconfigured connection could report as a mistyped table.** `DUCKDB_FILE_NOT_FOUND = /^IO Error: No files found that match the pattern/` fires on the Azure blob branch and on any parquet/csv path. A wrong container name, a bad mount, or a listing that an expired credential can't populate could each surface as "no files found" rather than as an auth error — and then *every* table on that connection answers 404, which looks like user error and alerts nobody.

This is the same risk you reasoned about carefully for extensions ("a 404 here would report a broken deployment as a mistyped table") and then drew the line on the other side of. Genuinely asking rather than asserting: do we know that Azure/httpfs raises a distinguishable error for a bad container or an expired SAS, rather than an empty match? If it doesn't, this one may belong on the 502 side of the line you already drew.

## Smaller notes

**A single-value closed enum is a future skew trap.** `api-doc.yaml` declares `reason` as `enum: [TABLE_NOT_FOUND]`. Generators render that as a *closed* enum, so the day a second reason is added, an older generated client deserializing the error body rejects the unknown value — a breaking change built into a field whose whole purpose is to grow. Suggest a plain `type: string` with the known values enumerated in the `description`, or an `x-extensible-enum`. Cheap now, expensive later.

**The evidence covers one test tier.** The reported run is `bun test packages/server/src`, which is `test:unit`. `test:integration` (`bun test tests`, needs a built dist) isn't reported. Since this changes status codes on live routes, `tests/integration/legacy_routes` and `duckdb_storage` look like the plausible places something asserts the old shape. Probably clean, but it's a cheap confirmation and it's the kind of change the unit tier can't fully speak for.

**`classifyDriverFailure` is an allowlist that has to be kept in sync.** It preserves exactly `TableNotFoundError | InvalidArgumentError` and rewraps everything else as `ConnectionError`. Correct for what the wrappers throw today, but the first wrapper that throws a `ConnectionAuthError` or a `BadRequestError` gets silently downgraded to 502. Inverting it — preserve anything publisher-authored, classify only genuinely-foreign errors — would make it self-maintaining.

(I chased whether `UnsupportedCatalogFormatError`'s 422 was already being eaten here; it isn't — the lazy DuckLake attach runs in `getMalloyConnection`, outside `fetchTable`'s `try`. No issue. Noting it so nobody else re-walks it.)

## Minor / non-blocking

- The spec now declares 400 alongside 404 on the four routes, but 502 remains undeclared anywhere in `api-doc.yaml`, including on the `ConnectionError` path this PR preserves. Not this PR's job to close, and the description is honest about it — just noting the conformance gap is narrowed rather than closed.
- `code` is added to the `Error` schema as optional, though `httpError()` always emits it. Harmless, and optional is the safer declaration if any other path constructs an error body.
- All the regexes are `^`-anchored against driver prose, which is the right strictness and fails in the safe direction. Not worth building anything for — just worth knowing that a reworded driver message is a silent revert to 502.

**Two to resolve before merge:** decide whether `IO Error: No files found` belongs on the 404 or the 502 side of the line you drew for extensions, and drop the closed enum on `reason`.

Two things I did not do, flagging as such: I didn't run the branch, and I read `service/connection.ts` and the controller at the PR head rather than from a local build.

housejester (approved):
Both points are addressed, and both went further than I asked. Approving — what's left is description-only plus one question.

**`IO Error: No files found` back on the 502 side** (`0710acb0`). Right call, and the test now also pins `IO Error: Could not resolve hostname…`, which shares the `IO Error:` prefix. That's the guard against someone later re-adding a broad `^IO Error:` pattern and quietly swallowing a network fault — the assertion fails if they do. Worth adding that a credential which expires mid-life on a file-backed connection lands in this same shape, and an expiring token silently turning every table into a 404 is exactly the failure this now prevents. The narrower match is the right trade.

**Open string instead of a closed enum** (`d71f4f3f`). Fixed, and the rationale is in the spec where the next person will read it. The asymmetry you landed on is the correct one and worth naming so nobody "tidies" it later: `ErrorReason` stays a one-value union in `errors.ts` — strict about what the server *emits* — while the wire contract is an open string with known values documented, liberal about what clients must *accept*. Right way round.

I diffed the two revisions line-for-line; those are the only two changes, and nothing else moved in the rebase.

## One thing left, and it's the description

The **"What changes" table** still reads `Publisher-authored not-found (Azure, DuckDB, DuckLake) | 502 | **404**`. That's now two-thirds true: DuckDB catalog errors and DuckLake catalog misses do become 404, but an absent file-backed table stays 502 — which is what `0710acb0`'s own test asserts. The table is the part people quote, so it's worth a row split.

Same section, **"Six new tests"** with seven bullets under it: the spec files now add **17** `it(...)` cases. The DuckDB catalog set, the extension guard, the file-backed guard, the four `getConnectionSqlSource` cases and the three in `errors.spec.ts` are all unlisted — and a couple of them are the most interesting tests in the PR.

## One question, non-blocking

Testing reports `bun test packages/server/src`, which is `test:unit`. Was `test:integration` (`bun test tests`, needs a built dist) run? Since this changes status codes on live routes, `tests/integration/legacy_routes` and `duckdb_storage` look like the plausible places something could assert the old shape. Probably clean — just not covered by the tier that was reported, and a line either way closes it.

## Suggestion for the follow-up section

Worth one sentence recording that **file-backed tables — an Azure blob, or any parquet/csv path — still answer 502 for a mistyped path**, because the file is addressed by path and an absent one is indistinguishable from a misconfigured mount. It's the right trade as argued above, but it's the natural "why didn't this fix my case" question, and better found in the follow-up section than re-derived.

Not re-litigating `classifyDriverFailure`'s two-class allowlist — correct for what the wrappers throw today, and a fine call to leave.

Sha-Bang:
Both of the two-to-resolve are done, and the IO Error one went somewhere different from where I expected.

## `IO Error: No files found` — removed, but not for the reason you asked about

I probed the pinned driver rather than reasoning about it, because the extension case taught me my intuition here is not reliable. Every failure mode I can construct is distinguishable:

| case | driver says |
|---|---|
| missing Azure credentials | `Invalid Input Error: No valid Azure credentials found!` |
| unresolvable host | `IO Error: Could not resolve hostname error for HTTP HEAD to '...'` |
| remote object absent | `HTTP Error: HTTP GET error on '...' (HTTP 404 Not Found)` |
| failed DuckLake attach | `IO Error: Failed to attach DuckLake MetaData ... at path` |
| failed Postgres attach | `IO Error: Unable to connect to Postgres at "...": could not translate host name` |

None match the anchored pattern, and there is a mechanism behind that rather than luck: `No files found that match the pattern` is glob resolution **succeeding and matching nothing**, whereas an auth or host failure fails the listing itself and words itself differently.

So your specific worry does not reproduce. I removed the pattern anyway, on a better argument that the probing surfaced: a file-backed table is DuckLake or an Azure blob, and an absent parquet under one of those is corruption or a bad mount, not a name someone mistyped. That is the same call already made for extensions, and I had drawn the line on the wrong side of it. The wrapper's own "Azure file not found" throw goes back to being unreachable, which is honest -- the driver rejects before the falsy guard it sits behind.

The test that pinned the old behavior now pins the new one, with the unresolvable-host string beside it so the shared `IO Error:` prefix cannot creep back into the match.

## Closed enum — dropped

`reason` is a plain `type: string` now, with the known values in the description and callers told to treat an unrecognized value as absent. Your reasoning applies with extra force downstream: the consumer mirrors this spec and diffs it against the release, and an enum present on only one side is a drift finding in its own right.

## Integration tier — ran it

`tests/integration/legacy_routes` and `duckdb_storage`: **15 pass, 0 fail**. Also swept the whole `tests/` tree -- `502` appears nowhere in it, and no integration test exercises these routes' error paths at all, so the unit tier was in fact speaking for the whole assertion surface here. Worth confirming rather than assuming, as you said.

## `classifyDriverFailure` allowlist — agreed in principle, not taking it here

You are right that it is maintenance-shaped rather than wrong. Checking reachability: inside `fetchTable`'s `try` the only publisher-authored throws are `TableNotFoundError` (both wrappers, and the falsy guard), and `getMalloyConnection` -- where `ConnectionAuthError` and `UnsupportedCatalogFormatError` originate -- is outside it, matching what you found for the 422. So nothing is being downgraded today.

Inverting it cleanly needs something to test *against*: every class in `errors.ts` extends `Error` directly, so "preserve anything publisher-authored" means either a shared base class introduced across all ~20 of them, or a list that is the same maintenance burden pointed the other way. That is a real change and not one I want to smuggle into a status-code fix, so I would rather leave the two-item allowlist honest and small.

## Also

502 is still undeclared in `api-doc.yaml`, as you noted -- still not this PR's job, still narrowed rather than closed.

Issue #1130, opened 2026-09-08
Semantic retrieval: `--init` drops the embedding cache, and readiness is unobservable without warming it


jswir:
Two related observations about the semantic retrieval index, found while
building a harness that measures retrieval quality across package versions.
Both are about the *embedding cache being treated as schema state rather than
as a derived cache*, and both make it easy to get answers from a retriever you
did not think you were using.

Line references are against `main` at the time of writing.

## 1. `--init` drops `entity_embeddings`

`dropAllTables` (`packages/server/src/storage/duckdb/schema.ts:795`) lists
`entity_embeddings` among the tables it drops
(`packages/server/src/storage/duckdb/schema.ts:805`). It runs whenever
`initializeSchema` is called with `force = true`, which is `--init` /
`INITIALIZE_STORAGE=true`.

`entity_embeddings` is not schema state that needs resetting with the schema.
It is an expensive derived cache, and it is already content-addressed: the
incremental sync in `packages/server/src/mcp/tools/embedding_index.ts`
re-embeds only documents whose `content_hash` changed, which is exactly the
behaviour an edit-and-reload workflow wants. Dropping the table defeats that
and makes the next `get_context` re-embed every facet of every loaded package.
For a package of any size that is a real embedding bill and a long warm-up,
repeated on every `--init` boot.

The two concerns are also coupled through the one flag, which is what makes
this hard to avoid: `--init` is *also* the only way to make the server re-read
`publisher.config.json` rather than load environments from the database
(`packages/server/src/service/environment_store.ts:603`, and the `reInit`
branch at `:623`). So an operator who needs a config change to take effect pays
a full re-embed as a side effect, with no way to ask for one without the other.

Two possible directions, either of which would help:

- leave `entity_embeddings` out of `dropAllTables`. `content_hash` already
  makes it self-correcting, so a stale row is repaired on the next sync rather
  than needing a drop.
- separate "re-read the config" from "drop the store", so the config path does
  not require the destructive one.

## 2. Retrieval readiness is not observable without warming it

The sync memo is a `WeakMap` keyed by the `Package` object
(`packages/server/src/mcp/tools/embedding_index.ts:432`). So it is lost not
only on a process restart but on **every package reload** — `reload_package`,
REST `?reload=true`, and each watch-mode recompile all call `Package.create`
(`packages/server/src/service/environment.ts:1918`), producing a new instance
with no memo.

After that, the first `get_context` for the package kicks a sync that is
[...]

PR #1114, opened 2026-09-04
fix(skills): make an eval run say what it measured, and fail loudly when it cannot

jswir:
Follow-up to the review on #1101, plus three rounds of findings since. Seven commits, one through-line: **a run should be able to say what it measured, and a set should fail loudly rather than publish a false number.** Every fix here started as a number that looked fine and was not.

Mirrored to ms2data/agent-skills#81; the set changes are credibledata/malloy-samples#23; the design record is ms2data/service#6159.

## Read the original review against this repo first

Every line number in #1101's review matches the `ms2data/agent-skills` copy of the scripts, not this one. The malloy-samples runbook sent the reviewer there, and that mirror was behind by 393 lines in `run_baseline.py` alone. Three findings were already fixed here: named-view capture, `contaminated` as a bool with write-time type validation, and `servedRevision` from the server. The runbook now names Publisher as the source of truth.

Replaying the reviewer's own saved transcripts settles the blocking item:

| case | now |
| --- | --- |
| `ecom_yoy` (their headline failure) | `run: order_items -> sales_summary_yoy`, source `declared` |
| `ecom_cumulative_4m_crossing_2022` | `run: order_items -> sales_by_month_2022`, source `last_ok` |
| the other three | still no query, correctly: those transcripts contain **zero** `execute_query` calls |

So the named-view fix accounts for two, and the other three are answerers that genuinely never queried. That is the real defect: the judge gate read `answer_text` while the schema described `submitted`, so those three were judged *and* labelled `not_submitted`. `not_submitted` now means neither prose nor query, which keeps a confident refusal on an answerable case inside the pass rate.

## A run can now say what it measured

- **Which retriever answered.** `get_context`'s response carries it; every call now records it, the run summarises `retrievalMode`, and `flip_table.py` exits 2 on a mismatched pair. Local retrieval degrades to lexical silently with no embedding key, which is why the reviewer had to infer their own run's mode.
- **Which build answered.** `--scope env/package@version` carries the version to `get_context` and `execute_query`. Both tools default an omitted version to the *pinned* one, meaning whatever the workspace serves at call time, so a run stamped `targetVersion: 0.0.58` was answering from whatever was current. `--target-version` fills it in when the scope omits one; two disagreeing pins are refused rather than silently overridden.
- **Which model bytes answered.** `modelGitSha` comes from `--model-repo` and is null without it, instead of naming the tree around Publisher's served copy.
[...]

PR #1131, opened 2026-09-09
fix(server): keep a reloaded package's warm embedding index

jswir:
Closes the reload half of #1130, and adds a regression test for the scoped-`get_context` cache deletion reported separately (already fixed on `main` by #1028 — see below).

## What was wrong

A reload never dropped a package's vectors — they are keyed by package name in `publisher.db`, and a test already pinned that nothing re-embeds across one. What a reload threw away was the *record* that they were current: a `WeakMap` keyed on the `Package` object, and every reload allocates a new instance (`reload_package`, REST `?reload=true`, and each watch-mode recompile all reach `Package.create`).

That cost two things.

**The first question after any reload was ranked lexically** while the diff re-discovered that every content hash still matched. With watch mode on, that is one degraded answer per save, including saves that changed nothing relevant.

**`embeddingIndex.status` reported `ready` the whole time.** It derived readiness from whether cached rows covered the current entity *names*, and rows outlive a restart and a reload — so the first poll returned `ready` while the next question was still answered lexically. That defeats the polling loop the field was added for: `package.controller.ts` puts `embeddingIndex` on the `?reload=true` response precisely because that is the request which invalidates the index. A harness following the documented "poll until `ready` before measuring retrieval quality" measured a lexical run and recorded it as a semantic one — a wrong number, not a slow start.

## The change

Instance identity was a proxy for "the entity set may have changed". The proxy is never wrong, just coarse. This records what the sync already computes instead: one fingerprint over the desired `(entity, facet, content_hash)` row set, stored on the per-name `syncMeta` entry that already survives a reload.

- A reload whose facet texts hash the same keeps the warm index and is ranked semantically on its first call.
- An edit moves the fingerprint and re-syncs, re-embedding only the facets whose text changed.
- `status` is decided by that same recorded sync, so `ready` now means one thing: the next question about this package will be ranked semantically.

`desiredFacets` becomes the single definition of the desired row set, shared by the diff and the fingerprint so the two cannot drift. The in-flight bookkeeping moves into `kickSync`, taking the search gate from ~45 lines to 15. Existing invariants are kept and still commented: the `generation` counter still invalidates a recorded sync across a purge, a failed sync still evicts and arms the per-package cooldown, and the per-name mutex still serializes every read-diff-write section.

**The fingerprint is computed fresh per call, not memoized on the `Package`.** A first cut did memoize it, and that read the cached value while ignoring its `entities` argument — correct only while every caller passes the same set for a given instance, which nothing enforces. Reproducing the scoped-sync bug below showed what that shortcut costs: the caller was handed the full set's fingerprint, read as synced, and a wrong argument became a wrong answer instead of a re-sync. Measured cost of dropping it: **3ms for a 1,269-entity package**, on a path whose next step is a network embedding call with a 5s timeout. If it ever needs caching it belongs beside the entity index, where the set is genuinely immutable.

## The scoped-`get_context` report: already fixed, now pinned

A separate report described one `sourceName`-scoped call deleting 2448 of 2603 cached vectors and taking the package lexical for the life of the `Package` instance. `syncPackageEmbeddings` reads every row for a package and deletes the ones absent from the desired set it is handed, so handing it a scope-filtered entity list means "make the cache match this subset" — and it obliges.

**That is already fixed on `main`, by #1028, not by this PR.** `getContext` passes `Array.from(byId.values())` — the whole package — and applies the scope inside the SQL scan via `trySemanticSearch`'s `sourceName`; the filtered list is a separate `inScope` used only for the enumeration tiers. The report's line numbers are against a pre-#1028 build.

Nothing pinned that separation, so this PR adds the test. Re-introducing the scoped sync input makes it fail with **half the package's rows deleted**, naming the harm rather than the symptom.

## Behaviour change worth flagging in review

`embeddingIndex.status` keeps its name and changes its basis, so it is a public-surface change even though the enum is untouched. It now reads `indexing` in two places it previously read `ready`: just after a restart, until the first question re-establishes the sync, and after a doc-only edit. Both clear on the next `get_context`. `embeddedRows` and `embeddedEntities` keep their meanings and are now documented as coverage rather than readiness, since they can equal `totalEntities` while `status` is `indexing`.

They also stop being scoped to the *configured* `EMBEDDING_DIMENSIONS`, which review caught. `dims` records the length a provider actually returned, and the sync diff deliberately ignores it, so a provider that ignores the request parameter (Ollama) writes rows the configured value never matches. Moving readiness to the recorded sync while leaving the counts on the configured value let the two halves of one response contradict each other: `status: ready` beside `embeddedRows: 0`, for a package whose vectors the search scan was reading happily by their observed length. That also contradicted the schema text this PR adds, which calls `embeddedEntities` the entities holding at least one usable vector. Both halves now use the sync's own currency rule.

Before this branch the same filter pinned `status` at `indexing` for those providers, because readiness was derived from that same dims-filtered coverage query. That was on the deferred list below; removing the clause closes it, so it is off the list rather than left open.

`api-doc.yaml`, `docs/configuration.md` and `RELEASE_NOTES.md` are updated for this.

## Verification

Every pin was checked by mutation rather than assumed to be meaningful:

| Mutation | Fails |
|---|---|
| Restore the per-instance gate | exactly the new reload tests |
| Restore name-coverage `status` | exactly the two new status tests |
| Re-introduce the scoped sync input | exactly the new scope test |
| Restore the configured-dims filter on the counts | exactly the new dims test, on the row count, after its `ready` assertion has already passed |

The integration test is the real oracle — real server, real DuckDB, stub HTTP embeddings endpoint. It calls `reload_package` for real, then asserts `retrieval: "semantic"` on the first call and **exactly one** embedding request (the query embedding, no bulk re-sync). It also fails under the first mutation.

`embedding_index.spec.ts` + `get_context_tool.spec.ts` 167 pass, `tests/integration/mcp` 31 pass, `tsc --noEmit`, `bun run lint` and prettier all clean.

Pre-existing failures, reproduced on pristine `main` by reverting this branch's files: `config.spec.ts` (28) and `config.theme.spec.ts` (5). `rate_limit.spec.ts` and `oom_guards.integration.spec.ts` flake under load — a different test in them fails on each run, and they pass in isolation.

## Not in this PR

The other half of #1130, `--init` dropping `entity_embeddings`, is working as intended and is left alone. `--init` resets the server root, and it is also the only stated reclaim path for rows orphaned by a config change — two docstrings in `embedding_index.ts` name it as exactly that. (The secondary suggestion to drop it from `dropAllTables` would make `--init` a partial wipe with no full option and delete that reclaim path.)

Three adjacent findings, all pre-existing, are worth their own issues rather than folding in: `--init` also drops `themes`, which has no config source and so is unrecoverable rather than merely expensive; there is no MCP surface for index readiness at all; and a non-`--init` boot that finds an environment gone from config cleans up nothing. (A fourth, the configured-versus-observed `dims` split, is fixed above rather than deferred.)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -245,24 +245,31 @@ What to know before turning it on:
   cache along with the rest of persisted storage; the only cost of a wipe is re-embedding. A server
   upgrading from a release that embedded one vector per entity discards its cache once, on the
   first boot after the upgrade, and re-embeds each package on its next question.
-- First query per package: the first question kicks off the embedding sync in the background and
-  answers lexically; once the sync lands, later questions are ranked semantically. Responses carry
-  a `retrieval` field (`"semantic"` or `"lexical"`) whenever the provider is configured, and a
-  lexical one adds `retrieval_reason` saying why: `indexing` (still building — clears on its own,
-  worth one retry), `cooldown` (a recent provider failure is being short-circuited),
-  `too-many-entities`, `provider-error`, or `unavailable`. Only `indexing` is worth retrying.
+- First query per package: the first question about content nothing has embedded yet kicks off the
+  embedding sync in the background and answers lexically; once the sync lands, later questions are
+  ranked semantically. A reload does not repeat that: `reload_package`, `?reload=true` and a
+  watch-mode save all keep the warm index as long as the saved files hash the same, so the next
+  question is ranked semantically on its first call. An edit re-syncs, and re-embeds only the parts
+  whose text changed. Responses carry a `retrieval` field (`"semantic"` or `"lexical"`) whenever the
+  provider is configured, and a lexical one adds `retrieval_reason` saying why: `indexing` (still
+  building — clears on its own, worth one retry), `cooldown` (a recent provider failure is being
+  short-circuited), `too-many-entities`, `provider-error`, or `unavailable`. Only `indexing` is
+  worth retrying.
 - Checking readiness without watching the log: `GET /api/v0/environments/{env}/packages/{pkg}`
   carries an `embeddingIndex` object with `status` (`indexing` / `ready` / `cooldown` /
   `too-many-entities`, the same words `retrieval_reason` uses), `embeddedRows`, `totalEntities`,
   `embeddedEntities`, and `lastSyncedAt`. Poll it until `ready` before measuring retrieval quality,
-  so you are not measuring a half-built index. `ready` means every entity the package exposes has a
-  vector under the model you have configured now, so a server pointed at a new `EMBEDDING_MODEL`
-  reports `indexing` until it has re-embedded rather than reporting rows retrieval would reject.
-  Two things worth knowing: the sync runs on a `get_context` question, so a package nothing
-  has queried stays at `indexing` rather than warming on its own; and the first read after a
-  package loads or reloads builds that package's entity index, which is work a plain metadata read
-  would not otherwise do. It is absent when no provider is configured, and reading it takes no
-  locks, so polling cannot slow an indexing run.
+  so you are not measuring a half-built index. `ready` means the next question about the package
+  will be ranked semantically, and nothing weaker: it is decided by the same completed sync the
+  search path gates on, so a server pointed at a new `EMBEDDING_MODEL` reports `indexing` until it
+  has re-embedded, and a restart reports `indexing` until the first question re-establishes the
+  sync, even though the vectors are still on disk. Do not read readiness off `embeddedEntities ==
+  totalEntities`: those count coverage by entity name, so they can be equal while a doc edit is
+  still unembedded. Two things worth knowing: the sync runs on a `get_context` question, so a
+  package nothing has queried stays at `indexing` rather than warming on its own; and the first
+  read after a package loads or reloads builds that package's entity index, which is work a plain
+  metadata read would not otherwise do. It is absent when no provider is configured, and reading it
+  takes no locks, so polling cannot slow an indexing run.
 - Failure behavior: if the endpoint is down, times out, or rejects the key, retrieval falls back
   to lexical (with a warning in the server log) and retries after a cool-down. A package with more
   than 5,000 entities stays lexical.
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -259,13 +259,17 @@ What to know before turning it on:
   carries an `embeddingIndex` object with `status` (`indexing` / `ready` / `cooldown` /
   `too-many-entities`, the same words `retrieval_reason` uses), `embeddedRows`, `totalEntities`,
   `embeddedEntities`, and `lastSyncedAt`. Poll it until `ready` before measuring retrieval quality,
-  so you are not measuring a half-built index. `ready` means the next question about the package
-  will be ranked semantically, and nothing weaker: it is decided by the same completed sync the
+  so you are not measuring a half-built index. `ready` means the index is warm, so the next
+  question about the package is ranked semantically: it is decided by the same completed sync the
   search path gates on, so a server pointed at a new `EMBEDDING_MODEL` reports `indexing` until it
   has re-embedded, and a restart reports `indexing` until the first question re-establishes the
-  sync, even though the vectors are still on disk. Do not read readiness off `embeddedEntities ==
-  totalEntities`: those count coverage by entity name, so they can be equal while a doc edit is
-  still unembedded. Two things worth knowing: the sync runs on a `get_context` question, so a
+  sync, even though the vectors are still on disk. It describes the index, not the next response —
+  a question whose own query embedding fails still falls back, with `retrieval_reason:
+  provider-error`. Do not read readiness off `embeddedEntities == totalEntities`: those count
+  coverage by entity name, so they can be equal while a doc edit is still unembedded. Nor off
+  `embeddedRows`, which counts every cached vector under the current model regardless of its
+  length, so a change to `EMBEDDING_DIMENSIONS` that no question has probed yet still counts the
+  old rows. Two things worth knowing: the sync runs on a `get_context` question, so a
   package nothing has queried stays at `indexing` rather than warming on its own; and the first
   read after a package loads or reloads builds that package's entity index, which is work a plain
   metadata read would not otherwise do. It is absent when no provider is configured, and reading it
--- a/packages/server/src/mcp/tools/embedding_index.spec.ts
+++ b/packages/server/src/mcp/tools/embedding_index.spec.ts
@@ -122,6 +122,15 @@ function entityOfKind(kind: string, name: string): EmbeddableEntity {
    return { kind, name, source: "src", modelPath: "m.malloy", embedDoc: "" };
 }
 
+/**
+ * A stand-in Package instance. Identity is all the index uses it for, so a
+ * fresh one models exactly what a reload does: replace the instance while the
+ * cached rows, and the package's name, stay put.
+ */
+function instance(): Package {
+   return {} as unknown as Package;
+}
+
 /** Poll through the cold-start "indexing" response until the sync lands. */
 async function searchReady(
    args: Parameters<typeof trySemanticSearch>[0],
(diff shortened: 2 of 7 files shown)


packages/server/src/mcp/tools/embedding_index.ts
```diff
@@ -874,56 +1093,19 @@ export async function trySemanticSearch(args: {
...
-      const tracked: SyncState = {
-         done: false,
-         providerKey,
-         generation: meta.generation,
-      };
-      // The sync promise is deliberately not stored: nothing may ever
-      // await it (cold starts answer lexically); completion is observed
-      // through `done` and failure through the handler below.
-      syncPackageEmbeddings(
+   const fingerprint = fingerprintFor(entities);
```
Sha-Bang, line 1096: `fingerprintFor(entities)` walks `entityFacets` + `prepareEmbeddingInput` + a sha256 per facet for every entity, and it's on the warm path — this line and `:1615` both run it before anything checks whether a sync is needed. on `main` the warm path was a `syncState.get(pkg)` WeakMap lookup, and `getEmbeddingIndexStatus` took `IndexedEntity[]`, which structurally couldn't do facet work.

measured: 1.3ms docless at n=1269 (the body's number), 32.8ms at 1500 doc chars, 130ms at 6144, 528ms at n=5000 × 6144 — and that's not a strawman, `:1155` names exactly that cap and `MAX_DOC_CHARS` is 6144 at `:197`. it's non-yielding CPU on the shared event loop, on the latency path of every retrieval call and on an endpoint we tell operators to poll in a loop, so a poll loop now starves the sync's own embedding calls.

which falsifies three sentences this PR adds: `:1532` "it takes no mutex and writes nothing, so calling it cannot perturb or serialize behind a sync in flight", `api-doc.yaml:3440` "Reading it takes no locks, so polling cannot perturb an indexing run", and `docs/configuration.md:276` "polling cannot slow an indexing run". no locks, but not free.

the comment at `:578` already names the fix — cache it beside the entity index, where the set is genuinely immutable. `getPackageIndex`'s per-`Package` index is that set, and both call sites already derive `entities` from the same `pkgIndex`. computing it once when the index is built keeps the anti-memoization argument intact (keyed on the set, not the instance) and drops the per-call cost. what if we did that here rather than as a follow-up? if it has to wait, i think those three sentences need to change in this PR.

docs/configuration.md
```diff
@@ -245,24 +245,35 @@ What to know before turning it on:
...
+  question is ranked semantically on its first call. An edit re-syncs, and re-embeds only the parts
+  whose text changed. Responses carry a `retrieval` field (`"semantic"` or `"lexical"`) whenever the
+  provider is configured, and a lexical one adds `retrieval_reason` saying why: `indexing` (still
+  building — clears on its own, worth one retry), `cooldown` (a recent provider failure is being
+  short-circuited), `too-many-entities`, `provider-error`, or `unavailable`. Only `indexing` is
+  worth retrying.
 - Checking readiness without watching the log: `GET /api/v0/environments/{env}/packages/{pkg}`
   carries an `embeddingIndex` object with `status` (`indexing` / `ready` / `cooldown` /
   `too-many-entities`, the same words `retrieval_reason` uses), `embeddedRows`, `totalEntities`,
   `embeddedEntities`, and `lastSyncedAt`. Poll it until `ready` before measuring retrieval quality,
```
Sha-Bang, line 261: `kickSync` has exactly one caller — `trySemanticSearch` at `embedding_index.ts:1098` — and `meta.synced` is process memory. so after a restart every package reads `indexing` and nothing but a `get_context` question can move it. this line says "Poll it until `ready` before measuring retrieval quality" and `api-doc.yaml:3433` says "poll a package you are about to query", which both read as poll-then-query. that's the order that spins.

the working order is query once, then poll — and `serve.py` in #1114 already does exactly that: `warm_retrieval` issues the ranking call at `:174` before the loop, with `:121` explaining why. worth saying it here in those words, and saying it applies after every restart rather than only to a package never queried?

the body names the break as pollers seeing `indexing` where they saw `ready`. the sharper break is that a poll-first harness never terminates.

packages/server/src/mcp/tools/embedding_index.spec.ts
```diff
@@ -384,9 +393,86 @@ describe("trySemanticSearch", () => {
...
+      // here would hide the regression this pins.
+      const reloaded = await trySemanticSearch({ ...base, pkg: instance() });
+      if (!("hits" in reloaded)) {
+         throw new Error(`expected semantic hits, got ${reloaded.unavailable}`);
+      }
+      expect(reloaded.hits[0]?.name).toBe("alpha");
+      expect(counts.get("alpha")).toBe(1);
+   });
+
+   it("re-syncs a reloaded package whose content did change", async () => {
```
Sha-Bang, line 440: strip `+ KEY_SEPARATOR + d.hash` from `desiredFingerprint` and exactly one test fails — `is not ready when a doc changed and nothing was renamed`, a status assertion. every `trySemanticSearch` test still passes. under that mutation an in-place doc edit that keeps the facet count never re-syncs, and the package serves the previous draft's vectors indefinitely — the fail-open the content hash exists to prevent.

this test looks like it covers that, but it goes from `entity("alpha","src")` to `entity("alpha","src","second draft")` — an *added* facet, so it survives a key-only fingerprint. its own comment at `:463` is honest about that. the case that needs a pin is `"first draft"` -> `"second draft"`: same facet count, different text, asserting the second search re-embeds and the new vector is what ranks. worth adding while you're here?

Sha-Bang (commented):
the fingerprint is the right call — the `(key, hash)` multiset is a sound proxy for the row set and i couldn't construct a fail-open against it. the generation and providerKey guards are intact, and `meta.synced` is only recorded after the writes land inside the mutex. i mutation-tested the table in the body: three of four rows are exact. row 1 is overstated — restoring the per-instance gate fails one of the new reload tests (`ranks a reloaded package semantically on its first call`) plus the integration test, not both unit tests.

one blocking, and it's the one this PR argues against itself on. `fingerprintFor(entities)` now runs on the warm path at `embedding_index.ts:1096` and again at `:1615`, walking `entityFacets` + `prepareEmbeddingInput` + a sha256 per facet for every entity, unconditionally. the body's "3ms for a 1,269-entity package" is the docless case. details inline.

two non-blocking inline: the "poll until ready" order in both docs now hangs after a restart, and a doc-text edit that keeps the facet count is pinned only by a status assertion.

what i ran: `tsc --noEmit` and `bun run lint` clean, `embedding_index.spec.ts` + `get_context_tool.spec.ts` 167 pass, `tests/integration/mcp` 31 pass, plus the mutations above on a scratch copy.

packages/server/src/mcp/tools/embedding_index.ts
jswir: Done here rather than as a follow-up, in 033d23ea.

The package index now builds one frozen array per `Package` (`retrievalEntities`), and both call sites pass it. `fingerprintFor` caches on that array in a WeakMap, and only when the array is frozen. A subset is a different array, so it misses and is computed fresh. That keeps the argument against keying on the instance. It also dedupes its own input now, so it matches what the sync records after main's `uniqueByEntityKey`.

Measured with your worst case, 5,000 entities × 6,144 doc chars, after a sync has created the meta: a status read went from 264–347ms to 12–14ms. The 12ms left is the two row-count queries.

I rewrote all three sentences. They no longer say it's free: a read takes no lock, the first read after a load builds the fingerprint, and later reads are two row counts.

docs/configuration.md
jswir: Agreed, and poll-first never terminating is the sharper way to put it. Fixed in 4ca3c9d0. `docs/configuration.md`, `api-doc.yaml` and the release note now all say: send one `get_context` question, then poll. They say it applies after every restart, not only to a package nothing has queried, and that a script which polls before asking waits forever.

packages/server/src/mcp/tools/embedding_index.spec.ts
jswir: Added in 5516232c: `re-syncs a doc edit that keeps the facet count, and ranks the new text`. It goes "first draft" → "second draft", then asserts that the second draft is re-embedded and that its vector is the one that ranks. The first draft's vector is orthogonal to the query, so a stale index returns no hit.

It uses frozen arrays, as get_context now does, so it also covers the fingerprint cache. Mutation-checked: it fails when `+ KEY_SEPARATOR + d.hash` is removed, and when the cache returns one value for any array.

jswir:
Tested locally end to end, after addressing the review. Setup: real Publishers built from this branch (5516232c) and from current `main` (99e5238b), with real OpenAI embeddings (`text-embedding-3-small`), serving the `storefront` example. Each server used its own server root. I drove them over MCP (`get_context`, `reload_package`) and polled `embeddingIndex` on the package resource.

**Reload with no file changes, same steps on both builds**

| | `status` after reload | first question after reload | second question |
|---|---|---|---|
| `main` | `ready` | lexical, `retrieval_reason: indexing` | lexical, `indexing` |
| this branch | `ready` | semantic, same top 3 and scores as before the reload | semantic |

So both bugs this PR describes still reproduce on `main`: the reload throws away the warm index, and `status` says `ready` while questions are answered lexically.

**Other cases, on this branch**

- **Doc edit, then reload.** I changed `total_sales`'s `#(doc)` from "Total revenue" to a wording about gross merchandise value. That keeps one doc facet, the same shape as the new unit test.
  - `status` went to `indexing`, while the counts still read 191/191. That's the case the docs now warn about.
  - The next question answered lexically and started the sync. `ready` about 1s later.
  - `total_sales`'s score for "gross merchandise value" went from 0.51 to 0.77, so the new text is what ranks.
- **Restart, then poll without asking.** Still `indexing` after 15s. It stays there, as predicted in review.
- **Restart, one question, then poll.** `ready` within about 1s, and the next question is semantic. This is the order the docs now give.
- **Status read cost.** 2–3ms per read. The first read after a load took 640ms, which is the one-time build of the entity index and fingerprint.

**Still open, deliberately not in this PR:** after a restart, each package gives one lexical answer before it's warm, because the completed-sync record is in memory and only a question starts a sync. Starting the sync when a package loads would let poll-only reach `ready`. That changes when embedding calls happen, so I'd rather do it as its own PR.

girishjeswani (approved):
Approving. All three findings from the last round are genuinely fixed, and I verified each by running rather than reading.

**The fingerprint cache does what it claims, and I measured it independently.** At the worst case from the review (n=5000 × 6144 doc chars): **556-588ms unfrozen, 6.5-8.1ms cached**. That reproduces the 528ms figure for the pre-fix burn and comes in under the 12-14ms claimed. One note for anyone re-measuring: a naive benchmark reports ~1.3ms and means nothing, because `getEmbeddingIndexStatus` short-circuits on `meta &&` before reaching `fingerprintFor` unless a sync has actually recorded `synced` — the measurement only becomes real after driving a full sync first.

The cache is correct on the points that mattered. `Object.isFrozen` is checked on both the get and the set, so an unfrozen array is never read from or written to it. A different array with identical contents genuinely misses — confirmed empirically rather than assumed, `Array.from(frozenArray)` recomputed at full cost on every call. That is the anti-memoization property intact: a scope-filtered subset is a different array and cannot inherit a superset's fingerprint. And passing `args.entities` rather than the deduped local at the search site is the detail that makes the whole thing work; the comment explaining it is worth keeping, since the copy would have been a 100% miss rate.

**The docs order is fixed on all three surfaces**, and the three falsified sentences are gone — a repo-wide grep for `perturb`, `cannot slow` and `does not slow` returns nothing. I also checked a surface the review did not cite: `get_context_eval.ts`'s `warmUp` already polls by re-querying `get_context` rather than polling status, so it terminates and needs no change.

**The new test is genuine and the vectors make it load-bearing.** First draft `[0,1,0]`, second `[0,0,1]`, query `[0,0,1]` — so a stale index returns no hit at all and the `toBeCloseTo(1.0)` assertion cannot pass on the old vector. Stripping `+ KEY_SEPARATOR + d.hash` now fails a `trySemanticSearch` test, where before it failed only a status assertion; it aborts at `counts.get("alpha: second draft")`, i.e. no re-embed happened, which is exactly the stale-vector case. The cache-returns-one-value mutation is caught by the same test. The older test is still there and the comment correctly explains the two are complementary rather than duplicated.

No regressions since the reviewed state: the generation and `providerKey` guards are unchanged, `meta.synced` is still recorded only inside the mutex after the writes settle, and the one semantics change — `fingerprintFor` deduping its own input — closes a latent mismatch with the sync's already-deduped list rather than opening one.

Two non-blocking things.

**The `Object.isFrozen` guard has no test behind it.** I removed it and all 183 unit tests stayed green. It is correctness rather than an optimization — the only thing stopping a stale fingerprint from an array mutated after caching — and the invariant currently holds only because the sole producer always freezes. A small test (unfrozen array, mutate an element's `embedDoc`, assert the fingerprint moved) would pin it.

**The docblock's freeze argument is a little incomplete.** It says an unfrozen array could be edited after its fingerprint was cached, which is true, but `Object.freeze` on an array is shallow, so a *frozen* array's element objects are still mutable. Not reachable today — `collapseAliases` mutates `embedDoc` in `collectEntities`, before `getPackageIndex` freezes — but it is the refactor that would silently reintroduce the stale-index bug this PR exists to fix. One sentence, or freezing the entities themselves.

PR #1132, opened 2026-09-09
chore(release): stamp 0.2.5 on main

Sha-Bang:
Stamps the 0.2.5 release notes heading and resets main's declared package versions to 0.2.5. Recovered per the publisher-release skill's step 3 recovery path — this branch was pushed by the gh-release job for v0.2.5 but never opened as a PR.


Sha-Bang:
Superseded by #1133, which stamps 0.2.5 and 0.2.6 together. This branch's heading rename is carried there, so nothing here is lost.

PR #1134, opened 2026-09-09
fix(get_context): report model paths a source actually resolves in

jswir:
`get_context` hands back a `model_path` the source does not resolve in, so an agent queries that file and gets `Reference to undefined object '<source>'`. On a fixture with selective imports, 5 of the 8 cards it returned named a file the source does not compile in.

Reproduced against a built server, before and after:

```
before: 8 cards, 5 of them `Reference to undefined object`
after:  7 cards, 0 failures, each under a file that resolves it
```

This was found by an eval run whose two arms scored 76% and 44% on identical package bytes. The gap was this bug, not model quality: cases that queried the defining file matched 73%, cases that queried some other file matched 20%.

## Three defects behind one symptom

**1. A model claimed sources it cannot resolve.** `collectImportedSourceInfos` re-loaded each imported *file* and pushed every source that file exports into the importing model's `sourceInfos`, never consulting the selective import list. A file importing one name from an 8-source file inherited all 8, and 7 of them resolve nowhere in it.

That walk existed for a reason: `modelDefToModelInfo` emits only `modelDef.exports`, and the import statement stores imported names `exported: false`, so the stable `ModelInfo` of an importing file shows none of its imports. The fix reads `modelDef.contents` instead — the file's actual namespace, its own declarations plus exactly the names an import selected, keyed by local (post-`is`) name. Same set as malloy's own `Model.explores`.

It also drops one full re-compile per import statement at package load, and a silent `catch {}` around it. A renamed import (`import { b is a }`) now reports `b`, the name that resolves, rather than `a`, which does not.

**2. Only one of a source's valid paths was reachable.** Entities were deduped on `kind|source|name` with no path. Sorting the model list (already on main) made the survivor stable but not correct: on the fixture the importing file sorts first, so the one surviving path was the one that does not compile.

**3. ~~`#(agent-hidden)` was ignored entirely.~~** Removed from this PR after review. It was new behaviour riding on a bug fix, and only half a surface: `get_context` honored the tag while the REST `/models/{path}` `sourceInfos` and the SDK ModelExplorer still listed a hidden source, so a model already carrying the annotation would go visible in the app and invisible to the agent on upgrade, with no release note. It needs the REST surface and a note, which is its own change.

## The decision to push back on

**This reverses a deliberate rule and rewrites the test that pinned it** (`indexes a re-exported source once, from the same model every time`, in a describe block whose docblock calls duplicates "a defect observed in a comparable retrieval service").

The case for reversing: that rule made the *other* valid paths unreachable, and it chose its survivor before a model's source list was trustworthy — so "first sorted model" could name a file the source does not compile in, which is exactly what the fixture shows. With defect 1 fixed, every path it would have discarded is a path that compiles, and returning them all matches the hosted API, which keys its result set the same way.

If you disagree, the alternative is to keep one card per source and take only defect 1 — that alone fixes the user-visible error. Worth saying out loud since it's the one judgement call here.

### The cards are not redundant copies, which strengthens the case

Worth adding, because it was not the argument originally made for the reversal
and it is a better one. The same source under two paths can carry a **different
contract**, not just a different `model_path`. On the bundled `storefront`:

| card | `givens` |
| --- | --- |
| `products` @ `storefront.malloy` | *none* |
| `products` @ `data_app.malloy` | `CATEGORY`, `BRAND`, `REGION`, `SINCE`, `MIN_SALE` |
| `products` @ `dashboards/overview.malloy` | the same five |

`givens.malloy` is imported by the two dashboard/app files and not by the model
that declares `products`, so only those paths carry the givens — visible in
`execute_query`'s metadata, where `data_app.malloy` reports
`##! experimental.givens` and `storefront.malloy` does not.

So one card per source was not merely lossy. Whichever path won advertised its
givens as if they applied to the source everywhere: an agent reading the old
card for `products` is told it can filter on `CATEGORY`, and finds it cannot the
moment it queries the defining file. Keying the card on (model path, source) is
what makes `givens`, `authorize` and `model_path` describe the same thing.

Verified through the MCP tools against a build of this branch: 15 cards
(`returned` = `total_available` = 15), a `products` drill-down 3-of-3, and
`run: products -> { aggregate: product_count }` returning 200 under each of the
three paths.

## Second commit: a view's definition

`get_context` never returned view code, and not for the reason the hosted service withholds it. There, code is gated on `scopes[].entity_name`. Here there was nothing to withhold: `code` is read off `field.code`, which malloy fills for a scalar expression only.

```
shouty     code=upper(state)    dimension
total      code=amt.sum()       measure
by_state   code=<undefined>     view
```

A view now takes its definition from the model file, sliced at the range in the turtle's own `location` — what the hosted service stores. That range covers the definition and nothing around it: not the enclosing source, not the `#(doc)` line above. Restricted to turtles, because a physical column's `location` points at the source's `table()`/`sql()` expression.

`scopes[].entity_name` now turns code on by itself, overriding an explicit `include_code: false`, so the same request means the same thing on both surfaces. It also *narrows* on both retrieval modes now: the semantic block filtered on `sourceName` alone, so wherever an embedding provider is configured a pinned `entity_name` narrowed nothing and the caller got the whole ranked set back with every expression attached.

## `queryableSources` is unaffected

The query gate is *defined as* the discovery surface (`isCuratedSource` → `ownCuratedSourceNames` → `getSources`), so discovery stays a subset of queryable in every mode. Under `explores`, curation to `modelDef.exports` still runs on top of the namespace.

## Two traps worth reviewing closely (obsolete: agent-hidden was cut)

Neither spelling of `agent-hidden` may fold, and each folds a different way by default:

- **File-level** reads `ownModelNotes`, not `modelAnnotations`, which folds the whole import closure. Folding would let one shared include hide every file that imports it. `##(authorize)` folds *deliberately* — a gate an import could shed would be no gate — so this must not copy that path.
- **Source-level** cannot be read off `struct.annotations`: malloy copies a base's whole annotations object *by reference* onto a derivation that adds none of its own, so `source: child is hidden_base` carries the base's note as if it were its own. Reading it naively hides every extension of a hidden base, inverting the tag. `considerNoteOwner` already answers "who wrote this note" for `#(authorize)`, so this reuses it rather than adding a second, weaker heuristic. Renamed from `considerAuthorizeNoteOwner`; nothing in it was authorize-specific but the name.

## Verification

- `source_namespace.spec.ts` and `view_code.spec.ts` compile real Malloy rather than using fixtures, because the namespace and the slice are the load-bearing parts.
- `get_context_attribution.spec.ts` covers the index wiring: both cards for a shared source, and identical output **in order** under a shuffled model listing. The order assertions previously sorted both sides, which asserted set equality only and stayed green with the sort deleted; they now compare unsorted and fail without it.
- End to end against a built server for both commits. `scopes[].entity_name` had no automated test anywhere in the server on either retrieval mode, which made the earlier "all four combinations" claim a manual one; both modes are now pinned, the lexical half in `get_context_attribution.spec.ts` and the semantic half end to end against the real server and vector cache.
- Full unit suite: 3735 pass, 35 fail. All 35 are `config.spec.ts`, and `main` at the same commit gives an identical 35 on the same machine (bun 1.4.0), so they are pre-existing and not from this PR. They are also bun-version-dependent: they do not reproduce on bun 1.3.14, which is why a reviewer saw none. Integration suite: 412 pass, 1 fail (`concurrent_package`, which times out identically on `main`). Lint and prettier clean.

## Review follow-up (commit 3)

Three places still keyed on a bare source name after the card model changed, so
`limit` and the paging envelope no longer described the response:

- `windowBySource` bucketed by source name while `toSourceResults` builds one
  card per (model path, source), so one bucket fanned out into several cards
  after the window had spent a single slot on it. A `limit` of N could return
  more than N cards, and `returned` (cards) was reported against a
  `total_available` counted in names. The per-target entity cap was shared
  across a source's paths too. Now bucketed on (model path, source).
- `collapseAliases` kept the last-walked file as every card's representative,
  and dropped a folded name from every card rather than the one the fold was
  computed against.
- The stale comment above the model sort still stated the "only the first is
  kept" rule this branch reverses.

`get_context_attribution.spec.ts` gains a case pinning `limit` and the envelope
against a multi-path source; it fails on the previous key.

**Measured on the bundled `examples/storefront`**, which imports across three
files: **7 source cards before this branch, 15 after** (`products`, `regions`
and `order_items` each resolve under `storefront.malloy`, `data_app.malloy` and
`dashboards/overview.malloy`). `get_context_eval.ts` is unchanged at **8/8
recall@5 with identical hit ranks** either way -- but it is saturated at 100%,
so it bounds the regression rather than showing there is none.

The tool description now tells an agent a source repeats once per resolving
`model_path`, which nothing said before. Room came from cutting reference
material, per the budget test's own rule: `filter_params` in Parameters (it
keeps its zod description and its place in the Response list) and two field
enumerations. It is 2129 chars, was 2141 -- correcting the claim below, which
was wrong: nothing had to be cut for the earlier commits, they were net +1.

*(Both `agent-hidden` notes below are obsolete: the tag was removed from this
PR after review, along with the `skills/README.md` change that went with it.
They are kept here as the record of what was decided before it was cut.)*

## Review follow-up (commit 4): the two cross-platform failures

Both were already red on 869f5764; neither came from commit 3.

- `view_code.spec.ts` read the model file with `url.pathname`, which on Windows
  keeps a leading slash (`/D:/Temp/.../m.malloy`) and does not open, failing the
  whole describe block in `beforeAll`. Now `fileURLToPath`.
- The drill-down E2E asserted one card for `products`. That stopped being true
  on this branch: `products` resolves in `storefront.malloy` and both files
  importing it, so a drill-down naming a source and not a file gets one card per
  resolving path. Rewritten to pin the contract rather than the count, so a new
  importing file does not break it. **This test is the consumer the change
  needed auditing against, and it is what caught the next bug.**

Fixing it exposed the same counter incoherence on a third path: the
listing/drill-down envelope does not go through `windowBySource` and hard-coded
a drill-down as 1-of-1, answering `total_available: 1` beside `returned: 3`. It
now counts distinct (model path, source) pairings. Verified against a built
server: drill-down 3-of-3, unknown source 0-of-0 and still silent, package
listing 15-of-15.

## Review follow-up (commit 5): the semantic path was collapsing the fix away

Auditing the bug class rather than patching the next instance turned up a
fourth site, and it is the one that mattered: **the two retrieval modes
disagreed about the response.**

The vector cache deliberately holds one row per `(kind, source, name)` -- the
embedded text is identical whichever file resolves the source -- and the scan
fans that hit back out to every live entity sharing the key. The merge one loop
later re-collapsed them, keyed on the same bare name, so only the last one
written survived.

On the bundled storefront, one question answered with **5 cards on semantic and
11 on lexical**, and `products` reported only `storefront.malloy` where lexical
reported all three resolving files. "Last written wins" is also the mirror image
of the "first sorted model wins" rule this PR removes -- it picks one valid path
arbitrarily and discards the rest. So **defect 2 was surviving intact wherever
an embedding provider is configured**, which is the deployed configuration.
Nothing surfaced it; both modes answered confidently and only comparing them
showed the gap.

lunr never had the bug, because its ref is the per-path entity id. That is why
the lexical path was already correct and this went unnoticed.

After: semantic returns the same 11 cards as lexical, `returned` and
`total_available` both 11. Eval unchanged at 8/8 recall@5.

**The audit is closed.** `entityRowKey` has two consumers. Every keying site in
them is now either per card, or per name deliberately -- the vector cache
itself, the fan-out lookup, and the scope filters, which name a source rather
than a card.

One test companion (commit 5): `mcp_get_context_semantic.integration.spec.ts`
asserted one card for a source-scoped semantic query, which is exactly what the
fix above changes. Rewritten to pin the contract rather than the count, like the
drill-down E2E. That is the third test written against one-card-per-source; the
source audit above should have been paired with a sweep of the test tree, and
was not.

Full server suite run locally on this head: integration 406 pass / 1 fail (a GCS
download the dev machine has no credentials for), unit 3618 pass / 33 fail (all
`config.spec.ts`, pre-existing on main under bun 1.4). CI green on all three
platforms.

---

## Review follow-up (commit 6): the blocking findings

**Embedding spend.** The dedupe key gained the model path, so the entity list
was no longer unique by the vector cache's own key and nothing downstream
collapsed it. Every duplicate missed the row cache on a cold sync, was sent to
the provider, and then upserted onto the same row: identical text, paid for
once per resolving path. It also inflated `MAX_EMBEDDED_ENTITIES` and
`totalEntities`, both counted in cards, so a package could flip from semantic
to lexical because someone added an importing file. Collapsed at the cache
boundary rather than at either call site, so the guarantee holds for any
caller; the response fan-out is untouched, since it maps a hit back onto the
live per-path entities itself. Pinned by a test that reports 2 embed calls
without the fix and 1 with.

**`entity_name` did not narrow where it matters.** `matchesScope` ran on the
listing and lexical paths only. Applied to the semantic fan-out's output now,
which is where the lexical path applies it. Pinning `total_sales` returned
`percent_of_sales` beside it before the fix; the integration test fails that
way with the filter removed.

**Ordering.** Both order assertions sorted their own output, so they asserted
set equality while the claim was identical output under a shuffled listing;
deleting the sort left them green. They compare unsorted now. The sorts
themselves used `localeCompare` with no locale, which follows the runtime's
collation and moves with `LANG`/`LC_ALL`. Codepoint compare at all three sites:
the model walk, the package file listing, and `collapseAliases` — the third was
not in the review, but it is the same bug class and it decides *which* alias
survives, so two servers could answer with different spellings of one field.

**`docs/ai-agents.md`.** Updated: it promised "that one source's card", which is
what this PR stops doing, and described `include_code` without mentioning that a
pinned `entity_name` turns it on.

**`skills/malloy-analysis`.** Carries the edit #1137 deferred to this PR. It told
the agent `get_context` does not return a view's Malloy source, which this PR
makes false, and a skill that says a capability does not exist fails silently
because the agent stops looking.

## Review follow-up (commit 7): the three from the approving review, plus the nit

**A view's `code` came off disk, its coordinates off the compile.** The two
disagree whenever disk is ahead of the compiled model, and Publisher serves
that state on purpose: a package whose most recent reload failed to compile
keeps answering from the model compiled before that save, and the index is
cached per `Package`, so a failed reload leaves it intact. A first
`get_context` in that window cut post-edit bytes at pre-edit coordinates and
returned text that is not the view -- mid-token, or another field's definition
-- beside an `execute_query` still running the old one.

Withholding `code` while stale was the reviewer's fallback and I did not take
it: an edit nobody reloads produces the same mismatch and is never marked
stale, so that gate lowers the probability rather than closing the hole. The
loader now keeps the bytes the compile read -- captured from the worker's URL
reader, so it is that snapshot and not a second read -- and ships them on
`SerializedModel`; `get_context` slices those, with no fallback to reading the
file. A model with no snapshot returns no view code, which is a missing field
rather than a confident wrong one. Incidentally the capture is not `file:`-only,
so a package served from a remote store gains view code it never had.

**The semantic scope filter narrowed the rows without rescoping the counters.**
`belowCutoffCount` and `totalEntities` came from a scan that scopes on
`entity_source` alone, so a pinned `entity_name` left both describing the
unpinned set. The whole scope goes into the scan now: `sourceName` stays a
column predicate, and `model_path` and `entity_name` arrive as the rows the
scan may consider, since a cache keyed on `(kind, source, name)` can express
neither. A scope that admits nothing answers 0 of 0 without embedding the query.

One correction to the report: a pinned `entity_name` that matches nothing does
not return empty `sources`, because the scope exempts source rows -- it returns
the source card with no entities under it. The incoherence was one step over:
`below_cutoff_count: 0` read against an unpinned `total_entities` of 4.

**Nothing pinned a view's `code` reaching a response.** `view_code.spec.ts`
gains that consumer over a real compiled model: the definition and its
boundaries on the card, a file edited after the compile, and a model with no
snapshot. The first two fail with the disk read reinstated.

**The `|` nit is closed.** `sourceContextKey` and the three card keys go
through `KEY_SEPARATOR`, the way `entityRowKey` already did.

Not reproduced on a running server: staging a cold index inside a stale window
is awkward, so the test does it deterministically instead. Verified live on
`examples/storefront` that view code is correct under all three resolving
paths, and still correct with a cold index after the model file was shifted
down two lines post-compile.

Unit suite on this head: 3740 pass, 35 fail, all `config.spec.ts` /
`config.theme.spec.ts` and pre-existing under bun 1.4. `tests/integration/mcp`
31 pass. tsc, lint and prettier clean.


packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -1348,12 +1472,13 @@ async function collectEntities(pkg: Package): Promise<CollectedModel> {
...
-   // model that extends another), which surfaces the same entity twice. Keep the
-   // first occurrence per (kind, source, name).
+   // One model surfacing the same entity twice (a re-export, say) is a
+   // duplicate. Two DIFFERENT models surfacing it is not: a source is queryable
+   // at every path that resolves it, and each of those is its own card — so the
+   // model path is part of the key.
    const seen = new Set<string>();
    const deduped = entities.filter((e) => {
-      const key = entityRowKey(e.kind, e.source ?? "", e.name);
+      const key = `${e.modelPath}|${entityRowKey(e.kind, e.source ?? "", e.name)}`;
```
Sha-Bang, line 1481: this key was `entityRowKey(e.kind, e.source ?? "", e.name)` on `main` — exactly the key the vector cache uses — so the list handed to the embedding sync was unique by construction. adding `${e.modelPath}|` makes it one entry per resolving path, and nothing downstream collapses them: `:2187` passes `Array.from(byId.values())` straight to `trySemanticSearch`, and `embedding_index.ts:673` flatMaps it with no dedupe. on a cold sync every duplicate passes `if (!row) return true` at `:707`, so `embedBatch` at `:724` embeds identical text once per path and the N upserts then land on the same primary key.

which is why the comment at `:2210` — "the text is identical for every model path that reaches the entity, so the vector is stored once" — is true of storage and false of cost. on the bundled storefront the body reports 7 -> 15 cards, so roughly half the first-sync embedding spend buys nothing. a shared include imported by ten dashboards is 10x.

the same multiplier hits `MAX_EMBEDDED_ENTITIES` at `embedding_index.ts:848`: the 5,000 cap is now counted in cards, so a package can flip from semantic to lexical because someone added an importing file rather than because the model grew. `totalEntities` over-reports for the same reason.

could we dedupe on `facetRowKey` when building `desired`, or dedupe `entities` on `entityRowKey` before it reaches `trySemanticSearch`/`getEmbeddingIndexStatus`? and pin it with a test that a two-path package embeds each key once.

worth knowing #1131 lands in this same function — it factors this into `desiredFacets` and calls it on every `get_context` and every REST status read, so the two changes compound.

Sha-Bang (commented):
the (model path, source) keying is right and it's carried consistently — dedupe, `windowBySource`, `collapseAliases`, `buildSourceContext`, both envelopes and the semantic merge, each with a test that fails on the old key. `collectSourceInfos` reading `modelDef.contents` is pinned against the real compiler for selective, renamed, bare and no-import cases. no argument with the fix itself.

two blocking, both about blast radius the body doesn't mention. the dedupe key change at `:1481` multiplies embedding spend — the entity list is no longer unique by the vector cache's own key, and nothing downstream dedupes it. and `entity_name` now forces `include_code` at `:900` while the semantic path, the one that runs wherever a provider is configured, applies neither `entityName` nor `modelPath` — so pinning an entity doesn't narrow and now returns code for everything. both inline.

one i'd cut rather than fix: `#(agent-hidden)` arrives in a bug-fix PR with one line of docs and half a surface. inline.

also `docs/ai-agents.md:41` is now false — it promises "that one source's card", which is exactly what this PR stops doing. pre-existing text from #1028 that this PR invalidates, and `git diff --stat -- docs/` is empty.

worth flagging across PRs: #1137 adds a skill rule saying `get_context` does not return a view's Malloy source. `:900` plus the turtle slice at `:1153` makes that false. whichever lands second leaves the other wrong.

what i ran: `tsc --noEmit` and `bun run lint` clean, `bun test src` 3651 pass / 0 fail, `tests/integration/mcp` 30 pass. the 33 `config.spec.ts` failures the body calls pre-existing did not reproduce on bun 1.3.14 — worth rechecking whether that claim still holds.

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -875,7 +891,13 @@ export function resolveRequest(params: GetContextParams): ResolvedRequest {
...
       searches,
       listingOnly,
-      includeCode: params.include_code ?? false,
+      // Pinning an entity by name turns code on, matching the hosted
+      // retrieval API: a caller who has narrowed to one entity is asking what
+      // it IS, and the definition is the answer. Deliberately overrides an
+      // explicit `include_code: false` rather than deferring to it, because
+      // parity is the point -- the same request must mean the same thing on
+      // both surfaces. The flag still decides every unpinned call.
+      includeCode: (params.include_code ?? false) || Boolean(scope.entity_name),
```
Sha-Bang, line 900: the comment above this line justifies it as "a caller who has narrowed to one entity is asking what it IS" — but on the path that actually runs in a provider-backed deployment, the caller hasn't narrowed to anything. `matchesScope` is the only thing that honours `request.entityName`, and it's called at exactly two sites: `:2018` (listing) and `:2386` (lexical). the semantic block at `:2209` applies neither `entityName` nor `modelPath`; `trySemanticSearch` filters on `sourceName` only.

so with an embedding provider configured, pinning `entity_name` does not narrow to that entity, and now serializes `code` for every ranked entity in the response via `:453`. the caller asked for one definition and gets the whole ranked set, expressions included.

the `modelPath` half is pre-existing — absent on `main` too — so this PR didn't introduce it. but this PR is what makes it load-bearing: the whole thesis is that `model_path` identifies the card, and the deployed retrieval mode ignores the caller's `model_path` scope while now returning all the paths they were trying to exclude.

there's no test for `scopes[].entity_name` anywhere — grepping `packages/server/src` and `packages/server/tests` only turns up SQL column names — so the body's "all four include_code/entity_name combinations" is a manual claim. could we apply `matchesScope` to the semantic ranked rows after the fan-out, and pin `entity_name` on both retrieval modes?

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -1195,16 +1299,35 @@ async function collectEntities(pkg: Package): Promise<CollectedModel> {
...
+      const agentHidden =
+         model.getAgentHiddenSourceNames?.() ?? new Set<string>();
 
       for (const sourceInfo of sourceInfos) {
          const sourceName = sourceInfo.name;
-         const provenance = readFieldProvenance(modelDef, sourceName);
-         // First model wins, matching the entity dedupe below, so a source's
-         // identity and its governance always come from the same model.
-         if (!governance.has(sourceName)) {
+         if (agentHidden.has(sourceName)) continue;
```
Sha-Bang, line 1317: `#(agent-hidden)` is new behaviour arriving in a bug-fix PR, and it's enforced only here. the REST `/models/{path}` `sourceInfos` and the SDK ModelExplorer still list a hidden source, so it's visible in the app and invisible to the agent — the shape that becomes a support ticket rather than a feature.

it also has exactly one mention outside code, `skills/README.md:33`, and no release note. a model that already carries the annotation — copied from the hosted engine, which is the stated parity motivation — silently loses discovery on upgrade.

i'd pull it into its own PR rather than fix it here. if it stays, it needs the REST surface and a note.

packages/server/src/mcp/tools/get_context_attribution.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+      expect(sorted(await cardsFor(twoFilePackage()))).toEqual([
+         { source: "mine", model_path: "uses.malloy" },
+         { source: "only_here", model_path: "defs.malloy" },
+         // Both, not one picked by file-walk order.
+         { source: "shared", model_path: "defs.malloy" },
+         { source: "shared", model_path: "uses.malloy" },
+      ]);
+   });
+
+   it("reports the same pairings whatever order the models are listed in", async () => {
```
Sha-Bang, line 144: both sides get sorted before comparing here and at `get_context_tool.spec.ts:2942`, so this asserts set equality, not order. delete the `.sort(...)` at `get_context_tool.ts:1272` and both tests stay green — while the body claims "identical output under a shuffled model listing". `expect(await cardsFor(reversed)).toEqual(await cardsFor(forward))` unsorted would pin what's claimed.

while you're in there: `package_load_worker.ts:527` uses `.localeCompare(b)` under a comment whose whole point is that ordering must be stable between two servers on identical bytes. no-locale `localeCompare` follows the runtime's collation, which moves with `LANG`/`LC_ALL`. a codepoint compare would deliver what the comment promises.

packages/server/src/mcp/tools/get_context_tool.ts
jswir: Fixed, and you were right about the mechanism: nothing downstream collapsed it. Deduped at the cache boundary rather than at either call site, so the guarantee holds for any caller instead of for the two we happen to know about.

`uniqueByEntityKey` now runs at the top of `trySemanticSearch` and `getEmbeddingIndexStatus`, which covers all three symptoms you named in one place: the embed spend, the `MAX_EMBEDDED_ENTITIES` cap, and `totalEntities`. The response fan-out is untouched, since it builds its own map from `byId.values()` and reads `modelPath` from the live entity.

Pinned as you asked, and I checked the test fails for the right reason rather than passing for free: with the dedupe removed it reports `Expected: 1, Received: 2` on the embed count for a two-path package.

On "first wins" — two entities sharing a key but carrying different doc text already collided on the row's primary key, so which one got embedded was decided by write order before any of this. Now it is read order, which is at least stable. Called out in the doc comment.

packages/server/src/mcp/tools/get_context_tool.ts
jswir: Fixed. `matchesScope` is now applied to the semantic fan-out's output, which is where the lexical path applies it too.

Your read of the call sites was exactly right, and the consequence was worse than the comment admitted: pinning `total_sales` on the storefront package returned `percent_of_sales` beside it, with `code` on both, because the pin turns `include_code` on. That is the failing assertion when I remove the filter again, so the new test earns its place.

You were also right that "all four `include_code`/`entity_name` combinations" was a manual claim — there was no test for `scopes[].entity_name` anywhere in `src` or `tests`, on either mode. Both are pinned now:

- lexical, in `get_context_attribution.spec.ts` ("narrows a lexical drill-down to the pinned entity")
- semantic, end to end against the real server and vector cache in `mcp_get_context_semantic.integration.spec.ts`

I left the `entity_name` → `include_code` coupling in place rather than dropping it. It only made sense to keep once the pin actually narrows; before this fix it was a way to return more, which is the opposite of what the comment claimed for it.

The `modelPath` half is fixed by the same line, since `matchesScope` reads both.

packages/server/src/mcp/tools/get_context_tool.ts
jswir: Agreed, and cut. It is out of this PR entirely: the two tag regexes and `agentHiddenSourceNames` in `source_extraction.ts`, the `Model` accessor and its memo, both enforcement points here, the `source_namespace.spec.ts` block, the two attribution tests, and the `skills/README.md` change that went with it. `skills/README.md` is back to what is on main.

The upgrade case is what settled it. Retrieval visibility is defensible as a design, but a model already carrying the annotation — copied from the hosted engine, which was the whole parity argument — silently loses discovery, and the REST `sourceInfos` and ModelExplorer keep listing it. Visible in the app, invisible to the agent, no release note. That is a support ticket, as you said.

The work is parked, so when it comes back as its own PR it starts with the REST surface and a note rather than from scratch.

This PR is now only the model-path fix it claims to be.

packages/server/src/mcp/tools/get_context_attribution.spec.ts
jswir: Both fixed, and your test of the test was right: I deleted the sort at `get_context_tool.ts:1272` and both files stayed green.

They compare unsorted now. `expect(await cardsFor(reversed)).toEqual(await cardsFor(forward))` here, and the `.sort()` is gone from `pathsFor` in `get_context_tool.spec.ts:2942`. With the sort deleted, this one fails.

On `localeCompare`: changed to codepoint compare at three sites, not two. The third is `collapseAliases` (`get_context_tool.ts:1623`), which used `a.name.localeCompare(b.name)` to pick which member survives a collapse. Same class as the two you found, and arguably worse, because it decides which spelling of a field the response carries rather than only what order cards arrive in. It is in code this PR already touches, so it seemed wrong to leave a sibling of the reported bug behind.

Comments at all three now say codepoint, and why: with no locale argument `localeCompare` follows the runtime's collation, which moves with `LANG`/`LC_ALL`.

jswir:
All six addressed, rebased onto current main, CI green on all three platforms. Replies inline on each thread; the three items that were only in the review body:

**`docs/ai-agents.md:41`** — you were right that it was false and that `git diff --stat -- docs/` was empty. Updated: it promised "that one source's card", which is what this PR stops doing, so it now says one card per resolving model path and points at `model_path` / `entity_name` for narrowing. Also fixed the `include_code` paragraph, which described it as a field-expression flag and did not mention that a pinned `entity_name` turns it on.

**The `config.spec.ts` claim** — worth doubting, and it does hold, but the body was not saying enough for you to check it. On this machine the file fails 30 tests on this branch **and** on `main` at the same commit, so it is pre-existing. The difference is bun: I am on 1.4.0, you ran 1.3.14, and they do not reproduce there. The body now says that rather than asserting "pre-existing" and leaving you to take it on trust. I also corrected the suite numbers, which were stale: 3735 pass / 35 fail, against an identical 35 on main.

**The cross-PR conflict with #1137** — #1137 merged first, so this PR carries the skill edit, as you suggested. `skills/malloy-analysis` no longer tells the agent `get_context` cannot return a view's source; it now says a pinned `entity_name` returns the view's `code` and to read that before the description, with a fallback clause for a server that returns none. Bundle regenerated.

One thing I did beyond what you asked, flagged so you can push back: the third `localeCompare` in `collapseAliases`. Details on that thread.

packages/server/src/mcp/tools/get_context_tool.ts
```diff
@@ -1062,8 +1156,25 @@ function readFieldProvenance(
...
             ? field.e.path[0]
             : undefined;
+      // Malloy fills `code` for a scalar expression only, so a view -- whose
+      // definition is a query pipeline -- arrives with none, exactly like a
+      // physical column. Its `location` does cover the definition though, so
+      // the text comes off the file instead. Restricted to turtles because a
+      // physical column's location points at the source's table() expression,
+      // which would slice to something that is not the field at all.
+      const viewCode =
+         field.type === "turtle" && field.location?.url && field.location.range
```
Sha-Bang, line 1223: the range here comes from the compiled IR, but the bytes come from current disk — `makeSourceTextReader` calls `readFileSync` at index-build time. AGENTS.md:92-95 documents the state where those two disagree: a package whose most recent reload failed to compile is `stale: true`, still listed and still answering queries from the model compiled *before* that save.

`indexCache` is keyed on the `Package` object, and a failed reload doesn't replace it, so the cached index survives too. if the index is still cold when that happens, the first `get_context` slices post-edit bytes with pre-edit coordinates — so the agent is handed text that isn't the view (mid-token, or another field's definition) while `execute_query` runs the old one. it presents as a confident wrong answer rather than a missing field, which is the part I don't like.

narrow: it needs a cold index inside a stale window. could the text come off the same snapshot the compile read, or failing that, could `code` be withheld while the package is stale? this is the one I'd want fixed before merge.

I confirmed the cache keying and the staleness contract by reading; I did not reproduce it.

Sha-Bang (approved):
the namespace fix is the right one, and `source_namespace.spec.ts` pinning it against a real compile rather than a fixture is what makes it convincing — `modelDef.contents` is the set a query addressed to that path can actually name, so it is the set discovery should report. the multi-path card model is coherent the whole way through: the four keying sites agree with each other, and `limit` and the envelope now count the same unit.

three worth a look before this lands. the first is inline; the other two are below because my tooling lost the thread anchors partway through.

**`get_context_tool.ts:2245` — the new scope filter narrows the hits without rescoping the counters that describe them.** `belowCutoffCount` and `totalEntities` still come from the scan a few lines up, which scoped on `sourceName` alone. pin an `entity_name` that matches nothing and `merged` ends up empty while the counters still describe the unpinned set, so the response is zero sources alongside a low or zero `below_cutoff_count`. that is the combination `SemanticSearchResult`'s own docblock says cannot occur ("never wait on that combination as a signal"), and the tool description tells the agent `no sources = nothing cleared the floor` — so an agent that pins an entity and misses is told the model had nothing above the floor, when the truth is its pin matched nothing, and it has no reason to retry with a different name. reachable wherever an embedding provider is configured. could the model/entity scope go into the scan so the counts and the rows describe the same set, or be recomputed here against the filtered set?

**`view_code.spec.ts` — nothing pins a view's `code` reaching a response.** these pin `sliceRange` and the compiler contract, which is the right instinct; the slice is the load-bearing part and a fixture would only encode our idea of what malloy emits. but the `include_code` block in `get_context_tool.spec.ts` uses a measure and a physical column, so the turtle branch in `readFieldProvenance`, the `sourceTextFor` wiring and the `isView` guard on `aliasOf` all rest on the manual end-to-end run in the body. break any of the three and both files stay green. worth a case that builds an index over a model carrying a view and asserts the card comes back with its definition? written against a file edited after the compile, it would cover the inline staleness finding at the same time.

one nit not worth a thread: `sourceContextKey` joins on `|`, which isn't injective — `("a|b","c")` and `("a","b|c")` produce the same key. it needs both a pipe in a filename and a backtick-quoted source name carrying one, so it is close to unreachable, but `entityRowKey` switched to `` for exactly this and its docblock says why. all seven call sites go through the one helper, so it is a one-line change if you want it closed.

approving — none of it blocks landing, and the three above are a follow-up commit rather than a re-review.

reviewed at 94f2ab07 by reading; I did not stand it up.

packages/server/src/mcp/tools/get_context_tool.ts
jswir: Fixed, and not by withholding `code` while stale -- that would have left the wider case open. The staleness window is the documented instance of a more general disagreement: an edit nobody reloads at all produces the same mismatch and is never marked stale, so a gate on `stale: true` lowers the probability rather than closing the hole.

The loader now keeps the bytes the compile actually read, captured from the worker's URL reader rather than read a second time, and ships them on `SerializedModel`. `get_context` slices those. There is deliberately no fallback to `readFileSync`: a model with no snapshot returns no view code, which is a missing field rather than a confident wrong one -- the failure mode you objected to.

Your read of the cache keying was right, and it is what makes the Model the correct place to hold the text: a reload replaces the Model wholesale, so the IR and the text it indexes cannot drift apart afterwards.

One thing I did not manage: I did not reproduce the corruption on a running server, because staging a cold index inside a stale window is fiddly. The test does it deterministically instead -- the file is edited after the compile so the old range lands mid-token, and it asserts the card's code is not what reading the file now would produce. Reinstating the disk read fails it.

Incidental: the capture is not `file:`-only, so a package served from GCS/S3 now gets view code it never had.

PR #1135, opened 2026-09-09
fix(runtime): coalesce publisher:resize into one message per frame

girishjeswani:
## Summary

An embedded data app visibly jitters while its tiles load. This is the publisher-side half of the cause. The host-side half is [ms2data/service#6355](https://github.com/ms2data/service/pull/6355); either fix helps alone, and they compose.

## Cause

`setUpEmbedResize` observes `document.documentElement` with an undebounced `ResizeObserver`:

```js
var ro = new ResizeObserver(postSize);
ro.observe(document.documentElement);   // fires on ANY layout change
```

`postSize` posted a `publisher:resize` whenever the measured height differed at all. A dashboard whose tiles resolve one by one therefore emits a height per tile, and a host that resizes the iframe on each report steps visibly through them.

The measurement itself is already careful — the `measureContentHeight` comment explains why it sums child bottom edges rather than trusting `scrollHeight`, to avoid a ratchet loop. This is the same class of problem one level up: not *what* is measured, but how often it is announced.

## Changes

**Coalesce to one message per animation frame.** Same `pending`-flag idiom as the SSE reload debounce further down this file. `requestAnimationFrame` rather than a timeout, because the next paint is exactly when a coalesced layout is worth measuring, with a `setTimeout(fn, 16)` fallback where rAF is unavailable.

Trailing edge, not leading: `measureContentHeight` re-reads the live DOM, so the deferred call reports the *settled* height rather than the first of the burst.

**Ignore changes under 8px.** Below that is sub-pixel reflow, which a host cannot act on without the resize itself becoming a layout change inside this frame.

**Initial reports still emit synchronously.** `DOMContentLoaded` and `load` call `emitSize` directly. The host shows a min-height placeholder until the first message arrives, so deferring that one by a frame would leave a collapsed frame for no benefit. Only the bursts that follow need coalescing.

No behaviour change for a page whose height is stable: it posts once and goes quiet.

## Testing

There are no tests around `packages/server/src/runtime/` — it ships as a browser script rather than through the module graph — so I verified the logic against a fake clock outside the tree:

| Scenario | Result |
|---|---|
| Burst of 9 layout changes in one frame | 1 message, carrying the settled height (690, not the leading 300) |
| Sub-epsilon noise (+3px, then +5px) | 0 messages |
| Real change (+200px) | 1 message |
| Exactly +8px | 1 message (boundary is inclusive) |

Worth recording: the first version of that harness advanced a queue on each measure instead of re-reading live state, and reported the *leading* height. The harness was wrong, not the runtime — but it is the kind of check that can quietly confirm the wrong thing.

## Reviewer notes

- **The 8px epsilon is a judgement call.** It needs to sit above sub-pixel reflow and below perceptible movement. If a page legitimately changes height by less than 8px, that change is now not reported until a larger one follows.
- **Happy to add a test** if the project wants coverage here; it would need a jsdom or Playwright harness for this file, which does not exist today, so I did not add one speculatively.
- **Not fixed here:** an embedded app's `position: fixed` drawer opens scrolled down, because the frame is sized to full content height and so has no internal viewport to anchor to. That needs a host-to-iframe scroll message, and the protocol is currently one-way (height up, nothing down). Separate change if wanted.

### DIFF
--- a/packages/server/src/runtime/publisher.js
+++ b/packages/server/src/runtime/publisher.js
@@ -251,27 +251,62 @@
             var pad = parseFloat(bodyStyle.paddingBottom) || 0;
             return Math.ceil(maxBottom + scrollTop + pad);
          }
-         function postSize() {
+         // Smallest change worth reporting. Below this is sub-pixel reflow, which
+         // the host cannot act on without the resize itself becoming a layout
+         // change inside this frame — the feedback loop.
+         var RESIZE_EPSILON = 8;
+         var resizePending = false;
+
+         function emitSize() {
             var h = measureContentHeight();
-            if (h !== lastHeight) {
-               lastHeight = h;
-               try {
-                  window.parent.postMessage(
-                     { type: "publisher:resize", height: h },
-                     "*",
-                  );
-               } catch (_e) {
-                  /* ignore */
-               }
+            if (Math.abs(h - lastHeight) < RESIZE_EPSILON) return;
+            lastHeight = h;
+            try {
+               window.parent.postMessage(
+                  { type: "publisher:resize", height: h },
+                  "*",
+               );
+            } catch (_e) {
+               /* ignore */
             }
          }
-         // Initial + observe content changes
-         if (document.readyState === "loading") {
-            document.addEventListener("DOMContentLoaded", postSize);
-         } else {
-            postSize();
+
+         // Coalesce a burst into one message per frame.
+         //
+         // ResizeObserver on documentElement fires on EVERY layout change, so a
+         // dashboard whose tiles resolve one by one used to post a height per
+         // tile. The host resized on each, which is what made an embedded app
+         // visibly jitter while it loaded. Same `pending`-flag idiom as the SSE
+         // reload debounce below; rAF rather than a timeout because the next
+         // paint is exactly when a coalesced layout is worth measuring.
+         //
+         // Trailing edge, not leading: the last measurement in a burst is the
+         // settled one, and reporting the first would send a stale height.
+         function postSize() {
+            if (resizePending) return;
+            resizePending = true;
+            var schedule =
+               typeof requestAnimationFrame === "function"
+                  ? requestAnimationFrame
+                  : function (fn) {
+                       setTimeout(fn, 16);
+                    };
+            schedule(function () {
+               resizePending = false;
+               emitSize();
+            });
          }
-         window.addEventListener("load", postSize);
+         // Initial + observe content changes
+         // The first report goes out synchronously: the host shows a
+         // MIN_EMBED_HEIGHT placeholder until one arrives, so deferring it by a
+         // frame would leave a collapsed frame for no benefit. Only the bursts
+         // that follow need coalescing.
+         if (document.readyState === "loading") {
+            document.addEventListener("DOMContentLoaded", emitSize);
+         } else {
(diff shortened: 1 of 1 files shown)


packages/server/src/runtime/publisher.js
```diff
@@ -251,27 +251,62 @@
             var pad = parseFloat(bodyStyle.paddingBottom) || 0;
             return Math.ceil(maxBottom + scrollTop + pad);
          }
-         function postSize() {
+         // Smallest change worth reporting. Below this is sub-pixel reflow, which
+         // the host cannot act on without the resize itself becoming a layout
+         // change inside this frame — the feedback loop.
+         var RESIZE_EPSILON = 8;
```
Sha-Bang, line 267: **Should-fix** - symmetric is the part i'd question rather than the 8. a frame 7px too tall is invisible; 7px too short clips, and the frame has no internal scrollbar.

in steady state the host's own epsilon is vacuous — its current height tracks `lastHeight`, so anything you send has already cleared 8. the two bands only compound once the two sides disagree, which the host's MIN clamp guarantees:

```
runtime sends 100   host clamps to 120,  lastHeight = 100
content 127         runtime sends (Δ27); host refuses, |127 - 120| = 7
content 134         runtime silent (Δ7 vs 127)
```

frame 120, content 134, and both sides think they're finished. same shape at the MAX clamp.

would gating on shrinks only work? `if (h < lastHeight && lastHeight - h < RESIZE_EPSILON) return;` — grows can pass at 1px precisely because `measureContentHeight` already `Math.ceil`s, with the host doing the same on its side.

which is also a note on the comment: it says the epsilon absorbs "sub-pixel reflow", but after that `Math.ceil` sub-pixel never reaches here. what 8px actually absorbs is a 1-2px rounding flip-flop — worth saying that instead, since it's the stated justification for the constant.

Sha-Bang (approved):
approving — the coalescing itself is sound. things i checked and found clean:

- rAF vs `ResizeObserver` ordering: RO callbacks are delivered after rAF callbacks within the same rendering update, and `measureContentHeight` forces layout, so the trailing-edge measurement is the current one and a change landing between frames gets picked up by the pending callback.
- `resizePending` against a hidden, offscreen or `display:none` document: a paused rendering update pauses RO delivery and rAF together, and the flag is cleared before `emitSize` runs, so it resumes rather than stalling — no new stall class versus the old synchronous post.
- the `setTimeout(fn, 16)` fallback, and the synchronous `DOMContentLoaded`/`load` emits alongside a pending rAF (they don't double-post — the rAF measures the same height and skips).

one should-fix inline, and it belongs to the pair rather than to this PR: the epsilon is symmetric, so once this side's `lastHeight` and the host's applied height disagree the two independent 8px bands compound into a dead band larger than either, and under-sizing is the direction that actually hurts.

one small behavior note, take it or leave it: with `lastHeight = -1`, a first measurement of 0-6px is now never sent, where the old `!==` check would have sent it. that leaves the host's `sized` flag unset and surfaces its "isn't displaying" notice after 8s. needs a body whose children bottom out in (0, 6], so unlikely — but it's a real change from the old check.

the host half is ms2data/service#6355; i've left a blocking finding there, on a deferred shrink that can go stale and strand the frame below its content. that one is independent of this PR and doesn't affect this approval.

PR #1138, opened 2026-09-10
fix(preaggregation): memoize the violation walk, which /status re-ran for every package on every poll

Sha-Bang:
`/api/v0/status` re-runs a full `#@ preaggregate` annotation walk for every package on every poll. On a 10s probe interval that is a few seconds of synchronous main-thread CPU per call, which blocks the event loop and adds seconds of queuing delay to every unrelated request the server process is serving.

This memoizes the walk.

## How it reaches `/status`

`Package.preaggregateAccessWarnings()` is spread into `getPackageMetadata()`, and `/status` walks every package in every environment through it:

```
GET /api/v0/status
  -> EnvironmentStore.getStatus() -> listEnvironments() -> Environment.serialize()
  -> listPackages() -> Package.getPackageMetadata()
  -> preaggregateAccessWarnings() -> preaggregateFindings()
  -> Model.preaggregateViolations()        <- no memo
  -> validateModelPreaggregation()         <- re-parses annotations on every
                                              field of every source
```

`getPackageMetadata()` was already on the poll path; what changed in 0.2.4 is that the preaggregate walk was added to it. The cost scales with total field count in a server process's loaded packages, so an idle server process is unaffected and one holding real packages is not.

## Measured

A customer cell, before and after 0.2.4 rolled out. Request volume was identical across the boundary — the load is a fixed-cadence prober.

| | 0.2.3 | 0.2.4 |
|---|---|---|
| `GET /api/v0/status`, worst pod | ~0.5ms | up to 5.8s |
| Server-process CPU, fleet | 0.62 cores | 2.6 cores |
| p95 on unrelated package-metadata routes | ~13ms | ~3.5s |
| Node heap, hot pods | flat | sawtooth 2.4 -> 3.14 GB |

Per-pod CPU tracks that pod's `/status` duration almost exactly: a pod spending 5.8s per 10s poll burns ~0.6 cores, while pods with nothing loaded stay at 0.04 cores and answer in 0.5ms.

## Why the memo needs no invalidation

`this.modelDef` is assigned once, in the constructor, and a reload replaces the `Model` object outright rather than mutating it. Nothing in the server writes to `modelDef.contents` either — every reference is a read. This is the same argument `getDeclaredQueryMetadata` already documents a few methods above; that method was given a memo for exactly this reason and the preaggregate path added later was not.

## `readonly`

Callers now share one array instead of each getting a fresh walk, so a caller that sorted or spliced the result would be editing every later caller's copy. The return type says `readonly` to make that a compile error. Type-level only — a runtime freeze would cost something on the path this change exists to make cheap, and a shallow freeze would not protect the entries anyway.

There is one caller today (`Package.preaggregateFindings`), and it only iterates.

## Tests

The validator has its own spec; these pin the memo.

Identity alone would also hold for an implementation that recomputed and refilled a retained array, so `does not re-walk when the underlying contents change` mutates the contents between two calls and asserts the answer does not move. `keeps each model's memo to itself` checks two models do not share an array, so a memo parked on the wrong scope cannot pass. The first test guards the fixture: if the synthetic annotation ever stops parsing, the walk returns `[]` and the identity assertions would pass while measuring nothing.

Removing the memo fails three of the five, each for the right reason — the two identity tests report equal content at a different reference, and the re-walk test picks up the mutated source.

Full server unit suite green (3645 pass, 3 skip, 0 fail); typecheck and lint clean.

### DIFF
--- a/packages/server/src/service/model.ts
+++ b/packages/server/src/service/model.ts
@@ -369,6 +369,11 @@ export function bindingsAllowDegradeToLive(
    );
 }
 
+/** The one empty result {@link Model.preaggregateViolations} hands back, so its
+ *  identity contract holds on the branch that never reaches the memo. */
+const NO_PREAGGREGATE_VIOLATIONS: readonly Readonly<PreaggregateViolation>[] =
+   Object.freeze([]);
+
 export class Model {
    private packageName: string;
    private modelPath: string;
@@ -460,7 +465,7 @@ export class Model {
       | undefined;
    /** Memo for {@link preaggregateViolations}. */
    private preaggregateViolationsMemo:
-      | readonly PreaggregateViolation[]
+      | readonly Readonly<PreaggregateViolation>[]
       | undefined;
    /** Given names (`$NAME`) referenced by any authorize gate reachable
     *  anywhere in this model -- every top-level source's own gate, and every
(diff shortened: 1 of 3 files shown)


housejester (approved):
Approving. Verified the claims the memo rests on against source, and ran the spec both ways.

- `this.modelDef` has exactly one write site, the constructor (`model.ts:663`), and nothing writes to `modelDef.contents`.
- A reload builds a fresh `Model` per file and swaps the whole map (`package.ts:2270`), so the memo on the old objects stays correct for the model they keep serving after a failed reload.
- `preaggregateFindings` is the only caller and only iterates.
- The spread of `preaggregateAccessWarnings()` into `getPackageMetadata()` landed in #1103 (0.2.4), and it is the only new per-poll work in that function between 0.2.3 and 0.2.4, so the attribution holds.
- Spec: 5/5 with the memo, 3/5 fail with it reverted, the same three you name.

Two non-blocking thoughts, either fine as a follow-up:

1. **Guard the class, not just this instance.** Nothing at `getPackageMetadata()` says it is on the `/status` poll path. The other computed spreads next to this one are cheap today, but the next validation walk someone spreads in there repeats this silently. A short comment on the method saying it is polled every few seconds per package and anything added must be memoized or trivially cheap would close most of that. Memoizing the composed warnings array at the Package level would be the stronger version.

2. **Confirm the attribution after it rolls.** The measurement pins the regression on `/status` duration and the diff supports pinning it on this walk, but there is no profile naming `validateModelPreaggregation`. Worth re-checking the worst pod's `/status` latency once this ships, so we know the whole 5.8s came from here.

One wording nit: "worker" in the description reads as a Publisher server process, but in this repo the same word names the compile worker behind `loadViaWorker`. "Server process" would remove the ambiguity for a reader here.

— Stoic Doreen

PR #1139, opened 2026-09-10
feat(docker): let a deployment bound how far the Snowflake driver reads ahead of a slow consumer

housejester:
The Docker image now installs a small shim in front of the ADBC Snowflake driver that can set
`adbc.rpc.result_queue_size` on every Snowflake statement from `ADBC_RESULT_QUEUE_SIZE`. **Opt-in:**
unset — the image default — the shim is a pass-through and the driver behaves exactly as upstream
ships it. The image itself *is* different (the extension loads the shim, which loads the upstream
driver beside it, on every Snowflake query), but with the variable unset the shim calls
`SetOption` zero times and forwards every call; a deployment sets `ADBC_RESULT_QUEUE_SIZE=1` to turn
the bound on. Non-Docker installs are unaffected either way: the `snowflake` extension has no way to
set this option, and the server process does not touch it.

### Why

The driver prefetches result chunks ahead of the consumer with no bound tied to consumption. Each
chunk gets its own buffered channel, and a chunk's goroutine releases its concurrency slot the
moment its download finishes while the decoded records stay queued — so `prefetch_concurrency`
bounds *downloads in progress*, not *records buffered and unconsumed*. Whenever a `snowflake_query()`
stream is consumed more slowly than the network delivers it — a `CREATE TABLE AS` into DuckLake on
object storage — the remaining result set accumulates in memory outside DuckDB's buffer manager,
where `PUBLISHER_DUCKDB_MEMORY_LIMIT` neither sees nor bounds it.

On a ~140M-row materialization that was an 8 GiB worker OOM-killed on every attempt, including
after the two DuckLake write bounds (#1112, #1116) were live. Those raise the consumer's
throughput and remain load-bearing; they could never close a gap whose producer side is unbounded.
The extension never calls `AdbcStatementSetOption` and exposes no secret field, `ATTACH` option,
setting or environment variable that reaches it, so a wrapper at the driver boundary is the only
place to set the option without forking the extension.

`result_queue_size = 1` makes every stream's goroutine block after one record until the consumer
reaches that stream, which re-couples the read-ahead to consumption.

### Measured

Same container (6 GiB / 4 CPU), same link, peak cgroup `memory.stat` `anon`,
`SNOWFLAKE_SAMPLE_DATA.TPCH_SF100.ORDERS LIMIT 20000000` (9 columns, ~165 B/row in Arrow):

| consumer | `result_queue_size` / `prefetch_concurrency` | peak anon | wall |
|---|---|---|---|
| fast (`count/sum`, 4 threads) | 100 / 5 (driver defaults) | 96 MiB | 13 s |
| slow (`threads=1`, Parquet zstd-19) | 100 / 5 | **3325 MiB** — the whole result, with 1% consumed | 598 s |
| slow, same | **1** / 5 | **286 MiB**, flat | 524 s |
| slow, same | 1 / 2 | 258 MiB, flat | 518 s |

Byte-identical output across the slow runs. On a fast consumer (100M rows, interleaved controls)
queue 1 with the default concurrency costs nothing measurable — 22 / 26 s against 21 / 25 s — and
even there removes the 408–1008 MiB the defaults buffered across four runs. Lowering
`prefetch_concurrency` to 2 *does* cost throughput (27–41 s), so the two knobs are exposed
independently and the image leaves that one at the driver default.

### What is in the image

`packages/server/adbc-shim/shim.c` (under 150 lines of C with comments) is compiled in a new `adbc-driver` build stage
and installed **as** `libadbc_driver_snowflake.so` beside every copy of the extension, with the
upstream driver renamed to `libadbc_driver_snowflake.real.so` alongside. The shim forwards every
ADBC call to the real driver, which it finds beside itself, and after each successful
`StatementNew` sets `adbc.rpc.result_queue_size` from `ADBC_RESULT_QUEUE_SIZE` and
`adbc.snowflake.rpc.prefetch_concurrency` from `ADBC_PREFETCH_CONCURRENCY`. Neither is defaulted in
the image, for the same reason `PUBLISHER_DUCKLAKE_*` have no defaults: an upgrade must not change
what the driver does until an operator asks. An unset variable sets nothing; a rejected option is
logged and the statement proceeds unbounded — the behaviour without the shim, never a failed query.
The shim logs one line per driver load — `[adbc-shim] wrapping … result_queue_size=<value|(unset)|(invalid …, ignored)>`
— saying whether the bound is on. Per driver load, not per process: the extension opens one ADBC
database per pooled Snowflake client and the driver manager re-runs `AdbcDriverInit` for each, so the
line repeats once per client and that is expected. The real driver is opened `RTLD_LOCAL` through the
shim's own handle, which is deliberately never closed (a Go c-shared library cannot be unloaded cleanly).

The driver download moves into that stage unchanged in substance (same pinned digest, same
fail-the-build posture) so that `selftest.c` can `dlopen` the shim/driver pair the way the driver
manager will and fail the **build** if the chain does not come up — no network, no credentials.
The runtime image gains two `.so` files; the compiler never reaches it.

`adbc.h` is vendored byte-for-byte from apache/arrow-adbc@7f35429e (`c/include/arrow-adbc/adbc.h`,
Apache-2.0, header retained), sha256-pinned and checked in the build stage so a re-vendor is a
deliberate edit: the `AdbcDriver` struct layout is the ABI being wrapped. Both wrapped slots sit in
the 1.0.0 region of the struct, so a 1.0.0 manager allocating the smaller struct is still safe.

### This is an interim

Every touchpoint carries the marker `ADBC-SHIM` — `grep -rn 'ADBC-SHIM'` returning nothing is
the definition of removed — and `packages/server/adbc-shim/README.md` carries the step-by-step
checklist (Dockerfile stage and final-stage hunks, the smoke-test line, the docs rows, the
directory, deployments, a retiring release note). It states the removal condition: when the extension exposes
the options ([iqea-ai/duckdb-snowflake#66](https://github.com/iqea-ai/duckdb-snowflake/issues/66))
the server issues the `SET` itself on the build and attach paths and this comes out; or when the
driver bounds its read-ahead by consumption as its documentation already implies
([adbc-drivers/snowflake#197](https://github.com/adbc-drivers/snowflake/issues/197)). Both filed
with the numbers above.

### Tests

- Build-time: `selftest.c` runs in the `adbc-driver` stage and fails the image build if the chain
  does not come up. It asserts `StatementNew` resolves into the shim by exact basename and
  `DatabaseNew` into the real driver, re-inits the same table to prove the self-wrap guard, and
  runs the env-var validator over an accept/reject matrix including the `strtol`-saturation
  overflow cases. It runs twice: against the real driver, and against `stub_driver.c`, a fixture
  that fills only empty table slots — the real driver overwrites every slot on init, so only the
  stub can hand the shim its own `StatementNew` back on a re-init; a build without the self-wrap
  guard fails that run (`second init re-wrapped the table`), verified by removing the guard. The
  header's digest is checked before compiling.
- CI's `docker_smoke_test` 4a now asserts **both** the shim and `libadbc_driver_snowflake.real.so`
  beside every extension copy — with the shim occupying the driver's file name, the original
  check no longer watched the file that can go missing — and the Dockerfile's final stage
  enforces the same as a count-match (one of each per extension directory, not "at least one
  somewhere"). 4b (a bogus-secret `snowflake_query()` reaches the login retry loop within 10 s)
  holds because the chain reaches the same point.
- The shim validates both variables as positive integers that fit a C `int`, checking `strtol`
  saturation: `0`, `-1`, `abc`, `9223372036854775808` all log `(invalid "…", ignored)` at load and
  set nothing, rather than printing as if the bound were on while the driver rejects them per
  statement. With both variables unset the shim is silent (no per-client log line); `ADBC_SHIM_DEBUG=1`
  restores it.
- `docker_smoke_test` builds and checks amd64 only (pre-existing workflow structure); the released
  image is multi-arch and the arm64 build (`docker-image.yml`, native runner) gets the build-stage
  self-test but not the functional 4b check.
- **Not automated:** that the option is actually *applied* on a statement. 4b's login fails before
  any `StatementNew`, and the self-test has no connection. The evidence is the real-credential runs
  above; the README's "what is and is not tested" table records the gap, and a deployment that turns
  the bound on should treat `result_queue_size=1` in the worker's `wrapping` line as its acceptance
  check.
- Against the built image, with real credentials: `snowflake_query('SELECT 1')` logs
  `result_queue_size=(unset)` with no env and `result_queue_size=1` with `-e ADBC_RESULT_QUEUE_SIZE=1`,
  and returns in both; the slow-consumer cell above re-run through the built image with
  `ADBC_RESULT_QUEUE_SIZE=1`: **288 MiB peak, 567 s, byte-identical parquet** (unshimmed: 3325 MiB).
- `node scripts/release-notes.mjs extract` accepts the new `[Unreleased]` section.


girishjeswani (approved):
Approving. I reviewed the C and the packaging separately, and compiled the shim against a stand-in driver rather than reading it alone — the core safety claims hold up under execution.

**The framing I'd change first, because it decides how much the C has to carry.** The rewiring is at image-build time: `Dockerfile:78` compiles the shim to `libadbc_driver_snowflake.so`, the exact name the DuckDB `snowflake` extension resolves via `dladdr`, and the upstream driver becomes `.real.so`. Nothing gates that, and the env var is only consulted inside the already-loaded shim. So for anyone taking the upgrade the image *is* different even with `ADBC_RESULT_QUEUE_SIZE` unset — the process dlopens a locally-compiled object which then dlopens the real driver, on every Snowflake query. `RELEASE_NOTES.md:41`'s "an image upgrade changes nothing" reads as though the image is untouched; the same paragraph already has the accurate version ("the driver behaves exactly as upstream ships it"). Worth the swap, because it is the difference between "inert" and "inert if the pass-through is perfect."

On that, the data-path inertness is real and I verified it: with both vars unset, `SetOption` is called **zero** times and `shim_StatementNew` is a straight forward. All `dlopen`/`dlsym` returns are checked, it **fails closed** (a missing real driver yields status 9, not a silent fallback, which is the right choice), path construction is explicitly bounds-checked before the `strcpy`, there is no double-release of `AdbcError`, and a failed `SetOption` leaves the statement valid and releasable. Clean under `-Wall -Wextra`.

Two things I'd fix.

**`shim.c:49` — the positive-integer check misses overflow.** It validates digits then `strtol(s, NULL, 10) > 0` but never checks `errno == ERANGE`, and `strtol` saturates to `LONG_MAX`. So `9223372036854775808`, `18446744073709551616` and `99999999999999999999999` all pass validation and get forwarded verbatim; the driver re-parses the original string and rejects it per statement, leaving a startup line that claims the bound is on when it is not. That is exactly the failure this function was added to prevent. `errno = 0` before the call and reject on `ERANGE` closes it. Everything else in the matrix is already right — `0`, `-1`, `abc`, `" 5"`, `"5 "`, `5x`, `+5`, `1e3` are all rejected and reported as `(invalid "…", ignored)`.

**`shim.c:141` — no self-wrap guard, and it segfaults.** The capture `real_StatementNew = driver->StatementNew` never checks the slot isn't already `shim_StatementNew`. If a manager reuses an `AdbcDriver` struct across inits and the driver preserves populated slots, the shim wraps itself into unbounded recursion — I reproduced it, exit 139, stack overflow. It doesn't fire today only because the real driver unconditionally overwrites that slot, which is undocumented driver behaviour the shim silently depends on and precisely the sort of thing a driver bump changes. One line:

```c
if (driver->StatementNew == shim_StatementNew) return ADBC_STATUS_OK;
```

**A correction to something I assumed going in.** I read `fe641843`'s "guard the real driver per extension directory" as a C-level load-once map and went looking for a first-call race on it. That's wrong — it's a Dockerfile/CI filesystem assertion, and there is no map, no guard and no mutex in `shim.c`; `AdbcDriverInit` dlopens unconditionally every time. The statics at `:36-40` are process-global and written unguarded on every init, which is formally a data race but benign in practice: TSan stayed quiet across 8 threads × 200 concurrent inits (with a positive control confirming TSan was live), because inits serialise through the loader lock and every one writes the same values. It only bites if two different real drivers are ever loaded in one process, where the last writer wins globally. Worth a comment stating that assumption, or `_Atomic`.

**Two things worth crediting specifically.**

The vendored header is byte-for-byte upstream — sha256 checked, `diff -u` produces zero lines, `struct AdbcDriver` field order intact with the 1.1.0 additions appended. I also checked a hazard that could have been nasty and isn't: both wrapped functions sit in the 1.0.0 region of the struct, before `ErrorGetDetailCount`, so the unconditional writes are safe even if a 1.0.0 manager allocates the truncated struct.

And the cost is measured rather than asserted. I came in expecting queue-1 to serialise download against consume and cost throughput broadly; `README.md:37-50` shows the opposite on the slow-consumer case — 286 MiB / 524 s against 3325 MiB / 598 s, byte-identical output, so *faster* — and 22-26 s vs 21-26 s on a fast consumer. The distinction that makes that true is drawn correctly: `result_queue_size` bounds records queued while `prefetch_concurrency` still governs parallel downloads, and the README explains why the image deliberately leaves the latter alone given it costs 1.3-1.6×.

**On the packaging side**, the `-35` is a clean relocation with nothing lost — every removed effective line reappears in the new `adbc-driver` stage, driver SHA256 pins are byte-identical so no version bump rode along, and the replaced verification is strictly stronger: the old `test -n "$(find … -print -quit)"` meant "a driver exists somewhere", the new one counts extension dirs, shim copies and real-driver copies and asserts all three match, closing the hole the old comment itself complained about. No compiler ships in the runtime image, the one network fetch is checksummed, and running `selftest` inside the build means a broken shim fails the build rather than the deploy — which is the right place for it.

**Three smaller things.**

The selftest wouldn't have caught either follow-up bug: I built at all three commits and it exits 0 with identical output each time, because it never sets an env var. Its `dladdr` assertion is also weaker than it looks — `libadbc_driver_snowflake.so` is a substring of `libadbc_driver_snowflake.real.so`, so it would pass if `StatementNew` resolved into the real driver, which is the one thing it exists to verify. Matching the basename, or asserting the pointer differs from the real driver's, would fix it.

The header tracks moving `main` with no pinned commit SHA and no checksum (`README.md:114-117`). It's 73 doc-only lines ahead of `apache-arrow-adbc-18` today, so nothing is wrong now, but this repo set the bar on provenance-plus-checksum for vendored artifacts during the chart.js review, and a re-vendor could silently pick up an ABI change.

And the `wrapping …` line at `shim.c:152` goes to stderr on **every driver load**, once per pooled client rather than once per process, unconditionally — including in the default image where the feature is off and there is nothing to report. Gating it on `ADBC_SHIM_DEBUG` when both vars are unset would keep the signal without the noise.

Two coverage notes rather than asks. `docker_smoke_test` is amd64-only while the released image is natively multi-arch, so the arm64 shim gets the build-time selftest but never the functional smoke check — pre-existing workflow structure, worth a line in the description. And `ADBC_REAL_DRIVER` lets an env var redirect which shared object gets dlopened; it's documented in the shim's README but not in `docs/configuration.md`, and being an env-controlled `dlopen` path it's worth a mention even as a test hook.

The `ADBC_*` naming break from the `PUBLISHER_*` family is right, incidentally, and I'd leave it: `PUBLISHER_*` means "read by `config.ts`", and this is read by a C library the server never touches, so keeping it distinct preserves that invariant rather than breaking it. One clause in the docs row saying so would stop the next reviewer re-deriving it.

PR #1140, opened 2026-09-10
fix(server): close the authorize bypass and the MCP bind by default

sagarswamirao:
## What

Two defaults in the server are open where they should be closed. Each is fixed independently, with its own commit and its own test:

- **The authorize bypass trusted a header's presence.** `x-publisher-bypass-authorize` disabled every `#(authorize)` gate with no secret to know. It now requires `PUBLISHER_BYPASS_AUTHORIZE_SECRET` and refuses when none is configured.
- **MCP bound `0.0.0.0` with bare `cors()`.** It now binds `127.0.0.1` via its own `MCP_HOST`, and CORS reads `MCP_CORS_ORIGINS`, defaulting to no cross-origin.

These came out of a twelve-finding multi-tenant security review (F-5 and F-7). They were grouped because each is **already unreachable or already blocked** on the deployment this review was run against -- an ingress strips the bypass header, and port 4040 is not on the container or Service -- so neither changes what that deployment does today. The value is for anyone running Publisher directly, where both are live.

## What is deliberately not here

**The `?reload=true` gate is not in this PR.** An earlier revision added one, keyed on a shared `x-publisher-reload-secret` env var. A shared static secret is not authorization: it has to live wherever the proxy runs so the proxy can attach it to forwarded requests, which means every caller that reaches the proxy still gets its reload through. It would have added a configuration step and a migration burden without changing who can recompile a package. Gating reload needs per-caller identity, not a static string every caller can borrow, so it belongs with authorization work rather than a default flip. `?reload=true` is untouched here.

The per-tenant caps half of F-8 (query-concurrency counter, SSE cap, memory governor) is also not in this PR. Re-keying those per tenant multiplies peak concurrency by the number of active tenants inside one fixed-memory pod; it needs the memory arithmetic and a load test, not a quick change alongside two default flips. `query_concurrency.ts` and `heap_check.ts` are untouched.

F-6 from the same review is a Credible-side router concern in a different repository and is not here either.

## One thing worth reading before approving

F-7 looks like a one-line default flip and is not. `PUBLISHER_HOST` drove **both** `mainServer.listen` and `mcpApp.listen`, so defaulting it to loopback would have moved the REST port to localhost and taken the deployment offline -- turning a change chosen for being non-breaking into an outage.

MCP therefore gets its own knob. Precedence is `MCP_HOST`, then an explicit `PUBLISHER_HOST` so `--host` still moves both together, then `127.0.0.1`. The REST default is unchanged. The `.mcp.json` the server writes and the `claude mcp add` line it prints both derive from `MCP_HOST`, so neither hands out a host the server is not listening on.

## Migration

Each flip is silent until something stops working, so each needs a deliberate step. `RELEASE_NOTES.md` carries the operator-facing version.

| Change | Symptom if you do nothing | Step |
| --- | --- | --- |
| Bypass requires a secret | A bypass that used to work stops bypassing | Set `PUBLISHER_BYPASS_AUTHORIZE_SECRET` and send it as the header value, or stop relying on the bypass |
| MCP binds loopback | A remote MCP client can no longer reach `:4040` | Set `MCP_HOST=0.0.0.0` (and put a gateway in front, per `docs/security-posture.md`) |

## Testing

Each finding has a test that fails without its production change -- verified by reverting each one and watching it go red, not by assertion:

| Spec | With the change reverted |
| --- | --- |
| `authorize_bypass_header.spec.ts` | 6 fail (`Received: true` where the bypass must be refused) |
| `mcp_exposure.spec.ts` | 3 fail (bind default, bind wiring, CORS registration) |

Full suite on the committed tree: **3678 pass, 0 fail** (server), **391 pass, 0 fail** (create-malloy-package). `format`, `typecheck` (all five packages), and `lint` all exit 0.

## One thing found on the way that is not fixed here

**A test helper's block-comment regex is unsound on `server.ts`.** `/\/\*[\s\S]*?\*\//g` treats the `/*` inside route literals such as `public/*` as a comment opener, which deletes 39,073 characters -- most of the file -- before the assertion runs. The pre-existing guards survive only because their anchors happen to fall outside the swallowed region. The specs added here strip line comments only, and say why.


housejester:
Reviewed all four commits at `cd9a93d`, traced both reload routes plus every in-repo `?reload=true` caller, and pulled the logs for the three red CI jobs to pin causes rather than guess.

Three genuinely independent fixes, each with a test that goes red when its production change is reverted, and the fail-closed direction is right on all three. The F-7 write-up — "looks like a one-line default flip and is not", because `PUBLISHER_HOST` drove both listeners — is the best part of this PR, and gating the legacy `/projects/…` alias alongside the modern route is exactly the thing that usually gets missed. Calling out the unsound block-comment regex in the existing test helper, rather than quietly copying it, is also worth saying out loud.

What isn't ready is the blast radius of the reload gate: CI is red in three jobs, one of them for a reason that reintroduces the exact bug its guard exists to catch, and a fourth failure wave is queued behind a job that never got to run.

## Mental-model delta

Three request-level capabilities stop being properties of *the deployment's edge* and become properties of *server configuration* — but only on the code path each one actually gates, and the MCP half is inert in the published Docker image.

| --- | --- | --- |
| Who decides whether `#(authorize)` holds | whoever strips `x-publisher-bypass-authorize` at the edge | whoever holds `PUBLISHER_BYPASS_AUTHORIZE_SECRET` |
| MCP reachability | one knob for both listeners | two knobs; MCP defaults loopback — *unless* `PUBLISHER_HOST` is set, which the image does |
| MCP CORS | reflect-any-origin | deny-by-default allowlist, `*` opt-in |
| `?reload=true` on a **package** | unauthenticated recompile | 403 without the secret |
| `?reload=true` on an **environment** | unauthenticated reload | **unchanged — still open**, and it is the bigger hammer |

Things a future reader now has to hold: there are **two** reload gates to keep in sync (`server.ts:1626`, `server-old.ts:655`) and a **third** reload path with none; MCP host resolution treats an *inherited* `PUBLISHER_HOST` as "the operator asked for a wider bind"; and `/health*` is served on the MCP port, so the bind flip moves those out of network reach with it.

## Worth acting on

**[HIGH] The reload gate 403s the Playwright fixtures.** `notebook-givens.spec.ts:60`, `notebook-authorize.spec.ts:65` and `notebook-readme-links.spec.ts:56` each `fetch(...?reload=true)` with no header and throw on `!res.ok`; `playwright.config.ts:63`'s `webServer` block passes no env, so `PUBLISHER_RELOAD_SECRET` is unset. The log is unambiguous — `Error: Package reload failed: 403 Forbidden` three times, 14 tests failed or not-run. Fix: set the secret in `webServer.env` and send the header, ideally from one shared helper instead of three copies.

**[HIGH] `--mcp_host` is missing from `SERVER_VALUE_FLAGS`, and the drift it hides is a real exposure bug.** The flag was added to `server.ts`'s argv chain but not to the mirror in `packages/create-malloy-package/src/scaffold.ts:1558`, so `scaffold.spec.ts:1112` fails on all three platforms with `- "--mcp_host"`. Not only a red test: `hostFlagOf` (`scaffold.ts:1627`) walks tokens *positionally* precisely so a value-flag with a missing value swallows the next token the way the server does. With `--mcp_host` unknown to it, `… --mcp_host --host 127.0.0.1` reads as loopback to the scaffolder while the server consumes `--host` as the MCP host and binds REST on `0.0.0.0` — the exposure warning suppressed on the one shape that is genuinely exposed, which is the case the comment at `scaffold.ts:1631` says the positional walk exists to prevent. That drift guard was the only thing enforcing the mirror.

**[HIGH] A second failure wave CI has not reached.** `cross-platform-tests.yml` runs `create-malloy-package tests` *before* `Run integration tests`, so the server integration tier never executed here. At least three specs reload over REST and assert the outcome: `concurrent_package.integration.spec.ts:170` (`expect(status).toBe(200)` on "all succeed"), `dashboards.integration.spec.ts:818` (`expect(reload.status).toBe(424)`), and `materialization_lifecycle.integration.spec.ts:331,368`, where unasserted reloads are used to *reset state* — those will fail silently and surface as confusing downstream failures. Worth running `bun run --cwd packages/server test:integration` before the next review pass, so fixing the drift test does not just unblock the next red step.

**[MED-HIGH] The closed door has a bigger open one beside it.** `GET /api/v0/environments/:env?reload=true` (`server.ts:1550`) is ungated and reaches `environmentStore.getEnvironment(name, true)` → `reloadEnvironmentManifest` → `addEnvironment` → `existingEnvironment.update()` / `loadEnvironmentIntoDisk` (`environment_store.ts:1496`, `:1607`): re-reads config from disk, re-fetches configured package locations (including `github`/`gcs`/`s3`, over on-disk edits) and writes the database. It is a superset of the per-package reload, documented in `api-doc.yaml:263` as "Load / reload the environment", and it is the cheaper hammer for anyone driving repeated recompiles. The PR body scopes it out; `RELEASE_NOTES.md` does not, so an operator reading only the release note will believe reload is closed. I would gate it here — it is the same eight lines and the same `authorizeReload` — or, failing that, add the sentence to the note.

**[MED-HIGH] The MCP bind flip is a no-op in the published image, while the docs now assert loopback.** `docker/production.docker:28` and `docker/malloy-samples.docker:18` both set `ENV PUBLISHER_HOST=0.0.0.0`, so `MCP_HOST` resolves to `0.0.0.0` in every container run — the distribution `AGENTS.md:60` and `docs/deployment.md:105` document, with `-p 4040:4040`. The precedence rule reads an image-baked ENV as a deliberate operator widening, which it is not. I think that ENV is load-bearing (loopback inside a container makes port publishing useless), so the honest fix is disclosure rather than a flip: state in `RELEASE_NOTES.md` and `docs/configuration.md:85` that the official image sets `PUBLISHER_HOST=0.0.0.0` and therefore still exposes MCP unless you set `MCP_HOST=127.0.0.1` and drop `-p 4040`. Relatedly, `packages/server/README.docker.md:74` — the image-facing env table, and the one that flags "the one env var the image sets for you" (now two) — lists neither `MCP_HOST`, `MCP_CORS_ORIGINS`, nor either new secret.

## Smaller notes

**[MED] The doc sweep stopped at three files.** `?reload=true` is still documented as needing nothing in `AGENTS.md:229/235/271` — including §7, whose whole premise is that REST is the supported interface when nobody can reconnect an MCP client, reload included — plus `packages/create-malloy-package/templates/AGENTS.md:159` (every scaffolded workspace's briefing), `skills/malloy-dashboards/SKILL.md:59`, `docs/packages.md:107`, `docs/dashboards.md:531`, `docs/configuration.md:59`, and `docs/persist-storage-tutorial.md:225/430/627/656`, where three of the four are `curl … >/dev/null` so the 403 is invisible and the next step reads stale state. `SECURITY.md:46` also still says the server binds `0.0.0.0` flat.

Worth pairing with a design question rather than only a doc fix: the risk being priced is *"an untrusted caller can reach this route"*, and the bind already answers that. As shipped, a loopback-only dev server with no secret loses reload entirely, and an unattended agent — MCP now loopback, REST reload now 403 — has no reload path at all. Would gating on effective reachability (require the secret only when the REST bind is not loopback) close the same hole with none of that fallout? If the always-on version is deliberate, that reasoning belongs in the release note.

**[MED] CodeQL is red on a production line, not only on specs.** `server.ts:400` draws *Permissive CORS configuration* because `getMcpCorsOrigins()` can return `"*"` from the environment; two further alerts land on the new specs. The wildcard is also a footgun on an unauthenticated endpoint — an operator who sets `*` reopens what this commit closes. Dropping `*` and requiring explicit origins clears the production alert and the footgun together; otherwise it needs a dismissal with a reason. Either way it wants a decision, since the check is currently failing.

**[MED] `/health` follows MCP to loopback, and that is not in the migration table.** `registerHealthEndpoints(mcpApp)` (`server.ts:391`) serves `/health`, `/health/liveness` and `/health/readiness` on the MCP port, so any external probe or load-balancer check against `:4040/health` starts failing. Belongs in the row next to "a remote MCP client can no longer reach `:4040`".

**[MED] A bypass rollout note.** The bypass now fails silently in the direction that matters: a data-management caller that keeps sending the old `true` gets gates enforced and an empty index, with nothing in the response saying why. Anyone whose indexer relies on it needs `PUBLISHER_BYPASS_AUTHORIZE_SECRET` provisioned on both sides *before* the server carrying this change rolls, not after. The release note says to set the variable; saying that the ordering matters, and that the symptom is silent, would help.

**[LOW-MED] One MCP-URL advertiser was missed.** The PR states that the `.mcp.json` and the printed `claude mcp add` line both derive from `MCP_HOST`. There is a third: `formatReadinessLine` (`environment_store.ts:427`) builds `PUBLISHER_READY … mcp=http://<host>:<mcpPort>` from `PUBLISHER_HOST`, and its own comment claims to mirror `server.ts`'s use-site defaults. `--host 10.0.0.5 --mcp_host 127.0.0.1` then advertises an MCP URL nothing is listening on, on the exact line `AGENTS.md` tells scripts to grep. One-line fix: reuse the same resolution.

**[LOW] Stale rationale in a package this PR has to touch anyway.** `scaffold.ts:1547` ("`PUBLISHER_HOST` falls back to `0.0.0.0` for both the REST and the MCP listener"), `:252` ("Both listeners take this one flag") and `:1163` ("Publisher's own default is 0.0.0.0 as well") are all now false for MCP. Nothing behaves differently — the generated script sets `--host 127.0.0.1` explicitly — but it is the reasoning the next reader inherits.

**[LOW, non-blocking]** `secretsMatch` is duplicated verbatim in `authorize_bypass_header.ts:41` and `reload_authorization.ts:38`. The "standalone, no transitive imports" argument is about not importing `server.ts`; a four-line `secret_compare.ts` with no imports satisfies both and stops the copies drifting. Also, on "Full suite on the committed tree: 3692 pass, 3 skip, 0 fail" — that is `packages/server`. The root `test` script chains `test:create-malloy-package`, where the drift spec lives, and `packages/server`'s own `test` chains `test:integration`; running the root script would have caught two of the three red jobs locally.

## Top three before merge

1. Get the reload gate's blast radius under control — Playwright fixtures, the three integration specs, and the seven un-swept docs — or rescope the gate to non-loopback binds.
2. `--mcp_host` into `SERVER_VALUE_FLAGS`, then actually run the integration tier so the second wave is not discovered post-merge.
3. Decide the two disclosure gaps: the image's `PUBLISHER_HOST=0.0.0.0` (which makes the MCP flip inert exactly where it ships), and the still-open environment-level `?reload=true` — gate it, or say so in the release note.

<sub>Review pass by an assistant session ("Stoic Gibbon") on behalf of a maintainer; findings verified against the tree at `cd9a93d` and the CI logs for runs 34520774765 / 34520775192.</sub>

sagarswamirao:
Thanks -- this was a careful pass, and it changed the shape of the PR rather than just adding commits to it.

The headline: **the reload gate is gone entirely.** Not rescoped, removed. That was the right call for a reason your own review points at, and it moots most of the list below.

## Why reload came out

The gate keyed authorization on a shared static secret in `PUBLISHER_RELOAD_SECRET`. A proxy sitting in front of the server has to hold that secret to attach it to forwarded requests -- so every caller who reaches the proxy still gets their reload through. It was a configuration step and a migration burden that did not change who can recompile a package.

Checking the deployment side made that concrete: our router refuses reload with `501 NOT_IMPLEMENTED` *before* dispatching, at all four call sites (`EnvironmentsController.getEnvironment`/`getPackage`, and the legacy `ProjectsLegacyController.getProject`/`getPackage`). No reload of any kind reaches the Publisher through it. So the gate was not protecting a reachable path there, and for a direct deployment it was protecting it with a secret every caller could borrow. Gating reload needs per-caller identity, which is authorization work, not a default flip.

The PR is now two independent fixes: the authorize bypass (F-5) and the MCP bind/CORS (F-7).

## What that resolves from your list

- **[HIGH] Playwright fixtures 403** -- gone. Those three specs and `playwright.config.ts` are back to `origin/main` verbatim; no secret, no header, no shared helper needed.
- **[HIGH] The second integration wave** -- gone as a reload concern, but I took the substance of it: I ran `test:integration` rather than assuming. It surfaced 9 failures, 8 of which were mine for running the suite without a prior `bun run build`. Current state on the branch: server unit 3678/0, create-malloy-package 391/0, integration 406/0, format/typecheck/lint all clean. Your closing note -- that the root script chains the two tiers I skipped -- was exactly right, and it is the process fix I actually needed.
- **[MED-HIGH] The environment-level `?reload=true`** -- moot as a gap in this PR, since the package gate it sat beside is gone too. Both are now equally ungated upstream, which is at least consistent, and the release note no longer claims otherwise.
- **[MED] The seven un-swept docs** -- moot; nothing in them is now false.
- **[MED] Rescoping to non-loopback binds** -- your design question ("would gating on effective reachability close the same hole with none of the fallout?") was the better version of the gate. It is worth revisiting whenever reload authz is done properly, so I have not lost it, but it belongs with per-caller identity rather than here.
- **[LOW] Duplicated `secretsMatch`** -- moot, only one copy remains now that `reload_authorization.ts` is deleted.
- **[HIGH] `--mcp_host` missing from `SERVER_VALUE_FLAGS`** -- fixed (`scaffold.ts:1566`), and the drift spec is green across all three platforms. Your explanation of *why* it mattered -- that `hostFlagOf` walks positionally, so an unknown value-flag makes `--mcp_host --host 127.0.0.1` read as loopback to the scaffolder while the server binds REST wide -- is the part I would have missed. That is the exposure warning being suppressed on precisely the shape that is exposed.

## What survives, and what I disagree with

**[MED-HIGH] The image's `PUBLISHER_HOST=0.0.0.0` makes the MCP flip inert -- confirmed, and still open.** `docker/production.docker:28` and `docker/malloy-samples.docker:18` both set it, so `MCP_HOST` resolves to `0.0.0.0` in every container run. I agree that ENV is load-bearing and that disclosure beats a flip. Not yet written; it needs `RELEASE_NOTES.md`, `docs/configuration.md`, and `README.docker.md` (which currently documents neither `MCP_HOST` nor `MCP_CORS_ORIGINS`). Happy to add it here or take it as a follow-up -- say which you prefer.

**[LOW-MED] `formatReadinessLine` -- confirmed, real.** `environment_store.ts:427` builds `mcp=` from `PUBLISHER_HOST`, so `--host 10.0.0.5 --mcp_host 127.0.0.1` advertises an MCP URL nothing is listening on, on the line `AGENTS.md` tells scripts to grep. A third advertiser I missed when I claimed there were two.

**[MED] CodeQL permissive-CORS on `server.ts:400` -- still red, still wants a decision.** Worth flagging that the inline-suppression route does not exist here: this repo uses CodeQL *default setup*, which ignores in-repo config, and `// codeql[...]` is Semgrep syntax while `lgtm[...]` is dead LGTM.com syntax. So it is either dropping `*` or an API dismissal. I lean toward your framing -- `*` on an unauthenticated endpoint is a footgun that reopens what the commit closes -- but it is a behaviour change from what is currently merged, so I would rather you call it.

**[MED] `/health` following MCP to loopback -- I do not think this one holds.** `registerHealthEndpoints` is called twice: `server.ts:390` on `mcpApp` and `server.ts:909` on the REST `app`, the latter commented as required for Kubernetes monitoring. So `/health`, `/health/liveness` and `/health/readiness` remain reachable on `PUBLISHER_PORT` regardless of the MCP bind, and an external probe or LB check against the REST port is unaffected. Only a probe pointed specifically at `:4040/health` moves, which seems like the intended consequence of the bind flip rather than collateral. Have I misread the setup you had in mind?

**[MED] The bypass rollout ordering note -- fair, and not yet written.** The failure is silent and in the direction that matters: a caller still sending the old `true` gets gates enforced and an empty index with nothing in the response explaining why, so the secret has to be provisioned on both sides before the server rolls. That belongs in the release note and is not there yet.

**[LOW] Stale rationale in `scaffold.ts`** (`:1547`, `:252`, `:1163`) -- correct, all three now false for MCP. Not yet fixed.

## Where that leaves it

Ready: F-5, F-7, `--mcp_host` drift, and a release note covering both flips. Outstanding and unwritten: the image-ENV disclosure, the readiness-line fix, the bypass ordering note, the scaffold comments, and a decision on the CORS wildcard. None are large; I did not want to bundle them in before you had seen the reload removal, since it changes what the PR is.

One correction to my own body while I am here: the "3692 pass, 3 skip" figure you flagged as `packages/server`-only is gone -- it counted the now-deleted reload specs.

housejester (approved):
Re-reviewed at `37bf6234` (the new head is a merge of main; the PR's own ten files are unchanged from `e3326acc`). Approving.

Removing the reload gate rather than rescoping it was the better call, and the reasoning is the part worth keeping: a shared static secret that the proxy has to hold is not authorization, so the gate bought a migration burden without changing who can recompile a package. Running `test:integration` rather than assuming, and reporting that 8 of the 9 failures were your own setup, is the answer I was hoping for rather than the one I asked for. The `--mcp_host` drift fix is in and green on all three platforms.

What remains is small, and most of it is documentation, which is yours to sequence.

## Before merge

**The scaffolder needs `0.0.15`.** `Require a version bump when the published content changes` is red: `packages/create-malloy-package/package.json` declares `0.0.14`, and `0.0.14` is already on npm (`npm view @malloy-publisher/create-malloy-package versions` → `0.0.11`–`0.0.14`). The check's own wording is the part that matters — *"a release would SKIP this package and stay green."* The only published change here is the `--mcp_host` entry in `SERVER_VALUE_FLAGS`, so the failure mode is that the drift fix never reaches a scaffolded workspace while every signal reads green. The `0.0.13` bump in `f8e9867b` was correct when you made it and was overtaken by the merge from main — staleness, not an oversight.

## The CORS wildcard — you asked me to call it

**Drop `*`.** Requiring explicit origins clears the production alert and the footgun in one move, and `*` on an unauthenticated endpoint that can read whatever the models connect to hands back the thing this commit closes. An operator reaching for `*` is almost always reaching for "let my console in", which a named origin serves.

Worth knowing either way: the `// codeql[...]` comments are inert, as you said, and the alerts confirm it — `gh api repos/malloydata/publisher/code-scanning/alerts?ref=refs/pull/1140/head` reports **157 open** (`mcp_exposure.spec.ts:36`) and **158 open** (`server.ts:404`); only the `js/missing-rate-limiting` one shows `fixed`, because its file left with the reload gate. So if `*` stays, the comments want rewording as plain prose without the directive-shaped prefix, plus an API dismissal carrying that reasoning — otherwise the next reader sees a suppression that isn't one.

## `/health` — you were right, I was wrong

`registerHealthEndpoints` is called twice, `server.ts:390` on `mcpApp` and `server.ts:909` on the REST `app`, so REST-port probes and LB checks are untouched and only a probe aimed specifically at `:4040/health` moves — the intended consequence of the bind flip, not collateral. Withdrawn; nothing owed, and it should not go in the migration table.

## Documentation — a fact to price, not a request

All of the below is non-blocking and entirely your call; I would not hold the PR for any of it.

One thing is worth surfacing because it looks deliberate and probably was not: **the doc sweep from `cd9a93d7` was reverted inside `095ce012`**, three commits before `revert(server)`, which is clean and says so explicitly.

```
$ git diff cd9a93d7 095ce012 --stat -- AGENTS.md docs/ api-doc.yaml
 9 files changed, 28 insertions(+), 74 deletions(-)
```

`095ce012`'s message describes only the scaffold mirror and the reload-guard contract, and the same commit also pulls in main's `Dockerfile` and `packages/server/adbc-shim/*` — which reads like a rebase resolution that took main's side on the doc files rather than a decision about them. Two of the nine (`api-doc.yaml`, `docs/api-overview.md`) were reload-only and are correctly gone. The other seven described F-5 and F-7, which are still in the PR.

The consequence is that `MCP_HOST`, `MCP_CORS_ORIGINS`, `PUBLISHER_BYPASS_AUTHORIZE_SECRET` and `--mcp_host` now appear in no markdown or spec file at all — `--help` is the only place the flag is described — and these read false against the shipped behaviour:

| File | Says |
| --- | --- |
| `docs/authorize-bypass-deployment.md:11` | `x-publisher-bypass-authorize: true`, in the page written for whoever operates this |
| `docs/authorize.md:349`, `:398`, `:425` | `: true` works; `:425` is the Known Limitations entry |
| `docs/security-posture.md:44` | `: true` skips gates outright |
| `docs/choosing-a-surface.md:150` | an HTML app can send `: true` and skip gates |
| `SECURITY.md:46`, `:54` | binds `0.0.0.0`; gate skipped on `: true` |
| `AGENTS.md:60` | "Both bind `0.0.0.0` by default" |
| `docs/ai-agents.md:108` | "The server binds `0.0.0.0` by default" — in the MCP section |
| `docs/configuration.md:84` | `PUBLISHER_HOST` binds "both the REST and MCP servers" |
| `packages/server/README.docker.md:74` | `PUBLISHER_HOST` = "Bind address", and no row for either new var |

If you want them back, `git checkout cd9a93d7 -- docs/authorize.md docs/authorize-bypass-deployment.md docs/security-posture.md docs/choosing-a-surface.md docs/configuration.md docs/ai-agents.md AGENTS.md` recovers text you already wrote; the reload-only bits to drop afterwards are the `PUBLISHER_RELOAD_SECRET` row in `configuration.md` and the reload-secret bullet in `ai-agents.md`. `SECURITY.md` and `README.docker.md` were never in the sweep.

Three smaller notes in the same bucket, all low:

- **The image ENV.** Confirmed unchanged (`docker/production.docker:28`, `docker/malloy-samples.docker:18`, and `docker-compose.example.yml:27` publishing `4040`). The mechanism worth a sentence wherever this lands: the precedence rule reads *any* `PUBLISHER_HOST` as "the operator asked for a wider bind" and cannot tell an operator's `--host` from a value baked into an image — which is also why `-e MCP_HOST=` falls through to wide rather than back to loopback. If prose feels like a lot for it, `ENV MCP_HOST=0.0.0.0` beside the existing line in both Dockerfiles is behaviour-identical today and makes the image's wide MCP bind a line someone can read and override.

- **The bypass rollout note**, if you write it, has a usable mechanism rather than just a warning: across an old/new overlap there is no single header value that works on both sides — old pods accept only `true`, new pods only the secret — so the bypass flaps and the symptom is an intermittently empty index with nothing in the response saying why. One value bridges it: set `PUBLISHER_BYPASS_AUTHORIZE_SECRET=true` for the roll, then rotate to a random value once every pod carries the change.

- `formatReadinessLine` and the three `scaffold.ts` comments you already confirmed — no need to restate them; the only thing I would add is that the docblock at `service/environment_store.ts:418` claims the line "mirrors server.ts's use-site defaults", so the comment is false alongside the value, and the same `0.0.0.0` claim ships to users in `packages/create-malloy-package/README.md:299,317,320` and `templates/AGENTS.md:102-115`.

## Checked and cleared

So you do not re-spend attention there:

- Promoting the bypass header to a credential does not leak it. There is no `morgan` and no header-bag logging anywhere in `packages/server/src`, and `readBypassAuthorize` has one use site (`server.ts:1936`).
- Adding `--mcp_host` does not open a new hole in the scaffolder's loopback claim. `--host 127.0.0.1 --mcp_host 0.0.0.0` would read as loopback to `hostFlagOf`, but `matchesGeneratedInvocation` declines on any flag absent from `GENERATED_FLAGS`, so such a script is never adopted and the caller falls back to the explicit loopback command. The `SERVER_VALUE_FLAGS` addition closes the swallowing bug without blessing a new shape.
- No IPv6 regression: `PUBLISHER_HOST` already defaulted to the IPv4-only `0.0.0.0`, so `127.0.0.1` does not newly strand a client resolving `localhost` to `::1`.
- `authorize_bypass_header.spec.ts:51` asserts `"true"`, `" True "`, `"1"`, `"yes"` are all refused with a secret configured — that would have failed before the fix, so it pins the change rather than restating it.

<sub>Re-review by an assistant session ("Gentle Stoat") on behalf of a maintainer; verified against the tree at `37bf6234`, the code-scanning alerts on `refs/pull/1140/head`, and the CI logs for run 35133037924.</sub>

PR #1143, opened 2026-09-10
fix(connections): finalize eligible ad hoc SQL so plain statements work on Postgres

### LINKED ISSUE #1141: sqlQuery fails for plain SQL on Postgres connections: the path does not apply the row_to_json wrapper the connector's row.row unwrap expects
## Summary

`POST /api/v0/environments/{env}/connections/{conn}/sqlQuery` fails for every plain SQL statement against a **Postgres** connection, including `SELECT 1`:

```
{"code":502,"message":"The \"string\" argument must be of type string or an instance of Buffer or ArrayBuffer. Received undefined"}
```

Reproduced on v0.2.4 and v0.2.6 with a local Postgres 18.

## Root cause

Malloy's Postgres dialect wraps every generated query as `SELECT row_to_json(...) AS row FROM (...)` (`@malloydata/malloy`, `dialect/postgres/postgres.ts`), and the Postgres connector's `runSQL` / `runSQLStream` unwrap each result row with `row.row` (`@malloydata/db-postgres`, `postgres_connection.ts`). That is the contract between dialect and connector. Statements that are not wrapped come back as `null` / `undefined` rows. Verified against `PooledPostgresConnection` 0.0.432:

| statement | `runSQL` rows | `runSQLStream` rows |
|---|---|---|
| `SELECT 1 AS x` | `[null]` | `undefined` |
| `SELECT row_to_json(t) AS row FROM (SELECT 1 AS x) t` | `[{"x":1}]` | `{"x":1}` |

The server already honours the contract in one place: `service/db_utils.ts` `getSchemasForPostgres` wraps its introspection SQL in `row_to_json`, with a comment explaining why. The ad hoc `sqlQuery` path (`controller/connection.controller.ts` → `stream_helpers.ts` `streamSqlWithBudget`) passes the caller's statement through unwrapped. Since #779 (row + byte caps, 2026-05) Postgres takes the streaming branch, where `Buffer.byteLength(JSON.stringify(undefined))` throws the error above; before that the buffered branch returned an array of nulls. So plain SQL against Postgres has never worked through this endpoint; it just failed quietly until the caps landed.

## Repro

```
curl -s -X POST localhost:4000/api/v0/environments/<env>/connections/<postgres-conn>/sqlQuery \
  -H 'content-type: application/json' -d '{"sqlStatement":"SELECT 1 AS x"}'
```

## Workaround

Callers wrap their own statement: `SELECT row_to_json(t) AS row FROM (<sql>) t`.

## Suggested fix

In the `sqlQuery` path, apply the same wrapper `getSchemasForPostgres` uses, gated on `Dialect.hasFinalStage` rather than on a hardcoded Postgres check, so any dialect sharing the `row.row` contract is covered. Take the wrapper text from the dialect's own `sqlFinalStage()` for the same reason. Add an integration test that sends `SELECT 1` through `sqlQuery` against a real Postgres: nothing in the current suite pushes a row through the streaming branch for this connector, which is how the byte-cap change shipped without noticing.

The wrapper cannot be applied unconditionally, and the reason it cannot is the more important half of this issue.

**The `publisher`-type proxy connection must not be wrapped.** `@malloydata/db-publisher` implements `runSQL` via this endpoint, and it reports the REMOTE connection's `dialectName` — so for a proxied Postgres connection the Malloy compiler on the caller's side has already finalized the statement before it is sent, and `db-publisher` does no unwrapping of its own. Passing the statement through verbatim is what makes that path work today: the remote's connector strips the one `row` level the compiler added. Wrapping adds a second level that nothing strips, and the caller reads nulls off `{"row": {...}}` — a wrong answer rather than an error, which is worse than the bug above. The Malloy CLI and the VS Code extensions reach a Publisher over this connection type, so their versions are not ours to coordinate.

**Statements that cannot sit in a subquery must not be wrapped either** — DDL, `SET`, `EXPLAIN`, `SHOW`, a data-modifying CTE, Postgres' `SELECT ... INTO`. They return no rows for the connector to unwrap, so they already work through this endpoint; wrapping turns a working statement into a syntax error.

Both groups are recognisable from the statement itself: one already mentions the dialect's final-stage call, the other does not open as a subquery-safe `SELECT`. Every test should fail toward passing the statement through, since a statement wrongly passed through can be reported, while a statement wrongly wrapped returns wrong data. A statement in neither group that still yields nullish rows should answer 400 naming the wrapper to apply by hand, rather than the 502 above.

housejester:
Fixes #1141.

`POST /api/v0/environments/{env}/connections/{conn}/sqlQuery` failed for every plain statement against a Postgres connection, `SELECT 1` included, with `The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received undefined` and HTTP 502.

Malloy's Postgres dialect finalizes a query by collapsing its result into a single JSON column named `row`, and `@malloydata/db-postgres` unwraps that column from every row it returns — the contract between dialect and connector. A statement the caller wrote does not project it, so the driver handed back nothing readable. It had never worked through this endpoint; before the response caps landed it returned an array of nulls rather than failing, so nothing said so.

## What changed

`service/final_stage_sql.ts` decides eligibility and builds the wrapper. Both are read off the dialect rather than hardcoded: `Dialect.hasFinalStage` gates it, `sqlFinalStage()` supplies the wrapper text, and the token to look for is extracted from `sqlFinalStage()` output rather than spelled `row_to_json`. Postgres is the only dialect with a final stage today, pinned by the existing tripwire in `incremental_compiler_contract.spec.ts`.

`db_utils.ts` and `incremental_apply.ts`'s `probeSelect` already honour this contract for SQL they author themselves. Neither is touched here; this is the same contract for SQL the caller authors, where the statement's shape is not known in advance.

## What is deliberately not wrapped

The endpoint carries two contracts at once, and nothing in the request distinguishes them, so eligibility is decided from the statement.

**Already-finalized SQL.** This is what a `publisher` proxy connection forwards: it reports the remote connection's `dialectName`, so the Malloy compiler on the caller's side finalized the statement before it was sent, and `db-publisher` does no unwrapping of its own. Passing it through is what makes that path work — the remote's connector strips the one `row` level the compiler added. Wrapping adds a second level that nothing strips, and the caller reads nulls off `{"row": {...}}`: a wrong answer rather than an error. The Malloy CLI and the VS Code extensions reach a Publisher over this connection type.

**Statements that cannot sit in a subquery** — DDL, `SET`, `EXPLAIN`, `SHOW`, a data-modifying CTE, Postgres' `SELECT ... INTO`. They return no rows for the connector to unwrap, so they already worked; wrapping turns a working statement into a syntax error.

Every test fails toward passing the statement through. A statement wrongly passed through can be reported; a statement wrongly wrapped returns wrong data. So a statement in neither group that still yields nullish rows now answers 400 naming the wrapper to apply by hand, rather than the 502 above — a nullish row is never data, and the guard fires whether or not the byte cap is enabled (with the cap disabled the old code returned a response full of nulls and said nothing).

## Tests

`hammer/scenarios/connections/01-raw-sql-postgres/` runs against a real Postgres: the reported `SELECT 1`, a projection read back by column name, a CTE, DDL that must pass through *and actually run*, and the compiled-query shape a proxy forwards, which must come back as plain columns.

Both halves were checked by reverting rather than assumed. Without the wrapper the scenario fails with the new 400. With the marker check removed — the blanket wrap the issue originally suggested — the proxy step fails `expected "US", got null`, which is the silent corruption that motivates the gating.

`final_stage_sql.spec.ts` covers eligibility; `connection.controller.spec.ts` pins that the controller applies the verdict and what an unfinalizable statement reports.

Server suite green: 3662 unit, 410 integration. Two hammer scenarios (`quoted-persist-name`, `quoted-persist-name-colocated`) are red on `main` and untouched here — they fail at `createMaterialization` with the 424 introduced by #1106, which deliberately changed the behaviour they pin.

## Docs

`api-doc.yaml` on both `sqlQuery` endpoints, a `[Unreleased]` note, and `docs/connections.md` writes down the proxy passthrough as load-bearing, which nothing recorded before.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -31,6 +31,39 @@ One behaviour change to know about: `skills-npm.yml` now publishes only from `ma
 
 ---
 
+## [Unreleased] — plain SQL through `sqlQuery` now returns rows on Postgres
+
+`POST /api/v0/environments/{env}/connections/{conn}/sqlQuery` failed for every
+plain statement against a Postgres connection, `SELECT 1` included, with
+`The "string" argument must be of type string or an instance of Buffer or
+ArrayBuffer. Received undefined` and HTTP 502.
+
+Malloy's Postgres dialect finalizes a query by collapsing its result into a
+single JSON column named `row`, and `@malloydata/db-postgres` unwraps that
+column from every row it returns. A statement the caller wrote does not project
+it, so the driver handed back nothing readable. It had never worked through this
+endpoint; before the response caps landed it returned an array of nulls rather
+than failing, so nothing said so.
+
+An eligible statement is now finalized on the way in, for any dialect whose
+connector expects it -- read off `Dialect.hasFinalStage`, which is Postgres
+alone today, rather than hardcoded. No request change: `SELECT 1` returns a row.
+
+What is deliberately NOT finalized, and still runs exactly as sent:
+
+- DDL, `SET`, `EXPLAIN`, `SHOW`, a data-modifying CTE, and Postgres'
+  `SELECT ... INTO`. None can sit in the subquery position the wrapper uses, and
+  they returned no rows to unwrap in the first place, so they already worked.
+- A statement that already projects the column. This is what a `publisher` proxy
+  connection forwards: it reports the REMOTE's dialect, so the Malloy compiler
+  on the far side finalized the statement before it was sent. Finalizing it
+  again would add a level the connector does not strip, and the caller would
+  read nulls off `{"row": {...}}` with no error to notice. The Malloy CLI and
+  the VS Code extensions reach Publisher this way.
+
+A statement in neither group that still produces rows the driver cannot read now
+answers 400 naming the wrapper to apply by hand, instead of the 502 above.
+
 ## [0.2.7] — bound how far the Snowflake driver reads ahead of a slow consumer
 
 The Docker image now installs a small shim in front of the ADBC Snowflake driver
--- a/api-doc.yaml
+++ b/api-doc.yaml
@@ -802,6 +802,16 @@ paths:
         Executes a SQL statement against the specified database connection and returns the results.
         The results include data, metadata, and execution information.
 
+        Some dialects finalize a query by collapsing its result into a single
+        JSON column named `row`, and their connectors unwrap that column on the
+        way back out (Postgres is the only one today). An eligible statement is
+        finalized for them on the way in, so a plain `SELECT 1` returns a row.
+        A statement that cannot be finalized -- DDL, `SET`, `EXPLAIN`, a
+        data-modifying CTE, or one that already projects the column, which is
+        what a `publisher` proxy connection forwards -- is run exactly as sent.
+        When such a statement produces rows the driver cannot read, the response
+        is 400 naming the wrapper to apply by hand.
+
         Rows returned are capped at PUBLISHER_MAX_QUERY_ROWS (default 100,000).
         The cap is forwarded to the connector as a rowLimit on RunSQLOptions;
         queries that return more rows than the cap fail with HTTP 413 rather
@@ -1170,8 +1180,8 @@ paths:
         Executes a SQL statement against the specified database connection,
         resolved in the context of the named package, and returns the results.
 
-        Subject to the same PUBLISHER_MAX_QUERY_ROWS row cap as the
-        environment-level sqlQuery endpoint.
+        Subject to the same PUBLISHER_MAX_QUERY_ROWS row cap, and the same
+        dialect finalization, as the environment-level sqlQuery endpoint.
       parameters:
         - name: environmentName
           in: path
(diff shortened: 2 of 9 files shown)


housejester:
Closing in favour of fixing this in the connector rather than working around it here.

The wrapper this PR adds only helps statements that can sit in a subquery, so `EXPLAIN`, `SHOW`, `INSERT … RETURNING` and data-modifying CTEs stay broken by construction. It also has to tell a proxied Malloy query apart from a caller's own SQL from the statement text alone, and that inspection has to be right every time or a proxied query silently returns nulls.

`@malloydata/db-postgres` is the only connector that post-processes rows at all: `runSQL` and `runSQLStream` unwrap a `row` column from every row. The connector shipped with a raw path alongside it — `runQuery`, calling the same internal helper with `deJSON: false` — which was removed in malloydata/malloy#1506. That commit also moved the connector's own `test()` from the raw path onto the unwrapping one, where it has been running `SELECT 1` and discarding the result since.

An opt-out there covers every case above with no statement inspection, and needs no change to this endpoint's API. #1141 stays open and tracks it.

PR #1144, opened 2026-09-11
docs(apps): design one editable .malloy app to replace notebooks, dashboards and workbooks

Sha-Bang:
A design doc, not code. It proposes one authored artifact — an **app**, a plain `.malloy` file — as the surface you reach for instead of a notebook, a dashboard, or a workbook, and works out the WYSIWYG editor, the pluggable place to save, the agentic "+" gesture, and the embedding contract that go with it.

## Why now

Publisher has four authored artifact types and exactly one is editable: the **workbook**, which is also the only one that isn't Malloy. It persists a bespoke JSON blob, exports to syntactically invalid Malloy, and is unreachable from the shipped Console. The declarative artifacts are the good ones — reviewable, agent-authorable — and nothing can write them back; no endpoint in `api-doc.yaml` persists a model, notebook or dashboard file. So every downstream consumer that wants an editor builds its own.

## The shape

`## app { layout=grid | flow }` makes reading mode a property of the document. `grid` is today's dashboard; `flow` is the narrative document. `# tile` on a top-level `query:` makes a tile. No new file format, no sidecar, no fork.

```malloy
## app { title="Storefront" layout=grid columns=3 }
import "ecommerce.malloy"

##" ### Revenue is up 12% this quarter
##" Narrative that introduces the page rather than any one tile.

#" ### Revenue trend
# tile { colspan=2 }
# line_chart
query: revenue_trend is orders -> by_month
```

That example is verified, not sketched: run through `MalloyTranslator` on the pinned `@malloydata/malloy` 0.0.432 with the `import` resolved, it translates with zero problems.

## Three findings the design rests on

**Prose needs no language change.** A tile's prose is its `#"` doc comment, arriving on the query's `annotations.blockNotes`. Prose belonging to *no* tile is a standalone `##"` block, arriving in `modelAnnotations[url].ownNotes.notes` — **with a source position per note**. So `flow` mode interleaves prose and tiles by author order, and a notebook's "markdown cell, then code cell" becomes "doc comment, then query". This was the open question; the compiler answered it.

**A stored app runs on the door that's already governed.** An app's `import` is a *binding*, not a compile-time import: each tile runs as `run: <body>` against the bound model, which is what `DashboardTile.tsx:119` already does. That path already compiles caller text through `loadRestrictedQuery`, already refuses caller-minted `#(authorize)`, and already carries a host's trusted-attribute injection. An earlier draft proposed a restricted authoring compile over a synthesized prologue — a bespoke trust tier layered on `/compile`, the one door that is construct-unrestricted. Routing tiles to the already-restricted endpoint deletes that work and its risk.

**Tag edits never rewrite a line the editor doesn't own.** Annotation lines fold last-wins per property *and* merge, so appending `## app { columns=9 }` beside an author's `## artifact { … } app { columns=3 }` wins on `columns` and leaves `artifact.title` intact. Ownership is decided from a line's *content* on every reload rather than recorded, which matters because `Tag.toString()` silently drops references — `# a=$ref b=1` serializes to `# -a b = 1` with an empty log. The rule keeps that bug unreachable.

## What it says it cannot do

Each of these is the kind of limit that otherwise gets found late, so they're stated in the doc rather than implied:

- **`explores` curation is not a security boundary for app authors.** Gates hold across the reachable source set (`assertAuthorizedForAllSources` runs unconditionally), but `assertQueryBoundaryCompiled` checks only the run target and never enumerates joined sources — so a tile can join an unexported source. That's the position Publisher already takes for `/compile` ("discovery curation, not access control"), but a store hands the capability to a narrower author than a modeler. A deployment relying on curation needs a join walk, and the doc puts it in P1.
- **An app cannot declare its own givens.** Controls come from the bound model.
- **A `.malloynb` whose cells build on each other does not convert, permanently.** The per-cell `extendModel` chain has no representation in a single document and no phase adds one. The four artifact types become three, not one.
- **Prose rendering is unbuilt** — free in the language, real work in the renderer.
- **The framing policy is a prerequisite for the first edit-capable route**, not a later nicety, because a builder's frame holds a write path. Console routes set no framing header at all today.

## Reversing a ratified principle

`malloyyo-dashboards-design.md` is normative on *"one engine, two document types"* and says merging the formats is a non-goal "now and later". This reverses that, and writes the reversal into that document rather than leaving two docs disagreeing. The v1 reasoning still holds on its own terms; what changed is that four capabilities — editor, storage, patch protocol, embedding — would each have to be built twice, and that cost wasn't on the table when the principle was written.

The reversal is claimed **only on the reading-mode axis**. §13 is explicit that it does not extend to notebooks used as incremental modeling scratch.

## Also in this PR

- The claim that Malloyyo's composite has no per-tile layout is **corrected** — false since their 0.2.19, where the client-side combiner lifts `# colspan`/`# break` onto a tile's card. Note `malloydata/malloyyo#160` does *not* track this; it's about `dashboard_columns` spelling and repeats the stale claim, so it needs amending rather than citing.
- The adjacent "one trust tier" sentence is **left alone deliberately** — Malloyyo removed the Malloy renderer from custom components but kept a sandboxed custom-component tier, so that divergence is real.
- `choosing-a-surface.md` gains a forward note; `docs/README.md` gains an index row.

## Review

Worth your attention in this order: **§4** (the store dialect and the curation gap — the security argument), **§6** (the edit algebra, and whether content-based line ownership holds), and **§13** (whether the supersession argument earns the reversal). §14's phasing and §15's open questions are where I'd most expect disagreement.

Prior art is covered in §12, including Malloyyo's recorded rejection of storing authored artifacts outside a repo — the doc answers that objection directly rather than walking past it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -82,6 +82,7 @@ there for the primitive, then follow the application you need.
 | [agent-skills/](agent-skills/)                                 | Author or contribute the bundled agent skills.                                                |
 | [../SECURITY.md](../SECURITY.md)                               | Report a security vulnerability, or check what's in scope.                                    |
 | [malloyyo-dashboards-design.md](malloyyo-dashboards-design.md) | _Design doc:_ the grammar and architecture behind native `dashboards/*.malloy` support.       |
+| [malloy-apps-v2.md](malloy-apps-v2.md)                         | _Design doc:_ one editable `.malloy` **app** replacing notebooks, dashboards and workbooks.   |
 
 ## Full public docs
 
--- a/docs/choosing-a-surface.md
+++ b/docs/choosing-a-surface.md
@@ -9,6 +9,13 @@ SPDX-License-Identifier: MIT
 > (`.malloynb`), **dashboards** (`dashboards/*.malloy`), and **HTML data apps** (`public/`), with
 > the pros, cons, and decision rules for each.
 
+> **This page describes what you can author today, and today has three surfaces.**
+> [malloy-apps-v2.md](malloy-apps-v2.md) proposes reducing that to two, by merging notebooks and
+> dashboards into one `.malloy` **app** where a `## app { layout=grid | flow }` tag decides the
+> reading mode — so the notebook-or-dashboard question below becomes one tag rather than two
+> formats. It is a design, not shipped code: keep choosing between the three surfaces here until it
+> lands, and read that document before authoring new guidance on this axis.
+
 All three are artifacts that live _inside a package_, ship with the model, and run on the same
 engine: the same governed query endpoints, the same [givens](givens.md), the same
 `@malloydata/render` renderer, the same query caps. Ask any of them the same question and the numbers
--- a/docs/malloy-apps-v2.md
+++ b/docs/malloy-apps-v2.md
@@ -1,0 +1,651 @@
+<!--
+Copyright (c) Credible Data Inc.
+SPDX-License-Identifier: MIT
+-->
+
+# Design: Malloy apps (v2)
+
+**Status: a design, not shipped code.** Nothing below describes behavior you can run today. Where it
+says a mechanism exists, there is a citation; everything else is a proposal. Written September 2026.
+
+One authored artifact — an **app** — becomes the surface you reach for instead of a notebook, a
+dashboard, or a workbook, and it gains a WYSIWYG editor, a pluggable place to save, an agent that can
+add a tile, and an embedding story. The app is a plain `.malloy` file. There is no new file format, no
+sidecar, and no fork.
+
+It does not absorb every existing artifact. `.malloynb` survives for the one thing an app cannot
+express — a document whose cells build on each other — and HTML data apps survive for pages that ship
+code. §13 draws both lines.
+
+**Related:** [choosing-a-surface.md](choosing-a-surface.md) (whose three-surface taxonomy this
+narrows to an app-or-code decision), [malloyyo-dashboards-design.md](malloyyo-dashboards-design.md) (whose "one engine,
+two document types" principle this reverses, §13), [security-posture.md](security-posture.md) (the
+trust boundary §5 and §9 argue against), [givens.md](givens.md), [dashboards.md](dashboards.md).
+
+## 1. Why
+
+Publisher has **four** authored artifact types, and only one of them is editable:
+
+| Artifact | Format | Editable | Lives in |
+| --- | --- | --- | --- |
+| Notebook | `.malloynb`, `>>>` delimiters | no | a package |
+| Dashboard | `dashboards/*.malloy` + `# artifact` | no | a package |
+| HTML data app | `public/*.html` | by editing files | a package |
+| **Workbook** | `JSON.stringify(WorkbookData)` | **yes** | a `WorkbookStorage` |
+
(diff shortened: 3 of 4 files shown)


Sha-Bang:
Closing this unmerged. Kyle's dashboard builder shipped while this sat, and on the load-bearing choices the code decides the opposite way — so rewriting the doc to agree would leave it mostly restating `docs/dashboard-builder-plan.md` (#1158), which already covers both halves of this: the reader/writer/round-trip half, and the notebook-format half in its §7.

Where the two disagree, the code wins:

- §14 has a model-level `## artifact` tag as a page view spelled as a tag, with the array form opening read-only. `readDocument.ts` refuses a file *without* that tag, and the array form is the editable one.
- §4 has an app carrying given values, never declarations. `planGivens` in `spliceDocument.ts` writes `given:` declarations and inserts `##! experimental.givens`.
- §3 has `WorkbookStorage` becoming `AppStore` with an etag. #1146 landed it as `DocumentStorage`, `saveDocument(locator, content): Promise<void>` — no etag, no version, no concurrency control.
- §3/§8 put the store outside the package with promotion as the bridge in. `DashboardEditor` treats the package dashboard as a read-only origin and "Export" hands the file back: copy-and-export, no server write path.
- §6 wants a line owned only when every property on it is managed. The writer decides on the first key and replaces the line whole. Known cost, tested, accepted.
- §7 runs the page as one query. `preview.ts` runs each tile's view separately with a `where:` refinement.
- §13 makes a notebook the same file under `## app { layout=flow }`. Kyle's §7 keeps `.malloynb` read-only for VS Code interop and puts the authored notebook in a new format, gated on a markdown-annotation grammar proposal to Malloyyo.

Four things here aren't covered anywhere else, and each matters *more* now that an edit-capable route ships, not less. The branch stays (`nathan/malloy-apps-v2`, head 7564837) as the only copy until they land somewhere:

1. **The gate/curation analysis (§4).** `#(authorize)` is evaluated at the entry point only and a `join_*` reaches around it at any depth; `assertQueryBoundaryCompiled` never enumerates joined sources. So `queryableSources: "declared"` is discovery curation, not access control. The plan doc has no security section, and the writer now puts `given:` declarations into the file it writes. This is the piece with a real reader — it belongs as a gap in the plan doc's G-list, retargeted from "the store dialect" to "a dashboard the builder writes".
2. **Promotion (§8).** Parked until the server write path is back in scope, but the analysis is done: run every tile before and after and compare result hashes; touch `explores` only when it already exists, because adding it unconditionally turns curation on and 404s every pre-existing target in that package; refuse a `given:` that shadows the bound model's, because binding is per declaration so a re-declaration shadows and breaks tenancy; refuse a traversing import specifier.
3. **Framing (§9).** Console routes set no framing header, `publisher:*` validates `event.source` but never `event.origin`, and the embed token is read by no server code. This was written as a prerequisite for the first edit-capable route. `/dashboards/<slug>/edit` is that route.
4. **`DocumentStorage` has no concurrency control.** `BrowserDocumentStorage.saveDocument` is an unconditional write; two tabs on one document silently last-write-wins.

Not carried forward: the `nest:`-page-view thesis in §14. That argument belongs to the shared-grammar discussion §8 of the plan doc already sequences, not to a Publisher design doc.

PR #1159, opened 2026-09-14
fix(boundary): admit a derivation that composes through a named query

sagarswamirao:
## What

The query boundary admits an ad-hoc run target when the submitted text declares it as
derived from a curated source. That walk read `buildSourceAliasMap`, which matches only
`source: NAME is BASE`, so it dead-ended on the standard pre-aggregate idiom that composes
through a named query:

```malloy
query:  agg     is <curated> -> { ... }
source: blended is agg extend { ... }
run:    blended -> { ... }
```

`blended` resolves to `agg`; `agg` is neither curated nor a `source:` alias; the walk fails
closed and the caller gets a bare `404 Query target is not queryable` on a query that only
ever reads a source they may plainly read. `/compile` is exempt from the boundary by design,
so the same text compiles clean a moment earlier -- the 404 contradicts the immediately
preceding success and leaves nothing to act on. Observed in production against a curated
forecast source.

## Why not just widen the pattern

Adding `query:` to `buildSourceAliasMap` in place is not the fix. That map reads raw text,
so it matches declarations written inside comments and string literals, and it is
last-declaration-wins, so a later declaration replaces a name's real base. Those properties
are tolerable on a DENIAL path and are not tolerable on an ADMISSION path: widening the map
in place adds edges a caller controls to the walk that decides what they may read.

## What this does instead

Read the pair the authorize gate already uses:
`buildDerivationBaseMap(stripMalloyCommentsAndLiterals(query))`. It follows `query:` hops,
blanks comments and string-literal bodies, skips backtick-quoted spans whole, reads Unicode
identifiers, parameter lists and parenthesised bases, and keeps a SET of bases per name
rather than the last one.

The quantifier is inverted from the authorize gate's use of the same map, and that inversion
is what makes the wider map safe here:

| | Extra edge means | Rule |
|---|---|---|
| Authorize gate | widens DENIAL | denied if ANY branch reaches a gated source |
| Query boundary (this) | widens ADMISSION | admitted only if EVERY declared base proves out |

Under every-base, an additional declared edge can only add an obligation, never discharge
one. For the ordinary one-base-per-name chain -- everything that compiles -- this returns
exactly the previous walk's answer. Cycles and over-long chains fail closed, as before.

`buildSourceAliasMap` stays: `resolveFilterSource` still walks it.

## Tests

- Derivation through a `query:` hop is admitted over a curated source and denied over a
  hidden one.
- A second declaration for a name adds an obligation rather than replacing the first, in
  both a form Malloy rejects at compile and a form it accepts -- the latter is what pins the
  every-base quantifier itself. Inverting that one rule to any-base leaves the rest of the
  suite green, so it is asserted directly.
- A derivation cycle fails closed.
- `query_text` unit coverage for the base map keeping every declared base per name.
- Five `filter_bypass` vectors covering the `resolveFilterSource` change below, added to the
  existing enforced-vector table so each asserts both the missing-filter rejection and the
  scoped result once filters are supplied. All five fail without that change.

Full server unit gate (`bun test src` in `packages/server`): **3752 pass, 3 skip, 0 fail**.
The three are environment-gated skips (BigQuery credentials, a win32 guard), not failures.

## Also fixed here: filter inheritance read raw text too

`resolveFilterSource` resolves which `#(filter)`-protected source's filters apply to an
ad-hoc query, and it read the caller's raw text through the same unhardened pair. Filters are
injected purely from that walk with no post-compile backstop, so a name the walk fails to
reach is served unfiltered and silently.

Four shapes reached that state on a query that really does read a protected source: a comment
between `is` and the base (which erases the edge the compiler still reads around), a
declaration forged in a string literal (which last-wins lets REPLACE the real base), and a
forged `run:` in either a literal or a comment (which re-points run-target extraction). This
scans stripped text before both reads, which closes all four.

It does NOT close a derivation composing through a named `query:`, which this walk still
cannot follow: it resolves a single source name, and a set-valued walk has to decide which
protected source's filters apply to a name with several bases. Tracked separately.

## Not addressed here

- Filter inheritance through a named `query:` hop, as above -- needs the set-valued redesign.
  Tracked in #1160.
- `derivesFromCurated` terminates on curated sources only, so a source composed over a
  model-exported named query is still denied while a caller writing the equivalent hop is
  admitted. Same "compiles clean, then a bare 404" shape this PR exists to fix. Tracked in
  #1161.
- Boundary denials still emit no log or metric, so a cycle or budget-exceeded denial is
  indistinguishable from an ordinary non-curated 404. Given this fix exists because a bare
  404 explained nothing in production, a debug log naming the denial reason is worth a
  follow-up.


housejester (approved):
Approving — verified against `ec13c319f`.

Fixing `resolveFilterSource` rather than documenting it was the right call, and the tests are the part worth singling out. I reverted only the `model.ts` hunk and confirmed the five new vectors produce exactly 10 failures (5 × 2 assertions), so each genuinely fails on the old behaviour rather than passing for an unrelated reason. Full unit gate on this head — `bun test --timeout 100000 src` → **3752 pass / 3 skip / 0 fail**, ten over the pre-fix baseline, no regression. Re-ran an independent probe set as well: the five closed mechanisms are refused, and the remaining `query:`-hop case behaves exactly as #1160 describes.

The every-base quantifier is properly pinned too — flipping `.every` to `.some` in `derivesFromCurated` fails exactly one assertion, (b2). Good that (b)'s comment was trimmed to credit the compile error rather than claim a verdict it doesn't produce.

Splitting #1160 and #1161 out instead of widening this PR is right; both change what the boundary admits and want their own tests.

One non-blocking thought for later: `buildSourceAliasMap`'s new closing invariant — "only as sound as its caller's input" — is stated but nothing enforces it, so a future second caller passing raw text silently reopens all four mechanisms. It has exactly one caller and that caller already strips, so stripping inside the function would be idempotent and free. Not for this PR.

<sub>Review assisted by Claude Code (session "Solemn Capybara").</sub>

Issue #1160, opened 2026-09-14
Filter inheritance is dropped when a derivation composes through a named query


sagarswamirao:
`Model.resolveFilterSource` walks a derivation chain to decide which `#(filter)`-protected
source's filters apply to an ad-hoc query. #1159 moved it onto
`stripMalloyCommentsAndLiterals`, which closed the shapes where a comment erased a real
derivation edge or a string literal forged one. One shape is left, and it is the same
`query:`-hop blind spot that #1159 fixed on the query boundary.

`buildSourceAliasMap` reads `source:` declarations only, so a chain that composes through a
named query dead-ends:

```malloy
query:  agg is <protected> -> { group_by: id }
source: a   is agg extend {}
run:    a -> { group_by: id }
```

`a` resolves to `agg`; `agg` is not a `source:` alias; the walk ends with no protected source,
so no filter clause is injected. Filters are injected purely from this walk -- there is no
post-compile backstop on that path -- so the query runs unfiltered rather than being refused.

This is not the boundary's `derivesFromCurated` problem and is not fixed by the same change.
That walk answers a yes/no question, so it could adopt the set-valued
`buildDerivationBaseMap` with an every-base quantifier. This one has to return WHICH source's
filters apply, and a set-valued map admits a name with several declared bases. The open
question is what to do when more than one resolves to a protected source with different
required filters: union the requirements, refuse the query, or something else. Picking that
rule is the substance of the fix.

Worth deciding alongside: whether `#(filter)`'s deprecation in favour of `given:` changes the
priority here. `given:` carries its own `where:` and is unaffected; `#(filter)` is still the
REST `filterParams` mechanism and still documented as enforced server-side
(`docs/givens.md`), so it is a supported control today.

Found during review of #1159.

PR #1162, opened 2026-09-14
refactor(query_text): strip inside buildSourceAliasMap rather than requiring callers to

sagarswamirao:
Follow-up to #1159, from its review. Raised there as a non-blocking note and deliberately
left out of that PR, since it changes a shared helper's contract rather than the boundary
behaviour #1159 was fixing.

## What

`buildSourceAliasMap` strips its own input instead of trusting the caller to have done it:

```ts
const text = stripMalloyCommentsAndLiterals(query);
...
while ((match = declRe.exec(text)) !== null) {
```

## Why

#1159 fixed `resolveFilterSource` by stripping comments and string literals before it walks a
derivation chain, which closed four ways caller text could misread as syntax: a comment
between `is` and the base erases a real edge, a declaration forged in a literal replaces one
(the map is last-declaration-wins), and a forged `run:` in either re-points the run target.

That fix left the guarantee stated as a precondition -- the docstring asked callers to pass
stripped text. A documented precondition holds only until the next caller, and this one fails
silently in the unsafe direction: the map decides which `#(filter)`-protected source's filters
an ad-hoc query inherits, filters are injected purely from that walk, and there is no
post-compile backstop. A second caller passing raw text would reopen all four mechanisms with
nothing to catch it.

Moving the strip inside makes the property the function's rather than the caller's.

## Cost

None in practice. `buildSourceAliasMap` has exactly one production caller
(`Model.resolveFilterSource`), which already strips, and
`stripMalloyCommentsAndLiterals` blanks to spaces, so it is idempotent -- verified against
backtick spans, escaped quotes, double-quoted literals and unterminated comments. That caller
pays one extra linear scan and nothing else, and it keeps its own strip because
`extractRunTargetSourceName` still needs stripped text and does not strip internally.

## Tests

Two cases in `query_text.spec.ts`, both asserted against RAW text on purpose, since the point
is that the guarantee no longer depends on the caller:

- a forged declaration inside a literal, and both comment forms between `is` and the base,
  each resolving to the real base;
- an idempotency assertion, because that is what makes the internal strip free.

Both fail without the change. Full server unit gate: **3754 pass, 3 skip, 0 fail** (the three
are environment-gated skips -- BigQuery credentials and a win32 guard -- not failures).

## Not addressed here

This does not close filter inheritance through a named `query:` hop, which is #1160:
`buildSourceAliasMap` reads `source:` declarations only, and that walk has to resolve a single
source name, so following `query:` hops needs a rule for which protected source's filters
apply when a name has several bases.


housejester (approved):
Approving. Moving the strip inside the function is the right shape, and both new cases fail on the old code: the literal case wins under last-declaration-wins, and in the comment case `\s+is\s+(\w+)` runs into `--`. The notes below are non-blocking.

**1. The docstring now contradicts itself** (`query_text.ts:48-55`). It still says the function "reads RAW text. It therefore misreads caller text in BOTH directions, and neither is theoretical". The new paragraph at `:62-67` says it strips its own input, which closes both of those misreads. Suggest rewriting `:48-55` to describe what's actually left: it reads only `source:` declarations, it's last-wins, and its identifier pattern is narrow (next note).

**2. Pre-existing, same family: the regex still drops edges the compiler reads.** `:73` misses three forms that `buildDerivationBaseMap` (`:199-207`) already handles:

```
source: mine is (protected extend {})  / run: mine  → alias map empty
source: mine(p::string) is protected   / run: mine  → alias map empty
source: café is protected              / run: café  → run target "caf", alias map empty
```

In each case `resolveFilterSource` returns `undefined`, so no `#(filter)` clause is injected, including a `required` one. This isn't an access bypass, since `bypass_filters` is already caller-controlled. It's a silent correctness gap: a caller who writes `is (x)` gets unfiltered rows with no error. Porting the `ident` pattern and the optional `\(?` and parameter list into this regex, while keeping it single-valued and last-wins, would finish the "the guarantee is the function's" framing. Fine here or as a follow-up next to #1160.

**3. The idempotency test doesn't pin idempotency** (`query_text.spec.ts:112-120`). `map(strip(raw)) == map(raw)` holds because both sides now strip, and it only exercises one single-quoted literal. The "no cost" argument depends on idempotency across backtick spans, escaped quotes, double-quoted literals and unterminated comments. A direct `expect(strip(strip(s))).toBe(strip(s))` over those shapes would pin it. I worked through them and believe it holds today.

**4. Consistency nit.** `buildDerivationBaseMap` still has the "expects text already stripped" precondition (`:195`). The stakes are lower there because its callers fail closed, but the same move would cost nothing there too, for the same idempotency reason.

— Mellow Muskrat (Claude Code review session)

sagarswamirao:
Thanks for the approval, and for the three cases in note 2 -- I checked each against both patterns before porting and they miss exactly as you described:

```
"source: mine is (protected extend {})"   narrow: MISS  | wide: mine -> protected
"source: mine(p::string) is protected"    narrow: MISS  | wide: mine -> protected
"source: café is protected"               narrow: MISS  | wide: café -> protected
"source: plain is protected"              narrow: OK    | wide: OK
```

All four notes are addressed in 8fef189c. Since you had already approved, here is what changed rather than making you re-read the diff.

**1. The docstring.** Rewritten to describe what is left rather than what the strip closed: `source:`-only, last-declaration-wins, reads stripped text. The paragraph claiming it reads raw text and carries both misreads is gone.

**2. The regex.** Ported `ident` and the two optional groups from `buildDerivationBaseMap`, keeping this one `source:`-only and single-valued since `resolveFilterSource` needs exactly one base. Load-bearing rather than decorative: reverting to the narrow pattern fails the new case and nothing else.

**3. Idempotency.** You were right that the old test could not fail -- both sides stripped. Replaced with a direct `strip(strip(s)) === strip(s)` over the shapes the argument depends on: escaped and double quotes, backtick spans, and unterminated comments and literals. Your reading of it holds.

**4. `buildDerivationBaseMap`.** Now strips its own input too. Both its callers already stripped, so nothing changes for them.

One thing worth surfacing from note 4, because it is the sort of test that reads as a contract and is not. `"links a declaration a comment split, once the comment is stripped"` asserted `buildDerivationBaseMap(raw).get("mine")` is `undefined` -- pinning the precondition as behaviour. Making the function strip its own input turned that red. The test was documenting the gap rather than a guarantee, so it now asserts both spellings agree, which is the idempotence the change rests on. Flagging in case you would rather it had stayed as a record of the old shape.

4032 pass / 0 fail on the unit suite, 350 / 0 on integration.

PR #1163, opened 2026-09-15
feat(authorize)!: #(authorize) is the lock and answers 403, #(access_filter) is the row filter

Sha-Bang:
Two annotations, one question each, and the name tells you which.

| annotation | the question | a denial is |
| --- | --- | --- |
| `#(authorize)` | may this caller reach this source at all? | **403** |
| `#(access_filter)` | which rows may they see, once they may? | **200**, with their rows |

Before this, `#(authorize)` was the row filter and every denial was a 200 with
zero rows. That is defensible for a filter and wrong for a lock: a caller the
gate excludes running `-> { aggregate: total is sum(salary) }` got back a row
of `NULL`, and a `count()` of `0` — a fabricated answer about a resource they
were refused. The plain word now means the thing that answers "may you", the
row filter takes the name the surrounding world already uses (Looker and Omni
both say `access_filter`; Spring Security splits `@PreAuthorize` from
`@PreFilter`), and the lock answers 403.

Also in here, because the rename is what made them reachable to state: the
body grammar is a narrow one publisher parses itself rather than an arbitrary
Malloy boolean, repeated notes on a route mean a conjunction, and `#(partition)`
is gone.

**The name declares the scope, and nothing is inferred from the body.**
`#(authorize)` takes `'literal' <op> $GIVEN` and the `true`/`false` sentinels;
`#(access_filter)` takes `field_path <op> $GIVEN`. A body on the wrong
annotation is refused at load naming the other one, both ways. The second
direction is new and is a fix, not tidiness: a source-shaped body on the row
route used to be legal and grafted as a constant predicate, which answered a
refused caller 200-with-zero-rows — the same fabricated answer, on the route
nobody was looking at.

## Migrating, and the two halves fail differently

A row-shaped body under `#(authorize)` is **refused at load** and cannot be
missed. A source-shaped one keeps loading and **silently starts answering
403** — that is the half that needs a grep before upgrading, and anything
keying its own logic on the status has to move to the message.

**Rolling the binary back is unsafe once a model uses `#(access_filter)`.** No
earlier release knows the route, and an unknown `#(...)` loads clean and serves
every row. Roll models back first.

## The 403 is its own commit

`4f5ea74e` is reviewable and revertible alone. It restores a decision this
codebase had until #1045 deleted it, but not the mechanism: the old one was a
synthetic one-row DuckDB probe. This reads the **compiled condition** —
`resolveGateShape` already resolves the graft target, lifts the condition
through the entry model's materializer, expands given ids to names and
memoizes, so the lock is a third arm of that union and `decideLock` is a pure
positive allowlist over the IR. Evaluating the annotation text instead would
have meant a second parser and a second string-literal decoder for text Malloy
has already decoded and type-checked.

Four decisions worth a reviewer's attention, all in that commit:

- **A throw is a deny, never a 500.** A `null` given binding reaches TS as
  `(null).includes(...)`. Every path through `decideLock` is caught.
- **Denials book to a new counter**, `publisher_authorize_lock_total`, not to
  `publisher_authorize_row_level_total{decision="denied_by_gate"}` — that one
  is documented as the fail-closed "could not apply" case operators alert on,
  and an ordinary 403 is not that.
- **Exact match, not warehouse collation.** A source-shaped body used to be
  compared by the warehouse; MySQL's default is case-insensitive, so
  `'Finance'` against `["finance"]` starts denying. Fail-closed, and pinned.
- **`/compile` truth-evaluates the lock.** It decides filters by presence, not
  truth, which left `includeSql` returning a locked source's SQL. If "may you
  reach this source" is what the word means, reading its SQL is reaching it.
  The cost is real: an author outside the group can no longer compile-check a
  locked source. Filters keep today's presence-only behavior.

Two properties that hold by construction and are tested anyway: a hidden
locked source answers **404**, not 403 (the lock throws from the same two
sites `rejected` already threw from, inside `denyHiddenAsNotQueryable`), and
storage/pre-aggregation routing is blocked on **any** collected gate
regardless of route, so a lock-only source never serves from frozen rows.

## Wire

`Source.authorize` is the lock, `Source.accessFilter` is the filter.
`sourceAuthorize` never shipped in a release and is deleted. MCP's
`get_context` reports both, each under the field named for its own annotation.

## Testing

`bun test src` 4059 pass / 3 skip / 0 fail; `bun test tests` 415 pass / 4 fail,
the four being the pre-existing watch-mode and GCS-download timeouts. Every
test that carries the risk goes through a real `Model.create` and real DuckDB
and asserts the **served response**, not a parse — including the aggregate
case, which is the one this change exists for. Each of the mutations that
would undo the change was run and turned a named test red.


Sha-Bang:
@claude review

jswir:
Reviewed the split end to end, adversarially on the security side. **No authorization bypass found.** The lock is a genuine fail-closed allowlist, every query-producing surface funnels through one `decideLock` path, hidden sources still 404 before the lock is consulted, and materialization refuses gated sources independently. I tried to smuggle `or`/`not` past the grammar through parens and unterminated strings, and to reach an admit through unknown IR node kinds, wrong-arity values, and thrown exceptions — all deny.

Findings split cleanly into four code changes and one documentation sweep.

## Code (4)

**1. Filter controls lose the givens they need to load their options.** When a dashboard filter control populates its dropdown, Publisher tells it which runtime parameters that dropdown query must send — the ones the source's own `where:` reads, plus the ones its gate reads. The gate half is fetched through a callback that `package_load_worker.ts:753`, `:989` and `model.ts:3550` still point at `?.authorize`, which after the split holds lock-route gates only. For a source migrated to `#(access_filter)`, it comes back empty, the dropdown query goes out without the parameter, and the control shows "Options unavailable" — the symptom `bcac565e` (#1152) fixed. `getAccessFilter()` is never wired into any suggest path, and nothing tests suggest over an `#(access_filter)`-gated source.

The comment above it calls this "a known, accepted gap" but names `#(authorize)` as the affected route — the opposite of the real gap. Flagging that specifically: it's why this reads as a doc problem and isn't. Fix the sentence alone and the behavior stays broken under a now-accurate comment.

**2. The admit-path test can't catch a regression of this PR's own bug.** `authorize_integration.spec.ts:818-826` asserts only `expect(result.data).toBeDefined()` against a fixture that defines `measure: c is count()`. Regress the split so an admitted caller also gets `WHERE false` grafted and the test still passes. Asserting the count is one line.

**3. `decideLock` returns one value for three different outcomes.** `authorize_lock.ts:33-46` returns `"admit" | "deny"`, so a policy refusal, an unresolved given, and a caught exception are indistinguishable at the return site; `model.ts:220-229` records all of them as `denied_by_lock`, the label the metric's own doc (`authorize_metrics.ts:275`) calls "routine." `denied_unresolvable` — documented as the only label signalling something is wrong — is reachable only from the load-time classification branch at `model.ts:232`, never from a per-request evaluation failure. An operator can't alert on "the gate is broken" vs "the gate is working." The new `publisher_authorize_lock_total` counter also has zero spec references.

**4. Five operations that now return 403 don't declare it.** `"403"` appears three times in `api-doc.yaml`, all on the `frozenConfig` admin routes. `execute-query-model` (1762), `execute-notebook-cell` (2120), `post-querydata` (810) and `post-querydata-in-package` (1178) declare 401 but not 403; `compile-model-source` (1822) declares 200/400/404/500/503. All five can throw `AccessDeniedError`, and this is the contract the generated clients are built from.

## Documentation sweep (10)

All one root cause: a route-name substitution applied to prose without a semantic check, leaving text that says the opposite of the code. Worth one pass grepping every prose mention of both route names.

Three of these actively teach the old behavior back:

- `docs/discovery-and-access.md:63-78` — "`#(access_filter)` is the identity gate", "to restrict *who* can query, gate the source with `#(access_filter)`". An author who follows this gets 200-with-zero-rows where they wanted a 403. Lines 10-11 of the same file are correct.
- `docs/authorize-bypass-deployment.md:8` and `security-posture.md:46,51-52` — documented as skipping "`#(access_filter)` gate evaluation", but `assertAuthorized` returns at `model.ts:1282-1283` before any gate walk, so the header short-circuits the lock too. An operator would wrongly conclude hard locks survive the bypass.
- `authorize.ts:619-631` — "Grafting and probing treat every group identically regardless of route." A maintainer who believes this grafts a lock as a `where:` and reinstates the fabricated response.

The rest are inverted or incomplete:

- `source_extraction.ts:87-113` — both field docs claim the same route; the file uses `ACCESS_FILTER_ROUTE` 6x in code and names `access_filter` 0x in prose.
- `gate_classification.ts:590` and `:643` — "every gate is a row filter now", the sentence `1f7c4591` removed from `model.ts` for being untrue after the split, still in the function that implements the split. Two more copies in `authorize_integration.spec.ts:3958` and `row_level_authorize.integration.spec.ts:1693`.
- `authorize.ts:29-31` — "a note is a gate iff Malloy routes it to `authorize` or ... to `authorize`"; `access_filter` is never named.
- `authorize_metrics.ts:209` — the row-level counter's shipped description says `#(authorize)`. This is what an SRE reads in a metrics catalog.
- `gate_dimension.ts:249,264,276,283` — four messages hardcode "row-level" on a validator that runs for both routes, pointing an author at the wrong annotation during a load failure.
- `AGENTS.md:26` — attributes the lock's job to `#(access_filter)`.
- `docs/ai-agents.md:93` — says `/compile` rejects only `#(access_filter)`; the guard is one regex matching both. `docs/authorize.md:389` has it right.
- `materialization_eligibility.ts:33-37,123-129` — an `#(access_filter)`-only source is refused with "protected by an authorize gate".

jswir (approved):
approving - findings in comment should be fixed before merging please

PR #1166, opened 2026-09-15
feat(skills): validate a model's definitions once, instead of every answer

jswir:
Follow-up to #1114. Two commits, one thesis: **a run should say what its pass rate rests on, and a check should never claim more than it measured.**

## The problem

The loop's standing rule is that a golden must never be derived through the model under test, because a model bug would certify its own golden. The rule protects something real -- in the ecommerce sample, `units_sold` excludes cancelled orders while `total_sales` and `total_gross_margin` include them, though the model's own `status` doc says cancelled lines "are not sales". A key computed with the model's own measures would have baked that in and scored it correct forever.

It also does not scale. On a model with real transformation, deriving every key independently means reimplementing the model: two implementations, no reason to trust either, and the eval task becomes *rebuild the model to verify the model*.

## The change

Validate **definitions** once instead of **answers** every time. A definition is checked against the layer directly beneath it; every case depending on it inherits that.

The rule that replaces "never through the model":

> A key must be independent of **the definitions the question tests**, not of the whole model.

And what makes it sound:

> A golden is trustworthy if it was derived independently, **or** if every definition it tests has itself been validated.

## A claim I made and then had to retract

An earlier draft argued this was affordable because semantic models are "wide, not deep" -- one or two hops from a raw column. **That is false, and measuring it is what caught it:**

| Model | Definitions | within 2 hops | deepest chain |
|---|---|---|---|
| ecommerce (sample) | 32 | 90.6% | 3 |
| storefront (sample) | 19 | 94.7% | 3 |
| a 492-definition benchmark | 492 | 100% | 2 |
| **a real 548-definition customer model** | **548** | **42.5%** | **12** |

The design survives for a better reason. On that customer model the twenty deepest measures close over **812** definition-checks with no reuse and **116** with a ledger -- an 86% saving. Reuse pays off *more* as chains deepen, because deep chains share their lower layers. The argument was never about shallowness; it is about definitions being **shared**, which is what depth produces.

## What it refuses to claim

This is the part worth reviewing hardest.

- **Anything reaching through a join records `unchecked`, never `agrees`.** A within-model check is blind to fanout: a join duplicating rows inflates the measure and the control expression equally, and the comparison stays green on a genuinely broken join. 4 of 28 on ecommerce are held back this way.
- Same for a definition spanning more than one line (only the first line was read), a failed query, no rows, and an all-null slice.
- **Building the ledger without a server exits 3, not 0.** It validated nothing.
- **Without `--definitions` there is no EVIDENCE block at all**, rather than an empty reassuring one. Absent is a different fact from checked.

## Two bugs a live run caught that reasoning did not

**Join detection matched any `word.word`**, so `total_sales is sale_price.sum()` was held back as "crosses a join" -- a method call on a column, not a traversal -- silently blocking the measure twelve of the set's cases depend on. Now detected by join *name*.

**The evidence classifier read `verifiedBy`,** which is free text. Against the real ecommerce set it reported 5 independent / 36 unchecked, when all 34 of those goldens read "authored and re-derived against ecommerce-truth" and the set is fully truth-derived. It now reads the structured `golden.verification` through the existing `has_second_derivation`, and reports 49 independent -- the true answer. A well-founded set reading as unvalidated is the same over-claim as an unfounded one reading as validated, pointed the other way.

## Supporting changes

`parse_definitions()` source-qualifies what `model_definitions()` flattened -- the flat `dict[name]` resolves a same-named field in two sources to whichever file was read first, so a ledger keyed on `kind:source:name` cannot use it. `model_definitions()` is now a thin flat view with its contract unchanged, **including its non-recursive glob**: `model_text()` next door does recurse, which reads like an oversight, but changing it here would quietly move `stale_rubric_claims`, so recursion is opt-in and the inconsistency stays visible.

`entity_id()` is extracted in `mcp_payload.py`, already the only place ids are built, so the ledger builds them through the same function rather than a second copy of the format.

Which definitions a case tests comes from `expectedEntities`, which already uses the ledger's id format -- no new hand-maintained field, since a wrong entity id is what cost a real set two days.

## Verification

- Against a live Publisher on ecommerce: **28 definitions, 15 agree, 13 unchecked**, 4 of those needing a raw check.
- **Negative control**, because a check that cannot fail is not a check: corrupting the served `total_sales` to `sale_price.sum() * 2` reports `disagrees` and exits 1. `total_sales_2022` and `total_sales_2021` correctly stay `agrees` -- defined in terms of `total_sales`, both sides move together, and the defect localises to the definition that owns it rather than reporting three failures.
- `exprSha` is a Merkle chain over a definition and its dependencies, so an edit beneath a measure invalidates it even when its own text is byte-identical.
- **565 python tests** pass, up from 516 (45 new). 178 skills tests and 18 bundle specs; `skills_bundle.json` regenerated.

## Deliberately not here

Raw checks for the join-crossing measures, and relaxing `verify_goldens`' gate to the composition rule -- the step that would actually let a set run without a truth package. Both wait until this shape is reviewed. **So the motivating bug above is not yet caught by anything**: catching it needs a raw check, and the three measures involved are recorded `unchecked`, not validated.

The six eval skills are shared, so this owes `ms2data/agent-skills` the same files once merged.


jswir:
Superseded by #1173, which carries both of these commits unchanged (definition ledger + evidence line) and finishes the leg: authored controls for the join-crossing measures, and the composition rule at the gate. The retraction of the shallow-models claim and of the mechanical-raw-check claim is in that PR's body.

PR #1167, opened 2026-09-15
fix(skills): a contaminated attempt must not reach the printed score

jswir:
Follow-up to #1114, found while running the loop against a customer set. Two
bugs, one of which makes the other dangerous.

## The one that matters

A 33-case run in which **every** attempt was flagged contaminated printed:

```
passed        12 of 21 decided (57%)
```

while every score event in its own `events.jsonl` carried `verdict: null`. The
contamination check fired correctly, the ledger recorded it honestly, and the
only line a human reads disagreed with both.

The nulling lands on `sc`, the copy bound for the ledger, while the aggregates
twenty lines below read `verdicts`:

```python
sc = {k: x for k, x in v.items() if k not in (...)}
tainted = bool(att.get("breaches"))
if tainted:
    sc["verdict"] = None          # the copy
...
scored = {q: v for q, v in verdicts.items() if q not in wrong_gold}
ok = sum(1 for v in scored.values() if v.get("verdict") == "match")
```

`run_baseline.py` already documents the identical mistake one field over.
From the comment above `gold_status`:

> Assigning it into the `sc` copy meant `wrong_gold` below, which reads
> `verdicts`, never saw a status the set had declared: a key marked
> `verified_wrong` in the case file stayed in the aggregates unless the judge
> said so too.

That one was fixed by writing into the verdict. This does the same for
contamination, rather than adding a second filter at the point of use, so there
stays one source of truth.

**Why it is worth the churn:** it defeats the check at the one moment the check
fires. A contaminated arm does not look broken, it looks like a result. I was
one step from reporting "Haiku 57% vs Sonnet 81%" off that line, on a run where
the answerer held tools it should not have.

### Reproducing, without a model

`--rebuild` re-derives the summary from saved transcripts and calls nothing:

```
python3 run_baseline.py --set <set> --out <run-with-breaches> --rebuild
before:  passed  12 of 21 decided (57%)
after:   passed   0 of 0 decided
```

## The trigger

`ALWAYS_BLOCKED` carried `Task`, `TaskOutput` and `TaskStop` but not
`TaskCreate`, `TaskGet`, `TaskList` or `TaskUpdate`. When the runner is launched
detached the host grants those four, and they alone flagged all 37 attempts of
a Haiku arm — the same shape as the three MCP resource tools directly above
them, whose comment records the same thing happening on 2026-09-01.

`--allowedTools` grants permission without restricting availability, so the
denylist is the only lever, and it cannot anticipate a host surface that did not
exist when it was written. That is precisely why the first fix matters: it turns
a future miss from a silent wrong number into `0 of 0 decided`.

## Tests

Three regression tests, each beside the thing it covers: two in
`run_baseline_test.py` pinning that a flagged attempt is not counted and that a
fully contaminated run decides nothing, one in `agent_harness_test.py` pinning
the blocklist.

```
run_baseline_test      82 pass
agent_harness_test     37 pass
```


jswir:
Superseded by #1173, which carries both fixes as its first two commits, now DCO-signed (they were not here). The model-free reproduction (--rebuild: passed 12 of 21 -> passed 0 of 0) is cited in that PR's verification section.

PR #1168, opened 2026-09-15
docs(skills): eval-import dead-ends on a question nobody sent an answer with

jswir:
Two files, one paragraph of substance.

## What broke

`eval-import` Step 2 classifies each question by **what arrived with it**. The `question only` row ends at *"a case with no `golden`, which measures coverage and discoverability"* and the skill never says what to do when the set needs a key and nobody knows the number.

That is the common case, not an edge one. A reviewer working from a bare question list read the skill, found no path from a question to a key, and concluded the agent was meant to judge the *path* the answerer took rather than its answer. Reasonable reading of what the skill says.

## The fix introduces no new principles

Every rule it needs was already written down, so it is cited rather than restated:

- `provisional` and why a key derived through the model cannot be `verified` -- the two paragraphs immediately above it
- `verify_goldens.py --promote` as the only way out -- same place
- what no query can settle -- Step 4's second kind, which scores on day one
- competing readings -- **Hold an ambiguous golden** in `skill:eval-loop`'s `reference/golden-side-door.md`, which already carries the procedure for recording candidates instead of inventing a number

The one thing stated here rather than cited: the author holds **full model access** while the agent tested later holds only `get_context` and `execute_query`. That asymmetry is the entire reason a key derived this way is worth anything, and nothing said it.

## Why the ambiguity pointer earns its line

"What did the sold goods cost?" has three defensible answers on the ecommerce sample:

```
6002288.38   every line
5788744.33   excluding Cancelled                 (status doc, units_sold)
5730526.91   excluding Cancelled and Returned    (lifetime_spend)
```

The model documents three conventions for which lines count as a sale. An author without the pointer picks one and sounds certain, and a confident wrong key sends people to fix a model that was already right.

`skills_bundle.json` regenerated. 178 skills tests and 18 bundle specs pass.

Shared skill, so this owes `ms2data/agent-skills` the same file once merged.


jswir:
Superseded by #1173, where this is commit 4 with one addition: the conductor deriving a key records the entities it used as expectedEntities.required -- the producer that field never had -- and case-format.md names that as the one exception to "leave it out".

PR #1172, opened 2026-09-15
fix(sdk): the writer leaves a given's unmodelled tags alone, and a label escapes like the free text it is

Sha-Bang:
Onto `sdk/dashboard-document`, so merging this lands the fixes in #1158. Rebased onto `55abcbc5`.

## The writer was deleting annotations it does not own

`planGivens` deleted every `#` line above a `given:` whenever that given changed, including a pure relabel. That is right for the control contract the builder writes and wrong for anything else. Credible puts `#(secure)` — the marker that makes a given an access control — directly above `given:` declarations, so renaming a control's label silently stripped it. The file still compiled, still round-tripped, and the gate could not see it, because `LocalGiven` never held the tag that was lost.

This is the blind spot `MODELLED_TAG_KEYS` was introduced for on tiles ("unticking one filter deleted its `# big_value` and the strip came back as a one-row table"). The same rule now applies to givens, over the keys `givenTagLine` actually writes.

A removed declaration still takes every tag with it, deliberately. A `#` line left behind does not lapse — it attaches to whatever is declared next, so an orphaned `#(secure)` would migrate onto a different given, which is worse than losing it. Measured against the pinned compiler, along with the reassuring half: a blank line or a `//` comment between a marker and its declaration does **not** detach it, so preserved tags need no repositioning.

## A label is free text, and four of them were not escaped

`quoted` covers the title and a starting given's value. The four tag values composed from a person's own typing did not use it: a tile's `label` and `subtitle`, and a given's `label` and `description`. A quote or a trailing backslash closed the value early and left the rest of the line parsed as tag syntax.

The gate caught all four, so nothing corrupt was ever written. The symptom was worse than it sounds anyway: a document that could not be saved again, reporting a reason that never mentions the character responsible. Type `Revenue (the "good" kind)` into a tile label before this and the save is refused forever.

## The no-edit path, twice

A splice with no edits returned `ok` without running the read-back gate, so an ask no planner could place reported a write that never happened — `delete document.drills` on a drill inside a shared tag line returns `ok` with the file untouched. It is now compared against the file on disk and refused when they differ.

That comparison then had to learn what "the same document" means: a caller that materialises an optional collection the reader omits (`drills: []`, `localGivens: []`, a tile's `filters: []`) asks for no change. `spliceDashboardDocument` is a public entry point and normalising that shape is a reasonable thing for a host to do, so the two spellings compare equal rather than reading as a change no planner can place.

## The tripwire

It counted `#` lines file-wide, which cannot see a tag that survived but moved onto a different declaration. It now compares unmodelled tags keyed by the declaration they sit above — including the enclosing source, since `view:x` repeats across sources and a colliding key would let a tag move between them unnoticed. A given-retag leg runs beside the colspan one.

## Known, and pinned by tests rather than fixed

Ownership is decided by the **first key on a line** and the line is replaced whole, so an unmodelled key sharing a line with a modelled one goes with it — `# label="Cat" audit_scope=finance` loses `audit_scope` on a relabel. This is the convention the tile path has had since `MODELLED_TAG_KEYS`, and it is not a regression: before this change the whole line went regardless. A routed annotation cannot be caught this way, having no key for `tagKey` to match, so `#(secure)` is never the first key on a shared line. There is a test for both halves.

Not changed here, each with a reproducing input if you want them picked up: `planDrills` cannot find a `drill` inside a shared tag line, so removing one is a no-op and retargeting leaves a stale duplicate; `planSettings` drops an unmodelled property from `dashboard { … }` and collapses whitespace inside an unmodelled quoted value; `blockAbove` stops at a blank line, which is narrower than Malloy's attachment boundary, so a marker above a blank line survives a removal as a migrating orphan — on tiles too.

## Tested

- `bun test` in `packages/sdk`: 718 pass, 0 fail.
- Both sets of new tests were run against the unpatched code first. The nine writer tests: eight fail, for the right reason — `#(secure)` is absent from the output, not a mismatched assertion; the ninth pins the removal decision and passes either way by design. The four escaping tests: all four fail.
- Both typecheck configs, prettier, eslint.

One honest note on the tripwire's real-file leg: of the editable dashboards in the repo, only `examples/storefront/dashboards/overview.malloy` declares givens and none carries a foreign tag on one, so that leg guards adjacency and future fixtures. Presence is carried by the unit tests.


Sha-Bang:
@claude review

PR #1175, opened 2026-09-15
skills(analysis): four rules an eval diagnose traced to repeated failures

jswir:
Four changes to three analysis skills. Each one is named by an `eval-diagnose` cluster, and each cluster cites the cases that produced it.

## Where the evidence comes from

One 24-question set over a `query_executions` semantic model, answered three times: haiku at default effort, sonnet at `low`, sonnet at `xhigh`. Each run was diagnosed separately. Nine clusters came back agent-skill-owned; these are the four that a skill change can actually move.

Worth stating plainly: **all of it is one package.** These are real failures with traced mechanisms, not a broad sample, and two of the nine clusters came back `sufficiency: insufficient`, meaning the diagnoser itself would not act on them yet. I have taken the ones marked sufficient and one that repeated identically across two arms.

## 1. The fallback rule was causing the failure it looks like it prevents

`malloy-analysis` §5 already says:

> When the exact ask is impossible, deliver the closest thing that works; do not stop at the explanation.

That rule is about the **shape** of an answer — fewer dimensions per chart, several charts, a table. All three arms read it as licence to swap the **field**.

The case: *"In August 2026, how many real execute_query calls came from the React app versus the MCP tools?"* The field that carries that distinction is `surface`, written only on a logging path that began 2026-09-01. It is null for all 2,033 real August calls. Every arm correctly proved this, then built the split from `log_layer` anyway — which records which process wrote the row, not which surface the user was in. Every August row is `mcp_tool` by construction, so "MCP tool calls: 2,033" is a fact about logging, presented as an answer about product usage.

Sonnet at `xhigh` produced the best version of this and still failed:

> **React app calls: not measurable** — the dataplane/HTTP logging layer wasn't turned on until September 1, 2026 … **MCP tool calls (real, `is_synthetic = false`): 2,033**

The refusal is correct. The number beside it is the substitution. So the fix bounds the existing rule rather than adding a competing one — a new rule next to it would have contradicted it.

## 2. The terminal rule is satisfiable by annotation

> Do not present a result that failed verification with a caveat: fix it, or tell the user you cannot confidently answer.

Reads as covering two computations disagreeing. It does not visibly cover *"the field is null for this period"* or *"the denominator is 11 rows"*, so both arrive as footnotes attached to an unchanged headline. The `xhigh` diagnosis put it exactly:

> the agent names the condition that invalidates its result and then prints the invalidated result as the headline anyway, because the rule says surface the caveat and never says the conclusion itself must move.

Widened to any invalidating condition the agent names itself.

## 3. Extremes of a rate need their denominator read first

"Which package is least reliable" ranked on a failure rate returns a package with between 1 and 11 queries. One failure in three is a 33% failure rate and outranks a group with thousands. Added to the §5 quick reference and to `malloy-analysis-pitfalls`.

The model under test documents no minimum volume, and the sonnet-low diagnosis owned that half as a model gap — correctly. The skill half is the part that generalises: read the denominator, and if the model sets no floor, choose one and say which.

## 4. Population qualifiers belong in the retrieval targets

"Real" in "how many **real** calls" is not filler; it names rows the model marks for exclusion. Agents searched the literal nouns and dropped it, so `is_synthetic` never entered context — despite ranking #1 under a direct on-target search. And the failure repeats when the question carries no qualifier at all, because a table of requests or logs holds synthetic rows regardless.

The cost of noticing late is visible: one arm hand-rolled `caller_id != 'prober'` and reported **2,035** where the model's documented flag gives **2,033**.

Added as decomposition pattern 8 in `malloy-phrase-detection`, plus one question in the `malloy-analysis` discovery checklist.

## What I deliberately left out

- **Ad-hoc query construction failures** (`qe-011`, `qe-018`, haiku only — 18 of 19 `execute_query` calls spent on syntax errors, and one arm concluded the tool "supports only pre-built named views"). Haiku-only, and `malloy-analysis` §4 already states ad-hoc queries are supported. This reads as a model-capability limit, not missing guidance.
- **Two `insufficient` clusters**, per the diagnoser's own sufficiency marking.
- **Model-owned findings**, which belong in the semantic model, not here. The largest: an exclusion rule duplicated inline across four views and cross-referenced from none of the measures.

## Testing

`bun test` in `packages/skills`: **178 pass, 0 fail.** Note that 7 of those fail on a clean checkout until `bun run build` populates `packages/skills/skills`; that is a missing build artifact, unrelated to this change. `frontmatter.spec.ts`, `manifest.spec.ts` and `workflow-exclusions.spec.ts` — the three that would catch an error here — pass either way.

No manifest changes: all three skills already ship.

## Honest caveat on verification

These are prose rules, so I have not re-run the eval to prove they move the numbers. The cases they target are held in the set (`qe-013`, `qe-014`, `qe-020`, `qe-021`, `qe-023`, `qe-024`), and `qe-022` is a holdout the fixes were not written against, so a re-run is the real test. I would rather land the diagnosis and measure than claim the fix works.


jswir:
Folded into #1195 with #1193 and #1194, which carries this change unmodified as its own commit.

PR #1177, opened 2026-09-16
fix(materialization): federate a proxied Postgres source through its SSH tunnel

housejester:
## What broke

A `storage=` build of a source on a Postgres connection that carries a `proxy` (an SSH tunnel to a bastion) fails on every attempt, each source after a full TCP connect timeout:

```
Failed to materialize source '<name>' into storage destination '<dest>':
IO Error: Unable to connect to Postgres at "host=<the database's own host> port=5432 … sslmode=allow":
connection to server at "<host>" (…), port 5432 failed: Connection timed out
```

The query path opens the connection's tunnel and connects the pool to its local endpoint (`buildProxiedPostgresConnection`). The build path handed DuckDB's `postgres` extension the connection's own host and port (`federatePostgres` via `buildPgConnectionString`), which is exactly the host the bastion exists to keep unreachable. `FederationConfig` carried no proxy at all. The passthrough had only ever been exercised on BigQuery and Snowflake, whose federation is secret-based with no network hop, so nothing had noticed.

## What this does

- `FederationConfig` carries the connection's `proxy`; the build session passes it through from the source connection.
- `federatePostgres` opens the tunnel for the build (`openProxy`, the same helper the query path uses), attaches DuckDB at the tunnel's local endpoint, and returns a `close` on the `FederatedHandle`. The build session calls it in its `finally`, after the DuckDB session is disposed. The tunnel is opened inside the same `try` as the attach, so any failure after it is open — a refused TLS mapping as much as a failed `ATTACH` — closes it. No listener outlives the build.
- The tunnel is build-scoped rather than shared with the query path's per-connection tunnel on purpose: that tunnel lives with the environment's connection generation and is closed when a connection update retires the generation, which a multi-minute build cannot depend on.
- TLS is mapped from the connection's `sslmode` into libpq's vocabulary, matching the query path's trust: unset or `no-verify` → `require` (encrypt, no verification, so a force-SSL target is not refused for plaintext); `disable` → `disable`; `verify-ca` → `verify-ca` with `sslrootcert=$NODE_EXTRA_CA_CERTS`, refusing when the bundle is absent, as the query path does; `verify-full` → `verify-full`, against the runtime's bundled roots plus `NODE_EXTRA_CA_CERTS` — the set `resolveProxiedTls` verifies against — written by the process into a private `mkdtemp` directory as the single file libpq takes as `sslrootcert` (never reused from a path that merely exists), so a publicly-trusted target needs no bundle; an unreadable `NODE_EXTRA_CA_CERTS` is skipped with a warning, as both runtimes do at startup. The attach string dials the tunnel endpoint as `hostaddr` and keeps the database's own name as `host`, which libpq uses for SNI and the certificate check, so the hostname half of `verify-full` holds through the tunnel and nothing is downgraded. Verified against a TLS Postgres through DuckDB's own attach: `host=db.test hostaddr=127.0.0.1` connects with `verify-full`, `host=wrong.test` is refused with `does not match host name`, and `host=127.0.0.1` alone is refused the same way.
- libpq conninfo values are quoted when they carry whitespace, a quote or a backslash (`pgConninfoPair`), and `redactPgSecrets` redacts a quoted password whole.
- A tunnel that fails to close after a failed federation is logged, not swallowed; the federation error is still the one raised.
- `buildSourceIntoStorage` takes a small `deps` seam (federate, read) so the one line that closes a proxied source's tunnel is covered: a test proves the close runs after a clean build and after a failed read, against a real plain-DuckDB file destination.
- The `sslmode` switch is typed against `PROXIED_SSLMODES`, so a mode added to that list without a mapping here is a compile error rather than a build-time throw.
- Unproxied Postgres, BigQuery and Snowflake federation are byte-for-byte unchanged.

`FederationDeps` is a small injection seam so the tunnel opener can be stubbed in tests; `ProxiedAttachTrust` does the same for the CA material. Production callers pass nothing.

## Verification

- `connection_federation.spec.ts`: a proxied source dials the tunnel to the database's own host and port, attaches at `127.0.0.1:<local port>` and never at the real host, defaults to `sslmode=require`, and returns a `close` that closes the tunnel; a failed attach closes the tunnel; a TLS refusal between opening the tunnel and the attach closes the tunnel and issues no `ATTACH`; an unproxied source still attaches at its own host with no `close`; the `sslmode` mapping including the `verify-ca` bundle requirement, `verify-full` without a bundle and without a downgrade, `host`/`hostaddr` on every proxied attach, and the quoting of a password carrying a space, a quote and a backslash; `ambientTrustBundlePath` writes the runtime's roots plus `NODE_EXTRA_CA_CERTS` to one file in a 0700 directory of its own, reuses it within the process, and skips an unreadable extra bundle.
- `pg_helpers.spec.ts`: a quoted password with an escaped quote redacts whole.
- `materialization_build_session.spec.ts`: unchanged, passing. 89 tests across the three files.
- `tsc --noEmit` clean; eslint clean on the touched files.
- Not in this PR: a live build through a real bastion. The failure above was observed on a deployment with a bastion-fronted replica; the fixed path is exercised there once a release carries it.

Release note under `## [Unreleased]`; one line in the storage tutorial's prerequisites; the proxied-TLS section of `docs/connections.md` now lists `verify-full` (implemented since #908) and the build-path caveat.


packages/server/src/service/connection_federation.spec.ts
```diff
@@ -339,3 +504,118 @@ describe("attachDuckLakeReadWrite", () => {
...
+         expect(pem).toContain(tls.rootCertificates[0]);
+         expect(pem).not.toContain("EXTRA-MARKER");
+      });
+   });
+
+   it("hands libpq a file this process wrote, in a directory only it can enter", () => {
+      const file = ambientTrustBundlePath();
+      // A private mkdtemp directory: nothing pre-existing at a predictable path
+      // can be picked up as the trust store.
+      expect(statSync(join(file, "..")).mode & 0o777).toBe(0o700);
```
Sha-Bang: **Blocking** - this fails on `windows-latest` and it's currently red on the head — `Cross Platform Tests / Platform: (windows-latest)`, `3776 pass / 1 fail`, `Expected: 448` (0o700). 27444af was green, fb28c55 is not.

windows doesn't report POSIX mode bits, so both this and the 0o600 check on the next line can only hold on posix. the property being pinned is a real one and worth keeping — a trust store picked up from a predictable path is exactly the thing `ambientTrustBundlePath` is written to avoid — so gating the two mode assertions on `process.platform !== "win32"` rather than dropping them seems right, keeping the `not.toBe(join(tmpdir(), ...))` check running everywhere.

packages/server/src/service/connection.ts
```diff
@@ -1203,28 +1229,223 @@ async function federatePostgres(
...
+            pgConninfoPair("sslrootcert", trust.caBundle),
+         );
+         break;
+      }
+      case "verify-full": {
+         logger.warn(
+            `Connection proxy on '${name}': the storage build verifies the certificate chain (verify-ca) but cannot verify the hostname through the tunnel, which terminates at ${endpoint.host}.`,
+         );
+         parts.push(
+            "sslmode=verify-ca",
```
Sha-Bang: **Should-fix** - the verify-full downgrade may not be necessary. libpq splits the two roles: `hostaddr` is what it dials, `host` is what it uses for SNI and the certificate hostname check. so `host=<pg.host> hostaddr=<endpoint.host> port=<endpoint.port> sslmode=verify-full` should keep both halves through the tunnel, and duckdb's postgres extension hands the conninfo string to libpq unchanged.

it matters because the query path went to real trouble to preserve exactly this — `resolveProxiedTls` takes the ssl-object route specifically so `servername` can be set to the real host, with the comment explaining that a connectionString sslmode param can't carry it. so a connection configured verify-full currently authenticates its host when queried and accepts any host with an allowed-CA certificate when built, which is a wider gap than the WARN log suggests.

if `hostaddr` does work here, it'd also let the docs note and the log line go away. worth trying before accepting the downgrade?

packages/server/src/service/materialization_build_session.ts
```diff
@@ -951,6 +956,16 @@ export async function buildSourceIntoStorage(params: {
       // nothing federated or read-write survives the build) and removes its
       // throwaway working directory.
       await dispose();
+      // Then the tunnel, after the attach that used it is gone. Best-effort: a
+      // listener that fails to close is logged, never raised over the build's
+      // own outcome.
+      if (federatedClose) {
```
Sha-Bang, line 978: **Should-fix** - this is the only thing standing between a proxied build and a stranded listener, and nothing covers it — `materialization_build_session.spec.ts` only exercises the gating paths that return before any I/O, so `buildSourceIntoStorage` never reaches this `finally` in a test.

the federation-side tests are thorough about the tunnel being closed on three different failures, but they all prove `federatePostgres` *returns* a close; the one line that proves anyone *calls* it is here. if the assignment at 854 were dropped in a later refactor, every test still passes and each build leaks a listener until the process restarts.

is a test that stubs the federate call to hand back a spy close, and asserts it runs on both a clean run and a throw, reachable without too much scaffolding?

packages/server/src/service/connection.ts
```diff
@@ -1203,28 +1229,223 @@ async function federatePostgres(
...
+      // re-attach of the identical source within one session (the alias is the
+      // connection name, the config is that connection's, so replacing is a no-op
+      // rebind to the same source). (bigquery/snowflake federate via CREATE OR
+      // REPLACE SECRET with no ATTACH; secrets are instance-scoped, so isolation
+      // covers them too.)
+      await connection.runSQL(
+         `ATTACH OR REPLACE '${escapeSQL(attachString)}' AS ${quoteIdentifier(alias, "duckdb")} (TYPE postgres, READ_ONLY);`,
+      );
+   } catch (e) {
+      await endpoint?.close().catch(() => {});
```
Sha-Bang: **Nit** - a close failure here is swallowed silently, so the one case where the tunnel actually does outlive the build is the case with no trace of it. the build session's own cleanup logs this (`Failed to close the SSH proxy a storage build federated through`) — worth the same `logger.warn` in a nested catch, keeping the original error as the one that propagates?

Sha-Bang (commented):
cross-model pass (claude + codex + cursor). the tunnel handling itself reads well — the build-scoped tunnel and closing it inside the same try are both right. three things below, one of which is the red windows job on this head.

packages/server/src/service/connection_federation.spec.ts
housejester: Fixed in 10abc3e1 as you describe: the two mode assertions run only where `process.platform !== "win32"`, and the unguessable-path check runs everywhere. Windows reported 0o666 (438) for the directory.

packages/server/src/service/connection.ts
housejester: You were right, and it holds through DuckDB's attach. Checked against a TLS Postgres with a certificate for `db.test`: `host=db.test hostaddr=127.0.0.1 … sslmode=verify-full` connects on TLS 1.3; `host=wrong.test hostaddr=127.0.0.1` is refused with `server certificate for "db.test" does not match host name "wrong.test"`; `host=127.0.0.1` alone is refused the same way. 48fc4045 emits `host=<real> hostaddr=<tunnel endpoint>` on every proxied attach, keeps `verify-full` as `verify-full` against the ambient bundle, and drops the downgrade, its warn, and the three doc caveats.

packages/server/src/service/materialization_build_session.ts
housejester: Reachable with a small seam. 48fc4045 gives `buildSourceIntoStorage` an optional `deps` (federate, read); two tests run the full session path against a real plain-DuckDB file destination with a stubbed federate that hands back a spy close and a stubbed read that supplies the SELECT. The close runs once after a clean build and once after a failed read, with the build's own error the one raised. Dropping the assignment at the federate call fails both.

packages/server/src/service/connection.ts
housejester: Done in 48fc4045: a nested catch logs the close failure with the connection name, and the federation error is still the one that propagates.

packages/server/src/service/connection_federation.spec.ts
Sha-Bang: closed — `build` is green on 48fc404, windows included. and the platform gate kept more than i asked for: asserting the directory carries the `mkdtemp` suffix holds the unguessable-path property on windows too, where the mode bits can't. better than just skipping the assertion there.

packages/server/src/service/connection.ts
Sha-Bang: closed. the part i checked hardest was whether the test went soft to accommodate the new string, since the real host now legitimately appears in the ATTACH — `not.toContain("db.internal.example")` had to go. inverting it to `not.toContain("host=127.0.0.1")` still discriminates (`hostaddr=127.0.0.1` doesn't contain `host=127.0.0.1`), and with the exact-string match beside it plus `not.toContain("verify-ca")` on the verify-full case, it ends up tighter than before rather than looser.

docs and the spec both dropped the downgrade language in the same commit, so nothing is left claiming a downgrade that no longer happens.

packages/server/src/service/materialization_build_session.ts
Sha-Bang: closed — and it's a real test: drop the `federatedClose` assignment and both cases fail, since it runs the actual `buildSourceIntoStorage` against a file destination rather than asserting on a stub.

one thing i checked because it's where this kind of change usually breaks: the `read` → `passthrough` rename that came with the injection seam. a missed `read.` would now resolve to the injected function instead of the result, silently. i went through the function body — `selectSQL` and `cost` are the only two, both renamed, none left.

Sha-Bang (approved):
approving — all three are closed and CI is green on 48fc404 (build, connection integration, playwright, python-sdk).

the `hostaddr` change is the one worth calling out: it turns the verify-full downgrade from something to document into something that doesn't happen, so the build path and the query path now agree on what a connection's `sslmode` means. that's a better outcome than the note-and-log version this started as, and the docs and spec moved with it.

PR #1180, opened 2026-09-16
fix(server): carry a materialized source's own where: onto the serve shape

housejester:
## The bug

A `#@ persist … storage=` source that carries a `where:` in its extend block is served from the storage tier **without that filter**. The query succeeds, `servedFrom` reports `storage`, and the answer contains every row the filter excludes.

```malloy
##! experimental.persistence

source: raw is conn.sql('SELECT order_id, amount FROM orders')

#@ persist name="high_value" storage=lake
source: high_value is raw extend {
  where: amount >= 100
}
```

`run: high_value -> { aggregate: total is amount.sum() }` answers `375` — the whole table — where the source means `300`. The same model served live, or persisted colocated, answers `300`.

Two facts combine.

**The build does not apply the filter, by design.** `PersistSource.getSQL()` for a persisted source is the persisted relation alone; an extend-block `where:` refines that relation when it is *read*. The colocated tier depends on this: its substitution swaps only the `FROM`, so the reading query keeps its own `WHERE` and filtered rows come back filtered.

**The serve shape did not re-apply it.** The storage tier does not use that substitution — it re-declares the source as `conn.virtual('handle')::Shape` and re-emits its refinements, which were dimensions, measures, joins and views. There was no filter among them, so `filterList` entries were dropped.

Nothing caught it in between. Every other refinement the shape cannot reproduce announces itself by failing the shape's **compile**, which is what triggers the fallback ladder; a dropped `where:` compiles perfectly and simply answers with more rows. Materialization eligibility refuses a given, a `#(partition)`, an authorize gate and an unbound parameter — it says nothing about a plain filter.

## The fix

Carry the source's `filterList` onto the serve shape, one `where:` line per entry.

Deliberately deferred rather than baked: the build is untouched, the artifact still holds the unfiltered relation, and the filter is applied when the table is read. That is the same division of labour the colocated tier already has — artifact unfiltered, filter at read — so the two tiers now agree and a source's meaning no longer depends on which one answers it. It also means the correction applies to tables that are **already built**; no rebuild is required.

One entry per `where:`, never combined: Malloy parenthesises each `filterList` entry and ANDs them, so joining two entries' `code` with `and` would reassociate a top-level `or`. Filters are emitted after joins and dimensions (a `where:` may reference either) and before views. `filterList` accumulates through `extend`, so a source's own entries already carry every filter it inherits — which is why a filtered source built on a filtered source needs no separate handling.

**Filters are never thinned by the serve-shape ladder.** The ladder drops the riskiest refinement category and retries (full → drop views → drop views+joins → base-only) so one un-reproducible refinement cannot cost the whole package its tier. Dropping a view costs the tier for queries that use it; dropping a filter answers with rows the source excludes. So `filter` is kept at every tier, the empty one included, and the floor is no longer "always compiles".

That rule needed enforcing in one place. The ladder had **two** floors built with an empty keep-set — the last entry of `keepKinds`, used when a package carries no pre-aggregation groups, and an inline one appended to the `tiers` literal when it does. Neither could drop a filter before this change, because the kind did not exist; adding it made both a hazard, and adding it to `keepKinds` closed only the first. The second was caught in review.

So the ladder is now built by an exported `buildServeShapeTiers()`: one place to construct a tier, an invariant that can be stated about all of them, and a test asserting every rung keeps the never-thinned kinds — which fails if a rung added later drops one.

That is the deliberate trade: an entry that **cannot** be reproduced — one reaching through a join whose target is not materialized, or one over a column the source hides with `except:`, since the declared `::Shape` is narrowed to the source's public columns — is still emitted, so the shape fails to compile. An entry whose expression text cannot be read fails closed the same way. (A filter referencing a given cannot arise: `assertMaterializationEligible` refuses a given-referencing source outright, as it does `#(partition)` and `#(authorize)`.)

The shape is one model text covering every binding, so that failure surfaces model-wide. It is answered per binding: the floor is probed, each binding is then compiled alone, the ones whose filters do not reproduce are **withheld**, and the ladder is re-entered once with the rest. Withheld, never thinned — thinning is the unfiltered serve this exists to prevent — and re-entered rather than frozen at the failing shape, so a source keeps its joins and views instead of losing them to a sibling's bad filter. Serving fewer queries from the tier is a cost; serving the wrong rows is not a trade worth making.

One limit is documented rather than fixed: the floor carries no joins, so a filter reaching through a join whose target *is* materialized fails the solo probe though it compiles at the richer tiers, and that binding is over-withheld. It fails safe, and closing it means probing each binding with its join-dependency closure — more machinery than the case has warranted.

## Tests

Six scenarios, each mutation-checked against the code it defends:

| scenario | pins |
|---|---|
| `filtered-source-serves-only-filtered-rows` | the rule, on the single-level shape |
| `…-colocated` | the control — colocated was and stays correct |
| `filtered-chain-serves-only-filtered-rows` | a filtered source built on a filtered source: both filters reach the answer |
| `filter-survives-the-serve-shape-ladder` | a filtered source whose view cannot be reproduced keeps its filter when the view is dropped |
| `unservable-filter-withholds-only-its-own-source` | a sibling's unservable filter costs that source the tier, and costs the others neither their rows nor their views |
| `filter-through-a-materialized-join-still-filters` | the over-withheld case still serves the rows its filter selects, whichever tier answers |

The first was written against the bug and carried `known-red`; the tag is retired here. Its colocated twin never did — colocated was correct all along, which is what makes it the control. The ladder scenario is `serve-shape-drops-view`'s arrangement with a filter added, so a source carrying one is forced through the thinning — reverting the ladder change turns it red with `expected 300, got 375`, which is the bug reappearing through the thinning path rather than the original one.

Unit coverage pins the emitted text, which the scenarios do not: one `where:` per entry, the no-combining rule, the emission order, that a source with no filters emits neither an `extend` block nor a `where:`, so a package declaring none reads as it did before, and the unreadable-entry fallback.

Unit coverage also pins the tier invariant directly: every rung of the ladder keeps every never-thinned kind, with and without rollup groups, so a rung added later cannot quietly drop one.

Verified: 91 hammer scenarios pass; the full server suite has the same failures before and after the change, as identical sets.

## Docs

`docs/persist-storage-tutorial.md` §7 listed what the serve shape carries. It now includes filters, and states the exception to its own per-query fallback rule — an un-reproducible filter costs the source its tier rather than serving it unfiltered.


girishjeswani (approved):
Approving. I verified the two load-bearing claims against the real compiler rather than taking them on reading.

**The no-combining rule is genuinely necessary**, which is the part I was most tempted to accept on faith. `code` is stored unparenthesised, so joining `"a = 1 or b = 9"` with `"b = 2"` on `and` parses as `a = 1 or (b = 9 and b = 2)`. Compiling confirms Malloy parenthesises each entry itself: `WHERE ((base."a"=1) or (base."b"=9)) AND (base."b"=2)`. Accumulation through `extend` also holds on a real three-level chain, parent-first, exactly one occurrence per entry, so there's no duplicate-`where:` hazard either.

**No residual path serves unfiltered from storage.** I enumerated every return from `compileServeShape` asking whether it could hand back a materializer that compiles *and* omits a filter. `filter` is in every rung's keep-set; the `nothingToEscalate` short-circuit can't be taken by a filtered source since a filter is itself a refinement; both degenerate branches return the already-failed floor materializer whose eager `getSQL` throws into `live_fallback`. Re-entry is bounded at exactly one pass and the empty-withhold set is handled explicitly. The over-withholding limit is fail-safe as documented.

Mutations all caught, including dropping `filter` from a *middle* rung rather than the floor, so the invariant test really does cover every rung. Nice touch at `spec.ts:1399-1402`: asserting `whereAt > -1` before the ordering comparison, with a comment saying why. Without it a dropped filter (`indexOf` = -1) satisfies the ordering assertion, which is exactly the trap an emit-order regression would slip through.

Two things to fix, both in the description rather than the code:

**1. The rollup framing is off in a way worth correcting, since it's the part a future reader will rely on.** There was one array literal, not two construction sites, and the empty keep-set floor existed on *both* the rollup and non-rollup paths — so "a package with any rollup plus…" understates it by half. More to the point it wasn't a pre-existing live bug: filters didn't exist as a refinement kind before this PR, so neither floor could drop one. It's a hazard this change would have *introduced* at two floors had it not fixed them, which is still worth fixing, just not "the same defect reached through the rollup path". Also `preaggregation_serve_bindings.ts` is a docstring-only change, not a tier call site — the only call site is `model.ts:4425`.

**2. The docs overpromise slightly.** §7 says siblings "keep the tier, and keep it at full strength". True on the normal re-entry path, but `servable.length === 0` and `withheld.length === 0` both send the entire model live (`model.ts:4472-4486`). Since the doc explicitly frames this as the exception readers should trust, it's worth a sentence.

Minor, no action needed: only scenario 76 carried `known-red` (77, the colocated control, never did), and the spec asserts absence of `extend {` / `where:` rather than byte-identity.

One gap worth a follow-up: **the over-withholding limit is documented but not pinned.** A filter reaching through a *materialized* join is fail-safe today, but nothing would catch a future change that made it fail-open. A scenario shaped like 80 with the join target persisted would close it.

For transparency on what I didn't check: hammer needs a live warehouse, so "90 scenarios pass" and "reverting turns 79 red with expected 300, got 375" are unverified on my side. Reading 79, the arrangement does look right — the three failure modes land on distinguishable numbers (300 correct, 375 thinned filter, 1300 lost tier), which is what makes the differential assertion worth having.

housejester:
Both corrections were right, and I checked them against the original code rather than taking them — thank you, the first one had me describing the change as something better than it was.

**On the rollup framing.** Confirmed on `origin/main~1`: two floors with an empty keep-set, one as the last entry of `keepKinds` (no-rollup path) and one inline in the `tiers` literal (rollup path). And `"filter"` had zero occurrences before this PR, so neither could drop one. So it was not a pre-existing defect reached through a new path — it was a hazard **this change introduced at two floors**, and adding the kind to `keepKinds` closed only the first. That is a worse story than the one I wrote and the description now tells it that way, including that the second was caught in review. You are also right that `preaggregation_serve_bindings.ts` is a docstring-only change; the only tier call site is in `model.ts`.

**On the docs overpromising.** Confirmed — `servable.length === 0` and `withheld.length === 0` both return the failed floor materializer, so the whole model serves live. §7 now says "normally only that source", names both cases, and notes they serve live so the rows stay right either way. It was the wrong sentence to leave absolute, since it is the one the section tells readers to rely on.

**On the follow-up.** Added, as `filter-through-a-materialized-join-still-filters`: a filter through a materialized join, a genuinely unservable sibling to force the floor failure, and a clean third source. It asserts the **rows, not the tier** — deliberately no post-mutation assertion on the over-withheld source, because a stale answer and a fresh one are both correct there, and pinning either would turn closing the limit into a red scenario. My first draft did assert the live number; I took it out for that reason. Mutation-checked: dropping the filter emission gives `expected 150, got 375`, which is the fail-open you were pointing at.

Your two "no action needed" notes were both inaccuracies in the description rather than in the code, so I fixed them: only scenario 76 carried `known-red` (77 never did — colocated was correct all along, which is what makes it the control), and the spec asserts the absence of an `extend` block and a `where:` rather than byte-identity, which is what it now claims.

On what you could not run: fair, and the numbers are `expected 300, got 375` for the ladder revert and `expected 100, got 1100` for the demotion revert. The three-way separation you spotted in scenario 79 is the point of the arrangement — 300 correct, 375 filter thinned, 1300 tier lost — so a failure says which mechanism broke rather than only that something did.

The enumeration of every `compileServeShape` return asking whether it could compile *and* omit a filter is the check I most wanted someone else to do, since it is the one I could not do adversarially on my own code.

PR #1181, opened 2026-09-16
test: anchor every bun test path filter so a runtime copy cannot be collected

Sha-Bang:
Adopts #1086 from Monty, who has left. The `./` fix is his and unchanged; the guard around it is rewritten to answer the two findings that PR was sitting on.

## The bug

`bun test <arg>` does not take a directory. Its own `--help` says it runs "all test files with "foo" or "bar" in the file name", so a bare argument is a **substring matched against the whole path**. `packages/server`'s `test:integration` passed `tests`, and once anything boots a server the gitignored `publisher_data/` holds a copy of the storefront example — its own `tests/` directory included. Those files get counted as the server's.

Still live on `main` today (`"test:unit": "bun test --timeout 100000 src"`), and reproduced on this machine against the exact shape:

```
$ bun test tests        # publisher_data/examples/storefront/tests/b.test.ts collected
Ran 2 tests across 2 files.
$ bun test ./tests
Ran 1 test across 1 file.
```

The inflated number is the harmless half. The copy is a snapshot, so it drifts from repo source, and **a drifted copy fails correctly while reporting the server's suite red for a reason outside the server.**

It reaches CI, which is easy to miss because a fresh checkout has no `publisher_data/` and collects the right files. But the integration step in `cross-platform-tests.yml` retries up to three times and attempt 1 boots a server — so attempts 2 and 3 measure something attempt 1 did not.

## What changed

A leading `./`, applied to every `bun test` in the repo rather than only the one that fires today: the server's `test:unit` and `test:integration`, the root `test:hammer` and `test:scripts`, the CLI's `test` and `test:integration`, `packages/skills`' `test`, `create-malloy-package`'s `test` and `test:e2e`, and the two `bun test` lines in `connection-integration-tests.yml`.

`scripts/test-path-filters.spec.ts` is what keeps it fixed.

## The guard, and why it is not the one from #1086

The review on #1086 raised two things, and both trace to one root cause: that version modelled bun's matching rule as a `collects()` predicate and asserted repo filters against the model.

**"The publisher_data loop is just 'does it start with ./?' written the long way."** Correct, and structurally so — checking an anchored filter against a predicate defined as "anchored filters match by prefix" can only restate the definition. The model is deleted. Every anchoring assertion now shells out to the bun running the suite: for each real filter it builds a scratch tree holding the filter's own target **and** a `publisher_data/` copy at the same relative path, then runs bun twice — once as written, once with the anchor stripped. The unanchored leg is what makes the anchored one mean anything, per filter.

That also covers the case a model cannot reach: a platform where `./` is not honoured would fall back to a substring and match anyway, so the anchoring would silently stop applying while CI stayed green. Hence the guard also running under `cross-platform-tests.yml`, where Windows is the platform that populates `publisher_data/` between retries.

**"It wouldn't notice if whole packages dropped out of the parse."** Also correct — the old non-vacuity check named three scripts and asserted a floor of one workflow. The roster is now pinned by name in full, scripts and workflow files both, so a package that stops being read fails rather than quietly shrinking a clean sweep.

It still does not walk the repo for `publisher_data/`: that is a runtime directory a fresh checkout does not have, so a walk would find nothing and pass while saying nothing. Every tree is built in a scratch directory instead.

### Tested

`bun run test:scripts` — 74 pass / 0 fail across 3 files. The guard itself is 34 tests, ~1s.

Mutation-tested, since a guard that cannot fail is worse than none:

| mutation | result |
|---|---|
| unanchor `packages/server` `test:integration` | ✅ red — `expected 1, received 2`, via real bun |
| `packageDirs()` drops the workspace members | ✅ red — 3 failures incl. the roster assertion |
| workflow scan stops matching `.yml` | ✅ red — 2 failures incl. the workflow roster |

Server unit suite: 3762 pass / 3 skip / 5 fail. **The 5 failures are pre-existing** — all in `build_targets_address_equality.spec.ts`, and they fail identically on clean `main` with this branch stashed. Same for the 12 `react-hooks` lint errors `bun run lint` reports; none are in files this branch touches.

Prettier clean on the CI gate (`./packages/**/*.{ts,tsx}`).

## Note

Monty is credited as co-author and his `Signed-off-by` is preserved. #1086 can be closed in favour of this.


housejester (approved):
Approving — the `./` fix is right and I'd land it as-is. Everything below is about the guard's edges, not the fix.

Rather than take the claims from the description, I checked them against the bun that would run this (1.3.14):

```
$ bun test ./src   → Ran 1 test across 1 file     # src/a.test.ts only
$ bun test src     → Ran 3 tests across 3 files   # + publisher_data/src/, foo/src/
$ bun test ./nope  → exit=1  "Test filter ... had no matches"
```

That last one closes the risk anchoring *introduces*: a typo'd anchored path fails loudly rather than quietly running zero tests. I also scanned every (cwd, filter) pair in the diff for test files that match the filter as a substring but live outside it — nothing is dropped by the anchoring in a clean checkout, so this is a pure narrowing to the intended trees. And all three platforms are green, which already answers the question the new cross-platform step exists to ask: Windows does honour `./`.

The model→real-bun move is what makes the guard worth having. The per-filter unanchored leg is genuine non-vacuity rather than an assertion the bug would also satisfy, and the roster being names instead of a count is the right call.

### The one thing I'd change: the roster can go quietly empty for a workflow line

`isBunTest` anchors at `^bun\s+test` and `commandSegments` splits only on `&& || ;`, so a `bun test` inside a shell construct is invisible to the parser. The workflow roster then pins the **file set** rather than the invocations, so nothing goes red. Reproduced by appending this to `connection-integration-tests.yml`:

```yaml
until bun test --timeout 100000 src/service/connection.spec.ts; do echo retry; done
```

`bun test ./scripts/test-path-filters.spec.ts` → 34 pass, 0 fail, unchanged. An unanchored filter, in the credential-carrying workflow this file singles out as "a worse surprise than most", passing green.

Not hypothetical shell trivia: `cross-platform-tests.yml:213` already wraps its integration run in exactly that `until <cmd>; do` retry idiom, and the connection suite is a plausible candidate for the same treatment.

Two cheap changes, `scripts/test-path-filters.spec.ts:117` and `:379`: let `isBunTest` tolerate a leading shell keyword (`until|while|if|then|do|else|!`), and pin the workflow roster by *invocation* (file + filters) rather than by file, so a dropped line shows up the way a dropped script already does. The script roster at `:362` is already the stronger shape — the workflow one is the weak sibling.

### Smaller

- **The new CI step does more than its name says.** `bun run test:scripts` runs the whole `scripts/` suite (74 tests), not the 34-test guard. That newly puts `set-version.spec.ts` and `release-notes.spec.ts` — both of which `Bun.spawnSync(["node", …])` — on Windows and macOS for the first time; they were ubuntu-only via `build.yml:99`. Green today, so no defect. But a future Windows quirk in the release tooling will redden cross-platform CI under a step labelled "Run test-path-filter guard", and the comment above it argues only for the guard. `cross-platform-tests.yml:135`. Either `bun test ./scripts/test-path-filters.spec.ts` (pleasingly, anchored), or keep the breadth and say so in the name — the disagreement is the part worth fixing.

- **The hardcoded counts are already stale.** `scripts/test-path-filters.spec.ts:11` says "40 files where the server has 37". `find packages/server/tests -name '*.spec.ts' | wc -l` → **41** today, so it is 44/41 before this merges. The repo already took a position on this three steps above the one this PR adds, in `cross-platform-tests.yml`: *"No count is written down: this branch has had four different totals, and every one of them was stale in a later commit."* Suggest "more files than the server has".

- **Minor, latent:** `packageDirs()` (`:173`) reads only the root manifest's `workspaces` globs, and its docstring frames that as closing the gap. There are two non-workspace manifests today — `e2e/package.json` and `examples/data-app/package.json` — and `e2e/` (currently `bunx playwright test`) is a plausible home for a future `bun test`. Neither has one now. Either scan every non-`node_modules` `package.json`, or name the workspace scoping as a deliberate gap the way `skills-npm.yml` does for `bun.lock`.

Nice side effect worth knowing: the guard runs green in a worktree with no `node_modules` at all, since it only touches node builtins and `bun:test`. That makes it genuinely self-contained, which is more than most guards manage.

<sub>Reviewed by Claude Code (session "Plucky Puffin") on James Estes' lenses.</sub>

PR #1184, opened 2026-09-16
fix(server): say why a server with no config is serving nothing

Sha-Bang:
Adopts #1085 from Monty, who has left. His change is unmodified; the review findings are fixed in later commits.

## What this fixes

The Docker image runs with `--server_root /publisher` and nothing mounted there, so no config resolves and the server reports `operationalState: serving` with `environments=0 packages=0 load_errors=0`. Every one of those numbers is accurate, which is the problem — nothing failed because nothing was asked for, so a forgotten config mount is indistinguishable from a healthy deployment, and nothing in the output at any level names the path that was checked.

**Still live on current `main`.** Measured against `30ba53fc` with an empty server root:

```
NO THROW
  environments: []
  frozenConfig: false
```

No throw, no environments, and nothing naming the path. (This is a different branch from the one #1049 covered — that PR's `${VAR}` case now fails the boot outright via `PublisherConfigError`, which is why #1049 was closed as superseded. The ENOENT path `main` deliberately treats as "not an error" is untouched, and that is this one.)

## The change

`getUnresolvedPublisherConfigPath(serverRoot)` returns the path that was looked for and not found, or null once a config resolved. It returns null for an explicit `--config` in either state, because `getPublisherConfig` already reports that case and two shapes of one complaint help nobody.

`environment_store.ts` logs the notice from `initialize()`'s success tail, immediately before the readiness line — once per process by construction, since `initialize` is private and called once from the constructor. It is `info`, not `warn`, because booting unconfigured is a supported mode: a deployment can mount no config and create environments over the API afterwards.

The Docker smoke job's `/api/v0/environments` assertion also gets stricter. It was `type == "array"`, which passes for an empty catalog **and** a full one — so an image that served nothing stayed green. It now pins empty for the unmounted container, and asserts the notice names the checked path.

## The gate is on what was declared, not what loaded

The first version gated on `this.environments.size > 0`, which asks *did anything load* where the sentence it prints claims *was anything asked for*. In the gap it is actively wrong: an environment the database holds whose directory has gone is skipped with a logged error and a bare `return` — no throw, so `failedEnvironments` stays empty too. The boot then reports `environments=0 packages=0 load_errors=0` and tells the operator to create a config file that was never the problem, directly beneath the one line naming the real cause.

Reachable the way the notice's own last sentence invites: boot unconfigured, create an environment over the API, restart onto a root where `publisher_data/<env>` is gone. Reproduced before fixing, and the count of declared environments — the manifest's plus the database's — is now carried out of `initialize()` rather than inferred afterwards. Only that scope sees both sources.

## The review finding on the negative test

`mock.module` is process-wide and every spec file shares one bun process, so a file without a storage mock of its own inherits whichever `StorageManager` the previously-run file installed. `environment_store_clone.spec.ts` installs one whose `initialize` throws its redaction fixture, so construction fails, `PUBLISHER_INIT_FAILED` fires, and the success tail carrying the notice never runs.

Reproduced before touching it — the file alone passed 6/6; run after the clone spec:

```
(fail) unconfigured boot notice > names the missing config once when nothing loaded
(fail) unconfigured boot notice > can still create an environment at runtime, as the notice claims
 10 pass, 2 fail
```

The third case is the one that mattered. `stays quiet when a config resolved` asserts the notice is **absent**, so under the leak it passed — for the opposite of the reason it names. A notice-absent assertion only means something once the boot it describes actually happened, so `assertInitialized` pins the success line logged immediately before the notice, in all boot cases.

### Tested

Server unit suite on a clean install: **3822 pass / 3 skip / 0 fail** across 173 files (verified by @housejester at `3e490efb`; a local tree resolving `@malloydata/malloy` 0.0.427 against the declared `^0.0.432` fails 5 tests in `build_targets_address_equality.spec.ts`, which is an install artifact and not a property of this branch).

Mutation-checked in two places: removing the storage mock while keeping `assertInitialized` turns all **three** boot cases red, including the negative one; restoring the `this.environments.size` gate turns the declared-vs-loaded case red.

Prettier and eslint clean on the CI gate.

## Note

Monty is credited as co-author and his `Signed-off-by` is preserved. #1085 can be closed in favour of this.


housejester:
Reviewed at `3e490efb`. Good, small change, and the second commit is the interesting one — the finding about the negative test is exactly right, and I reproduced both claims it rests on rather than taking them on trust.

**Verification** (clean `bun install` in a worktree at this head):

- Pre-fix spec (`29854a7d~1`) run after `environment_store_clone.spec.ts`: `10 pass, 2 fail`, with `stays quiet when a config resolved` green. The leak reproduces exactly as described, including the negative case passing for the wrong reason.
- Mutation check — current spec with the `mock.module` block stripped, run after the clone spec: `9 pass, 3 fail`. All three boot cases red, including the negative one. `assertInitialized` does the work claimed for it.
- `bun test --timeout 100000 src` → **3822 pass / 3 skip / 0 fail** across 173 files.

That last one is worth folding back into the description: the "5 pre-existing failures in `build_targets_address_equality.spec.ts`" caveat is a local `@malloydata` install artifact, not a property of this branch. On a fresh install the suite is clean.

---

### One thing worth acting on

**The gate is a weaker proposition than the sentence it prints.** `environment_store.ts:867` gates on `this.environments.size > 0`, which asks *did anything load*, not *was anything asked for*.

In the database branch of `initialize()` (`existingEnvironments.length > 0`), an environment restored from `publisher.db` that fails to load never reaches `this.environments`. The missing-directory-with-no-config-entry case at `environment_store.ts:690` is the sharper one: it logs an error and `return`s, recording nothing in `failedEnvironments` either. So with no config file on disk you get `environments.size === 0`, `failedEnvironments.size === 0`, a readiness line of `environments=0 packages=0 load_errors=0` — and the new notice, as the last thing before `PUBLISHER_READY`, saying *"no publisher.config.json was found at /publisher/publisher.config.json. Create one there…"*, pointing the operator at a mount that was never the problem. The one true line (`Environment "x" not found in config and files missing`) is further up the log and reads as noise beside it.

The reachable path is the mode this notice itself advertises: boot unconfigured, create an environment over the API (as its last sentence invites), restart onto a root where `publisher_data/<env>` is gone.

Suggest carrying the fact out of `initialize()` rather than inferring it — a flag for "the manifest declared none **and** the database held none" — and passing that to `logUnconfiguredNotice`. Adding `failedEnvironments.size === 0` to the early return closes the load-failure case but not the silent-skip one. Worth a spec alongside the other three: a database holding one environment whose path is gone, no config file ⇒ notice absent.

### Smaller

- **The smoke assertion is satisfiable by the bug it guards against.** `build.yml:490-494` greps the log twice, independently — once for `Serving with no environments` and once for `/publisher/publisher.config.json`. A notice naming the *wrong* path still passes as long as some other line carries the right one, and `config.ts:1086`'s bundled-default message has exactly that shape (`No publisher.config.json found at <root>/publisher.config.json; falling back…`). It isn't reachable in this container, which is the only reason the check currently means anything. One grep of the joined substring pins both halves to the same line.

- **The description outruns the diff.** Two paragraphs describe the second configured container and the `malloy_getContext` MCP round-trip that `3e490efb` removed — that commit message argues the removal well, but the body still presents the leg as shipped. `Measured against 30ba53fc` has also moved on. Trimming it to what ships would help; it undersells an otherwise strong write-up.

- **The two halves of the message have different reach.** `emitReadinessLine` writes straight to `process.stderr`, so `PUBLISHER_READY` survives any `LOG_LEVEL`; the notice goes through winston at `info` and disappears under `LOG_LEVEL=warn`. The reader being served here is a production Docker operator, who is the most likely to have set it. `info` over `warn` still looks right to me and the reasoning in the JSDoc is sound — noting the asymmetry, not asking for a change.

- `build.yml:474-476`: *"Nothing here regressed by saying so: the check this replaces could not show it either."* That defends a removal that isn't in the diff, so it has no referent for a later reader; `3e490efb`'s message already carries it. The first half of the comment — what the assertion pins, and why `type == "array"` wasn't enough — is worth keeping.

- The mock repository at `environment_store_unconfigured.spec.ts:44` has no `listStorageDestinations`, so the runtime-creation case logs `Error syncing storage destinations for "created-at-runtime"` on every run and passes regardless. The assertion still holds; one more stub would make the case exercise what it claims to.

— Zesty Gecko

housejester (approved):

Sha-Bang:
Fixed at `060fe7af`. The gate finding is right and I reproduced it before changing anything: database holding `created-at-runtime` with its directory gone, no config file, and the log carries `Environment "created-at-runtime" not found in config and files missing` *and* the notice, in that order.

Your note that `failedEnvironments.size === 0` wouldn't close it holds for the reason you'd expect — that branch is `logger.error` plus a bare `return`, so it never reaches `recordEnvironmentInitializationFailure` and nothing is recorded anywhere. Inferring "was anything asked for" from any post-hoc state was the mistake; the declared count (manifest entries plus database entries) is now carried out of `initialize()`, since only that scope sees both sources. Mutation-checked — restoring the `this.environments.size` gate turns the new case red. The spec you asked for is there.

The smoke grep is one joined substring now. I checked the collision you described and it's exactly as you said: `config.ts:1086` emits `No publisher.config.json found at <root>/publisher.config.json`, so the two independent greps would both pass on a notice naming the wrong path.

`listStorageDestinations` stubbed, so that case stops logging a sync error and passing anyway. The comment defending the removed leg is gone — you're right that it had no referent once the leg left the diff. Description trimmed to what ships.

On the 5 `build_targets_address_equality` failures: I can name the cause now. This worktree's own `node_modules` is empty, so resolution walks up to the parent repo, which holds `@malloydata/malloy` 0.0.427 against the declared `^0.0.432` — `getBuildTargets` postdates 0.0.427. Install artifact, as you said; I've put your clean-install number in the description instead of my caveat.

Not changing the `info` level, agreed on your reasoning. Worth recording that the asymmetry is real though: `PUBLISHER_READY` goes straight to stderr and the notice goes through winston, so `LOG_LEVEL=warn` keeps the readiness line and drops the explanation — the two are adjacent by design, and only one survives that setting.

PR #1185, opened 2026-09-16
fix(server): refuse a colocated persist whose persisted query references a given

housejester:
## The bug

A colocated `#@ persist` source whose **persisted query** filters on a given with a default builds with that default substituted, and then serves that slice to every caller — ignoring the value each one supplies.

```malloy
##! experimental { persistence givens }
given: ORG_ID :: number is 1
source: raw is duckdb.sql("""SELECT * FROM (VALUES (1,7),(2,9),(1,8)) AS t(org_id, user_id)""")

#@ persist name="inside"
source: inside is raw -> { where: org_id = $ORG_ID; select: * }               // baked

#@ persist name="outside"
source: outside is raw -> { select: * } extend { where: org_id = $ORG_ID }    // honoured
```

| source | build SQL | artifact rows | `count()` with no given | `count()` with `ORG_ID = 2` |
|---|---|---|---|---|
| `inside` | `… WHERE org_id = 1` | 2 | 2 | **2** — live answer is 1 |
| `outside` | no org filter | 3 | 2 | 1 |

A caller asking for org 2 is served org 1's rows. Silently — `count()` simply returns the wrong number.

## Why

A given's value is substituted when the compiler compiles. Inside the persisted query the only value available at build time is the declaration default, so it is baked into the relation. At read, persistence substitutes the stored table for the source's `FROM` and nothing re-applies a filter that lives *inside* that relation — there is nothing left to apply it to.

The extend-block placement is the opposite, and is the documented form (`docs/row-level-access.md`): absent from the build, applied over the materialized rows at read with the caller's own value.

Without a default the baked shape cannot build at all (`Given 'ORG_ID' has no value and no default`), so this needs the default to exist.

## The fix

Refuse the shape on the colocated gate, with a message naming the placement that fixes it — the safe form is one move away from the refused one.

The compiler already answers the question directly. A query source's `query.givenUsage` is the transitive summary of the givens its pipeline actually reads, so a non-empty one means a value was substituted while producing the persisted relation. Checked against `getSQL()` across twelve shapes, it agreed with whether the build SQL carried the predicate in every one:

| | baked into the build |
|---|---|
| `where:` inside the persisted query | yes |
| a given in a `group_by` | yes |
| the input source's own `where:` | yes |
| a dimension, measure or join on the INPUT source that the query **reads** | yes |
| `where:`, dimension, measure or join in the persist source's own extend block | no |
| the same declared on the input source and **not** read by the query | no |

The last two rows are why this reads the marker rather than walking the IR. `query.structRef` inlines the input source's whole definition, so a structural walk refuses every derivation of a base that merely OFFERS a given-filtered join — a base joining a visibility source, with persisted derivations over it using different subsets, is an ordinary shape. And the walk cannot be narrowed to fix it: for a dimension the query reads, the given sits in `structRef.fields`, exactly where an unread one sits.

The walk survives as the **fail-closed fallback**, for a source with no `query` at all, a `query` carrying no marker, or IR that cannot be read. That branch is what a future compiler dropping the marker would take, so it is pinned by its own test rather than left to inference.

**Givens only.** A parameter cannot appear in a persisted source's own query pipeline — the compiler answers `'<name>' is not defined` — so there is no parameter equivalent of this shape to refuse. (A parameter *is* frozen when a source is derived from a parameterized one, which is a different shape and a different gate.)

**The storage tier is deliberately untouched.** It already refuses every given-referencing source, whichever placement, so this check would be unreachable there — and its remedy would be wrong advice, since the extend-block form is refused on that tier too.

## Tests

A scenario in two packages, because the refusal fails its build: `gpq` carries the refused shape, `gpqb` the honoured one. It pins both halves — that the baked shape is refused, and that the honoured one still answers **per caller from the artifact**: 125 by default, 50 for `ORG_ID=2`, and still 125 after a mutation that would take a live answer to 1125. A refusal alone would not show the second, which is the property that makes refusing the first acceptable.

**One existing spec case was pinning the defect.** `accepts a colocated persist source that references a given but carries no gate` used the baked shape and asserted the colocated gate admits it. Its purpose — that the gate does not apply `referencesGiven` wholesale — is unchanged and still demonstrated, since `assertMaterializationEligible` refuses the extend-block shape while the colocated gate admits it. Rewritten with that placement rather than deleted.

Unit cases pin the pair that decides the check: the same join, dimension and measure declared on an input source are **admitted** when the persisted query does not read them and **refused** once it does. Plus the extend-block placements, a rollup named by the annotation its author actually wrote, the fail-closed fallback, and an unreadable definition.

The refusal message is asserted as a rule rather than a spelling — keep the given out of the persisted query, and apply it at read as a `where:`, a dimension, a measure or a join — because prescribing one placement told authors to make a move they had often already made.

Mutation-checked on each branch: disabling the refusal makes the build succeed and turns the scenario red; falling back to the structural walk turns the unread-declaration case red; and trusting the marker with no fallback turns the fallback case red. 92 hammer scenarios pass; the server suite has the same failures before and after, as identical sets.


girishjeswani:
Really good work — the diagnosis is right, the fail-closed handling is careful, and the mutation coverage is better than most. I probed the central claim hard because the twelve-shape table is what the whole gate rests on, and my prime suspects all came back clean: nested views, `having:`, join `on:` conditions and selected-but-not-filtered calculations are all correctly covered. About twenty shapes agreed with the build SQL.

One doesn't, and I think it needs a fix before this lands.

## A given passed as a source argument bakes, and is admitted

```malloy
##! experimental { persistence givens parameters }
given: ORG_ID :: number is 1
source: raw is duckdb.sql("...")
source: pp(x::number) is raw extend { where: org_id = x }

#@ persist name="p1"
source: p1 is pp(x is $ORG_ID) -> { select: * }
```

| source | `query.givenUsage` | build SQL | colocated gate |
|---|---|---|---|
| `p1` (given as argument) | `[]` | `… WHERE base."org_id"=1` | **ADMIT** |
| control (given inside query) | `[{id: ORG_ID}]` | `… WHERE base."org_id"=1` | REFUSE |

The two build SQL strings are byte-identical. Same baked default, same frozen slice, opposite verdicts — solely because the given reached the pipeline through `structRef.arguments` rather than a segment reference. So this is the same row-level-access leak the PR exists to close, surviving for this shape: the artifact holds org 1's rows and a caller supplying `ORG_ID=2` gets them, silently.

The compiler-side reason: `expandRefUsage` collects from segment `refSummary`, join `on:`/filterList, and atomic field `refSummary`. A source argument is none of those — it binds at `structRef` construction, so nothing ever summarises it onto the query. The marker isn't mishandled; it's empty when it shouldn't be.

What makes this worth fixing rather than documenting: **the fallback walk gets this pair right and the marker gets it wrong.** On `p1` vs the safe shape, the walk refuses the first and admits the second, both correct. It's the `if (Array.isArray(usage)) return usage.length > 0` short-circuit at `materialization_eligibility.ts:472-478` that loses it, since `[]` is an array, so the marker is trusted and the walk never runs.

## A fix that doesn't cost you the argument for using the marker

Check the argument holders before trusting an *empty* marker:

```ts
const u = q.givenUsage;
if (Array.isArray(u) && u.length > 0) return true;
// A given can bind into the relation as a source ARGUMENT without appearing in
// any segment's refSummary, so the marker reads empty while the build bakes it.
for (const a of [q.sourceArguments, q.structRef?.arguments, q.structRef?.sourceArguments]) {
  if (a && walkForGiven(a, new WeakSet(), 0)) return true;
}
if (Array.isArray(u)) return false;
```

Run across the decisive cases, all six land correctly:

| case | want | result |
|---|---|---|
| given as source argument (bakes) | REFUSE | REFUSE |
| extend-block `where:` (safe) | ADMIT | ADMIT |
| **unused_join** — the marker's raison d'être | ADMIT | ADMIT |
| unused_dimension | ADMIT | ADMIT |
| reads_dimension (bakes) | REFUSE | REFUSE |
| parameterized with a **constant** argument (bakes, no given) | ADMIT | ADMIT |

It keeps both unread-declaration cases that justify preferring the marker over a walk, and correctly ignores a constant argument — that bakes too, but it isn't a given, so it stays materializable. Worth adding the argument shape as a spec case; the current suite structurally can't see it, which is why the mutations all pass.

## Everything else checks out

- **Fail-closed is genuinely fail-closed.** `undefined` marker, missing `query`, unreadable IR all route to the walk, and `catch { return true }`. I mutated the fail-open variant specifically (treat a missing marker as "no givens") and the fallback test catches it.
- **The rewritten test is preserved, arguably strengthened.** Both assertions survive verbatim; only the model changed, and it now demonstrates the contrast on a shape that should be admitted. Candidly disclosed in the commit message, and I think it was the right call over deleting it.
- **Storage-tier and parameter claims both verified.** The gates dispatch mutually exclusively at `build_plan.ts:1030-1040`, and `referencesGiven` walks the whole `_sourceDef` — every probe returns `storage=REFUSE`, including the argument shape the colocated gate misses. And a parameter genuinely can't appear in a persisted source's own query pipeline: `'t' is not defined`, reproduced.
- Mutations (a)/(b)/(c) all behave as the description says, including (b) confirming the over-refusal that justifies the marker approach. Docs handle the storage-tier trap correctly — a reader is told the extend-block remedy works for colocated and that storage refuses either way, so no live trap. API additive, DCO 6/6, all 10 deletions accounted for with no test or assertion dropped. The scenario's three numbers do separate artifact-served-per-caller from live-served and from baked.

## Two smaller things

**A reason-code shift the release notes gloss.** A preaggregate rollup that also declares `storage=` runs both gates (`build_plan.ts:1081-1091`), colocated first, so it'll now report `given_in_persisted_query` where it previously reported `given`. Harmless, but "the `storage=` tier is unchanged" isn't quite true of the reason code.

**The `buildSubstitutesAGiven` doc comment asserts the twelve-shape survey as settled.** Once the argument shape is closed, I'd soften it to name what the marker provably covers (segment refs, join on-clause/filterList, field refSummary, turtles) rather than implying exhaustiveness. The next gap will be found the same way this one was, and a comment that reads as complete makes that harder.

Open question I couldn't close: whether argument binding is the *only* channel that binds a given without appearing in a segment's `refSummary`. Worth a word with the compiler folks, since the same reasoning applies to any future construct that binds at `structRef` construction time.

I didn't run hammer (no live warehouse), so the build SQL and gate verdicts above are from compiling the shapes directly; the leak at read is inferred from your own premise that the read path swaps only the `FROM`. A scenario in the shape of 82 with the argument placement would close that loop empirically.

housejester:
Found and fixed — you were right, and I reproduced it before changing anything: `bakes=true, colocated=ADMIT, storage=REFUSE, givenUsage=[]`, with the build SQL byte-identical to the control. Pushed as `62d8150e`.

This is a fail-open, which is worse than the over-refusals the earlier rounds found, and it is the specific risk I took on when I moved from the structural walk to the marker. I noted that risk at the time and then let twelve agreeing shapes talk me out of it. Twelve shapes agreeing is evidence the marker is usually right, not evidence it is complete — and the shape that breaks it was reachable the whole time.

**The fix.** An empty marker is now trusted only once the argument holders are clear. I traced where the binding actually lands rather than guessing at holder names: `$.query.structRef.arguments.x.value`. The walk searches any `arguments`/`sourceArguments` holder anywhere under `query`, not just the outermost, since a chain of parameterized sources nests one inside the next. Your six decisive cases all land correctly, including the constant argument staying admitted — that one bakes too, but a concrete instantiation has no per-caller binding to lose, and refusing it would have quietly broken the shape `parameter-eligibility` exists to admit.

Mutation-checked: removing the argument check turns the new case red on its own.

**Both smaller points taken.** The reason-code shift is real — I confirmed a rollup declaring `storage=` runs the colocated gate first at `build_plan.ts:1081`, so the release note now says what the tier accepts is unchanged while one reported value moves, rather than "unchanged" flat. And the doc comment no longer reads as a settled survey: it names what the marker provably covers, points at the argument channel as the thing it does not, and says the list is what has been shown rather than what exists.

**And the empirical loop is closed.** Scenario 82 gained a third package carrying the argument placement, so the refusal is proven through the real build path for this shape too, not only in the unit gate — 7/7.

On your open question, whether argument binding is the only channel that binds without appearing in a segment's `refSummary`: I do not know, and I have not tried to answer it from inside the publisher. The honest position is that the marker plus one known exception is what has been shown, which is why the doc comment now says exactly that instead of implying closure. Worth the conversation with the compiler folks you suggest — any future construct binding at `structRef` construction time would land in the same blind spot.

92 hammer PASS, 33 eligibility cases, lint and server typecheck clean, no new failures in the full suite.

girishjeswani:
Depth 1 is genuinely closed. Measured on `62d8150e`: `arg_given` now `REFUSE(given_in_persisted_query)`, and I confirmed the leak was real before the fix (build SQL byte-identical to the control). No regressions on the cases that were most at risk — `unused_join` and `unused_dimension` both still ADMIT, constant argument still ADMIT, 33/33 spec pass. The binding lands exactly where you said it did.

The nesting claim doesn't hold, though, and it's hiding another instance of the same leak.

## A given bound inside a wrapping parameterized source is admitted by both tiers

```malloy
given: ORG_ID :: number is 1
source: raw is duckdb.sql("SELECT 1 AS org_id, 100 AS amt UNION ALL SELECT 2, 200")
source: L1(x::number) is raw extend { where: org_id = x }
source: L2(t::number) is L1(x is $ORG_ID) extend { where: amt > t }
#@ persist name="wrapped"
source: wrapped is L2(t is 0) -> { select: * }
```

```
colocated : ADMIT
storage   : ADMIT
marker    : []
BUILD SQL : … WHERE (base."org_id"=1) AND (base."amt">0)
ROWS      : [{"amt":100,"org_id":1}]     <- org 2's row absent from the artifact
```

Depths 2 and 3 both admit; depth 1 refuses. So the fix covers exactly one level.

**The recursion isn't the problem — the code does recurse, at `:489-506`.** The problem is that the compiler never puts a nested binding in an `arguments` holder at all: it constant-folds the given during the inner source's construction. An IR scan for any `given` node under `query` returns nothing, and the only holder present is `$.query.structRef.arguments keys=[t] given=false`. The given is already the literal `1` by the time the gate sees it. Recursing finds nothing because there's nothing left to find.

Two consequences worth separating out:

**`storage=` doesn't catch this one either.** That breaks the "the storage tier refuses a given reference in any position" property — which both `RELEASE_NOTES.md:50` and my own previous round treated as settled. It holds for every placement either of us probed except this one.

**Nothing in the suite tests the nesting claim.** I mutated the walk to only check the outermost holder, leaving the recursion otherwise intact: **33 pass / 0 fail**. The deep descent's only measured effect across my probe set is an over-refusal (below), not a catch. Removing the argument check entirely does fail the new case alone, so your claim 6 is verified — it's specifically the deep-descent behaviour that's unpinned.

## One over-refusal, from the descent reaching where the marker was meant to stop it

```malloy
source: J(x::number) is joinee extend { where: org_id = x }
source: withjoin is raw extend { join_one: j is J(x is $ORG_ID) on org_id = j.org_id }
#@ persist name="join_arg_unread"
source: join_arg_unread is withjoin -> { group_by: user_id }   // does not read j
```

REFUSE, but the build SQL carries no given predicate at all. The holder sits at `$.query.structRef.fields[3].arguments` — the recursive descent reaches into `structRef.fields`, which is the inlined-input-source region the marker approach was introduced to avoid walking. Narrower than it could be (a constant argument over a base offering an unread given-filtered join correctly ADMITs, so it needs a *parameterized* joined source bound to a given), but it is the traded-defect risk showing up for real.

## What I'd suggest

I don't think the wrapped case is patchable in the gate. Since the fold happens during source construction, there's no IR node left to walk, so this likely needs a compiler-side signal — the marker propagating through source-argument substitution — rather than more walking on the publisher side. That reads to me like a Malloy-side issue rather than more work in this PR, which is your call to make.

If you split it, two things should be softened before merge so the gap isn't recorded as closed:

- `RELEASE_NOTES.md:50` — "still refuses a given reference in any position" isn't true for the wrapped argument.
- `materialization_eligibility.ts:481-483` — the `argumentBindsAGiven` doc asserts the nesting property, which is the one thing here that's disproven. That one matters more than the release note, since it's what the next person reads before deciding whether to look.

The alternative, if you'd rather close the class now, is to refuse any persisted query whose `structRef` chain instantiates a parameterized source, accepting the broader over-refusal. That would cover `join_arg_unread` too, at the cost of shapes that are genuinely fine.

## The rest verified clean

Fail-closed on the new path holds — cycles under a holder, inside an argument value, in `structRef`, a 400-deep chain, and a throwing getter all refuse. Prior fail-closed behaviour not regressed (undefined marker, missing `query`, non-array marker, unreadable def all still refuse). The new spec cases compile real Malloy through the `persistSources` helper rather than asserting synthetic IR, and assert both arms; not a change-detector. Scenario 82's third package genuinely carries the argument placement with the `cites:` matching the other refusal's pin, and 7/7 is plausible from the file (I didn't run hammer — no live warehouse). The doc comment at `:443-449` does now read as "what has been shown rather than exhaustive", which is exactly the change I was asking for. DCO clean on `62d8150e`, spec diff 28 added / 0 removed.

Two small ones: `argumentBindsAGiven` lacks the `WeakSet` its sibling walks carry (safe today via the depth bound and the fail-closed catch, but inconsistent), and the constant argument is admitted because `walkForGiven` finds no given under the holder — correct at depth 1, but that's the same mechanism that silently admits the wrapped case, so I wouldn't lean on it as evidence the approach generalizes.

housejester:
I tried your suggested direction and it opens a hole, so I have not taken it. Two of your three findings I could not reproduce as described — details below, and I would like your version before we settle this.

## The over-refusal is real, and narrowing to fix it is a fail-open

`join_arg_unread` refuses with no predicate in the build SQL — confirmed, exactly as you described. So I narrowed the walk to the query's own argument holders. That fixed it, and broke its sibling:

```
join_arg_read    marker=[]  bakes=true   narrow: ADMIT   descent: REFUSE
join_arg_unread  marker=[]  bakes=false  narrow: ADMIT   descent: REFUSE
```

A given binding a joined source the query *does* read sits under `structRef.fields` too, and `query.givenUsage` is **empty** for it. So the only thing catching it is the descent. Narrowing admits a shape whose build SQL carries `org_id=1` — the leak this PR exists to close, reintroduced through the join.

Your mutation evidence pointed the same way it pointed for me: outermost-only left 33/33 green because nothing in the suite covered the read-join case. That case is now pinned, and removing the descent fails it plus the depth-1 argument case.

So the descent stays, and its cost is stated on the walk rather than asserted in a test — a future signal precise enough to separate read from unread would flip `join_arg_unread` to admitted, and that should land as an improvement, not a red test. Refusing a source that does not bake loses a tier; admitting one that does loses a tenant's isolation.

## The wrapped case does not compile here

Your model, verbatim, on the pinned Malloy:

```
Parameter 'x' has no value supplied.
```

Four variants of the same idea fail identically — binding an argument while defining an intermediate named source does not seem to bind at all in this version. The three wrapped forms I could get to compile all REFUSE on both tiers, with the given node still present in the IR:

```
mid is L1(x is $ORG_ID) -> { select: * } ; wrapped is mid -> {…}   colo=REFUSE stor=REFUSE
…same with an extend on mid                                        colo=REFUSE stor=REFUSE
persist directly on L1(x is $ORG_ID) -> {…}                        colo=REFUSE stor=REFUSE
```

So I cannot confirm the constant-folding claim, the depth-2/3 admits, or that `storage=` misses it. Which means I have not softened `RELEASE_NOTES.md:50` or the nesting sentence: I am not recording a gap as real on a repro that does not reproduce, and I am not recording it as closed either.

What would settle it: your Malloy version, and the exact model that compiled. If it reproduces on something we ship, it is a genuine hole and I will take it seriously — including the compiler-side framing, which I think is right if the fold really happens during construction.

## Taken

`argumentBindsAGiven` now carries the doc it was missing — that the descent is load-bearing rather than defensive, with the measured reason, and that the over-refusal is its accepted cost.

On your last note: agreed, and it is why I did not lean on the constant argument as evidence the approach generalises. It is admitted because no given node sits under the holder, which is the same mechanism that would miss a folded one.

92 hammer PASS, 34 eligibility cases, lint and typecheck clean.

girishjeswani:
I re-ran both disputes myself. **You're right on both, and I withdraw the wrapped-argument finding.** Details, because the reason it looked like a leak is worth recording.

Same Malloy on both sides, so that wasn't it: `^0.0.432` pinned, `0.0.432` resolved, and the clone my earlier measurements came from also had 0.0.432. No version mismatch.

## Dispute 1 — withdrawn. The wrapped shape cannot build, so there is no leak

The confusion is that your error and mine are both real, at different stages. The model *does* compile — that part of my report was accurate, and it's why I saw ADMIT:

```
d2_wrapped  compiles=true  colo=ADMIT  stor=ADMIT  marker=[]
d3_wrapped  compiles=true  colo=ADMIT  stor=ADMIT  marker=[]
```

But `getSQL()` on the build relation fails:

```
d2_wrapped  buildSQL: FAILS: Parameter 'x' has no value supplied.
d3_wrapped  buildSQL: FAILS: Parameter 'x' has no value supplied.
```

So the gates admit it and then **no artifact is ever produced**. Nothing is frozen, nothing is served, no tenant's rows go anywhere. That is not a leak, and my report claiming baked SQL and artifact rows for this shape was simply wrong — I reported a build relation that this shape cannot produce. Your `Parameter 'x' has no value supplied.` was the same wall, hit one step earlier than I hit it.

Your three variants reproduce exactly as you said, and they bake because they genuinely build:

```
jesse_mid_arrow   colo=REFUSE  stor=REFUSE  build=OK  bakes=true
jesse_direct      colo=REFUSE  stor=REFUSE  build=OK  bakes=true
```

**So `RELEASE_NOTES.md:50` stands as written** — I have no counterexample to "refuses a given reference in any position", since the shape I offered never reaches a position. You were right not to soften it, and right not to mark it closed on my say-so either. The nesting sentence stands for the same reason.

The one piece worth keeping from that finding is narrow and not a gap: an ADMIT on a shape that then fails at `getSQL()` is fail-safe today only because the build stage refuses it. Not worth a change.

## Dispute 2 — confirmed. Narrowing is a fail-open and my suggestion was wrong

Measured at `1a8d09d4`, then with the descent narrowed to the query's own holders:

| case | marker | bakes | descent | narrowed |
|---|---|---|---|---|
| `join_arg_unread` | `[]` | false | REFUSE | ADMIT |
| `join_arg_read` | `[]` | **true** | REFUSE | **ADMIT** |

`join_arg_read` admits a source whose build SQL carries `org_id=1`. That's the leak this PR exists to close, reintroduced through the join, exactly as you said. Your reading of my own mutation evidence was also correct: outermost-only left the suite green because nothing covered the read-join case, and I mistook that silence for the descent being inert.

I checked `1a8d09d4` catches it — under the narrowing mutation, `refuses a given bound into a joined source the query reads` (`:812`) is the one failure, 33 pass / 1 fail. Restored, 34/34 green. Real behaviour assertion, not a change-detector. DCO signed, 28 spec lines added, zero removed.

And I agree with where you put the cost. Documenting the `join_arg_unread` over-refusal on the walk rather than pinning it in a test is the right call: a future signal precise enough to separate read from unread should land as a green improvement, not a red test someone has to decide whether to delete. The doc comment at `:481-483` now says exactly that, including that the descent is load-bearing rather than defensive — which is the thing I'd have gotten wrong reading the code cold.

## Where that leaves it

Of my three findings: the depth-1 argument channel was real and you closed it; the over-refusal is real but its fix is a fail-open, so keeping it is correct; the wrapped leak does not exist. Nothing outstanding from me. Sorry for the round-trip on the wrapped one — I should have measured `getSQL()` rather than stopping at the gate verdict.

girishjeswani (approved):
LGTM!

PR #1187, opened 2026-09-17
feat(connections): configEtag, an opaque writer-owned tag echoed on reads

housejester:
Adds one optional string to `Connection`. Publisher stores it, returns it, and never looks at it.

## Why

Credentials are write-only: they are accepted on create and update, and no read returns them. That is the right default, and it leaves one question unanswerable. A system that distributes the same connection to several Publishers cannot confirm from a read that any given one is still holding the credential it was last sent. `withheldFields` distinguishes a Publisher holding *some* password from one holding *none* — it cannot distinguish one holding last month's password from one holding the current one.

So a Publisher that misses a rotation and then restarts stays stale indefinitely, and nothing on the read path can see it. The push that delivers the rotation is the primary path and it works; what is missing is any way to notice when it did not land.

## What

`configEtag` is an opaque string the **writer** owns, on exactly the terms `fingerprint` beside it already establishes. Compute a tag over the configuration you are about to send, send the two together, and compare the tag each Publisher reports against the one you would send now — the same equality-only comparison an HTTP `ETag` supports, and the only operation defined on it. A Publisher reporting a different tag, or none, was not given that configuration.

Publisher does not compute, parse, validate or interpret the value. Its format is entirely the writer's.

## Not `fingerprint`

`fingerprint` identifies the *data* a connection reaches and deliberately excludes credentials, so that rotating one does not re-address the artifacts built through it. Two configurations differing only by a password therefore share a fingerprint — which is precisely the case this field exists to catch. The two answer different questions and neither substitutes for the other.

## Scope

| File | |
|---|---|
| `api-doc.yaml` | declares the field |
| `connection_public_view.ts` | one line in the allowlist |
| `connection_public_view.spec.ts` | 6 tests |
| `docs/connections.md` | a section under "Credentials in API responses" |
| `RELEASE_NOTES.md` | `[Unreleased]` |

No server logic changed. `mergeConnectionUpdate`'s existing shallow spread already gives the intended semantics — a patch carrying a tag replaces it, a patch omitting it keeps the stored one — so a client that never sets the field is unaffected, and a client that does never has to resend it alongside an unrelated edit.

**A write that does not carry a tag clears it.** The tag describes the configuration the writer that set it sent, so an update replacing that configuration without supplying one has invalidated it — inheriting it would assert that a config nobody has seen is still the one the tag names. It is also what keeps a writer convergent when it *stops* sending tags: it pushes its own config, the stale tag drops, and its next comparison agrees. Inherited instead, the tag outlives every config it described and that writer re-pushes on every poll forever. A null or non-object patch writes nothing and clears nothing.

**Why it is returned rather than withheld.** It names no secret, and a response that dropped it would report every Publisher as diverged on every poll, which is the opposite of the field's purpose. The contract-parity test forces that choice to be explicit: removing the single allowlist line fails 4 tests, including `classifies every Connection property as public or secret`, so this cannot later be reclassified as a credential by accident.

## What it does not tell you

Whether a Publisher is still *behaving* the way the configuration it echoes describes. An upgrade that changes how a stored config is parsed or defaulted moves the effective configuration without moving the tag. It reports what was delivered, not what is in effect. Documented as such.

## Testing

- `bun run test:server` — unit 3928 pass, integration 350 pass, 0 fail
- `bun run lint:server`, `bun run typecheck:server` — clean
- `bun run test:scripts` — 75 pass; `release-notes.mjs extract` parses the new section
- `packages/python-client/scripts/build-python-sdk.sh` — completes, generates `config_etag`

Verified the new tests fail without the change: dropping the allowlist line turns 4 of them red.

**Exercised against two running servers**, not only in unit tests: one built from this branch and one from
`main`, side by side, driven with the wire bytes a real distributing writer produces.

| Writer | Server | Echoed tag | `withheldFields` |
|---|---|---|---|
| sends a tag | this branch | byte-identical | `password`, `privateKey` |
| sends none | this branch | absent | `password`, `privateKey` |
| sends a tag | `main` (no such field) | absent | **`configEtag`**, `password`, `privateKey` |
| sends none | `main` | absent | `password`, `privateKey` |

The third row is the one that matters for a staged rollout: a server that does not declare the field still
reports it as withheld, because the public view is an allowlist and an unlisted field is hidden whole. A writer
that subtracts what a server says it withheld therefore sees no difference, so distributing the tag ahead of
the rollout is inert rather than a permanent mismatch.

Clearing was verified the same way: a server holding a stamped connection, given an update that omits the tag,
drops it on the first write and stays dropped on the second.

Re-run end to end against this branch after the update-connection change, on a proxied connection with a
pinned bastion host key: a create round-trips the tag byte-identically with `password` and `privateKey` still
withheld; a `PATCH /connections/{name}` **carrying** a tag sets it; the same `PATCH` **omitting** one clears it
while the rest of the update lands.


api-doc.yaml
```diff
@@ -4807,6 +4807,48 @@ components:
             environments on the same project with different
             impersonateServiceAccount values but the same fingerprint would
             share persisted-table ids.
+        configEtag:
```
Sha-Bang, line 4821: this lands on the shared `Connection` schema, so create (POST, which `$ref`s `Connection`) and read both carry the tag. the PATCH at `update-connection` doesn't — its request body is a hand-rolled inline schema listing `proxy` plus the ten per-type sub-objects and nothing else (api-doc.yaml:511-531).

combined with the clear-on-omit rule this PR adds, a writer on a generated client can't send `configEtag` on PATCH, and every PATCH it makes clears the stored one. the credential-rotation loop this field exists for is a PATCH, so through a generated client the feature's main path can only destroy the tag.

raw HTTP still works — the controller takes the body as `Partial<ApiConnection>` and passes it straight through, and nothing sets `additionalProperties: false` — so this is the declared contract rather than runtime behavior. confirmed by reading the spec and the controller; I didn't run it.

add `configEtag` to the PATCH schema, or model that body as a partial `Connection`?

packages/server/src/service/connection_public_view.spec.ts
```diff
@@ -597,6 +597,103 @@ describe("withheldFields", () => {
    });
 });
 
+describe("configEtag", () => {
```
Sha-Bang, line 601: these cover the two pure functions and stop there. the claim in the release notes and docs — Publisher stores it with the connection and returns it on reads — rests on `cloneApiConnection` being `{ ...connection }` (packages/server/src/service/connection_config.ts:182-184), which this PR doesn't touch and nothing here asserts.

it does hold today: every connection write replaces `this.apiConnections` wholesale with `malloyConfig.apiConnections`, those are the clones, and the tag never enters the Malloy pojo since that's assembled field by field. but break the round trip in your head and ask which assertion fails — none of these do. the plausible break is someone hardening `cloneApiConnection` into a field-by-field rebuild for exactly the allowlist reasoning that motivated `PUBLIC_CONNECTION` in the file beside this one, at which point the feature dies silently with all of these still green.

worth one test going connection-in → `assembleEnvironmentConnections` → `listApiConnections` → `toPublicConnection` → tag-out to pin it? traced by reading; I didn't run the suite.

Sha-Bang (approved):
two inline, neither blocking. the PATCH one is the one I'd want fixed before this gets used in anger — as declared, the endpoint the rotation loop actually calls can only clear the tag, never set it. approving.

one note on the name rather than the code: `configEtag` reads as an HTTP ETag, and a reviewer I ran this past immediately expected `If-Match` and a 409 on mismatch. it's documented as equality-comparison-only so that's a naming question, not a defect — but it'll come up again.

all from reading; I didn't run the suite.

api-doc.yaml
housejester: Confirmed, and it was worse than an omission: with the clear-on-omit rule this PR adds, a writer on a generated client could only ever *destroy* the tag through the endpoint it rotates credentials on. Fixed in d0381a6 — `configEtag` is declared on the `update-connection` body.

I took that over modelling the body as a partial `Connection`: that pulls in `name`, `type`, `fingerprint` and the metadata bags and changes the update contract well past what this PR is for. The one-field addition makes the rotation path expressible without touching anything else, and if the body is ever generalised the field is already there.

Verified against a running server rather than just the spec — a `PATCH /environments/{env}/connections/{name}` carrying a tag now sets it, and the same request omitting one clears it while the rest of the update lands.

packages/server/src/service/connection_public_view.spec.ts
housejester: Right, and the reasoning about how it would break is the part I'd missed. I'd checked the round trip by hand against a running server, which protects against nothing a refactor does later.

Added in d0381a6, going connection-in → `assembleEnvironmentConnections` → `toPublicConnection` → tag-out. Rather than trust that it covers the case, I applied the break you described — replaced `cloneApiConnection` with a field-by-field rebuild — and confirmed it is the single test that goes red, with the other 91 in the file still green.

Also pinned that the tag does **not** reach the Malloy pojo. That holds today only because the pojo is assembled field by field, and it is worth asserting for the same reason as the clone: the value is derived from a credential, and the engine config is not somewhere anyone audits for one.

PR #1190, opened 2026-09-17
fix(sdk): let Malloy's parser locate every dashboard edit, so the builder stops maintaining a second grammar

Sha-Bang:
The dashboard builder **splices**: it rewrites only the bytes it owns in a `dashboards/*.malloy`
file, so comments, formatting and unmodeled Malloy survive an edit. That contract is right and is
not what changes here. What changes is how it decides *which bytes it owns*.

Until now it answered that itself, by re-deriving Malloy's grammar from raw text. That is the
decision this PR reverses, and the reason is structural rather than a list of bugs: **a text scanner
is a second, unversioned copy of Malloy's grammar, and a copy drifts.** It drifts from the compiler
as Malloy gains syntax, and — worse — the copies drift from *each other*. On `main` the reader and
the writer each answer "where does this declaration end" with their own code, so any disagreement
between them is a corruption nobody wrote and nobody can see.

Nobody can see it because **neither gate can**. The read-back gate compares the *projection* — tiles,
tags, filters — so text stranded beside a clause it rewrote is invisible: the orphan is not a filter,
so the comparison never looks at it. The parse gate catches output Malloy cannot parse, not output
that parses and means something else. Both of those hold no matter how carefully any individual
scanner is written, which is why this is not fixable by patching them.

Two consequences were live on `main`:

| Input | Edit | Result |
|---|---|---|
| `view: kpis is vx + vy + { where: a ~ $A }` | change the tile's width | the refinement is **deleted** |
| `where: created_at::date >= $SINCE` | bind `$SINCE` | a second `where:` is written; the tile filters twice, and only one is removable |

Seven local review rounds on this branch each found defects, and nearly every one was the same
thing: two scanners disagreeing about where something starts or ends. That is the signal that the
approach, not the code, is the problem.

## Why the parser is the right substrate, permanently

**The grammar stops being ours.** `malloyTree.ts` is the only module that touches the parse tree, and
it calls the compiler the product already depends on. When Malloy gains syntax, the locator follows
for free; there is no second grammar to keep in step, and no class of Malloy the builder silently
mis-locates because nobody taught the scanner about it.

**Reader/writer disagreement becomes impossible rather than unlikely.** Both locate through that one
module. There is no second opinion available to disagree with, so the defect class that produced
seven rounds of findings has nowhere left to live. The rule that keeps it that way: *locators come
from the tree; recognizers may read text over a tree-bounded span; emitters take positions only from
a span or a token.*

**Depth, identity and references become structure instead of heuristics.** A `nest:`'s `where:` lives
under its own node, so it can never surface as the enclosing block's however the braces are written.
A managed filter is a comparison whose right side is a given reference — so `where: a ~ $A and c = 1`
is an *and*, structurally not a binding, and is left alone with no isolation heuristic and no
masking. And every `$GIVEN` reference is known exactly, so binding a given a `where:` already filters
on is refused rather than silently doubled; the text scan for `$NAME` it replaces was fooled by an
apostrophe in a comment.

**The premise that kept the parser out does not hold.** `docs/dashboard-builder-plan.md` justified the
scanners by saying the parser "misreads a refinement spelled `+ { limit: 5, where: … }` (it compiles)
by dropping the next view declaration". That parenthetical is false — a comma after `limit:` is a
syntax error, and the builder's own writer was *emitting* that spelling, with specs asserting it as
correct. Measured on the repo: 86 declarations across the 46 valid files, zero dropped. The scanner
layer was working around a defect the builder was manufacturing. That doc is corrected here.

**What was closed off, so the next person does not re-open it.** A sidecar or manifest is ruled out by
the storage contract (`docs/dashboard-builder-plan.md:34-39`) and, decisively, because a
`where: … $GIVEN` binding is not metadata but the query — the server derives each tile's `givenNames`
from the *compiled model* (`resolveTileGivens` in `packages/server/src/service/dashboard.ts`), so a
binding held elsewhere would not exist to the compiler and the control row would not know it. A tree-sitter
grammar means owning a second grammar again. `prettify` reformats. Regenerating from the AST loses
comments and formatting. Upstream has no AST-edit API; Malloy's own VS Code quick-fix does exactly
what this does — the compiler hands it `{range, replacement}` and it writes bytes.

**Cost.** Parsing happens on open and save, never per keystroke — measured at 6-9 ms on this repo's
real dashboards and models. Consolidating on one translator also takes a save from about four parses
to two, since the tree and the tokens now come from the same one.

## What that costs in code

This is not a smaller diff, and it should not be sold as one. `malloyTree.ts` is 846 lines of
explicit locator, and `spliceDocument.ts` grows because its refusals are now explicit rather than
implicit in a scan that quietly did the wrong thing. What shrinks is the surface that re-derives
Malloy's grammar: `malloyText.ts` goes 151 → 46 lines, and `declarationsUnder`, `declarationLine`,
`givenDeclarations`, `sourceBase`, `viewBody`, `BINDING_CLAUSE`, `declarationEnd` and `literal` are
gone. What is left in `malloyText.ts` parses `tiles=[…]` entries and tag lines — grammars that are
genuinely ours, not Malloy's.

Two hazards the adapter is written around, both silent rather than crashing:

- **The parser counts code points; JS strings are UTF-16.** On a line containing an emoji the parser's
  offset for the next clause is two units early, and slicing by it cuts a surrogate pair in half. One
  code-point map per parse converts every offset once.
- **An error-recovered tree is confidently wrong** — it invents a phantom view, reports an inverted
  range, and drops every later declaration, reporting none of it. A file with any syntax error is
  never edited through the tree: the editor opens it read-only with the compiler's message and line.

Node identity is by accessor, never `constructor.name`: `@malloydata/malloy` is a peer dependency, so
the host bundler minifies it and every class name becomes a single letter. A walk keyed on class
names passes `bun test` and fails in production. Because it is a peer dependency, a CI spec cannot
protect the host either — so the adapter asserts on first use that the tree is shaped as expected and
refuses with the reported `malloyVersion` if not. That runtime assertion is the only protection a
host that resolved a different version ever gets.

## Fourth tile kind: `opaque`

A view declared *in this file* whose body the builder will not rewrite — a chained `+ {…} + {…}`
refinement, or a `->` pipeline from a named view — used to read back as `inherited`, so its settings popover said "Declared on its source,
so its title and layout are set in the model." It is declared right in front of you, and its tags are
editable; only a filter has nowhere to go. It now has its own kind carrying the reason, and the menu,
the filter dialog and the writer each say what is true about it.

## Comments are nobody's to delete

A comment is trivia to the parser: it sits outside every span. So a writer that deletes a range
wider than one node's own text takes the comment with it, and **neither gate notices**. Nine of
those turned up on this branch, eight returning success.

Four were found by reading the code: collapsing a `+ { where: … }` whose base carried a trailing
comment; inserting a filter after a statement whose trailing comment then slid onto the new
`where:`, saying something its author never wrote; removing a clause whose separator spanned a
comment; and removing a tile, which widened the cut to whole lines and took a second declaration
sharing that line — one that is not a tile of this dashboard, so the read-back gate cannot see it
go.

The other five are about **which spellings are comments at all**. Malloy lexes `//` and `--` as
`COMMENT_TO_EOL` and `/* … */` as `BLOCK_COMMENT`, and the builder knew only about `//`:

| Input | Edit | Result on `main` |
|---|---|---|
| `where: a ~ $A, /* keep */ b ~ $B` | unbind the tile | the block comment is **deleted**, and the guard that exists to stop that cannot see it |
| `where: a ~ $A` alone in a body, every clause the builder's | unbind the tile | the statement is cut with **no comment guard at all** — the sibling clause path got one, this one was missed |
| `# colspan=6` above a `--` or `/* … */` line above the view | change the tile's width | a **second** `# colspan` is written below the comment |
| the same, with the `/* … */` running over several lines | change the tile's width | as above — found by the new sweep *after* the single-line case was fixed |
| `# label="…"` written **inside** a `/* … */` | change the tile's width | the comment's prose is read as a tag, so the reader reports a label the tile never had and the writer rewrites a line in the middle of a comment |

The last one is the quiet one. The reader takes a tile's tags from the parser, which reads straight
past a comment; the writer walked the same block as text and stopped at anything that was not `//`.
So the tag above the comment was visible to one and invisible to the other, the writer added rather
than patched, and the reader read the lower copy back — leaving the read-back gate satisfied by a
file now carrying two. That hand-written walk is deleted rather than taught a third spelling: where
a tag block begins now comes from the lexer, the comment index registers every line a comment
touches, and both sides skip those lines rather than reading a `#` inside a block comment as an
annotation. The last row is the one to note: it is a hazard the *fix* introduced, found by probing
the fix rather than by the specs written for it, and it is why the sweep is the deliverable here
rather than the three patches.

**One rule, where there were three.** Tile removal deleted a comment trailing its last line,
`where:` removal left one stranded above the closing brace, refinement collapse refused. All three
now answer the same way: a comment goes only with a declaration you asked to delete **outright** —
removing a tile still takes its own comments, and the builder shows that diff before a structural
save — while a filter edit, which rewrites a declaration that stays, refuses rather than destroying
or stranding a comment it was not asked about, naming the comment in the reason.

## One invariant, in place of one spec per corruption

Each of the defects above got a hand-written spec of its shape, and then the next sibling of that
shape shipped anyway. That is the real lesson of this branch, so the fixes come with the thing that
would have caught them as a class: **an edit changes the declaration it names, and nothing else.**

`__test__/inventory.ts` takes an inventory of a file from a **fresh parse** — declarations with
their tags, blocks and indents, every comment, the code each trailing comment sits beside, and the
statements nothing models (`run:`, `query:`) — and says what moved. Two things it deliberately is
not: it never reads the writer's own spans, since every defect here *is* a wrongly computed span and
an oracle phrased in those numbers certifies the mistake; and it is not "the bytes outside the
planned edits are unchanged", which is vacuous, because the writer applies exactly the edits it
planned.

Two details carry most of its weight. Trailing comments are keyed **per statement**, not per
declaration — the comment that slides lives on a line *inside* a view body, so asking the
declaration about its own first line looks at `view: kpis is {`, which never had one. And modelled
`#` tag lines are **masked to their key rather than dropped**: `# colspan=6` and `# colspan=3` both
read as `#colspan`, so a legitimate retag moves nothing while a duplicated tag moves the multiset.
The verbatim lines are kept beside the masked ones, because masking is exactly what cannot see
`# label="A"` become `# label="B"` on a declaration nobody asked to retag.

`invariants.spec.ts` runs it over one fixture carrying every shape at once, and **every row declares
whether the writer must write or refuse**, with a mismatch either way failing. That is not
ceremony: reverting the two-line `given:` fix turns a success into a *reasoned refusal*, and the
corpus sweep — which rightly treats a refusal on a real file as a legitimate answer — would have
stayed green over the defect it exists for.

The corpus sweep stays as a tripwire and gains the removal edit kinds, but measured it cannot be the
guarantee: three of the five real dashboards declare every tile on their source, so removal never
reaches the deletion path in them, filter and control removal have inputs in one file each, and none
of the five contains a single trailing comment — the shape behind one of the corruptions.

## Refusals barely widen

On a valid file the tree locates everything, so what is left is product shapes: a pipeline starting
from a named view, a compound `{…} + {…}` body, a tile inherited from the model. Several old refusals
go away — a `where:` clause list carried onto a second line is now edited exactly, a `{…} -> {…}` body
takes its binding in stage one, and a given on a continuation line rebinds without doubling.

## Acceptance: twelve reverts, twelve reds

Red is not the bar; red **in `invariants.spec.ts` alone, on the named field** is. Reverting the
trailing-comment anchor turns `statementExtent.spec.ts` red on its own, so a whole-suite run would
look like a pass while the new sweep sat green and blind. Each fix was therefore reverted one at a
time, typechecked (a revert that fails to compile proves nothing), and the new file run **in
isolation**:

| Reverted | Failure in `invariants.spec.ts` |
|---|---|
| `planReferenceFilters` comment guard | row expected `refuses`, got a write; `// the trend, on the right` gone from the output |
| `planStageFilters` clause guard | as above, over the block comment |
| trailing-comment anchor | `attached` — `+where: cat ~ $CATEGORY`, `-aggregate: n is count()` |
| `planRemovedTiles` `sharesLine` | `declarations` — `-view:a.sidekick`, a declaration nobody asked to remove |
| `planGivens` `lastLine` | row expected `succeeds`, got a refusal |
| whole-`where:` comment guard | row expected `refuses`, got a write; the block comment gone |
| `BLOCK_COMMENT` in the comment index | three rows, two flipping refuse→write |
| the hand-written tag-block walk | `tags` — `~view:a.kpis` and `~view:a.counts`, each carrying two `# colspan` |
| block-comment continuation lines | `tags` — `~view:a.counts` |
| the writer's skip of comment lines | `tags` — a `# label` rewritten inside a comment |
| the reader's skip of comment lines | `tags` — a label the tile never had, written out for real |
| a comment line meaning "nothing else on it" | `declarations` — `~dimension:a.helper`, a bystander whose `# label` the retag deleted |

Twelve for the eight the plan called for, because each fix turned up the next: after `--` was fixed
a `/* … */` between a tag and its declaration still hid it; after that a multi-line one still did;
once the walk reached past a block comment, a `#` line *inside* one became a tag; and review found
that a line holding a comment **and** a declaration read as wholly comment on both sides, so the walk
above a tile stepped over the declaration and rewrote the `#` tag belonging to *it*. Asking "is there
a comment on this line" is what all four have in common; the index now blanks comments out and asks
what code is left, which answers the block-above and trailing questions from one place.

The oracle carried the same wrong idea, and the honest result is worth recording: it was a **false
positive**, not a blind spot — with the writer correct it called a correct retag damage. What let the
defect through was the fixture, which put no comment and code on one line. Both are closed.

## Tested

- `bun test` in `packages/sdk`: 1044 pass, 0 fail (987 before), against a tree merged with the
  current `main`. `typecheck:sdk` (both tsconfigs, with `src/**/__test__/**/*` added to the spec
  config so the oracle is checked rather than merely excluded) and `lint` clean.
- The twelve-row gate above, run per row and reported.
- The corpus files still round-trip byte for byte on a no-op, and the four removal cases that have
  inputs write cleanly: the inventory reports only the declaration that was asked for. Measured, not
  assumed — the removal sweep early-returns on the files that have no input for it, and the four that
  do were confirmed to write rather than take the refusal escape hatch.
- 22 specs pin the parser adapter directly, including the astral-character offset, a `//` inside a
  string literal, a `"""` block containing `where:`, a backtick field name with an apostrophe in it,
  and refusal on a syntax error. The lexer probe behind the two token kinds is in
  `malloyTree.spec.ts`.
- A browser run against a local server, diffing the file the server actually serves. Changing the
  width of a chained-refinement tile now touches **only** the `# colspan=` line, leaving both
  refinement blocks and the comment between them intact. Adding a filter to a tile whose `where:`
  carries a `::` cast appends one clause beside it; removing it returns the file byte for byte. The
  comment fixes are writer-internal and covered by byte assertions rather than a second browser
  pass.


jswir (commented):
The approach is right, and the case for it in the description is the correct one: the scanner was a second copy of Malloy's grammar, and the reason to delete it is structural, not the bug count. The `invariants.spec.ts` oracle is the right shape for the problem — an inventory of what moved beats one spec per corruption, and the eleven-revert table is a real acceptance check rather than a claim.

Two defects need fixing first, and both are the old failure mode living in the new code.

**1. A comment sharing a line with code makes that line invisible as a declaration — and the resulting write reports success.**
[`malloyTree.ts:842`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/malloyTree.ts#L842)

`commentIndex` decides a line is wholly a comment by looking only at what precedes the comment token:

```js
if (r.text.slice(r.lineStarts[line], start).trim() === "") lines.add(line);
```

So `/* keep tidy */ dimension: helper is 1` lands in `commentLines` despite carrying a declaration. `blockStart` then walks straight over it instead of stopping there, and collects a `#` line belonging to something else.

Reproduced against the real code. Editing only `revenue`'s label:

```malloy
# label="Foo Label"
/* keep tidy */ dimension: helper is 1
# label="Revenue Tile"
view: revenue is by_category
```

rewrites **both** `# label=` lines via `planTilePresentation` ([`spliceDocument.ts:1011-1033`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/spliceDocument.ts#L1011-L1033)) and returns `ok: true`. `DashboardDimension` carries no tag field, so the readback gate has nothing to compare and cannot see it. With a `given:` as the victim it's caught — as a spurious refusal of a valid edit.

There's a second site with the same mistake, two lines below:

```js
for (let l = line + 1; l <= r.line(end - 1); l++) lines.add(l);
```

This marks the closing line of a multi-line block comment as wholly comment too, so `*/ dimension: x is 1` fails the same way. Worth closing both together — and worth a fixture in `invariants.spec.ts`, since the oracle should have caught this and no existing case puts a comment and code on one line.

**2. `readDimensions` splits on the word "is" when the tree already hands you both halves.**
[`malloyTree.ts:579-603`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/malloyTree.ts#L579-L603)

`FieldDefContext` exposes `fieldNameDef()` and `fieldExpr()` directly. Running this PR's own `parseMalloy` on ``dimension: `is` is 1`` — valid, parses clean — returns ``{ name: "`", expression: "` is 1" }``. It reaches the drill dialog's rows and the by-name lookup at [`spliceDocument.ts:757`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/spliceDocument.ts#L757). `malloyTree.spec.ts` has no dimension coverage. Small blast radius, but it's the last "re-derive it from text" in the module built to end that.

**Worth doing now, same cost:** the shape assertion covers the tree but not the token stream. `commentIndex` reads `tokenStream.getTokens?.() ?? []` with no equivalent check ([`malloyTree.ts:819-849`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/malloyTree.ts#L819-L849)). Since `@malloydata/malloy` is a peer dependency, this is the scenario the PR itself names — and by its own argument the runtime assertion is the only protection a mismatched host gets. An empty index means every comment guard passes clean while comments are deleted, and `readTags` starts reading a `#` inside a `/* */` as a real tag. The readback gate can't help; `comparable` only compares tiles, drills and givens, as the comment at L1486 says.

**Follow-ups, not blockers:**

- `docs/dashboard-builder-plan.md:272-276` still describes the architecture this PR replaced — `malloyText.ts` as the shared textual reading, and no mention of `malloyTree.ts` in the module list. §1 was corrected; this section looks missed.
- `opaque` has no release note and no mention in `docs/dashboards.md`. It's a user-visible classification with its own popover copy and its own refusal, and the false "set in the model" message it fixes clears the release-notes bar.
- Of the two shapes `document.ts:109-111` says produce `opaque`, only the chained refinement is tested end to end. The `->` pipeline is asserted once at tree level as `kind: "unsupported"` and never traced to a document.
- `TreeSource.statement`'s doc comment claims the `#` tag block is included; `readSources` doesn't do that and `TreeSource` has no `tags` field. Nothing reads it today, which is exactly why it's worth correcting now.
- `syntaxErrors` ([`spliceDocument.ts:34-48`](https://github.com/malloydata/publisher/blob/cb150017065f89451820181d46e992f280616ca3/packages/sdk/src/components/DashboardBuilder/spliceDocument.ts#L34-L48)) keeps a second hand-written cast of the translator's problem shape. The need is legitimate — it wants the message list, not a single reason — but export a shared helper rather than keeping two `as` casts in step by hand.

**On the `resourceUri`/`versionId` work.** It's bundled in here with its own release note, docs section and spec, and isn't mentioned in the description. I reviewed it separately and it's good: the pin reaches all six reads including tile and suggest queries, it's in every cache key, and Save is genuinely off rather than a disabled button. I reverted the `12ec04e5` fix and confirmed the spec catches it. One fix — `docs/dashboards.md:731-738` and the release note list what the pin covers and omit tile queries and suggest queries, which *are* pinned. The `<Dashboard>` docs one section up have the right wording. No reason to hold it up; it just reads and reverts more easily on its own.

**Checked and clean:** code-point→UTF-16 conversion applied at every site; no `constructor.name`/`instanceof`; both entry points gate on `parseRefused()` and the splicer re-parses after applying; `opaque` is exhaustive across `TileMenu`, `FilterDialog`, `controls.ts`, `readDocument.ts`, `spliceDocument.ts`; no per-render parsing; `inventory.ts` genuinely avoids the writer's spans and uses the real compiler.

jswir (approved):

PR #1192, opened 2026-09-17
feat(materialization): a tenant-scoped source can be materialized into a storage destination, and laid out by partition=

housejester:
A source whose `where:` reads a given — `where: org_id = $ORG_ID` — could not be materialized at all. An author with a multi-tenant model had two options: give up the storage tier, or hand-build one package per tenant with the values written in as literals. There was no middle.

This adds the middle. **One artifact instead of N, or instead of none.**

It is two capabilities, independent of each other, that compose:

- **Caller-varying predicates** — such a source is built **once**, holding every caller's rows, and each caller is served exactly their own. This is the admission: it widens what may be materialized.
- **Layout** — `#@ persist partition="org_id"` writes the stored table one directory per distinct value. Independent of the first: it works on a static source, and a caller-varying source with no `partition=` builds flat and still serves correctly, losing only pruning.

The invariant, stated once: **a routed query returns exactly the rows the live query would have returned for that caller, read from a frozen artifact.**

Author-facing semantics live in `docs/materialization.md`, which this adds to.

## Why these sources were always eligible

A caller-varying predicate was never part of the relation being stored. It is part of the question being asked. Materialization stores relations, not questions — so a predicate that varies per caller was never a candidate for freezing, and a gate refusing the source was refusing it for a property it does not have.

Malloy's own behaviour is the evidence: the SQL that builds a persist source is the persisted relation alone, and the source's extend-block `where:` is not in it. It refines the relation when the relation is *read*.

So "stripped" below is describing what the build already omits, **not an edit Publisher makes to the model**. There is no rewrite step, and therefore no rewrite to get wrong — the work here is deciding which sources qualify, and putting the terms back at read time.

The useful part is that the boundary is **mechanically decidable**. If changing a caller's value changes the build SQL, the predicate is in the artifact and the source must be refused. If it does not, the predicate is a read-time property and is re-applied per caller. That is not a description of the rule, it is the rule, and it is executable: `persist_bakes_oracle.spec.ts` compiles twelve shapes twice, changing only a given's *default*, and compares the SQL. Every shape whose SQL moves is refused. It asserts one direction, since the gate over-refuses deliberately, plus a case proving the corpus exercises both branches so it cannot pass by testing nothing.

Worth having because the gate decides by position, using a marker that summarises what a pipeline *references* — and that marker has already drifted from what the build *bakes*, at a source argument bound but never referenced.

## Partition-per-tenant, and what it is worth

```malloy
given:
  ORG_ID :: number is 1

#@ persist name="orders" storage=lake partition="org_id"
source: orders is raw -> { select: * } extend {
  where: org_id = $ORG_ID
}
```

One table holding every org's rows, written `org_id=1/`, `org_id=2/`, … and caller 1 reading only `org_id=1/`.

The property to note is an asymmetry: **read cost tracks the caller's own partition; build cost tracks the artifact.** A caller's query does not get slower as tenants are added; the build does, since it is one pass over everyone's rows. The pruning holds because the scoping term is an equality on a partitioned column, and partition count is not free — a very high tenant count is the many-small-files case, which the cardinality warning flags.

That asymmetry is the argument for the composition. The obvious alternative, one artifact per tenant, has the same read cost — but N builds, N schedules, N manifest entries, and N things that go stale independently. One partitioned artifact has **one** build and **one** freshness story at per-tenant read cost.

Three sets make it work. **D**, the dynamic terms: every conjunct of the extend-block `where:` that reads a given, each left out of the build and re-applied at read. **P**, the partition columns: a declared, ordered list, free to choose within what D offers — `partition="org_id"` is legal for a source that also filters by `user_id`, and the user term filters inside the org's files. **R**, the residual: D minus what P covers, re-applied as ordinary filters over the pruned partitions, and free to be any expression the live path accepted, since it compiles against the virtual source with the same givens the live query had.

P decides which stripped terms prune files and which filter rows. It decides nothing about which rows a caller sees.

## Two things this unlocks that are easy to miss

**Joins to caller-scoped dimensions.** The serve transform emits a join only when the joined source is itself materialized. So persisting a caller-scoped dimension table is what makes a join *to* it servable — without it the query does not merely lose the join, it loses the tier and falls back live.

**Cross-warehouse joins.** Two sources on different connections cannot be queried together live. Materialized into one lake they are siblings, and the join compiles.

## What the gate will tell you, and what to do about it

Each refusal names a position and the move that fixes it.

| where the given sits | in the artifact? | what you get |
|---|---|---|
| the source's own `where:` | no | **materialized**, term re-applied per caller |
| inside the persisted query, incl. via the source it reads | yes | `given_in_persisted_query` — move the term to the extend block |
| bound as a source argument | yes | `given_in_persisted_query` |
| a declared `dimension:` or `measure:` | no | `dynamic_projection` |
| a join's `on:` | no | `dynamic_join` |
| a given-scoped source reached through a join | no | `dynamic_joined_where` — enter through a non-persisted extension |

The bottom three are refused although the build leaves them out. None reaches the artifact, so none is a leak; they are refused because whether the serve shape reproduces them is a separate question this change does not answer, and the gate refuses what it has not proven. Coverage is recoverable later; isolation is not.

**And the author now hears about it.** A refusal used to be computed, written to `refusedSources`, and read by nobody — the package published, the source served live, and whoever wrote the annotation was told nothing. Each refusal is now a package warning carrying the gate's own message. It is the one materialization finding the build plan cannot be read for instead, since a refused source is left out of `sources` entirely.

**One shape moves the other way.** A source that only *reaches* a given-filtered source through a join its query never reads is now admitted, because Malloy prunes that join from the build SQL. That rests on Malloy's pruning, so `materialization_eligibility.spec.ts` asserts the build SQL on both halves — the filter present where the join is read, no `JOIN` at all where it is pruned. A Malloy version that stops pruning turns that red rather than turning the gate unsafe.

## Incremental refresh over a shared artifact

A caller-varying source can be incremental too, and `merge_key=` keeps meaning what the author wrote.

It needs help to. A merge key is chosen against the source as written — `order_id`, unique within one org — while the stored table holds every org's rows, so the key is ambiguous over it. The merge therefore matches on the declared key **plus** the columns the stripped terms filter on, which is the relation the author chose the key against. Nothing changes in the model. Scoping is all-or-nothing: a term resolving to no column of the source refuses it (`merge_key_scope_unresolved`) rather than scoping partially, which would be narrower than the bare key, still wider than the author's relation, and look like it works.

**This also fixes a live bug on the colocated tier**, whose gate has never refused a given and whose artifact is widened the same way, so an unscoped merge is reachable there on released code. On the storage tier it is reachable only once this branch admits such a source, which is why the scoping ships with the admission rather than after it.

Measured before the fix on both tiers, because they fail differently and only together show it:

| tier | unscoped merge |
|---|---|
| colocated, Postgres | build fails: `MERGE command cannot affect row a second time` |
| `storage=`, DuckLake | build **succeeds**, one caller's row destroyed and another's duplicated, no error |

Postgres refusing the ambiguous match is a property of the target warehouse rather than anything Publisher guarantees, and DuckDB does not do it. Both are pinned by `incremental-mz/a-merge-key-must-not-match-rows-outside-the-callers-scope` and its `-on-storage` sibling.

## Rollups stay out

A rollup over a caller-varying source is refused on its own grounds (`preaggregate_over_dynamic_source`). A persisted source may carry a caller's predicate because the term is put back when the rows are read; a rollup is stored pre-grouped and answered from directly, so there is no read for a term to be re-applied at.

Naming it matters beyond the message. A rollup deliberately skips `serveBindingsWithRefinements`, because its generated name matches nothing in the author's model — so it gets no filters re-applied, no givens declared, and no shape compile that could fail closed for it. If the rollup build ever learned to strip its base's read-time terms, which looks like an optimization rather than a security change, a rollup over a caller-varying source would start aggregating across every caller with nothing on that path to catch it.

## A no-op for everything already in a lake

`scopeColumns` is empty for any source that is not caller-scoped, so its merge is the merge it always was. A source declaring no `partition=` addresses byte-identically to before — asserted against the old formula rather than against another unpartitioned source — so nothing already materialized rebuilds. A build with no partition list issues the same single statement it always did.

## Serving

The serve shape declares the givens the author's model declares, and a routed query keeps its given values instead of having them withheld — withholding was correct only while the shape was built from sources mentioning no givens. A secure given has no default, so the shape's origin probe binds the request's givens; without that such a source fails to compile and falls back to live.

## Entry points above materialized sources

Where a caller's scope comes from a joined source's own filter, the arrangement is a non-persisted entry point over materialized sources. The shape can now carry such a source: rebound over its materialized bases, cut down to what it adds, and refused outright if it declares a join reaching anything not materialized.

Cutting it down is the fiddly part. When you write `source: B is A extend { … }`, Malloy copies A's fields onto B, so B's compiled form looks like it declares every dimension, measure, view and join that A does, and A's `where:` clauses sit at the front of B's filter list. Re-emitting any of those is `Cannot redefine`. The shape is one model, so that does not just drop B — it fails the whole shape and every source in the model loses the tier. The lift keeps only what B adds: fields A does not have by name, and filters past where A's stop.

It fails closed twice. A source is carried only if its base is already on the shape, so a base that is not materialized — or whose binding the freshness gate withheld — leaves it off and the query falls back to live. And a source declaring a join the lift cannot carry is dropped entirely rather than emitted without it: dropping a join quietly is safe on a materialized source, where an unreferenced join is pruned and a referenced one fails to compile, but here the source's own `where:` may be reading the alias. Lifts are attempted only on the richest rung of the escalation ladder, so every rung below is the shape the package compiled before.

**It is not reachable yet**, and that turns on `#@ -persist` rather than on the lift. A derived source over a persisted base has two states and needs a third: written plainly it inherits the base's `#@ persist` and becomes a build target of its own; written with `#@ -persist` it is documented as recomputed "instead of using the pre-built table", which `opt-out-persist-recomputes` pins. The lift honours that annotation by excluding it. `partitioned-mz/a-query-term-over-materialized-sources-still-scopes` stays known-red with the detail.

## `partition=`, mechanically

DuckLake applies a layout only to files written after it is set, so a partitioned build creates the table empty, `ALTER … SET PARTITIONED BY`, then inserts — in one transaction. The physical name is self-assigned and stable across generations, so a rebuild writes to the table currently being served; without the transaction, `WITH NO DATA` would empty that table under live traffic while routed queries still reported `servedFrom: storage`, and a failed insert would leave the stranded-table cleanup deleting a table the manifest still points at.

The layout joins the content address, so changing `partition=` rebuilds. A layout also survives later writes, which is what lets `partition=` coexist with an incremental refresh: measured on a real DuckLake with `ducklake_default_data_inlining_row_limit=0`, the setting the build session applies, a delta `INSERT` and a `MERGE` both write partitioned files, including into a partition directory absent at seed time.

## Tests

- `model_storage_serve_givens.spec.ts` — one artifact of three rows across two orgs answers 30 for org 1, 7 for org 2, 30 unbound, 0 with the tier off. The per-caller split shows the numbers came from the stored table rather than the live relation.
- Wholeness — a term the shape cannot reproduce withholds the binding and serves live, never the unfiltered artifact. Mutation-checked against a lenient emission, which returns the whole artifact.
- Partitioned build — statements and order, a rollback leaving the previous generation intact on a real DuckLake, a committed rebuild still writing one directory per value.
- `serve_shape_derived_lift.spec.ts` — compile-backed against a real `Runtime`, since the selection turns on what the compiler does to an extending source. Mutation-checked: removing the field subtraction, the filter subtraction, or the all-or-nothing join rule each turns exactly the assertion meant to catch it red.
- `hammer/scenarios/partitioned-mz/` — six scenarios; `incremental-mz/` gains two.
- 3994 unit tests. Hammer 99 PASS / 2 KRED; the single FAIL (`quoted-persist-name`) is pre-existing on `main`.

`## Query` in the hammer grammar gains `servedFrom:`. A correct tier and a live fallback return identical rows, so rows cannot tell them apart, and neither can `expect binding:`, which shows a binding exists rather than that it was used.

## Where this departs from the design

- **`ManifestEntry` did not gain the new fields.** The serve side does have the model — `serveBindingsWithRefinements` reads `modelDef.contents[…]` at serve time, already where filters and refinements come from — so they went on `PersistSourcePlan`.
- **The wholeness precondition has no check of its own**, because the fail-closed shape compile already enforces it. It has a test instead.

## For consumers

`StrippedTerm` is a new schema. A drift check comparing schemas present in both specs cannot ask for one missing from the consumer's copy, so it has to be added by hand at the version bump. The enum extension is caught normally.


girishjeswani (approved):
Reviewed this across three passes — isolation and the gate, partition/incremental, and tests/docs/API — because a bug here leaks or writes another tenant's rows. The isolation property holds, and it holds structurally rather than by care, which is the answer you want on a feature like this.

## What I verified

**28 of 28 adversarial served-vs-live cases came back equal, per caller.** Two callers against one artifact, a caller matching no rows (empty, not everything), a disjunction, multiple conjuncts, non-equality operators, a mixed static/dynamic `where:` in both directions, and the no-value-with-default case. Compiled and run against real models, not reasoned about.

**No path serves an artifact without its scoping term.** `filter` is in `NEVER_THINNED`, which `buildServeShapeTiers` puts in every rung including both floors; an unreproducible filter is emitted as deliberately-invalid Malloy so the shape fails to compile and the binding is *withheld* rather than thinned; and the givens are declared at every tier, so no rung carries a term without its declaration. `model_storage_serve_givens.spec.ts` demonstrates the whole path by observation — the shape fails at tiers 0/1/2 and the floor, logs `serving live`, and the assertion is that the answer is 0 rather than 37.

**The #1180 reassociation hazard cannot arise here by construction.** `collectDynamicTerms` takes each `filterList` entry's `code` whole and verbatim, never splitting a conjunct or joining two entries, and the transform emits one `where:` per entry. That is also why non-equality operators work: the D-set never parses the predicate, so there is no operator-specific extraction to get wrong.

**The oracle is a real independent check.** I removed the `argumentBindsAGiven` call — reintroducing exactly the #1185 source-argument leak — and `persist_bakes_oracle.spec.ts` failed cleanly, naming the shape. After `givenUsage` turned out to be an unsound oracle in #1185, replacing an argued boundary with an executable one is the right move.

**Nothing from #1180 or #1185 was lost in the `-218`.** `argumentBindsAGiven` and `buildSubstitutesAGiven` moved verbatim into the new module, the blanket `referencesGiven` was deliberately retired (it is what this PR replaces), and five refusals are new. A reason-by-reason census against the merge base subtracts nothing.

**Merge-key scoping is safe, with three independent layers.** The emitted `ON` really does carry the scope columns NULL-safe-OR'd in. A mixed resolving/non-resolving shape sets `scopeIncomplete`, both gates refuse, and `incrementalLineage` returns `undefined` — so even with both gates bypassed the merge is never issued and the refresh seeds instead. Expression terms and computed-dimension terms refuse rather than mis-scope.

**`partition=` does not affect row visibility.** `extractSourceFilters` re-emits every `filterList` entry unconditionally; the word `partition` appears in that file only in comments. Probed with two given-terms and `partition="org_id"` — both survive, P names one, R still applies.

Also: `api-doc.yaml` is fully additive (the `-12` is description prose only, no path, response code, enum value or `required:` removed, and `given` correctly retained rather than deleted); every documented refusal reason exists in code with the exact string; refusals are now observable both as a labeled counter and as an author-facing package warning, with three tests pinning the warning; all 27 commits are DCO-signed with matching authorship.

The best thing in the diff is the test design in `model_storage_serve_givens.spec.ts`. Making the live relation behind the scoped source deliberately empty means any non-zero row proves the answer came from the stored table, and the three rows have distinct sums (30/7/37), so tier, isolation and fallback are all decidable from row values alone. When I omitted one re-applied per-caller filter in the serve transform, eight assertions failed, led by `never serves the unfiltered artifact` with `Expected: not 37`. The suite detected the cross-tenant leak by name.

## Two things I would like before merge

**1. The merge-scope property has no unit test.** `merge_key_scope_unresolved` appears in the gate, the metrics enum, `api-doc.yaml` and the docs — and in zero spec files. No test passes a non-empty `scopeColumns`; every fixture uses `[]`. So the highest-severity behaviour in the PR is proven only by two hammer scenarios that need a live Postgres and DuckLake. Two small pure tests close it: one asserting `deltaStatements` puts the scope column in the `ON` text, one asserting the mixed resolving/non-resolving shape refuses on both gates and yields `undefined` from `incrementalLineage`.

**2. `partitioned-mz/04` can pass for the wrong reason.** It asserts only `cites: partition`, and all three refusal details in `persist_partition.ts` contain the literal `'partition='` — so the scenario passes just as happily on `partition_column_unknown` or `partition_column_not_public` as on the one it is named for. Scenario 05 does it properly with a `## Build refusals` table pinning the reason; 04 wants the same.

## Three factual corrections

**The PR body claims a cardinality warning that does not exist.** There is no cardinality check anywhere in `persist_partition.ts`, `materialization_service.ts`, `build_plan.ts`, `materialization_build_session.ts` or `package.ts`. The docs say the opposite and are correct: *"nothing checks the column's cardinality for you — a partition per caller across very many callers is the many-small-files case, and it is your judgement to make."* The docs are right; the body needs fixing.

**"Four positions are refused", followed by three.** `docs/materialization.md:155` and `RELEASE_NOTES.md:231` both say four and then list `dynamic_projection`, `dynamic_join`, `dynamic_joined_where`. Probably miscounted from the body's table, where `given_in_persisted_query` occupies two rows.

**The metric description is stale.** `materialization_metrics.ts:252-256` still enumerates only the five original labels and still advertises `given`, which `api-doc.yaml:6009` says is no longer produced. The enum three lines above was updated; the human-readable description beside it was not, so an operator reading metric metadata to build an alert gets a list that is both incomplete and misleading.

## Questions rather than findings

**An OR-composed dynamic term over-narrows the merge.** `where: org_id = $ORG_ID or user_id = $USER_ID` collects both columns and matches on `key AND org_id AND user_id`, which is narrower than the relation the author wrote. Running it, restating a row whose `user_id` moved produced a duplicate rather than touching another tenant — so the error direction is the safe one. Worth either documenting the duplicate hazard or refusing a non-conjunctive term when `merge_key=` is declared. The module header says an `or`-composition "re-applies whole" and that only pruning depends on shape, which is true of the read path but not of the merge scope.

**Does the partitioned build issue the source read twice?** The partitioned path emits `selectSQL` textually twice — `CREATE … WITH NO DATA` then `INSERT INTO … (selectSQL)`. DuckDB does not evaluate the local `WITH NO DATA` select, but on the non-split path `selectSQL` is `wrapPassthrough(...)`, and those table functions have to bind a remote schema, which on BigQuery and Snowflake generally means issuing the query. If so that is two billed source reads, one discarded — against a path that elsewhere goes out of its way to avoid exactly that. Could not prove it without live credentials.

**Malloy's join pruning is load-bearing for tenant isolation.** The newly-admitted shape is safe only because Malloy drops the unreferenced join from the build SQL — verified, no `JOIN` in the emitted SQL, byte-identical across defaults. If a future Malloy stopped pruning, that shape would bake `org_id=<default>` and the gate would still admit it. The oracle would catch it and there is a direct `not.toContain("JOIN")` assertion, so it is guarded — but the docs present the admission as a property of the gate rather than of the compiler, and one sentence naming the upstream assumption would help whoever bumps Malloy next.

**The colocated merge bug is reachable on released code.** On `main` the gate never refuses an extend-block given, the lineage carries no scope, and `deltaStatements` builds the `ON` from the bare key. What has masked it is the target warehouse rather than us — Postgres raises `MERGE command cannot affect row a second time`, while DuckDB has no such guard, which is exactly why scenario 06's note records the pre-fix storage build succeeding with org 1's restated row gone and org 2's duplicated. The fix here is self-contained and does not depend on `partition=`, so it may deserve its own backport decision rather than riding this PR.

## Smaller notes

- `persist_partition.spec.ts:180` — `describe("the wire plan reports what the build and the read need")` does not touch the wire plan; both cases duplicate what `:68-82` already asserts.
- `build_plan.ts:194 planDynamicFields` has no test, so `partition` and `strippedTerms` are never asserted on an actual plan object despite being documented wire contract. Low risk since the build re-resolves rather than reading them; fixing this and the item above is one small test.
- The oracle's vacuity guard is corpus-wide rather than per-case, so an individual case could silently flip branches while the suite stays green.
- The structural fallback for non-query-shaped sources survives a mutation (replacing it with `return false` leaves the suite green). I compiled five such shapes and none bake, so it is a corpus gap rather than a live leak — but the path is currently unfalsified.
- Partition value encoding is inherited DuckLake behaviour and is neither tested nor documented: null becomes `__HIVE_DEFAULT_PARTITION__`, `/` becomes `%2F`, a float renders as `f=1.50`. All round-trip correctly, but the float formatting is dialect behaviour we do not control, which makes partitioning on a float a foot-gun with nothing warning about it.

What I could not check: hammer was read rather than run, so scenario assertions are judged statically, and I did not run the full server suite — only the specs in scope.

housejester:
Thanks — the mutation tests you ran (pulling `argumentBindsAGiven` to see if the oracle caught it, omitting a re-applied per-caller filter to see if the suite named the leak) are the part I'd have most wanted someone to do, and they checked the two things I was least able to check myself.

All of it is addressed except where noted. Rebased onto `main`.

## Your two before-merge items

**The merge scope now has unit cover.** You were right that the highest-severity behaviour in the PR was proven only by scenarios needing a live Postgres and DuckLake. Four pure tests: the emitted `ON` carries the scope column NULL-safe-OR'd in; a source with no scope emits a merge byte-identical to the one before scoping existed; a scope column already serving as a merge key is not repeated; and a partly-resolvable shape (one term local, one reaching through a join) refuses on **both** gates with `merge_key_scope_unresolved`.

**`partitioned-mz/04` is pinned by reason.** It now carries a `## Build refusals` table asserting `partition_without_storage` on the colocated tier, the way 05 does. Your diagnosis was exact — all three refusal details quote `'partition='`, so `cites:` alone passed on any of them.

## Your three corrections

All three were wrong in the source, not just in the prose.

- **Three positions, not four** — fixed in `docs/materialization.md` and `RELEASE_NOTES.md`. Miscounted from the body's table, where `given_in_persisted_query` occupies two rows, exactly as you guessed.
- **The metric description** — rewritten to list all fourteen labels, and it now says why `given` is retained on the enum but no longer produced. You're right that an operator reading metric metadata to build an alert was the person this misled.
- **The cardinality warning** — the docs are the correct statement, as you said; there is no such check. The PR description has not been updated and still claims it. Flagging that rather than leaving you to re-notice it.

## Your questions

**No, the partitioned build does not issue the source read twice.** Measured, against a real Postgres with `postgres_query` wrapping `pg_sleep(3)`:

| statement | elapsed |
|---|---|
| `CREATE … AS (select) WITH NO DATA` | 27 ms |
| `ALTER … SET PARTITIONED BY` | 15 ms |
| `INSERT INTO … (select)` | 3025 ms |
| plain CTAS, unpartitioned path | 3030 ms |

With `log_statement=all`, the only user query Postgres received across the whole partitioned build was the single `COPY` during the INSERT; the `WITH NO DATA` phase sent nothing but connection keepalives. So the bind resolves the result schema without executing, and the partitioned path costs the same one source read as the CTAS it replaces.

Two limits worth stating: `log_statement=all` does not log extended-protocol Describe round-trips, so a cheap metadata exchange could be invisible to it — but the timing rules out execution. And BigQuery and Snowflake remain unverified, as you said; all I can add is that the mechanism is a schema bind rather than a run, which on BigQuery would normally be a dry-run.

**The OR-composed term over-narrows, and the module header was wrong.** It claimed a term "re-applies whole" and that only pruning depends on shape. True of the read path; not true of `DynamicTerm.columns`, which the merge scope reads and which collects from field usage without regard to composition. Header corrected to say exactly that, and both duplicate hazards are now documented for authors — the `or` case and the row whose scope column value changes between refreshes — with the advice to rebuild rather than refresh. I left them as documented hazards rather than refusals, on your reasoning that the error direction is a duplicate and never another caller's row.

**Malloy's pruning is load-bearing, and the docs now say so.** `docs/materialization.md` states that the admission rests on the compiler dropping an unread join, not on the gate, and flags it for whoever bumps the version.

**The colocated backport.** Agreed the fix is self-contained and does not depend on `partition=`. It ships here rather than separately.

## Two more from a separate pass, since they touch what you reviewed

**The scope was derived from the positional classification, which returns no terms when it refuses.** The two gates refuse different sets — the colocated one deliberately admits the positional cases — so a source one gate admitted while the classifier refused it for an unrelated position came back with an empty scope, and an empty scope is indistinguishable from "not caller-scoped". It now reads the source's own filters instead, so it does not depend on which gate ran.

**Scope columns are resolved against the stored schema.** A term naming an extend-block `dimension:` reported that name as written, which the build never materializes — so it would have reached the `ON` as a column the table does not have. And a term reading a joined field now contributes no columns at all rather than its local ones, since scoping by part of what a term constrains is narrower than the bare key and still wider than the author's relation.

**Your rollback-test instinct generalises, and caught a second one.** The real-catalog test used a select that failed to *bind*, so `CREATE OR REPLACE` errored first and the INSERT never ran — and a failed `CREATE OR REPLACE` leaves the old table alone with or without a transaction, so it passed against the unwrapped code. It now fails partway through the INSERT, and asserts the layout reverted as well as the rows. Verified it goes red with the transaction removed.

Relatedly: the schema read-back now happens **before** `COMMIT`. `describeOrDrop` drops the table when the read-back fails, and with a name stable across generations that deleted the generation the previous manifest names — the same defect the transaction exists to prevent, one statement later.

## Smaller notes I have not taken

Recorded rather than silently skipped: the `persist_partition.spec.ts:180` describe name and its duplicate coverage, a test for `planDynamicFields` on an actual plan object, the oracle's corpus-wide rather than per-case vacuity guard, the unfalsified structural fallback for non-query-shaped sources, and partition value encoding (the float foot-gun especially). The last two are the ones I would want back first — an unfalsified path and an undocumented dialect behaviour are different in kind from the others.

PR #1093, opened 2026-08-29
refactor(skills): retire malloy-analyze and slim malloy-modeling

jswir:
Stacked on [#1088](https://github.com/malloydata/publisher/pull/1088) (`manifests/publisher-local.json`). Do not merge this before 1088; the tree-coverage and symlink tests fail until the manifest matches this tree.

## Summary
- Retire `malloy-analyze`. Its exploration loop (profile, hypothesize, triangulate) now lives in `malloy-analysis` as **When there is no specific question**. Hosts auto-invoked on `malloy-analyze`'s description (`analyze this data`, `what's interesting`, `what's driving X`); those triggers moved with it, so deleting the skill is what makes the routing change take effect.
- Split setup out of `malloy-getting-started` into `malloy-publisher-setup` (server, MCP client, diagnosis table, tools table, edit-and-run loop). Getting-started keeps discovery and a first query. The `malloy` index is generated from frontmatter (`packages/skills/scripts/build-index.ts`) so a new skill cannot ship unlisted.
- Slim `malloy-modeling` to the shared 8-step workflow (217 → 121 lines). Duplicated syntax (reminders, SQL-to-Malloy, critical rules, anti-patterns, reserved words) is a pointer to the gotcha skills. Publisher tool names are gone; the host's setup skill names those. A question that arrives before a model routes to `malloy-model-as-you-go`.
- Register `malloy-publisher-setup` in `publisher-local.json` and drop `malloy-analyze`, so 1088's tree-coverage / symlink tests agree with this tree.

This is handoff tasks 01 and 02 on one branch. They both edit `malloy-modeling` and the setup skill, so they review more cheaply together.

## Test plan
- [x] `bun run --cwd packages/skills test` — 156 pass (index in sync, every skill grouped, manifest matches the tree, no `malloy-analyze`)
- [x] `cd packages/server && bun test src/mcp/skills` — 18 pass (bundle matches `skills/`, no em dashes)
- [ ] `cd packages/create-malloy-package && bun test src` (scaffolder still pins `@latest`, now also scans `malloy-publisher-setup`)
- [x] Confirm nothing under `skills/` still names `malloy-analyze`
- [ ] Read `malloy-modeling` end to end: pauses still read as real stops; no `malloy_*` tool names; edit-and-run loop is in `malloy-publisher-setup`
- [ ] Ask an agent "what's interesting in this data?" and confirm it loads `malloy-analysis`, not a missing skill
[...]

PR #1193, opened 2026-09-17
skills: hold a description to the budget its consumers read it under, and lead with the trigger

jswir:
A skill's `description` is the only text a host reads before deciding whether to load it. Two things consume it, with different budgets, and neither was asserted.

## Mechanism

The binding constraint was invisible from this repo. Nothing here truncates a description — the MCP prompt bundle, the npm pack, and the scaffolder all copy it verbatim. But a downstream plugin build (`ms2data/service` `plugins/builder.py`) rewrites the `description:` line **in place** at 200 characters and appends an ellipsis, silently, at build time.

Six of the eight shared skills that build packages were over. `malloy-analysis` shipped for several releases as:

> `description: Workflow for answering data questions against Malloy semantic models over MCP - structured discovery with get_context, query construction with execute_query, verification, and answer delivery. Use…`

The clause that says *when* to load the analysis workflow is cut off, on the surface that most needs it. Everything past 200 was written for an audience that never read it.

## Deliverables

1. **Two budgets, both asserted in `packages/skills/src/manifest.spec.ts`.**

   | Budget | Applies to | Why that number | Headroom today |
   |---|---|---|---|
   | 200 | the 8 shared skills a downstream plugin packages | that build truncates at 200 and appends `…` | 182–197 after this change |
   | 1024 | every shipped skill | the host loader's frontmatter budget | longest is ~633 |

   The ceiling is deliberately **not** 200. Nothing in this repo truncates, and 31 of 38 skills exceed 200; a cap belonging to one downstream packager does not belong on skills that packager never sees.

2. **The eight packaged descriptions rewritten to fit, each leading with its trigger condition** rather than its subject.

   | Skill | Before | After |
   |---|---|---|
   | `malloy-analysis` | 299 | 195 |
   | `malloy-patterns` | 294 | 197 |
   | `malloy-charts` | 266 | 197 |
   | `malloy-queries` | 220 | 184 |
   | `malloy-phrase-detection` | 208 | 192 |
   | `malloy-gotchas-queries` | 206 | 191 |
   | `malloy-analysis-pitfalls` | 200 | 195 |
   | `malloy-gotchas-rendering` | 195 | 182 |

   `malloy-phrase-detection` was the one written as a topic — *"How to construct search targets for the get_context tool"* — where every sibling opens with an imperative plus a condition. It reads as reference material for when you are stuck, and confidence is exactly the state where that loses. It is also the skill an analysis agent skipped before calling `get_context` with no `search_text`.

3. **`skills/README.md`** gains the budget as an authoring rule, beside the bundle-regeneration rule.

4. Bundle regenerated (8 description lines) and `@malloy-publisher/skills` bumped to 0.1.20, per that README.

## Decisions and tradeoffs

- **`PACKAGED_SKILLS` is a literal list, not derived.** This repo cannot see the downstream manifest. The companion test in `ms2data/agent-skills` *is* scoped to that manifest, so a skill **added** to the packaged set is caught there at the moment it is added; this list holds the ones already in it from growing back where they are authored. Complementary, not duplicated — and `manifest.spec.ts`'s own header already says a rule holding in one repo and not the other is how the corpora drift apart.
- **The test measures the raw `description:` line, not `frontmatter()`.** That helper strips outer quotes; the build that truncates does not. A quoted description two characters over would read as compliant and still ship cut.
- **Not done: a 200-char rewrite pass across the shared set.** 21 shared skills exceed 200 and nothing currently cuts them. That is a real rewrite, and it should follow a decision about whether the shared set as a whole targets the tightest consumer.

## Risks

| Risk | Mitigation |
|---|---|
| A shorter description triggers less accurately | Each now leads with its trigger condition, which is the part a host matches on; the removed text was the trailing enumeration of contents |
| `PACKAGED_SKILLS` goes stale if the downstream manifest changes | A removal only leaves a skill over-constrained, not broken; an addition is caught by the agent-skills-side test, which reads the manifest itself |
| Conflicts with in-flight skill PRs | #1175 touches all three of `malloy-analysis`, `malloy-analysis-pitfalls`, `malloy-phrase-detection`; #1093 touches `manifest.spec.ts`. Both conflicts are description lines or an isolated test block |

## Testing

`packages/skills`: **224 pass, 0 fail** (209 in `manifest.spec.ts`, up from 193).
`packages/server/src/mcp/skills`: **18 pass, 0 fail** after regenerating the bundle.
`bunx eslint ./src ./scripts --fix` made no changes; `prettier --check` passes.

**Both guards are load-bearing** — each defect they exist to catch, reintroduced:

| Mutation | Reported as |
|---|---|
| `malloy-queries` description +30 chars (215) | `(fail) malloy-queries: description survives the plugin build unchanged` — `"length": 200` vs `+ "length": 215` |
| `malloy-charts` description quoted and +3 chars (203 with quotes) | same test fails at `"length": 203` — confirming quotes are counted, as the build counts them |
| `malloy-publish` description +1000 chars (1147) | `(fail) malloy-publish: description fits the loader budget` — `"length": 1024` vs `+ "length": 1147` |

## Not in scope

Two content edits to `malloy-phrase-detection` and `malloy-analysis` — an explicit "do not enumerate" rule, and making the phrase-detection reference an imperative that leads its bullet rather than a trailing clause. They belong here rather than downstream, but they overlap #1175's edits to the same sections, so they follow once that lands.


jswir:
Folded into #1195 with #1175 and #1194, which carries this change unmodified as its own commit.

PR #1194, opened 2026-09-17
test(skills): the index has to account for every skill that ships

jswir:
`manifest.spec.ts` already asks that every `skill:` reference point at something the manifest ships. Nothing asked the other direction: that every shipped skill is reachable **from** the index.

That direction is the one that fails quietly. The index is how an agent holding one skill finds a sibling, so a skill it never names is installed and effectively undiscoverable. Nothing complains, because the skill loads correctly whenever someone asks for it by name, and nobody does.

## It had already drifted

`eval-judge` and `eval-import` arrived with #1114 and the index was never told. Two of the six eval skills have been shipping unreachable since then. This adds the check and the two rows.

## The check counts a mention, not a `skill:` reference

That distinction is the whole design, and I got it wrong twice before getting it right.

A group must be closed under its own `skill:` references, so that excluding a group cannot strand a pointer. The index lives in `modeling`, while some skills it should still account for live in `analysis` and `eval`. Naming those in prose backticks is how the index stays complete without dragging three groups into one.

I wrote the strict version first, converted seven prose rows to the invocable form to satisfy it, and watched **"keeps each group's references inside it"** go red. That is the existing suite catching the new test being wrong, which is a good sign about the existing suite. The prose form is deliberate; the comment now records why, so the next person does not repeat the mistake.

Before that I had an even weaker version that exempted anything in a `groups` entry. Every shipped skill is in a group, so that version could never fail, and it passed against a tree that already had this drift in it. Worth saying out loud since an unfailable test is worse than none.

## Verification

Mutation-tested rather than assumed. Deleting a single table row is **not** enough, because the index names several skills twice; removing every mention of one skill turns it red.

- `bun test src` in `packages/skills` — `manifest.spec.ts` 164 pass, 0 fail
- `bun test src/mcp/skills` in `packages/server` — 18 pass, bundle regenerated per `skills/README.md`
- `bun run check-pack` fails on this uninstalled worktree with `TS2688: Cannot find type definition file for 'node'`, **identically on unmodified main**, so it is untouched by this change and wants a reviewer with a normal checkout

## Context

Found while checking a downstream consumer that assembles these skills alongside its own. That consumer's problem is its own and is being fixed on its side; this is the half that belongs here, because a repo that publishes a routing index should be the one that knows the index is complete.


jswir:
Folded into #1195 with #1175 and #1193, which carries this change unmodified as its own commit.

PR #1195, opened 2026-09-17
skills: what an agent reads, and the tests that keep it honest

jswir:
Four changes that all answer the same question — does the agent reach the right guidance, and does it say the right thing — landed as one review because they conflict with each other as separate PRs. Replaces #1175, #1193 and #1194, which are closed pointing here.

## 1. Four rules an eval traced to repeated failures (was #1175)

One 24-question set over a `query_executions` model, answered three times: haiku at default effort, sonnet at `low`, sonnet at `xhigh`. Each diagnosed separately. Nine clusters came back agent-skill-owned; these are the four a skill change can actually move.

The interesting one is that an existing rule was **causing** the failure it looks like it prevents. "When the exact ask is impossible, deliver the closest thing that works" is about the shape of an answer. All three arms read it as licence to swap the **field** — proving a column was null for the period, then building the split from a different column that means something else, and presenting the result as an answer about the product. So the fix bounds the existing rule rather than adding a competing one.

Also: the terminal rule was satisfiable by annotation, so agents named the condition invalidating their result and printed the invalidated result as the headline anyway. Widened to any invalidating condition the agent names itself.

## 2. Descriptions inside the budget their consumers read them under (was #1193)

A description is the only text a host reads before deciding whether to load a skill. Nothing here truncates one, but a downstream plugin build rewrites the `description:` line in place at **200 characters** and appends an ellipsis, silently, at build time. `malloy-analysis` shipped for several releases as:

> `description: Workflow for answering data questions against Malloy semantic models over MCP - structured discovery with get_context, query construction with execute_query, verification, and answer delivery. Use…`

The clause saying *when* to load the analysis workflow was cut off, on the surface that most needs it.

Eight packaged descriptions rewritten to fit, each leading with its trigger condition rather than its subject. Two budgets asserted: **200** for the skills a downstream plugin packages, **1024** for every shipped skill. The ceiling is deliberately not 200 — nothing here truncates, and 31 of 38 skills exceed it, so a cap belonging to one downstream packager does not belong on skills it never sees.

**The first pass of those rewrites broke the YAML on five skills**, and the fix is the last commit here. Separating clauses with an unquoted `: ` ends the plain scalar, so the whole frontmatter block raises `mapping values are not allowed here` and a host that parses frontmatter rejects the skill rather than loading it — `malloy-analysis-pitfalls`, `malloy-gotchas-modeling`, `malloy-gotchas-queries`, `malloy-gotchas-rendering`, `malloy-queries`. The separator is now ` - `, which the packaging build's truncation also leaves valid; quoting would not, because that build rewrites the raw rest of the line and would drop the closing quote.

All 225 tests stayed green on five unloadable skills because `frontmatter()` was a hand-rolled line regex, which returns a value where YAML refuses one. It now parses with `Bun.YAML` — the oracle, not a reimplementation of it — and a per-skill "frontmatter parses as YAML" test names the failure directly. `rawDescriptionLength` still reads the raw line, because that is what the build measures.

## 3. The index has to account for every skill that ships (was #1194)

`manifest.spec.ts` already asks that every `skill:` reference point at something shipped. Nothing asked the other direction: that every shipped skill is reachable **from** the index. That direction fails quietly, because the skill loads fine when asked for by name and nothing ever asks.

It had already drifted. `eval-judge` and `eval-import` arrived with #1114 and the index was never told, so two of six eval skills have been shipping unreachable.

**The check counts a mention, not a `skill:` reference, and that distinction is the design.** A group must be closed under its own references so excluding it cannot strand a pointer, and the index sits in `modeling` while some skills it should account for sit in `analysis` and `eval`. I wrote the strict version first, converted seven prose rows to the invocable form to satisfy it, and watched **"keeps each group's references inside it"** go red. That is the existing suite catching the new test being wrong. An earlier draft exempted anything in a `groups` entry, which made it unfailable, since every skill is in one.

## 4. The `search_text` gate, from the copy that stops being served

`ms2data/agent-skills` documented a failure this repo's copies do not defend against: an agent answered a question without loading the phrase-detection skill, called `get_context` with no `search_text`, got a capped catalog instead of a search, and concluded the model lacked fields the cap had dropped. The rule was written there. Once `ms2data/service` assembles `malloy-*` from this package, that copy stops being served — so the fix reached nobody.

- `malloy-analysis` step 2 now leads with loading `malloy-phrase-detection` and requires a `search_text` on every target, restated in the step's checklist.
- `malloy-phrase-detection` gains the "Always send `search_text`" section.

Two exceptions, not one: the bare "what data is here?" listing, and reading an entity you already have the exact name of — scope to its source, set `entity_name`, pass `search_text: null`, and get its docstring and code without spending a search. The rule below it ("still describe it, don't just repeat the name") is about how to *phrase* a search, and now says so, because as written it routed that exact-entity lookup the opposite way.

`malloy-scope` also comes back the other way: its options went in a numbered list written before hosts had an `ask_user` card, so an agent with one rendered them twice.

## Why one PR

They conflict. #1175 and #1193 both edit `malloy-analysis`, `malloy-analysis-pitfalls` and `malloy-phrase-detection`; #1193 and #1194 both edit `manifest.spec.ts`, the bundle, and the same version number. Landing them separately costs two rebases and settles `0.1.20` twice.

## Sequencing

`ms2data/service` pins `skills_version: 0.1.19`, and six of the eight skills in `PACKAGED_SKILLS` are over 200 characters in that published version — `malloy-analysis` is 299. This PR fixes all eight (longest is 194). So this is a prerequisite for that pin bump rather than an independent cleanup: **publish `0.1.20`, then move the pin.**

## Testing

```
packages/skills            263 pass  0 fail
packages/server mcp/skills  18 pass  0 fail
```

225 before; the 38 new ones are the per-skill frontmatter-parses check. `packages/create-malloy-package` shows 6 failures, **identically on unmodified main** — pre-existing and untouched here. Note `bun run build` in `packages/skills` is required before its suite passes, since the built copy carries the stamped version.

Mutation-tested rather than assumed: for the budget, `malloy-queries` +30 chars fails; for the index, deleting one table row is **not** enough, because several skills are named twice, so removing every mention of one is what turns it red; for the YAML guard, restoring the `: ` in `malloy-queries` turns four tests red, including the new one.

Not done: a 200-character pass across the other 21 shared skills that exceed it. Nothing truncates them today, and whether the shared set should target its tightest consumer is a separate decision.


Sha-Bang (commented):
the description rewrites in §2 broke the YAML on five skills. unquoted `: ` mid-scalar ends the plain scalar, so `yaml.safe_load` on the frontmatter raises `mapping values are not allowed here` for `malloy-analysis-pitfalls`, `malloy-gotchas-modeling`, `malloy-gotchas-queries`, `malloy-gotchas-rendering` and `malloy-queries`. a host that parses frontmatter rejects those skills rather than loading them.

it's new: the same five parse clean on the published `0.1.19`, and all of agent-skills parses clean too. the old descriptions separated clauses with `.` and these switched to `: `.

the reason all 225 tests stay green on five unloadable skills is `frontmatter()` at `manifest.spec.ts:109` — a hand-rolled line regex rather than a YAML parse. it returns a value where YAML refuses one, so every assertion built on it passes. the budget work itself is right, and reading the raw line is correct for what the build truncates; what's missing is anything asserting the frontmatter *parses*. comment inline.

one cross-repo note, since it decides sequencing rather than correctness: service#6226 pins `skills_version: 0.1.19`, and six of the eight skills in `PACKAGED_SKILLS` are over 200 characters in that published version — `malloy-analysis` is 299. this PR fixes all eight (longest is 194). so #6226 can't stop shipping truncated descriptions until `0.1.20` publishes and that pin moves, which makes this PR a prerequisite for that one rather than an independent cleanup.

on the two nits about de-indented sub-bullets in `malloy-analysis` and `malloy-phrase-detection`: those read like deliberate flattening to me, not accidents, so ignore unless you disagree.

i did check the version bumps and they're fine — `packages/skills` is at `0.1.20` against `0.1.19` on npm, and `create-malloy-package` at `0.0.16` against `0.0.15`, so the publish gate is satisfied as-is.

skills/malloy-analysis-pitfalls/SKILL.md
```diff
@@ -1,11 +1,8 @@
 ---
 name: malloy-analysis-pitfalls
-description: Common data analysis pitfalls to watch for during query construction and result interpretation. Reference this checklist when verifying queries and results to catch errors before presenting an answer.
+description: Checks to run on a query and its numbers before a result is presented: wrong grain, fan-out, invented field names, mismatched filters, count vs distinct. The common data-analysis pitfalls.
```
Sha-Bang: **Blocking** - this description is not valid YAML. the unquoted `: ` before `wrong grain` ends the plain scalar, so `yaml.safe_load` on the frontmatter fails with `mapping values are not allowed here`, and a host that parses frontmatter rejects the skill instead of loading it.

same breakage in `malloy-gotchas-modeling`, `malloy-gotchas-queries`, `malloy-gotchas-rendering` and `malloy-queries` — all five are new here, and all five parse fine on the published `0.1.19`, so this is a regression rather than something pre-existing.

worth noting the old descriptions used `.` as the separator and these switched to `: `, which is what introduced it. quote the whole description, or go back to a separator that doesn't open a mapping?

skills/malloy-gotchas-modeling/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: malloy-gotchas-modeling
-description: Common Malloy modeling mistakes and how to avoid them. Read BEFORE writing source definitions, dimensions, measures, or joins. Covers reserved words, NULL checks, date functions, type casts, field management (extend except/accept/rename vs include public/internal/private), and query-based source gotchas.
+description: Common Malloy modeling mistakes, for writing source definitions, dimensions, measures or joins: reserved words, NULL checks, date functions, type casts, field management (extend except/accept/rename vs include public/internal/private), query-based source gotchas.
```
Sha-Bang: **Blocking** - same invalid-YAML break as `malloy-analysis-pitfalls`: unquoted `: ` mid-scalar, frontmatter fails to parse, skill doesn't load.

skills/malloy-gotchas-queries/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: malloy-gotchas-queries
-description: Common Malloy query and view mistakes. Read BEFORE writing views, queries, or notebooks. Covers chart constraints, aggregate filters, joined field aliasing, method syntax, and time truncation vs extraction.
+description: Common Malloy query and view mistakes, for writing views, queries or notebooks: chart constraints, aggregate filters, joined field aliasing, method syntax, time truncation vs extraction.
```
Sha-Bang: **Blocking** - same invalid-YAML break as `malloy-analysis-pitfalls`: unquoted `: ` mid-scalar, frontmatter fails to parse, skill doesn't load.

skills/malloy-gotchas-rendering/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: malloy-gotchas-rendering
-description: Common Malloy renderer annotation mistakes. Read BEFORE adding chart annotations, formatting tags, or building dashboards. Covers tag syntax, scale rules, sparkline setup, and big_value patterns.
+description: Common Malloy renderer mistakes, for adding chart annotations and formatting tags or building dashboards: tag syntax, scale rules, sparkline setup, big_value patterns.
```
Sha-Bang: **Blocking** - same invalid-YAML break as `malloy-analysis-pitfalls`: unquoted `: ` mid-scalar, frontmatter fails to parse, skill doesn't load.

skills/malloy-queries/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: malloy-queries
-description: Malloy query patterns, syntax rules, and chart annotation reference. Consult before writing or debugging any query. Covers dates, aggregates vs dimensions, join paths, filters, string matching, and common error patterns.
+description: Malloy query patterns and syntax, for writing or debugging a query: dates, aggregates vs dimensions, join paths, filters, string matching, chart annotations, and the errors a SQL habit produces.
```
Sha-Bang: **Blocking** - same invalid-YAML break as `malloy-analysis-pitfalls`: unquoted `: ` mid-scalar, frontmatter fails to parse, skill doesn't load.

packages/skills/src/manifest.spec.ts
```diff
@@ -167,6 +228,27 @@ describe("shipped skills", () => {
...
+      const description = frontmatter(name).description ?? "";
+      expect({ name, length: description.length }).toEqual({
+         name,
+         length: Math.min(description.length, DESCRIPTION_CEILING),
+      });
+   });
+
+   // Spread: PACKAGED_SKILLS is `as const`, and it.each takes a mutable array.
+   it.each([...PACKAGED_SKILLS])(
+      "%s: description survives the plugin build unchanged",
```
Sha-Bang, line 273: **Should-fix** - the reason the five broken descriptions go green through all 225 tests: `frontmatter()` at line 109 is a hand-rolled `^([a-z_]+):\s*(.+)$` line regex, not a YAML parse. it happily returns a value for a scalar that `yaml.safe_load` refuses, so every test built on it — this one, the loader-budget one, `frontmatter names its own directory` — passes on a skill no host can load.

the budget work here is good and the raw-line reading is right for what the build truncates. but nothing in the suite asserts the frontmatter is *parseable*, which is the property that actually decides whether the skill loads at all. add a `yaml.safe_load` assertion per skill alongside the budget checks?

skills/malloy-analysis-pitfalls/SKILL.md
jswir: Fixed in `48b24a24`. All five, plus the reason the suite stayed green on them.

Separator is now ` - ` rather than quoting. Quoting looked like the obvious fix and is the wrong one here: `plugins/builder.py` in ms2data/service truncates by rewriting the raw rest of the `description:` line, so a quoted description cut at 200 loses its closing quote and breaks the YAML in the built plugin instead of in the source. ` - ` survives truncation.

`frontmatter()` now parses with `Bun.YAML` instead of the line regex, and there is a per-skill `frontmatter parses as YAML` test. 225 -> 263. Mutation-tested: restoring the `: ` in `malloy-queries` turns four tests red, including the new one.

skills/malloy-gotchas-modeling/SKILL.md
jswir: Fixed in `48b24a24` along with the other four. See the thread on `malloy-analysis-pitfalls` for why the fix is a separator change rather than quoting.

skills/malloy-gotchas-queries/SKILL.md
jswir: Fixed in `48b24a24` along with the other four. See the thread on `malloy-analysis-pitfalls` for why the fix is a separator change rather than quoting.

skills/malloy-gotchas-rendering/SKILL.md
jswir: Fixed in `48b24a24` along with the other four. See the thread on `malloy-analysis-pitfalls` for why the fix is a separator change rather than quoting.

skills/malloy-queries/SKILL.md
jswir: Fixed in `48b24a24` along with the other four. See the thread on `malloy-analysis-pitfalls` for why the fix is a separator change rather than quoting.

packages/skills/src/manifest.spec.ts
jswir: Agreed, and this was the actual defect — the five broken descriptions were a symptom of it.

`frontmatter()` now parses the block with `Bun.YAML` and throws on anything it refuses, so every assertion built on it stops being reachable for a skill no host can load. Added `%s: frontmatter parses as YAML` per shipped skill so the failure names itself rather than showing up as four confusing downstream failures.

`rawDescriptionLength` is unchanged and still reads the raw line, for the reason your comment gives: that is what the packaging build measures.

Both in `48b24a24`.

jswir:
All five YAML breaks fixed, plus the reason the suite could not see them — `48b24a24`.

The separator is now ` - ` rather than quoting the descriptions. Quoting is the obvious fix and the wrong one here: the downstream plugin build truncates by rewriting the raw rest of the `description:` line, so a quoted description cut at 200 loses its closing quote and breaks the YAML in the built plugin instead of in the source. ` - ` survives truncation.

`frontmatter()` now parses with `Bun.YAML` — the real parser rather than another line regex — and a per-skill `frontmatter parses as YAML` test names the failure directly. `rawDescriptionLength` is unchanged, for the reason you gave: the raw line is what the build measures. 225 -> 263 tests, and restoring the `: ` in `malloy-queries` turns four red.

Two things arrived since your review.

**A fourth commit mirrors the `search_text` gate in from ms2data/agent-skills** (`fab06089`). The failure it fixes is documented there: an agent answered without loading the phrase-detection skill, called `get_context` with no `search_text`, got a capped catalog instead of a search, and concluded the model lacked fields the cap had dropped. That rule lives only in the Credible repo today, and once ms2data/service assembles `malloy-*` from this package, that copy stops being served — so the fix would have reached nobody. `malloy-analysis` step 2 now leads with loading `malloy-phrase-detection` and requires a `search_text` on every target; `malloy-phrase-detection` gains the `## Always send search_text` section, with two exceptions rather than one (the bare "what data is here?" listing, and pinning a known entity with `entity_name` + `search_text: null`).

**`malloy-scope` comes back the other way** (`61c20654`): its options went in a numbered list written before hosts had an `ask_user` card, so an agent with one rendered them twice.

Your sequencing note is right and is now the plan of record across all three PRs: this publishes `0.1.20` first, then the service pin moves. I confirmed the numbers against the published tarball rather than the source tree — at `0.1.19` it is six of eight over 200 (299, 294, 266, 220, 208, 206; `malloy-analysis-pitfalls` is exactly 200, so it does not truncate), and zero at this head, longest 195.

`packages/skills` 263 pass, `packages/server mcp/skills` 18 pass. The `create-malloy-package` failures are still identical on unmodified `main`.

girishjeswani (approved):
Approving. The budget fix is real and I measured it from both ends rather than trusting the numbers: published `0.1.19` has six of eight packaged skills over 200 (299/294/266/220/208/206, with `malloy-analysis-pitfalls` at exactly 200), and at this head all eight fit with the longest at 195.

The part worth calling out is your pushback on the suggested fix. Quoting the descriptions looks like the obvious answer, and it's wrong for a non-obvious reason — the downstream build rewrites the raw `description:` line, so a quoted description cut at 200 loses its closing quote. I tested it: simulating truncate-at-200, all eight packaged descriptions still parse; a quoted 250-char description cut the same way dies with `while scanning a quoted scalar`. Fixing the line-regex frontmatter parser to use `Bun.YAML` addressed the actual cause, which is why 225 tests had been passing over five unloadable skills.

Bundle verified byte-identical to a fresh regeneration, and `skills_bundle.spec.ts` goes red on a stale one. All four mutations reproduce. DCO on all 13.

One non-blocking thing, and I'd push back gently on the earlier read that it was deliberate flattening. In `skills/malloy-analysis/SKILL.md` the sub-bullets under "Check the common failure modes:" and "Quick reference by query type:" lost their indent and are now flat siblings of their own headers, leaving two labels ending in colons with nothing under them. I think it's an unintended side effect of the unwrap script rather than an authoring choice: that commit's message says the script "refuses to run unless the word stream is byte-equal before and after", and a word-stream guard is exactly the check that passes while leading whitespace changes. Whitespace-only fix, 9 lines — and worth knowing before that script runs on another file.

One thing I'd flag for the sequence rather than this PR: nothing in any of the three repos measures the published npm artifact that ms2data/service#6226's `skills_version` resolves to. This PR makes the source correct and guards it going forward, but the pin stays guarded by sequencing discipline rather than by a test. The natural home for that check is the service side, since that's the repo that owns the pin.

jswir:
Took the indent fix — `d441b0d1`. Your read of the cause is right and mine was wrong: I had accepted the earlier "deliberate flattening" reading, and the unwrap script's guard is exactly the check that would not notice. Word-stream equality passes while leading whitespace changes, so ten sub-bullets became siblings of their own headers and two labels ending in colons were left with nothing under them.

Restored to `  - ` for both groups, matching what they were before the unwrap commit. Bundle regenerated, 263 pass. Also synced into ms2data/agent-skills, since `malloy-analysis` is one of the byte-shared files — the same script runs there.

On the npm-artifact gap: half of it is closed on the service side, in the repo that owns the pin, as you suggest. `check-malloy-versions` there now unpacks the pinned tarball and asserts every skill's frontmatter parses — that is what would have caught the five broken descriptions at the pin rather than at review.

The 200-char budget deliberately is **not** checked there, and it is worth saying why so it does not look like an oversight. The budget applies only to the skills a given plugin manifest packages; that manifest lives in a private repo the service CI job cannot read; and a hardcoded list would be a third copy to keep in sync with `PACKAGED_SKILLS` here and `test_manifests.py` there. So the enforcement went into `plugins/builder.py` instead — the one place that knows the real packaged set — where it now **fails** the build naming the skills rather than truncating in place. A pin bump to a tarball with an overrun is a red plugin build, not a silent cut.

What remains guarded by sequencing rather than a test is the ordering itself, and that is now narrower too: the service check mints a read-only token and verifies the agent-skills tag exists, and warns when manifests at that tag name the bare `malloy` index.

PR #1196, opened 2026-09-17
fix(server): redact every connection credential from a failed connection test, not only Postgres

sagarswamirao:
A failed connection test returns the driver's error text to the caller, and drivers build that
text by quoting back the configuration they were handed. That text is redacted before it is
returned, but only for Postgres: `redactPgSecrets` covers URI userinfo
(`scheme://user:pass@host`) and the libpq `password=` keyword, which is every credential a
Postgres connection carries and none of the credentials any other connection type carries.

So a connection test that fails while opening an SSH tunnel, or while authenticating to a
warehouse, answers with the private key, passphrase, service-account JSON, bearer token or
storage secret key it was given -- in a response body that API clients routinely print into
their own logs.

## What leaks today

Each row is a credential the `Connection` schema accepts, in the shape a driver or a serializer
reports it when it echoes the config back:

| Credential | Redacted before |
|---|---|
| Postgres URI password (`postgres://u:pw@h`) | yes |
| `password=` keyword | yes |
| SSH `privateKey` (PEM block in prose) | **no** |
| SSH `privateKeyPass` | **no** |
| Snowflake `privateKey` | **no** |
| BigQuery `serviceAccountKeyJson` | **no** |
| Databricks `token` / `oauthClientSecret` | **no** |
| S3 `secretAccessKey` / `sessionToken` | **no** |
| Azure `clientSecret` / `sasUrl` | **no** |
| MotherDuck `accessToken`, Trino `peakaKey`, GCS `secret` | **no** |

## The change

`redactConnectionSecrets` in `pg_helpers.ts`, used by `testConnectionConfiguration` in place of
`redactPgSecrets`. It covers the secret-bearing field names across every connection type the API
accepts, then defers to `redactPgSecrets` for the two shapes that function already handles.
`redactPgSecrets` itself is unchanged, so its other callers keep their current behaviour.

Three shapes are matched, all shape-based rather than parsed, for the same reason the existing
function is: the input is arbitrary error prose that may embed a serialized config, so there is
nothing to parse.

- **Named fields** -- `name: value`, `name=value` and `"name": "value"`. The value class stops at
  the separators that end a field in JSON, in libpq keyword form and in ordinary prose, so a
  match cannot run past the field it started in and swallow the rest of the message.
- **Either casing** -- the same field arrives as `privateKey` from the API schema and
  `private_key` from a driver that snake-cases its config before reporting it.
- **PEM blocks** -- a private key is the one secret with no field name in front of it when a
  driver reports it as prose (`could not parse key: -----BEGIN RSA PRIVATE KEY----- ...`), so it
  is matched by its own delimiters.

Over-redaction is the safe direction and is preferred to a leak: a field merely named `token` is
masked whether or not it held a credential.

## Testing

`packages/server/src/pg_helpers.spec.ts` -- **51 pass, 0 fail**.

The new cases are load-bearing rather than tautological. With `redactConnectionSecrets` reduced
to its previous behaviour (delegating straight to `redactPgSecrets`), **17 of them fail**, each
naming the credential that survived:

```
error: expect(received).not.toContain(expected)
Expected to not contain: "s3cret-value-not-a-real-credential"
Received: "privateKey=s3cret-value-not-a-real-credential and password=***"
```

Coverage: one case per credential in the table above, a PEM block in both RSA and OPENSSH form, a
secret quoted inside a JSON object (asserting the neighbouring non-secret keys survive), the
snake_cased spelling, the two Postgres shapes still delegating correctly, a non-secret message
passing through byte-identical, and idempotence.

No test asserts against a hand-written copy of the regex; every case goes through the exported
function.

## Notes

- The sentinel values in the tests are generated strings. Nothing here is or resembles a real
  credential.
- `testConnectionConfiguration`'s catch block remains untested for the reason recorded in the
  comment above it: the service resolves every failure rather than throwing, so reaching it needs
  a module mock, and bun shares one process across spec files. The redaction it calls is tested
  directly instead.


housejester (changes requested):
The diagnosis here is right and the write-up is unusually good — the credential table, the "keeps a driver's prose" case, and leaving `redactPgSecrets` byte-identical so no existing caller changes behaviour are all the right calls.

But the swap is wired into a path that never executes, and the controller test that would have caught that passes with the change reverted. Details below, with the commands.

## Mental-model delta

Redaction stops being "two libpq/URI grammars" and becomes "a hardcoded list of credential *field names* × 3 syntaxes × 2 casings". Two consequences worth holding:

1. `redactConnectionSecrets` now names **two different exported functions** in `packages/server` — this one in `pg_helpers.ts` (name-based, 1 arg) and the one in `service/materialization_service.ts:405` (value-based, variadic configs). The import path decides which behaviour you get.
2. Adding a credential field to the `Connection` schema silently leaks it until someone also adds it to `SECRET_FIELD_NAMES`. Nothing enforces that link.

Also worth noting: `pg_helpers.ts` opens by explaining that it lives at `src/` root so neither `service/` nor `storage/` takes a dependency on the other. It now encodes knowledge of the SSH / Snowflake / BigQuery / Databricks / S3 / Azure / MotherDuck / Trino connection schemas, which is a different kind of coupling than that header anticipated.

---

## The fix is on a dead path

`connection.controller.ts:962` sits in a `catch` that only fires if `testConnectionConfig` *throws*. It doesn't — `service/connection.ts:2848` wraps its whole body in a catch-all that resolves every failure into `{status: "failed", errorMessage: redactPgSecrets(...)}` at **`service/connection.ts:2949`**, and the controller returns that object verbatim at `connection.controller.ts:942`.

The PR's own comment states the premise ("the service resolves every failure rather than throwing") without drawing the conclusion from it.

The new controller spec's own log output traces the failure straight through the *service* catch:

```
error: Connection test failed {"error":"Error: Cannot parse privateKey: Unsupported key format
    at openSshProxy (.../service/proxy.ts:96:15)
    at testConnectionConfig (.../service/connection.ts:2885:59)
    at async testConnectionConfiguration (.../controller/connection.controller.ts:942:23)"}
```

That `logger.error("Connection test failed", …)` is `service/connection.ts:2936`. The controller catch never ran.

So a caller whose SSH tunnel or warehouse auth fails still gets a message redacted by the Postgres-only passes — the exact scenario the description opens with. `service/connection.ts:1839` also rethrows an attach error unredacted into that same path.

**Recommend:** move the swap to `service/connection.ts:2949` (the response) and `:2937` (the log line — that one goes to operator logs, a wider audience than the caller, and is arguably the stronger case for this PR). Keeping the controller line as defence in depth is fine; it just isn't the fix.

## The controller test would not have failed before the fix

`connection.controller.spec.ts:900` asserts three sentinels are absent. I reverted the controller to `redactPgSecrets` and re-ran it:

```
$ bun test src/controller/connection.controller.spec.ts -t "secret redaction"
 1 pass  0 fail  5 expect() calls
```

Two independent reasons it cannot fail: the assertion lands on the *service's* output (above), and ssh2 rejects the malformed key with `Cannot parse privateKey: Unsupported key format` before it ever embeds one — there is no secret in the string either way. The spec's comment half-concedes this ("which shape of driver text reaches this path depends on how far the driver gets").

**Recommend:** after the first fix, either re-point it at the service boundary with a driver that does echo (a Postgres auth failure against a *reachable* host with a wrong password echoes the DSN), or drop it and rely on the `pg_helpers.spec.ts` cases — those are genuinely load-bearing. Neutering `redactConnectionSecrets` to `redactPgSecrets(s)` gives **18 fail / 35 pass** on the branch as it stands (the description's "51 pass / 17 fail" predates the two follow-up commits).

## `sasUrl=` redacts the harmless prefix and leaves the signature

`pg_helpers.ts:116` — the bare value class `[^,;&\s}"']+` stops at `&`. An Azure SAS URL's credential *is* the `sig=` query parameter, which comes after one:

```
in : auth failed: sasUrl=https://acct.blob.core.windows.net/c?sv=2021-08-06&ss=b&sig=SIGNATURE_IS_THE_CREDENTIAL&se=2030
out: auth failed: sasUrl=***&ss=b&sig=SIGNATURE_IS_THE_CREDENTIAL&se=2030
```

The test passes only because its fixture (`sasUrl=${SECRET}`) has no `&`. Same shape would hit any URL- or DSN-valued field echoed bare.

## A truncated or unterminated PEM block is not redacted at all

`pg_helpers.ts:124` requires the `-----END …-----` terminator, and drivers truncate long values routinely. Both of these pass through unchanged:

```
could not parse key: -----BEGIN RSA PRIVATE KEY-----\nMIIEowIBAAKCAQEA_KEY_BODY\n... (truncated)
ssh: -----BEGIN OPENSSH PRIVATE KEY----- b3BlbnNzaC1rZXktdjEAAAA_KEY_BODY
```

**Recommend:** make the terminator optional with a fallback to end-of-string —
`-----BEGIN [A-Z0-9 ]*PRIVATE KEY-----(?:[\s\S]*?-----END [A-Z0-9 ]*PRIVATE KEY-----|[\s\S]*)`.
Over-redacting the tail of a message that demonstrably contains a raw private key is the right trade.

---

## Worth a decision, not just a patch

`service/materialization_service.ts:405` already exports a `redactConnectionSecrets(message, ...connections)` that walks the config object, collects the values under sensitively-named keys, and removes those exact strings (plus their SQL-escaped forms). At this boundary you have the config in hand — `testConnectionConfig(connectionConfig)` at `service/connection.ts:2849`. `redactConnectionSecrets(err.message, connectionConfig)` would cover every row of the table in the description, plus the three misses above, plus shapes nobody has thought of — with no name list to keep in sync with the API schema, and no over-redacting of driver prose that happens to contain `token=`.

It isn't a free win: it only catches secrets echoed verbatim, so a base64- or URL-encoded echo slips through where the shape passes would catch it. The honest answer is probably **both** — value-based first, shape-based as the backstop — which is a small change to the same three lines.

Whichever way that goes, two exported functions with the same name and different signatures in one package is a trap worth avoiding. Renaming this one (`redactSecretFields`?) is worth doing independently.

## Minor

- Bare `privateKey: MIIEow…` (unquoted, colon) is deliberately unmatched, per the comment at `pg_helpers.ts:97`. The reasoning is sound and the test pinning it is a good catch — but it is the one place the change picks legibility over safety, for the highest-value secret. A value-based pass dissolves the dilemma rather than trading against it. Noting, not pushing.
- Redaction leaves the message non-JSON: `{"a":1,"privateKey":"s"}` → `{"a":1,"privateKey":***}`, since the value's quotes are inside the match. `$1"***"` keeps it well-formed.
- No `## [Unreleased]` section. Optional for a routine patch per `RELEASE_NOTES.md`, but a credential-disclosure fix is arguably narrative-worthy once it lands on the live path.

Checked and clear: alternation backtracking resolves `privateKeyPass` correctly despite `privateKey` preceding it in the list; snake_case handling works; idempotence holds; neither pattern is a ReDoS risk; and no existing `redactPgSecrets` caller changes behaviour.

<sub>Reviewed with Claude Code (session "Sleepy Tapir"). All experiments above were run on `86440cf` and reverted.</sub>

packages/server/src/controller/connection.controller.ts
```diff
@@ -950,12 +950,16 @@ export class ConnectionController {
...
+            // driver error embeds the config it was handed, credentials
+            // included. Covers every connection type, not only Postgres -- an
+            // SSH tunnel failure carries the private key and its passphrase,
+            // and a warehouse auth failure carries the token or key JSON, none
+            // of which the Postgres-shaped passes recognise. Untested on
             // purpose: the service resolves every failure rather than throwing,
             // so reaching this needs a module mock, and bun shares one process
             // across spec files -- mocking this module breaks connection.spec.ts.
-            errorMessage: redactPgSecrets(
+            errorMessage: redactConnectionSecrets(
```
Sha-Bang: **Blocking** - i think this wires the new redactor into the one branch that doesn't run.

line 942 is `return await testConnectionConfig(connectionConfig)`, and that function catches its own failures and *returns* `{status: "failed", errorMessage: redactPgSecrets(...)}` — `service/connection.ts:2946-2951`, where the comment says "this message goes into the REST response body". So the failed-test body a caller actually receives is built there, on the Postgres-only redactor, and comes back through line 942 untouched by anything here.

this `catch` only fires if `testConnectionConfig` throws. the comment just below is the proof, read the other way round: "the service resolves every failure rather than throwing". that's also why the new spec can't fail — the `.invalid`-host test returns through line 2949, so it never reaches this function.

net effect is that the SSH private key, the passphrase and the warehouse tokens this PR is about still reach the response body on the live path.

would it work to move the call to `service/connection.ts:2949` (and the sibling `error:` at 2937), and leave this catch delegating to the same helper? that'd also make the new controller spec load-bearing.

packages/server/src/pg_helpers.ts
```diff
@@ -61,3 +61,91 @@ export function redactPgSecrets(s: string): string {
...
+// Quoted value after `:` or `=`, e.g. "privateKey": "..." or privateKey='...'.
+const SECRET_QUOTED_PATTERN = new RegExp(
+   String.raw`(["']?(?:${NAME_ALTERNATION})["']?\s*[:=]\s*)(?:"[^"]*"|'[^']*')`,
+   "gi",
+);
+
+// Bare value after `=` only. The value class stops at the separators that end a
+// field in libpq keyword form, in a query string and in ordinary prose, so a
+// match cannot run past the field it started in.
+const SECRET_BARE_PATTERN = new RegExp(
```
Sha-Bang, line 127: **Blocking** - the bare value class stops at `&`, and `sasUrl` is the one field whose credential lives after one.

`api-doc.yaml:5104` defines it as "Full SAS URL including token", example `...?sp=rl&st=...`. so the signature is in the query string, past the first `&`. i ran it:

```
in:  sasUrl=https://acct.blob.core.windows.net/c?sv=2021-08-06&ss=b&sig=SIGSECRET123&se=2030
out: sasUrl=***&ss=b&sig=SIGSECRET123&se=2030
```

the `sig` is the credential and it survives. the JSON form is fine — `{"sasUrl":"..."}` redacts whole — it's only the bare `=` form that splits.

could the value class keep `&` and `;` when the value starts with a scheme, or is it simpler to add `sig` and `AccountKey` to the name list so the tail gets caught on its own?

packages/server/src/pg_helpers.ts
```diff
@@ -61,3 +61,91 @@ export function redactPgSecrets(s: string): string {
...
+// colon is the explanation rather than the key -- matching it would redact the
+// diagnosis and leave the caller with `Cannot parse privateKey: *** key
+// format`. Quoting, or an `=`, is what distinguishes a value from a sentence.
+const NAME_ALTERNATION = SECRET_FIELD_NAMES.map((n) =>
+   // Accept camelCase and snake_case for the same field.
+   n.replace(/([A-Z])/g, "[_-]?$1"),
+).join("|");
+
+// Quoted value after `:` or `=`, e.g. "privateKey": "..." or privateKey='...'.
+const SECRET_QUOTED_PATTERN = new RegExp(
```
Sha-Bang, line 113: **Should-fix** - `"[^"]*"` isn't escape-aware, so a JSON-escaped quote inside a value leaves the tail behind:

```
in:  config={"password":"ab\"TAILSECRET","host":"h"}
out: config={"password":***TAILSECRET","host":"h"}
```

the match ends at the escaped quote rather than the closing one. `redactPgSecrets` already uses the escape-aware form for single quotes (`'(?:\\.|[^'\\])*'`) — worth using the same shape for both here?

packages/server/src/pg_helpers.ts
```diff
@@ -61,3 +61,91 @@ export function redactPgSecrets(s: string): string {
...
+// match cannot run past the field it started in.
+const SECRET_BARE_PATTERN = new RegExp(
+   String.raw`(["']?(?:${NAME_ALTERNATION})["']?\s*=\s*)(?:[^,;&\s}"']+)`,
+   "gi",
+);
+
+// A PEM block is the one secret shape with no field name in front of it: a
+// driver reporting "could not parse key: -----BEGIN RSA PRIVATE KEY----- ..."
+// carries the whole key as prose. Matched by its own delimiters instead.
+const PEM_BLOCK_PATTERN =
```
Sha-Bang: **Should-fix** - `[\s\S]*?` needs the `-----END-----` delimiter to match, so a truncated message keeps the whole key body:

```
in:  could not parse key: -----BEGIN RSA PRIVATE KEY-----\nMIIEpAIBAAKCAQEA_KEYBODYSECRET\n... (truncated)
out: (unchanged)
```

truncation is pretty ordinary for driver text that embeds a PEM. would an alternation that also matches BEGIN-to-end-of-string cover it, or does that over-redact more than you want?

packages/server/src/pg_helpers.ts
```diff
@@ -61,3 +61,91 @@ export function redactPgSecrets(s: string): string {
...
+// as they appear when a driver or a serializer echoes the config it was handed.
+//
+// `password` is included even though redactPgSecrets has a `password=` pass:
+// that pass only matches the libpq `=` form, so a serialized config reporting
+// `"password":"..."` in JSON goes straight through it.
+//
+// Matched case-insensitively and in either casing convention, because the same
+// field arrives as `privateKey` from the API schema and `private_key` from a
+// driver that snake-cases its config before reporting it.
+const SECRET_FIELD_NAMES = [
```
Sha-Bang, line 75: **Should-fix** - this is a second hand-maintained list of the same thing — `logger.ts:97` already exports `SENSITIVE_KEY_NAMES`, and its "connection config credentials" block is this list almost exactly. nothing links them, so they drift on the next connection type.

worth noting `connection_public_view.ts:9-15` argues the case directly: "A denylist of secret-looking names puts the risk on the wrong case: the next connection type adds a credential field, nobody adds it to the list, and it ships readable." that's the allowlist rationale for the read path, but the drift half applies here too.

can this derive from `SENSITIVE_KEY_NAMES` instead? over-redacting the header names it also carries (`authorization`, `cookie`) seems harmless in error prose, and it'd leave one list to maintain.

packages/server/src/controller/connection.controller.spec.ts
```diff
@@ -880,6 +880,78 @@ describe("ConnectionController.testConnectionConfiguration name validation", ()
...
+               privateKey: SSH_PRIVATE_KEY,
+               privateKeyPass: SSH_PRIVATE_KEY_PASS,
+            },
+         },
+      } as never);
+
+      // The attempt cannot succeed against an unresolvable host. If it somehow
+      // did, this is no longer exercising the error path and must not pass on
+      // the strength of an empty message.
+      expect(status.status).toBe("failed");
```
Sha-Bang: **Nit** - flagging this only because it's tied to the controller finding: as written this exercises the returned-failure path in `service/connection.ts`, not the `catch` the PR changed, so it passes on `redactPgSecrets` and can't fail if `redactConnectionSecrets` is wrong.

the docblock is honest that it isn't the load-bearing test, so this isn't a complaint about the test itself — it just becomes real evidence once the redactor moves to the path this actually drives.

Sha-Bang (commented):
the redaction logic itself is a clear improvement over the Postgres-only passes, and the spec coverage in pg_helpers.spec.ts is genuinely thorough. main thing: i think it is wired into the wrong branch — the controller catch it upgrades only fires when testConnectionConfig throws, and that function resolves every failure instead. details inline, along with three shapes that escape the redactor (i ran them). reviewed with claude + cursor; codex could not run here.

sagarswamirao:
Fixed in 256f39be. You are right on every point, and the first one is the whole PR: the change was wired into a path
that does not execute, so nothing about a real caller's experience would have changed.

## The dead path

Confirmed. `testConnectionConfig` wraps its body in a catch-all that resolves every failure into
`{status: "failed", errorMessage: ...}` and returns it; the controller returns that object
verbatim. The controller `catch` only fires if the service throws, which it does not.

The part I have no defence for is that the comment I wrote on that line states the premise --
"the service resolves every failure rather than throwing" -- and I did not follow it to its
conclusion. I wrote down the reason the fix could not work and shipped it anyway.

Moved to `service/connection.ts`: the response at `:2949` and the log at `:2937`. Agreed that the
log is the stronger case of the two -- operator logs are read more widely than one caller's
response body -- so it is not the consolation half of this change.

The controller line stays as defence in depth, with a comment that now says it is not the
redaction point and why it cannot be tested, rather than implying it is.

## The controller test

Deleted rather than re-pointed. You showed it passes with the change reverted, for two
independent reasons, and its own comment half-conceded the second -- a test whose comment argues
against its own coverage is worse than no test, because it reads like assurance.

Replaced with three cases on `redactTestFailure` at the service boundary, driven with driver text
rather than through a live attempt: the failures reachable against an unroutable host
(`Cannot parse privateKey: Unsupported key format`, `getaddrinfo ENOTFOUND`) resolve before they
embed anything, so a test asserting on those passes either way -- the same trap as before.
Neutering the function gives 2 fail / 1 pass; the one that still passes asserts a clean message is
left untouched, which is correct behaviour for a neutered redactor too.

## Value-based, with the shapes as backstop

Took this. `redactTestFailure` collects the credential values out of the config this very request
supplied, removes those exact strings and their SQL-escaped forms, then runs the shape passes over
what is left.

Your framing of the trade is the reason for both rather than either: the value pass cannot miss a
field the schema gains, which is the failure mode of a name list; the shape pass catches an echo
the value pass cannot see, because it was re-encoded or reached the message from somewhere other
than the config in hand.

I did not import `materialization_service.redactConnectionSecrets` directly -- it imports
`escapeSQL` from `connection.ts`, so the dependency would have been circular. The collector is a
local one in `connection.ts` for now. Hoisting both to a shared module is the right follow-up and
I did not want it riding in this change.

## The name collision

Renamed mine to `redactConnectionSecretShapes`. Two exported functions called
`redactConnectionSecrets` in one package, where the import path silently decides which behaviour
you get, is a trap regardless of this PR. The name now says which of the two it is, and its
doc-comment points at the other one as the preferred choice where the config is available.

## sasUrl and the truncated PEM

Both confirmed and fixed.

The bare-value class no longer treats `&` as a terminator: a URL-valued field carries its
credential in a query parameter, so stopping there redacted the prefix and left the signature. The
cost is that a bare `a=1&b=2` tail following a secret-named field is swallowed whole, which is the
safe direction.

The PEM terminator is now optional with a fallback to end-of-string, as you suggested. Drivers
truncate, and a truncated key still contains most of the key.

Each has its own case now, carrying the shape that was failing -- a `&sig=` tail, and a PEM block
with no `-----END-----`. Both fail against the previous patterns, which the existing fixtures did
not, because neither contained the distinguishing shape.

## One thing your review made me find

Running the full server suite after the rewire -- which I should have been doing all along --
turned up two existing failures that were mine, not flake. The value pass removed the whole
`connectionString`, so a DuckDB attach failure came back as `Unable to connect to Postgres at
"***": connection refused`. No credential, and no host or port either.

`errorMessage redaction` asserts both halves: the password absent AND `127.0.0.1` still present.
It was right to fail. `connectionString` and `sasUrl` are now excluded from the value pass for
that reason -- they carry a credential inside a larger value whose remainder is the most useful
part of the message -- and the shape passes handle them in place instead.

It is the same property your "keeps a driver's prose" case is about, broken from the other
direction, which I would not have caught from my own specs alone.

## On the header comment

Fair, and worth saying plainly rather than working around: `pg_helpers.ts` now knows the shape of
eight connection schemas, which is not what "a Postgres/libpq helper at `src/` root so neither
layer depends on the other" anticipated. The value-based pass reduces how much that list has to
carry, but it does not make the coupling untrue. If you would rather the shape list live beside
the connection schemas and `pg_helpers` keep only the libpq grammars, say so and I will move it --
that is a better home and it is a small change.

---

Full server suite: 3871 pass, 3 skip, 1 fail. The one failure is
`dashboard tile normalization > stays linear ...`, a ~13s timing assertion that passes in
isolation and does not touch this path. `prettier:check` and `lint:server` both exit 0.

housejester (approved):
Re-reviewed `ed669678`. Everything from the first pass is fixed, and the two things you found on your own are better than what I asked for. Clearing my block.

## First, a correction I owe you

I wrote "neither pattern is a ReDoS risk" in my first review. That was wrong, and CodeQL was right. I checked the alternation backtracking and the value classes and never looked at `[A-Z0-9 ]*PRIVATE KEY` — a star class whose contents overlap the literal that follows it, which is the textbook polynomial shape. If my all-clear read as counter-evidence to the alert while you were working out whether to act on it, disregard it.

I verified the fix rather than the framing. CodeQL's exact stated shape, `"-----BEGIN PRIVATE KEY-----a".repeat(20000)`: **1ms** for the PEM pair, 1ms for the full pipeline. The `(?!-----BEGIN )` body guard is what makes the lazy body linear, and splitting into two patterns rather than offering an alternation is what stops the engine retrying the same input down two branches. Your 3ms-against-404ms measurement reproduces.

## What I checked

The three leak shapes are dead:

```
sasUrl=…?sv=2021&ss=b&sig=SIGSECRET&se=2030   →  sasUrl=***
-----BEGIN RSA PRIVATE KEY-----…(truncated)    →  ***
two PEM blocks in one message                  →  a *** b *** c
```

151 pass / 0 fail across `pg_helpers.spec.ts` and `connection.spec.ts`.

The `connectionString` / `sasUrl` exclusion from the value pass is the part I'd have missed. Removing a whole DSN passes a "no credential present" assertion while destroying the host and port that make the message worth reading, and the existing spec asserting `127.0.0.1` *survives* was right to fail. That is the property the "keeps a driver's prose" case is about, broken from the other direction, and running the full suite is what surfaced it. Worth keeping that habit.

The value-then-shape layering is the right resolution of the trade rather than a compromise between the two, and the `redactConnectionSecretShapes` rename with a doc-comment pointing at the value-based counterpart closes the name trap properly.

---

## Still open

**@Sha-Bang's escape-aware quote gap is not fixed.** `SECRET_QUOTED_PATTERN` at `pg_helpers.ts:108` still uses `"[^"]*"`. Confirmed live on this head:

```
in : config={"password":"ab\"TAILSECRET","host":"h"}
out: config={"password":***TAILSECRET","host":"h"}
```

Their suggestion is the right one — `redactPgSecrets` already carries the escape-aware form for single quotes (`'(?:\\.|[^'\\])*'`). This is the only remaining *leak* in the diff and it is one line, so I would land it here rather than defer it.

**The value pass misses `peakaKey`** — the one failure mode it exists to prevent. `CONFIG_SECRET_KEY` is `/pass(word)?|secret|private_?key|service_?account|access_?key|token/i`, and `peakaKey` matches none of those. Measured, with `privateKey` as the control:

```
{ peakaKey: PEAKA_SECRET_VALUE }   →  { peakaKey: PEAKA_SECRET_VALUE }
{ privateKey: PRIVKEY_SECRET }     →  { privateKey: *** }
```

The shape pass catches the `peakaKey=` form, so the residual is the bare-colon and re-encoded shapes — narrow. But the doc-comment says the value pass "cannot miss a field the schema gains," and it misses one the schema already has. The same gap exists in `materialization_service`'s collector, so it was inherited rather than invented.

**A pre-existing quadratic in `redactPgSecrets`, which this PR neither introduced nor made more reachable.** While isolating the CodeQL shape I found the genuinely slow pattern is two lines above the new code, in untouched territory. `([a-z][a-z0-9+.-]*:\/\/…)` — `[a-z0-9+.-]*` consumes a long alphanumeric run, then backtracks one character at a time looking for `://`, at every start position. No PEM marker required:

```
input "A".repeat(n):   n=50k → 1,352ms    n=100k → 5,705ms    n=200k → 23,936ms
```

Clean 4x per doubling, and it blocks the event loop. Reachable the same way everything else here is, since `error.message` derives from caller-supplied config. `redactPgSecrets` was already on this exact line before your change, so this is not a regression and I would not grow this diff with it. Bounding the scheme fixes it and is behaviour-identical on every shape the existing specs pin:

```
[a-z][a-z0-9+.-]{0,31}:\/\/        n=200k: 23,936ms → 8ms
postgres://u:***@h/d   x_postgres://u:***@h/d   amqp://guest:***@rabbit:5672/v   POSTGRES://U:***@H/D
```

## On your question about the header comment

Moving the shape list next to the connection schemas is better than where it sits, but I do not think it is the change worth making. @Sha-Bang's point is the sharper one, and the evidence got stronger with this commit — there are now four hand-maintained answers to "which keys are credentials":

| list | where | form |
|---|---|---|
| `SENSITIVE_KEY_NAMES` | `logger.ts:97` | name array |
| `SECRET_FIELD_NAMES` | `pg_helpers.ts:75` | name array |
| `SENSITIVE_KEY` | `materialization_service.ts:338` | regex |
| `CONFIG_SECRET_KEY` | `service/connection.ts` (new) | regex |

and the `peakaKey` gap above is that drift, already happened, in the list this PR adds. So the outcome I would push for is one list in one place that all four sites consume. I would leave the mechanism and the home to you, because you have already hit the constraint that shapes it — the `escapeSQL` circular import — and that is the same shared module you proposed for hoisting the collector. Both belong in one follow-up rather than in this PR.

Worth reading `connection_public_view.ts:9-15` before picking the shape: it argues the allowlist-over-denylist case directly for the read path, and may say the shared thing should not be a denylist at all.

---

Approving. I would land the escape-aware quote fix here since it is one line and it is the last live leak; the single credential-key list (carrying `peakaKey`) and the `redactPgSecrets` scheme bound are both worth their own issues. Happy to open those two if it is useful.

<sub>Reviewed with Claude Code (session "Sleepy Tapir"). Measurements run on `ed669678`; scratch changes reverted.</sub>

sagarswamirao:
Both of the open items are in, as d024ca96. I reproduced each against this head before changing anything.

## The escape-aware quote gap

Confirmed exactly as you had it:

```
in : config={"password":"ab\"TAILSECRET","host":"h"}
out: config={"password":***TAILSECRET","host":"h"}
```

Fixed with the form `redactPgSecrets` already carries, applied to both quote styles:
`"(?:\\.|[^"\\])*"` and `'(?:\\.|[^'\\])*'`. Agreed it belonged here rather than in a follow-up --
it was the last live leak in the diff and the fix is one line.

The test is load-bearing rather than decorative: against the naive `"[^"]*"` it fails with the
tail still present, and passes with the escape-aware form. Credit to @Sha-Bang for finding it; I
had read that pattern twice and not seen it.

## peakaKey

Also confirmed, with `privateKey` as the control:

```
CONFIG_SECRET_KEY matches peakaKey:        false
CONFIG_SECRET_KEY matches privateKey:      true
CONFIG_SECRET_KEY matches sasUrl:          false   (excluded on purpose)
CONFIG_SECRET_KEY matches connectionString: false  (excluded on purpose)
```

Added `peaka_?key`. Your framing is the uncomfortable part and it is right: the doc-comment says
the value pass "cannot miss a field the schema gains", and it was missing one the schema already
had. A claim about a mechanism, contradicted by the mechanism, in the same commit.

`sasUrl` and `connectionString` stay out deliberately -- they carry a credential inside a larger
value whose remainder is the diagnosis, which is the thing the earlier round broke.

## On your correction

Noted, and thank you for writing it down rather than quietly moving on. It did not cost me
anything: CodeQL flagged the shape, I could not reproduce an exploit from it, and I fixed the
structure anyway because the measurement on the *terminated* pattern was real. If anything your
retraction is the useful half -- `[A-Z0-9 ]*PRIVATE KEY` overlapping the literal that follows it
is the shape I would not have looked at twice either.

## The two follow-ups

Agreed on both, and agreed they do not belong in this diff.

**One credential-key list.** Your table is the argument -- four hand-maintained answers, and the
`peakaKey` gap is that drift already happening in the list this PR adds. I will file it. The
`escapeSQL` circular import is the constraint: `materialization_service` imports it from
`connection.ts`, so the shared home has to sit below both, which is the same module the collector
would hoist into. Both in one issue.

I will read `connection_public_view.ts:9-15` first, as you suggest. If the read path's case for
allowlist-over-denylist generalises, the shared thing should not be a denylist at all, and that
changes what the follow-up is rather than just where it lives.

**The `redactPgSecrets` scheme bound.** 23,936ms at 200k on an untouched line is worth its own
issue regardless of this PR, and it is reachable the same way everything else here is. Your
`{0,31}` bound is behaviour-identical on every shape the specs pin. Please do open it -- you have
the measurements, and it is cleaner attributed to the pass that found it than folded into mine.

## Verification

`pg_helpers.spec.ts` 57 pass / 0 fail; the redaction cases across it and `connection.spec.ts`
are 67 pass / 0 fail after rebasing onto the main merge that landed while I was working. Full
server suite 3873 pass / 3 skip / 1 fail -- the failure is
`dashboard tile normalization > stays linear`, an ~11s timing assertion that passes in isolation
and does not touch this path. `lint:server` and `prettier:check` both exit 0.

PR #1198, opened 2026-09-18
fix(authorize): carry a #(secure) marker through package load so a deployment can act on it

sagarswamirao:
Carries a `#(secure)` marker through package load, so a deployment in front of Publisher can act on it, and pins that Publisher does not drop it.

## The reported finding

From a multi-tenant security review, rated MEDIUM and described as mitigated in deployment with one real footgun:

> Row-level gates trust caller-supplied givens. A proxy in front can strip the caller's value and inject an identity-resolved one, but it only strips names registered as trusted. A gate written on a `$NAME` that is not registered can be satisfied by a caller-supplied given.
>
> **Fix:** Ensure every gate-referenced given is a registered trusted name; ideally validate this at model publish.

## What was already there

Most of that fix exists, which is worth stating so the remaining gap is not read as larger than it is.

**A gate referencing an undeclared given already fails model load.** The gate is an ordinary expression, so Malloy's compiler refuses `$NOPE` as undeclared before any gate-specific validation runs. `authorize_integration.spec.ts` has pinned this since row-level gates landed.

**"Registered" is not a concept this server has.** A trusted-name registry is a deployment-side facility: it needs an identity to resolve a value from, which Publisher does not have. So "validate at publish that every gate-referenced given is registered" cannot be implemented here in the form the finding states.

## What this PR does

Publisher's part of that contract is narrow and it is the part this PR pins: **carry the marker so the deployment can see it.** A `#(secure)` annotation that does not survive package load is the real fail-open, because a deployment can only strip a caller-supplied value for a given it can see is secure. Every gate over a dropped marker would be silently caller-satisfiable, with nothing in any response saying so.

The diff is a test and a documentation correction. **No production file changes**, and no behaviour changes: the marker already survives, and this is what holds it that way.

| Change | Why |
|---|---|
| Two tests asserting the marker reaches `getSources()` on the real load path | The shape a deployment actually receives, not an internal field |
| One test asserting it is absent when the given is annotated otherwise | Pins the predicate in both directions |
| Four `it.each` guards over every given shape | Each must keep loading |
| `docs/givens.md` | Two corrections, below |

The tests run on `Package.create` -> `PackageLoadPool` -> `Model.fromSerialized` rather than on a bare model. That distinction matters: `Package.create` throws on a per-model compile failure, so `resolves.toBeDefined()` genuinely proves the model compiled, where the permissive pool path would have made the same guards vacuous.

## Why not refuse a scalar `#(secure)` given

An earlier revision of this work refused, at package load, a `#(secure)` given declared with a scalar type. That was reverted, and the reasoning is worth recording because the refusal reads as the safer option.

It removed a working configuration, and the remedy it prescribed does not work. Measured against the engine:

| Shape | Loads | Queries |
|---|---|---|
| `ROLE :: string`, gate `$ROLE = 'admin'` | yes | correct: `admin` returns rows, `viewer` returns none |
| `ROLE :: filter<string>` | yes | `Parser Error` -- emits `WHERE ((INTERNAL ERROR FILTER EXPRESSION VALUE SHOULD NOT BE USED='admin'))` |
| `ROLES :: string[]`, gate `$ROLES = 'admin'` | yes | `Conversion Error: Type VARCHAR ... can't be cast` |
| `ROLES :: string[]`, gate `role in $ROLES` | yes | runs, but returns a row for both `admin` and `viewer` |

So the only shape that both loads and answers correctly is the scalar the refusal would have rejected, and the closest multi-valued alternative changes what the gate asks rather than retyping it.

The deeper reason is a layering one. Whether a scalar `#(secure)` given is usable depends on the attribute store in front of Publisher -- a constraint belonging to that deployment, not to the `given:` grammar. Enforcing it here would make the open-source server refuse a model on a rule it cannot see and does not own. Publisher's standing is to carry the marker faithfully; acting on it is the deployment's.

## Documentation

`docs/givens.md` gains a `#(secure)` section, and two things in it were wrong:

**The array-given claim.** The file said array and record givens are "not among" the accepted types, citing an empty-array default as proof. Those are two different claims. `given: ROLES :: string[]` compiles and the given surfaces with type `array`; only the `is []` default fails, with `unexpected ']'`. Verified by compiling both against a DuckDB-backed runtime.

**The tense.** The new paragraph described a deployment-side limitation as permanent fact in a reference document. Whether that limitation is a design property or a gap is an open question, so the paragraph is now tensed and points at the deployment's own documentation for the current requirement. Either answer leaves it true.

## Testing

`row_level_authorize.integration.spec.ts`, the `#(secure)` block: **7 pass, 0 fail.**

The predicate is pinned in both directions, which took two attempts and is worth recording:

- Replacing `routeOf(...) === "secure"` with a filter that always returns true makes **the over-match test fail**, and nothing else.
- An earlier version of that test used a given with no annotation at all. It does not work: an unannotated given carries an empty annotation array, so filtering it proves nothing and an always-true predicate passes it unchanged. The test now annotates the given `#(insecure)` -- non-empty, and adversarial, since it contains the substring the predicate must not match on.

`#(insecure)` is also why the predicate reads Malloy's own routing rather than matching text: it contains "secure" and means the opposite, and an earlier revision's `note.includes("secure")` accepted it.

## Not in scope

- Whether the deployment-side multi-valued rule is a design property or a gap. Nothing here enforces the constraint, so this PR is correct either way; the one paragraph sensitive to the answer is tensed for it.
- A gate over a multi-valued given compiles clean and then fails at query time (the second and third rows above). That is upstream of Publisher and wants its own issue.


housejester:
The bug this closes is real, and the framing is the right one — a marker that protects nothing is worse than no marker, and refusing the load is consistent with `assertNoMisplacedAuthorizeAnnotations` next door. The three-tier test split is a good instinct too.

The problem is that as wired, none of it runs. Findings below are validated against `a47005da`; commands and outputs included so they're cheap to re-check.

---

## 🔴 The refusal is unreachable in production

`assertNoScalarSecureGivens` is wired only into `Model.create` (`model.ts:3008`). Packages are not served through `Model.create`:

- Production load is `Package.create` → `getPackageLoadPool()` → `package_load_worker` → `Model.fromSerialized`.
- `Model.create` has exactly **one** non-test caller in the repo — `environment.ts:692`, the `/compile` gate model, taken only for a file on disk that isn't in the cached package model.
- And that caller can't observe the refusal either: `Model.create` catches every throw and returns a `Model` carrying the error (`model.ts:3137-3162`) rather than throwing. `environment.ts` never reads it.

So the check fires only in tests that read `getNotebookError()` / `compilationError` directly — which is every test in this PR.

Verified. A package on disk with `#(secure) ROLE :: string`, loaded through the real pool:

```
Package.create threw = undefined
notebookError       = undefined
compilationError    = undefined
```

Control on the same harness, a refusal that *is* mirrored into the worker (gate given with a default) — throws as expected:

```
#(authorize) on source "X" references `$ROLE`, which is declared with a default…
```

And the `/compile` branch, instrumented to prove it's genuinely taken:

```
[env] hasExactGateModel= false source? true model= added.malloy
[env] TAKING Model.create branch for added.malloy
compileSource returned problems = []
```

Every sibling authoring check is mirrored into the worker at **two** sites with an explicit `// See the identical check in Model.create` — `assertPartitionAnnotationsValid` (`package_load_worker.ts:749`, `:979`) and `assertNoMisplacedAuthorizeAnnotations` (`:755`, `:981`). This one is in neither.

Worth flagging that the PR description says the `shipped_example_authorize.spec.ts` tier exists "because the refusal has to fire on the package path rather than only on a bare model" — but that spec calls `Model.create` too, so it exercises the same path as the integration spec. That's the tier that would have caught this.

**Recommend:** mirror both worker call sites, and re-point the package-tier test at `Package.create` + `PackageLoadPool`. `row_level_authorize.integration.spec.ts:406` — *"CRITICAL — loads cleanly through the REAL worker-pool fromSerialized path"* — is that harness, and its comment says it exists for exactly this reason: *"the only way to pin that the fromSerialized path stays in lockstep… rather than trusting that a fix to one path also covers the other."*

## 🔴 `filter<string>` is classed scalar, and it's the documented multi-value type

`model.ts:3012` renders a filter given as `filter<string>`; `isSetValuedGivenType` accepts only `…[]` or `array`, so it's refused:

```
A `#(secure)` given must be set-valued, but these are scalar:
  - `CATEGORY` declared `filter<string>`
… Declare each as a list (for example `string[]` rather than `string`) …
```

`docs/givens.md:76-78` names `filter<string>` as *the* way to let a caller pass several values. And `#(secure)` enters this repo only through the dashboard builder — `git log -S "#(secure)"` gives #1158 and #1172 — where every occurrence is on a `filter<string>` given, and `spliceDocument.ts:154`/`:588` are production code written to preserve that exact shape.

This is latent only because of the finding above. Fix that without fixing this and every package carrying `#(secure) … :: filter<string>` stops loading — and it's the whole package, not a degraded gate, since an authorize-family throw from the worker aborts `Package.create` outright.

**Could you confirm with the registry side whether `filter<T>` is registerable?** If it is, it belongs in `isSetValuedGivenType`. If it genuinely isn't, the refusal still needs a migration note, because retyping is a real authoring change.

## 🟡 The matcher is a strict subset of Malloy's own routing

`SECURE_ANNOTATION_PATTERN` (`authorize.ts:1133`) misses 6 of the 7 spellings Malloy routes to `secure`. Probed with `routeOf` from `@malloydata/malloy` — the same call `noteRoute` (`authorize.ts:67`) already makes, in this file:

| note text | `routeOf` | regex |
|---|---|---|
| `#(secure)` | `secure` | ✅ |
| `#(secure) keep this server-side` | `secure` | ❌ |
| `##(secure)` | `secure` | ❌ |
| `#\|(secure)` | `secure` | ❌ |
| `#[secure]` / `#{secure}` / `#<secure>` | `secure` | ❌ |
| `#(SECURE)` | `SECURE` | ❌ (correctly — different route) |
| `# secure data` | `""` | ❌ (correctly) |

This module already states the rule in the opposite direction, ~1000 lines up: *"A rejecter that accepts MORE than the parser is a 400 on odd caller input, which is safe. **One that accepts LESS is a forged-gate bypass.**"* It notes the regex form exists only because `assertNoCallerAuthorizeAnnotation` reads pre-compile text, *"while the parser gets to ask `noteRoute`."* Here the parser has run. And `#(secure) <prose>` isn't exotic — it's the natural form given that Malloy reads a route to the first whitespace (`docs/givens.md:96`).

**Recommend:** `noteRoute(note) === "secure"` over the folded note list, and drop the regex. That also picks up `AnnotationsDef.inherits` (`malloy_types.d.ts:1171`), which the `blockNotes` + `notes` mapping at `model.ts:3018` skips and which `malloyGivenToApi` folds via `annotations.forRoute(undefined)` — I couldn't construct an inherited-given case to demonstrate, so treat that half as reading rather than a shown gap.

## 🟡 Three stated justifications don't hold in this repo

- `authorize.ts:1152` — *"Exported for the two load paths to share… both hand their assembled form here."* There's one call site; `grep findScalarSecureGivens src/ | grep -v spec` returns only the definition. The comment describes the design the first finding asks for.
- `authorize.ts:1139` — *"the rendered form used in the API and by the trusted-name registry is `string[]`."* `malloyGivenToApi` (`given.ts:249-252`) falls through to bare `type.type` exactly as `model.ts` does. Confirmed: `getModel()` returns `[{"name":"ROLES","type":"array",…}]`. The `endsWith("[]")` branch is unreachable from both producers — only the synthetic case at `authorize.spec.ts:246` exercises it. Keeping it as registry-facing defensiveness is fine; the justification just sends the next reader looking for an API that renders `string[]`.
- The message prescribes `string[]`, and `docs/givens.md:75-78` says that's a compile error. Half-stale: `ROLES :: string[]` compiles now (rendering `array`), but `is []` really is still `unexpected ']'` — so the remedy is reachable only with **no default at all**. Relatedly `given.ts:239-247` says *"If the grammar expands to allow array or record givens, the bare `type.type` discriminator… will land in the wire response with no element info — revisit when that happens."* It has, and this PR makes `array` mandatory for secure givens. Worth updating both in this PR, since it's the change that makes array givens load-bearing.

## 🟡 No RELEASE_NOTES entry, and `#(secure)` is undocumented

`RELEASE_NOTES.md:15`: *"For releases that warrant narrative — redesigns, **breaking changes**, migration steps — write a `## [Unreleased]` section below, in the PR that changes the behaviour."* Once this actually fires it refuses packages that load today, with a remedy (retype, drop the default) that needs stating.

Separately, `#(secure)` appears nowhere in `docs/` and has never had a release note — so this is the first place the project takes a position on what the marker requires. A short entry in `docs/authorize.md` or `docs/givens.md` would give the refusal something to point at.

## 🟢 Minor

- "fails closed by resolving to the empty list" is the registry's behaviour, not the server's. On its own an unvalued array given throws `Given 'ROLES' has no value and no default` — same kind of hard failure as the null scalar the message contrasts it against. A caller-supplied `{ROLES: []}` does work (zero rows), and `{ROLES: "admin"}` is correctly rejected `expected array, got string`. One clause in the message would save someone debugging that.
- `assertNoScalarSecureGivens` takes the already-found list, so the "names every finding" case at `authorize.spec.ts:253` never exercises `findScalarSecureGivens`.

---

## Rollout

No risk as it stands, for the two reasons above — nothing that loads today stops loading. Nothing in-repo is affected either: all 54 `.malloy`/`.malloynb` files scanned for any secure-routed spelling, zero hits. `bun run test:unit` on the branch is **3928 pass / 3 skip / 0 fail**.

The risk arrives with the fix. Before wiring it into the worker, it'd be worth (a) settling `filter<T>` with the registry side, and (b) sweeping real published packages for a secure-routed annotation on a non-array given — an in-repo scan can't stand in for that, since the marker is undocumented here and anyone using it learned it deployment-side.

If that sweep is slow, one option is to land the detection now as a load-time **warning** on the worker path — surfacing through `loadErrors` / `get_status`, where a broken or stale package already shows up — then flip it to a refusal with a migration note once the population is known. That gets the signal out of the unread server log, which is the actual complaint in the description, without pricing a breaking change on an unknown population.

**Top three before merge:** 1) wire into both `package_load_worker.ts` sites and re-point the package-tier test at the real pool harness; 2) settle `filter<string>`; 3) swap the regex for `noteRoute(…) === "secure"`.

---

<sub>Reviewed by Claude Code (session "Mellow Otter") on behalf of @housejester. Findings validated against `a47005da`; scratch harnesses were removed after running.</sub>

sagarswamirao:
Three things to flag on the wiring before this merges. The rule itself looks right, and the unit cases over `findScalarSecureGivens` are good -- but as wired, a scalar `#(secure)` given still loads clean in the cases that matter most. All three fail open, which is the direction this change exists to close.

**1. The check does not run on the package-load path.**

`assertNoScalarSecureGivens` is called only from `Model.create` ([model.ts#L3017](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/model.ts#L3017)). A published package never goes through it: `Package.create` calls `loadViaWorker` unconditionally ([package.ts#L474](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/package.ts#L474)), and so does `reloadAllModels` ([package.ts#L2186](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/package.ts#L2186)). There is no in-process fallback -- the pool throws rather than degrading. `Model.create`'s only live non-test caller is the ephemeral gate model in `/compile` ([environment.ts#L692](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/environment.ts#L692)).

Every sibling check is mirrored into the worker -- `assertPartitionAnnotationsValid` at [package_load_worker.ts#L749](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/package_load/package_load_worker.ts#L749), `assertNoMisplacedAuthorizeAnnotations` at [#L755](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/package_load/package_load_worker.ts#L755), plus the legacy-string-gate and at-most-one-gate checks. This one is not; `grep secure packages/server/src/package_load/package_load_worker.ts` returns nothing.

Fix: add the same block at `package_load_worker.ts:~749` (before `assertNoMisplacedAuthorizeAnnotations`, matching `Model.create`'s ordering) and at `:~978` for the notebook path. `modelDef` is already in scope at both points, so no new plumbing. The doc on `findScalarSecureGivens` ([authorize.ts#L1151](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/authorize.ts#L1151)) already says it was exported "for the two load paths to share" -- worth making that true.

Worth one worker-path test too: all three added specs call `Model.create`, including the package-on-disk one, so the suite can't currently tell "wired" from "not wired."

**2. The matcher is narrower than Malloy's router.**

`SECURE_ANNOTATION_PATTERN` ([authorize.ts#L1133](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/authorize.ts#L1133)) is an exact-line match on `#(secure)`. Running Malloy's own `parsePrefix` against it, these all route to `secure` and are missed:

| annotation | route | matcher |
|---|---|---|
| `#(secure)` | `secure` | match |
| `#\|(secure)` (block form) | `secure` | **miss** |
| `#[secure]` / `#<secure>` / `#{secure}` | `secure` | **miss** |
| `#(secure) role list` (trailing comment) | `secure` | **miss** |

`BRACKET_PAIRS` is `()`, `<>`, `[]`, `{}`; the sigil strip `^(#{1,2})(\|?)` admits the block form; and the prefix ends at the first whitespace, so a trailing comment leaves the route intact. This is the case the module header calls out directly -- "What counts as the tag is Malloy's answer, not a regex of ours... missing one Malloy DOES route here serves every row from a source its author locked." Every other matcher in the file goes through `noteRoute`; this is the only raw-text positive detector.

Fix: `noteRoute(text) === "secure"`, mirroring `authorizeNoteContent` at [authorize.ts#L137](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/authorize.ts#L137).

**3. `modelDef.givens` is the wrong source.**

The mapping at [model.ts#L3019](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/model.ts#L3019) hand-rolls `blockNotes.concat(notes)`, which skips the `inherits` chain that `notesInOrder` walks -- so a `#(secure)` inherited from an import is invisible. It also includes inline and `finalizeGivens`-bound givens that `compiledModel.givens` filters out, so a finalized scalar given carrying the marker would fail the whole load despite not being caller-suppliable, and the message would name the declaration-site name rather than the surface name.

Fix: feed the check from `malloyGivens` / `malloyGivenToApi` output, already computed a few lines above at `model.ts:2991`. That picks up the inheritance walk via `forRoute` for free and drops the duplicated type-rendering expression. The comment right there already says it: "Malloy's `Model.givens` already collapses inheritance from imports and applies any `finalizeGivens` runtime config. Just read it."

**One doc correction.** The note at [authorize.ts#L1136-L1142](https://github.com/malloydata/publisher/blob/77c823d314de9b53fb4e3f233949348058f0ee6b/packages/server/src/service/authorize.ts#L1136-L1142) says the API renders a list given as `string[]`. `malloyGivenToApi` emits the bare `type.type` -- `"array"` -- identical to the model.ts path, so there is no second producer emitting `string[]`; the only renderer that does is Malloy's `typeDefToString`, which isn't called here. The `endsWith("[]")` branch is harmless, but the stated reason for it doesn't hold. Separately, a `record`-typed given (`{org::string}`) renders `"record"` and is currently refused -- defensible since a record isn't a set, but the "Declare each as a list" wording doesn't fit it, so worth making that call deliberately.

Could you wire the worker path and swap the matcher to the route check? Those two are what make the refusal actually hold.

sagarswamirao:
All three are fixed, and the third was a real bug I had not seen. Taking them in
the order you raised them.

**The check now runs on the package-load path.** Wired at both
`package_load_worker.ts` sites, before `assertNoMisplacedAuthorizeAnnotations`,
matching `Model.create`'s ordering. You were right that the suite could not tell
wired from unwired: every spec called `Model.create`, which catches its own
throws and returns a Model carrying the error, so `getNotebookError()` reads the
same either way.

The package-tier spec is re-pointed at the real pool harness in
`row_level_authorize.integration.spec.ts` -- `Package.create` -> `PackageLoadPool`
-> `Model.fromSerialized`, the path a published package actually takes. Verified
by deleting the worker call site: that spec goes red, and green again when
restored. It could not do that before.

**The matcher is `noteRoute(note) === "secure"`.** Your table is the argument --
the block form, the other bracket pairs and a trailing comment all route to
`secure` and the exact-line regex missed every one. This module already states
the rule in the direction that matters: a rejecter accepting LESS than the parser
is the dangerous one. There is no reason for the only raw-text positive detector
in the file to be the exception.

**`modelDef.givens` was the wrong source, and it fails both ways.** This is the
one I had missed. The hand-rolled `blockNotes.concat(notes)` skips the `inherits`
walk, so a `#(secure)` inherited from an import was invisible -- fail open. And it
includes `finalizeGivens`-bound givens that the api surface filters out, so a
finalized scalar carrying the marker would have failed a whole package load for a
given no caller can supply -- fail closed on a non-problem, and naming the
declaration-site name at that.

Now fed from the `malloyGivenToApi` output computed a few lines above, which is
what the comment there already says to do. One thing I checked before switching,
since it would have silently broken the detection: `malloyGivenToApi` drops
reserved routes, but `isReservedRoute` excludes only empty or non-alphanumeric
ones, so `secure` survives.

**The doc correction is taken.** Nothing here emits `string[]` -- both producers
render `array` -- so the comment claiming an API renderer no longer says that.
The branch stays as registry-facing defensiveness and now says so.

On the record-typed given: `filter<string>` is now accepted, which is the case
that mattered, since every `#(secure)` given in this repository is
`filter<string>`. A `record` is still refused and I have left it that way
deliberately -- it is not a set, so the sentinel argument does not reach it -- but
you are right that "declare each as a list" does not fit, and the message now
prescribes `filter<string>` rather than `string[]`, which `docs/givens.md` says
is a compile error anyway.

## One thing worth settling before this merges

This PR refuses at the publisher layer, and that may be the wrong layer.

`#(secure)` arrived in ms2data/service PR-D4, whose first cut hard-failed the
publish and was reversed after review, for two recorded reasons: existing models
already gate over unmarked givens, so a hard fail breaks their re-publish; and
`#(authorize)` is an open-source Publisher construct that is forgeable only under
middleware resolving secure givens, so a publisher-layer check must not
hard-require a middleware concept. It concluded that hard enforcement belongs in
the deployment's own layer.

The scalar case is consequently already detected there as a non-blocking advisory
in `malloy-code-server`, whose rule is the same one this PR implements as a
refusal. So the sweep you asked for in the first review has an answer: the
population is whoever authored against that advisory without acting on it.

Happy to convert this to a load-time warning surfaced through `loadErrors`, which
was your own suggestion and matches that decision, or to close it if the advisory
is judged sufficient. The `filter<string>` finding survives either way -- the
existing advisory checks for `[]`, so it misses the one type every secure given
actually uses.

housejester:
All three fixes land, and I re-validated each on the real path rather than on the description — `Package.create` → `PackageLoadPool` → `Model.fromSerialized`, 11/11, plus `bun run test:unit` at 3932 pass / 3 skip / 0 fail.

| case | before | now |
|---|---|---|
| scalar `string` + `#(secure)`, pool path | loaded | **refused**, names `ROLE` and `string` |
| `filter<string>` + `#(secure)` | refused | **loads** |
| `#(secure) resolved from identity` (trailing text) | missed | **refused** |
| `#[secure]`, `#<secure>` | missed | **refused** |
| `# secure data` (route `""`) | — | loads (correctly ignored) |
| `.malloynb` with a scalar secure given | — | **refused** (second worker site works) |

I also checked the claim the rewrite rests on instead of taking it: `isReservedRoute("secure")` is `false` (`annotations.ts:38` excludes only the empty route and punct-only routes), so the marker does survive `malloyGivenToApi`'s filter. And `givens` is bound at `model.ts:2994`, ahead of the assert at `3017`, so all three call sites genuinely share one reading rather than one of them quietly seeing `undefined`.

Two notes on my own earlier comment. Swapping the regex for `noteRoute` was the half-fix — I read it as a matcher bug when the input set was wrong in *both* directions, and sourcing from `malloyGivenToApi` is the change that actually fixes the cause. And `##(secure)` turns out to be a Malloy syntax error inside a `given:` block, so that row of my table was never reachable in that position.

---

## The remedy the refusal prescribes does not produce a working gate

This is the one thing I'd want settled before merge, and it holds whatever enforcement level this lands at.

Taking the exact shape the refusal targets — a scalar given behind an `#(authorize)` gate — and applying the migration the release note prescribes:

| | loads | every query |
|---|---|---|
| **A** `ROLE :: string`, gate `$ROLE = 'admin'` — *the shape now refused* | ✅ | ✅ **correct** (`admin` → all rows, `viewer` → none) |
| **B** `ROLE :: filter<string>`, same gate — *what the release note prescribes* | ✅ | ❌ `Parser Error: syntax error at or near "ERROR"` |
| **C** `filter<string>` + `role ? $ROLE` (idiomatic filter use) | ❌ `Invalid #(authorize) annotation` | — |
| **D** `ROLES :: string[]`, same gate | ✅ | ❌ `Conversion Error: Type VARCHAR with value 'admin'` |
| **E** `ROLES :: string[]` + `role in $ROLES` | ✅ | ✅ but **different semantics** |

So the refusal rejects a model that works and points the author at one that emits broken SQL on every query. The only migration that runs is **E**, and it isn't a retype — it changes the gate from "is this caller an admin", where an admin sees every row, into "which rows match this caller's set". **A** returned both rows for `admin`; **E** returns one.

Related, and worth a look on its own: **B** and **D** both compile clean and fail at runtime. A gate over a multi-valued given should either work or be a compile error — emitting `ERROR` into the SQL is neither. Whatever happens to this PR, that looks like its own bug.

It also means `authorize_integration.spec.ts`'s positive case — `ROLES :: string[]` with `#(authorize) $ROLES = 'admin'`, asserted to load clean — is shape **D**, which throws on every query. As written it can't tell a correctly-authored package from an unusable one, which is the property that would make the refusal safe to trust.

## A question on layer, for @Sha-Bang

You raised the layering point yourself, and I think it's the crux, so I'd rather put it to Nathan than have us settle it between us.

**Is "a `#(secure)` given must be multi-valued" a property of the design, or a gap in the current registry that you'd expect to close?**

It matters because the two answers want different artifacts. If it's the design, this PR is right and `docs/givens.md` is the right home for the rule. If it's a gap, then a hard load-time refusal is the most expensive possible way to encode it:

- it writes the constraint into the language reference, so when the gap closes both the doc and every model written against it are wrong;
- the migration isn't free to undo — per the table it changes the caller's API contract, and in the only working form the gate's meaning;
- and it ships in an OSS release marked ACTION REQUIRED, so removing it later is a second breaking change.

A warning costs nothing to delete.

## Either way

Some of this is independent of where that lands, and I'd take it regardless:

- Keep the `noteRoute` matcher and the `malloyGivenToApi` sourcing. Both are strictly better at any enforcement level, and the `inherits` / `finalizeGivens` fix was a real bug.
- Keep `filter<string>` in the type rule. That one matters most wherever the check finally lives — a rule keyed on `[]` alone gets the one type every secure given actually uses wrong.
- Fix the prescribed remedy. `filter<string>` and `string[]` both fail at query time against an equality gate, so the release note and the refusal message are currently pointing authors somewhere that does not work.
- `docs/givens.md:74` still reads "**Array and record givens are not among them**", citing `string[] is []`, about fifteen lines above a new section recommending an `array` type. The declaration `ROLES :: string[]` does compile and renders `array`; only the `is []` default is `unexpected ']'`. Worth splitting those two sentences, since this PR is what makes array givens load-bearing.

And if the call is to ship the refusal anyway as an interim control while the end state gets worked out — a legitimate choice, since the fail-open it closes is real today — then I'd only ask that it be framed as exactly that: the docs section worded as a current deployment limitation rather than a language rule, and the remedy text replaced with one that produces a working gate. A stopgap that says it's a stopgap is easy to retire. One that has been written into the reference as design is not.

---

<sub>Reviewed by Claude Code (session "Mellow Otter") on behalf of @housejester. Validated against `fd773b6b`; scratch harnesses removed after running.</sub>

packages/server/src/service/authorize.ts
```diff
@@ -1116,3 +1116,134 @@ export function assertNoLegacyStringGate(
...
+      found.push({ name: given.name, type: given.type });
+   }
+   return found;
+}
+
+/**
+ * Refuse a model load that marks a scalar given `#(secure)`.
+ *
+ * A secure given is set-valued by design, and the reason is the fail-closed
+ * sentinel: an unvalued set-valued attribute resolves to the empty list, which
```
sagarswamirao, line 1217: Small drift left over from the rename: this doc still says "set-valued by design" and "resolves to the empty list", while the predicate is now `isMultiValuedGivenType` and the thrown message below says "carry more than one value" / "the fail-closed sentinel is an empty set". Since `filter<string>` is the form this actually accepts (and the one `docs/givens.md` points authors at), "set-valued"/"empty list" now reads as if a list type were required -- which the grammar rejects.

Could you align the wording here with the message body? Same for the summary line above it, which still says "marks a scalar given `#(secure)`".

packages/server/src/service/model.ts
```diff
@@ -3006,6 +3008,16 @@ export class Model {
             // authorize checks below.
             assertPartitionAnnotationsValid(modelDef);
 
+            // A `#(secure)` marker on a scalar given fails OPEN the same way
```
sagarswamirao, line 3011: Same wording drift as the JSDoc: "a `#(secure)` marker on a scalar given" predates the widening to `isMultiValuedGivenType`. Worth matching the phrasing the refusal message uses now ("a given that cannot carry more than one value") -- the worker's two call sites already say it that way, so this is the one spot left out of the rename.

sagarswamirao:
Re-ran the gate table independently against a local Malloy + DuckDB build rather than take it as read, and it reproduces. Two results worth adding, because they go a bit further than the table.

**A discriminates -- it is not a no-op gate.** The shape the refusal now rejects works correctly:

```
ROLE :: string, gate $ROLE = 'admin'
  ROLE='admin'   -> 2 rows
  ROLE='viewer'  -> 0 rows
```

So this is a correct, functioning gate being refused, not one that merely happens to load. And the prescribed replacements fail where you said:

```
B  filter<string> + $ROLE = 'admin'   -> Parser Error: syntax error at or near "ERROR"
C  filter<string> + role ? $ROLE      -> Cannot use the '=' operator with a filter expression
D  string[]       + $ROLES = 'admin'  -> Conversion Error: Type VARCHAR ... can't be cast
E  string[]       + role in $ROLES    -> runs
```

E confirms your semantics point from the other side: it returns 1 row for `admin` and 1 row for `viewer`. A returned every row to an admin. So E is a change to what the gate asks, not a retype.

**The fail-closed premise does not reproduce at this layer.** The rationale is that an unvalued multi-valued given resolves to an empty set that matches nothing, whereas a null scalar is rejected outright. Testing each empty form:

```
string[] unvalued       -> "Given 'ROLES' has no value and no default"
string[] is []          -> compile error: unexpected ']'
filter<string> is f''   -> Parser Error / operator error, depending on the gate
```

Neither multi-valued type has a writable empty form here, and the unvalued array is rejected the same way the doc says only a scalar is. That does not disprove the registry-side asymmetry -- Publisher has no identity source, so it could hold there and not here -- but nothing in this repo demonstrates it, and `docs/givens.md` now states it as a language rule. Which is exactly the design-or-gap question, so I'd defer to @Sha-Bang on it too.

One thing I would take regardless of where that lands: **the positive test should run a query, not just load.** `authorize_integration.spec.ts:3850-3871` asserts `getNotebookError()` is undefined for shape D -- which passes, and D throws on every query. That is the same gap as the original `Model.create`-only tests: asserting the right thing at the wrong layer. A test that ran one query would have caught B and D before either of us looked.

And a small one, independent of all the above: `docs/givens.md:74` still reads "Array and record givens are not among them", about fifteen lines above the new section recommending an array type. `ROLES :: string[]` does compile and renders `array`; only the `is []` default is `unexpected ']'`. Worth splitting those two sentences since this PR makes array givens load-bearing.

Separately, B and D compiling clean and failing only at query time looks like an upstream bug in its own right, whatever happens to this PR.

packages/server/src/service/authorize.ts
sagarswamirao: Fixed in `a3e977e7`. Both were leftovers from widening `isSetValuedGivenType` to `isMultiValuedGivenType` -- I renamed the predicate and the message and left the prose describing the older, narrower rule.

The summary line, the reason paragraph and the call-site comment now all say "cannot carry more than one value", matching the refusal message and the two worker call sites. Worth fixing rather than a nit, as you say: "set-valued" and "empty list" point at a list type, and `ROLES :: string[] is []` is the declaration the grammar rejects -- so the prose was steering authors at the one form the message does not prescribe.

packages/server/src/service/model.ts
sagarswamirao: Fixed in `a3e977e7`. Both were leftovers from widening `isSetValuedGivenType` to `isMultiValuedGivenType` -- I renamed the predicate and the message and left the prose describing the older, narrower rule.

The summary line, the reason paragraph and the call-site comment now all say "cannot carry more than one value", matching the refusal message and the two worker call sites. Worth fixing rather than a nit, as you say: "set-valued" and "empty list" point at a list type, and `ROLES :: string[] is []` is the declaration the grammar rejects -- so the prose was steering authors at the one form the message does not prescribe.

housejester:
This is the right shape. Inverting the invariant — Publisher's job is to *carry* the marker, not to enforce a type rule belonging to someone else's attribute store — puts the check where this layer actually has standing, and the refusal is cleanly gone: no trace in `authorize.ts`, `model.ts` or `package_load_worker.ts`, and the release note went with it.

**The new tests are load-bearing, which I checked rather than assumed.** The realistic regression here is the marker being filtered out of the served surface, so I mutated `isReservedRoute` to treat `secure` as reserved and re-ran:

```
6 pass                    -> 4 pass, 2 fail
```

Exactly the two marker tests go red; the four `it.each` guards stay green, correctly, since they only assert loading. And those four would have failed against the previous revision, where "scalar, marked" was refused — so they pin the reversal as well as the behaviour. Full unit suite on the branch: 3962 pass / 3 skip / 0 fail.

Thanks for re-running the gate table independently rather than taking mine as read — and the empty-form result goes further than anything I had. That neither multi-valued type has a writable empty form *at this layer* is the sharpest version of the argument, because it shows the premise was never demonstrable here in the first place.

## One thing I would still change

`secureNotesOf` matches `note.includes("secure")`. Plain `#` tags are safe — reserved route, filtered before they reach `annotations` (checked: `# secure data` yields `[]`). But a routed annotation carrying the substring is not:

```
#(insecure)  ->  notes = ["#(insecure)\n"]  ->  both marker tests pass
```

An annotation meaning the opposite satisfies the assertion. Narrow in practice, but "the marker survives package load" is now the whole deliverable, so that predicate *is* the deliverable. `noteRoute(note) === "secure"` fixes it, and this PR already wrote that two revisions ago — it went out with the check. It would also make the assertion match its own doc comment, which says "as SERVED".

## Docs

Scoping it as "a property of the deployment, not of the `given:` grammar" is the right register, and naming the registry as an example rather than a rule is honest about whose constraint it is.

Two small things:

- The paragraph states that limitation as present fact in a reference doc. If it is a gap rather than a design property — the open question for @Sha-Bang — closing it silently falsifies the paragraph. Tensing it ("today", "currently"), or pointing at the deployment's own documentation for the current requirement, costs nothing and saves a follow-up.
- `docs/givens.md:74` still reads "**Array and record givens are not among them**: `given: categories :: string[] is []` is a compile error". `ROLES :: string[]` does compile and renders `array`; only the `is []` default is `unexpected ']'`. You flagged this yourself, and it is in the file this PR edits.

## Worth not losing

Neither belongs in this PR, and both were owed before the pivot:

1. **The `filter<string>` type rule.** It now lives nowhere in Publisher, which follows from the layering call — but a rule keyed on `[]` alone still mis-handles the one type every secure given actually uses. That correction moved rather than landed.
2. **The runtime failures behind B, C and D.** A gate over a multi-valued given compiles clean and then emits `ERROR` into the SQL or throws a conversion error. That is what makes "declare it multi-valued" unactionable today, and it deserves its own issue whatever the design answer turns out to be.

## On merging

Worth stating plainly, since the design question is still open: **this no longer depends on it.** Nothing here enforces the constraint, so whichever way it is answered, carrying the marker through package load is correct. The only part sensitive to the answer is the one docs paragraph, and tensing it removes even that. I do not think this needs to wait.

---

<sub>Reviewed by Claude Code (session "Mellow Otter") on behalf of @housejester. Validated against `9f3e98f0`; mutations reverted and scratch harnesses removed after running.</sub>

packages/server/src/service/row_level_authorize.integration.spec.ts
```diff
@@ -4316,3 +4316,120 @@ source: Gated is duckdb.table('parent') extend {
...
+      const sources = pkg.getModel("m.malloy")?.getSources() ?? [];
+      for (const source of sources) {
+         const givens = (source.givens ?? []) as {
+            name?: string;
+            annotations?: string[];
+         }[];
+         const given = givens.find((g) => g.name === givenName);
+         if (given) {
+            return (given.annotations ?? []).filter((note) =>
+               note.includes("secure"),
```
sagarswamirao: This filters on `secure` as a bare substring, so the assertion is really "some note mentioning the word survived" rather than "the `#(secure)` marker survived". Checked what it accepts:

```
#(secure)               ACCEPT
#(insecure)             ACCEPT
# not secure at all     ACCEPT
# secure data           ACCEPT
#(doc) mentions secure  ACCEPT
```

No false green today, since nothing in the fixture emits those. But it is the same shape the module header calls out -- "What counts as the tag is Malloy's answer, not a regex of ours" -- and the production side already moved to `noteRoute` for exactly that reason, so the guard is now looser than the thing it guards.

Could you switch it to `noteRoute(note) === "secure"`? That makes the helper assert what its doc comment says it does, and it would keep catching a dropped marker if a future change started emitting an unrelated note with the word in it.

PR #1200, opened 2026-09-18
Add malloy-gx skill: bridge Malloy queries and Great Expectations

TakeAim82:
## Summary

Adds `skills/malloy-gx`, a skill that bridges Malloy query results with [Great Expectations](https://github.com/great-expectations/great_expectations) (GX), a Python data-quality/validation framework. It covers three workflows:

1. **Generate a starter GX expectation suite from a Malloy query.** Runs a query (via the REST API or, in an agent session, the `execute_query`/`get_context` MCP tools), profiles the actual result (null rates, observed ranges, low-cardinality value sets, primary-key-looking uniqueness), and saves a suite. Deliberately conservative heuristics (padded ranges, `mostly=` for observed null rates) so a fresh suite catches gross breakage without being a brittle exact-match snapshot of one sample — see `skills/malloy-gx/reference/malloy_type_mapping.md` for the reasoning and an override table.
2. **Validate a Malloy query against a saved suite** as a pass/fail data-quality gate (`scripts/validate_query.py`), with exit codes distinguishing a data failure (1) from a setup problem (2) so it drops into CI.
3. **General GX authoring help** — a cheat sheet mapping plain-English asks ("this shouldn't have nulls", "should be one of these values") to the right `gxe.*` expectation class, for when the ask isn't specifically the Malloy round-trip.

### Why row-level profiling instead of parsing `.malloy` source

A Malloy `dimension:`/`measure:` declaration doesn't carry an explicit type in the source text most of the time — it comes from the expression and the underlying table's schema. Profiling the actual query result is simpler and more honest about what's being checked than re-implementing Malloy's type inference. `reference/malloy_type_mapping.md` covers this in more depth.

### What I deliberately left for maintainer judgment

- **Not added to any `groups` entry** (`analysis`/`modeling`) in `manifests/publisher-local.json` — only registered in `auto_discovered`, since data-quality validation doesn't obviously belong to either existing role and I didn't want to presume.
- **Not added to the `skills/malloy` catalogue index** — that table is a curated taxonomy (Discovery / Modeling / Analysis / …) and this skill introduces a new category rather than extending an existing one; adding a row felt like a product decision for the committers rather than something to bundle into this PR unreviewed.

### Testing

Verified end-to-end against a live Publisher server (`npx @malloy-publisher/server@latest`) serving the bundled `storefront` example, and a real `great_expectations` install:
- `malloy_client.py` pulled real rows from `storefront -> by_category`.
- `generate_expectations.py` profiled the result and saved a 6-expectation suite (caught and fixed a real ordering bug: GX v1's `ExpectationSuite.add_expectation` needs an active `gx.get_context()` before it's called).
- `validate_query.py` passed against matching data (exit 0) and correctly failed with a clear report against a deliberately bad sample (exit 1).

Also ran the repo's own checks before opening this PR:
- `bun test packages/skills/src/manifest.spec.ts` — 167/167 pass
- `bun test packages/server/src/mcp/skills/skills_bundle.spec.ts` — 12/12 pass (bundle regenerated via `build_skills_bundle.ts`)
- `node scripts/add-license-headers.mjs --check` — clean

## Test plan

- [ ] Maintainer call on whether/where this fits in `manifests/publisher-local.json`'s `groups` and the `skills/malloy` catalogue table
- [x] `bun test packages/skills/src/manifest.spec.ts`
- [x] `bun test packages/server/src/mcp/skills/skills_bundle.spec.ts`
- [x] `node scripts/add-license-headers.mjs --check`
- [x] Live end-to-end run against `storefront` example + real `great_expectations` install


jswir:
Thanks for this — I ran it end to end against `examples/storefront` and the round trip works. `by_category` profiled 10 rows into a 6-expectation suite, and validating re-ran clean at exit 0. The REST client's error handling is good too: a missing parquet file and some invalid Malloy I wrote both surfaced the server's own message with no swallowed traceback.

Most of what follows comes from your branch predating #1195, which added checks this PR now has to satisfy. Your "167/167 pass" run was honest; the suite has grown since.

## Blocking

**1. The frontmatter doesn't parse as YAML, so no host can load the skill.** `description:` is an unquoted scalar with a bare colon-space at `Covers three things: (1)…`, which ends it early — `Bun.YAML.parse` throws, and `manifest.spec.ts` runs that parse on every shipped skill. Don't fix it by quoting: a downstream build truncates the raw line at 200 chars and a quoted string loses its closing quote. Use `--`, like the other skills. This replacement parses clean and also narrows the trigger (see below):

```yaml
description: Generate or run data-quality checks on a Malloy query result using Great Expectations (GX), a Python validation library. Use when the user is working with a Malloy model or query and asks about data quality, validation, or expectations -- generating a starter expectation suite by profiling a query result, or validating a query against a saved suite as a pass/fail gate.
```

**2. Bump `packages/skills/package.json`.** `skills-npm.yml` fails a PR that changes skills content without one.

**3. Add a `malloy-gx` row (in backticks) to `skills/malloy/SKILL.md`.** You flagged this as a maintainer call, which was fair — #1195 made it a test.

## Bugs, all reproduced

**The value-set check never fires on a current install.** I ran `pip install great_expectations pandas` as SKILL.md says and got pandas 3.0.6, where string columns are dtype `str`, not `object`:

```
status    dtype=str    ==object? False
```

`generate_expectations.py:117` gates on `series.dtype == object`, so `ExpectColumnValuesToBeInSet` can't be produced for any string column at any row count. Verified with a 25-row constant column that satisfies both thresholds — nothing generated. You likely tested on pandas 2.x. Worth pinning the dependency either way.

**A zero-row result produces a suite that passes anything.**

```
$ generate_expectations.py --query "... where: category = 'NO_SUCH_CATEGORY'" --suite-name empty_suite
Profiled 0 rows x 0 columns.
Saved suite 'empty_suite' with 0 expectations

$ validate_query.py --query "run: order_items -> by_category" --suite-name empty_suite
All 0 expectations passed against 10 rows.      # exit 0
```

`pd.DataFrame([])` has no columns either, so the row-count check and the column loop both skip. Good data passing a suite built from a broken query, silently — worse than no gate, because it looks like one. Probably wants to be a hard error.

**Any nested result crashes.** `TypeError: unhashable type: 'list'`, raw traceback, exit 1 — at `generate_expectations.py:68`, where `series.nunique()` runs *before* the name check on line 71, so every column gets hashed and nesting dies regardless of column name. Nested views aren't an edge case in Malloy.

**Booleans get a numeric range.** `is_numeric_dtype` is `True` for a bool Series, so they take the numeric branch. Generated: `{'column': 'is_active', 'min_value': -0.1, 'max_value': 1.1}`. `reference/malloy_type_mapping.md` says `ExpectColumnValuesToBeInSet` — the doc is right.

**`PK_NAME_HINTS` over-matches.** `"id"` with `endswith` fires on `amount_paid`, `is_valid`, `grid`. A money column in my test picked up `expect_column_values_to_be_unique: amount_paid`.

**Setup failures can exit 1.** `validate_query.py:68,75-76` — `run_query` and `suites.get` map to exit 2, but `gx.get_context` and `batch.validate` aren't wrapped. A nested result or a suite referencing a renamed column then reads as a data regression rather than a broken harness. The 0/1/2 split is the best thing here; worth making it hold at all three points.

## Our calls, no action needed

Keeping the name. It'll go in a new `quality` group (groups are exclusion filters, so a group of one lets a consumer drop just this). Listed Publisher-specific in `skills/README.md` — lives only here, not mirrored upstream. The description above also drops "use it even when Malloy isn't involved"; we'd rather a Publisher-bundled skill not fire on an unrelated CSV.

## Two notes

`malloy_client.py` is the third hand-rolled Publisher REST client here — our debt, not yours. If you want the cheap win, `skills/eval-answer/scripts/publisher_rest.py` is importable across skills today and its `rows_from_result` handles the nesting problem above.

This is strongest against a `#@ persist` source — materialized, stable, big enough for the statistics to mean something, and a rebuild is a natural trigger (`malloy-pub materialize --wait` then `validate_query.py`). Worth a short section saying so, and saying a ten-row aggregate isn't what this is for.

No tests — every other skill shipping Python here pairs each script with a `*_test.py` (22 for 22). `test:skills-python` skips a skill with none, so this passes silently. Not blocking, but every bug above is one a test would have caught.

PR #1201, opened 2026-09-18
fix(server): confine the per-package DuckDB sandbox to the package directory

sagarswamirao:
> **Draft: this does not pass its own suite.** It breaks package load and reload
> for the reason in "The blocker" below, and one of its failures is an existing
> test that asserts today's behaviour as intended. It is open for the finding and
> the two corrections it establishes, not as a merge candidate.

## The reported finding

From a multi-tenant security review, rated CRITICAL:

> DuckDB is created with `enable_external_access` left on and no
> `allowedDirectories`. `duckdbConnection` is a user-creatable type, so a tenant
> creates one and runs `SELECT * FROM read_text('/etc/passwd')` -- reading files,
> or writing them with `COPY ... TO`, or `httpfs` SSRF, on a worker that also
> hosts other orgs.
>
> **Fix:** Set `enableExternalAccess:false` (or `allowedDirectories:[packagePath]`)
> on the DuckDB sandbox and user DuckDB connections -- the single
> highest-leverage change.

## Confirmed, against the engine rather than the config

`securityPolicy` appears nowhere in `packages/server/src`, and the connector
defaults to `none` when it is unset, so every Publisher DuckDB session ran
unrestricted. A default-constructed session read `/etc/hosts` and returned its
882 bytes.

## First correction: the recommended fix is a silent no-op

Setting `allowedDirectories` alone restricts nothing. Per DuckDB's documentation
it is a carve-out FROM a denial -- directories "ALWAYS allowed to be queried,
even when `enable_external_access` is false" -- and `enable_external_access`
defaults to **true**. Probed directly: with `allowedDirectories` set and nothing
else, `read_text('/etc/hosts')` and `COPY ... TO` both still succeeded.

A test asserting the config shape would have passed on exactly that no-op, which
is why the spec here drives a real session instead.

## Second correction: the mechanism already exists upstream

Rather than hand-rolling the `SET`s, this uses the connector's own
`securityPolicy` (`none` / `local` / `sandboxed`), whose introduction upstream
records fixing this same class of bug -- a sandbox that set `allowed_directories`
without disabling external access enforced nothing.

`local` was measured and rejected: it kills external access wholesale, including
the package's own CSV and Parquet files, which would break every package that
ships data. `sandboxed` keeps package-local reads working and denies escape, so
that is what the per-package sandbox uses.

## Scope

**The per-package sandbox only.** Environment-level DuckDB connections are left
alone deliberately: they require `attachedDatabases` pointing at BigQuery,
Snowflake, Postgres, GCS, S3 or Azure, and a restricted policy kills the httpfs
those federations need. Two existing tests assert that external access stays
available for federated and MotherDuck entries. They are also admin-authored
rather than tenant-reachable, a different trust tier from the sandbox a tenant's
own package gets.

## The blocker

`securityPolicy: "sandboxed"` forces `lockConfiguration: true`, and the connector
pools DuckDB instances by a share key that includes the directory. Building a
second sandbox for one package path therefore reaches an instance the first
already locked, and re-applying the setting is refused:

```
Invalid Input Error: Cannot change configuration option "allowed_directories"
- the configuration has been locked
```

Package reload takes that path, so `package.spec.ts` goes from 23 pass / 0 fail
to 17 pass / 6 fail. Different package paths are unaffected, having different
share keys.

Directions, none verified: an upstream change so a pooled instance is reattached
rather than reconfigured; a per-package `workingDirectory` forcing distinct share
keys; or a Publisher-side cache so one package path builds its sandbox once.

## One failure is a contract question, not collateral

`does not treat package-root paths as filesystem isolation` asserts that a model
reading `../outside-package.csv` succeeds -- today's behaviour, written down as
intended. Reversing it is a deliberate decision rather than a side effect of this
change, and wants ratifying separately from the mechanism.

## Tests

`package_sandbox_filesystem.spec.ts`, four cases against a real DuckDB session
through the real `buildPackageMalloyConfig`: a package-shipped CSV reads; a read
outside the root is refused; a write outside the root is refused and the file is
asserted never to have been created; a network URL is refused.

Unmodified code: 1 pass / 3 fail, the package-local read being the one that must
pass. With the change: 4 pass / 0 fail.

The SSRF case originally pointed at `169.254.169.254` and actually issued a
request, timing out at 5s. It now uses an RFC 2606 name that never resolves, so
the assertion distinguishes a config denial from a request failure rather than
conflating them.

## Platform caveat

`sandboxed` is POSIX-only upstream and throws on Windows, so the policy is
applied only where `path.sep === "/"`. Windows therefore stays unhardened --
deliberate, to avoid failing package load on a Windows checkout, but it means the
fix is partial by platform and that is worth deciding on rather than inheriting.


sagarswamirao:
Closing this in favour of #1203, which documents the finding instead of enforcing
it here. Two reasons, and the second is the one that decided it.

**The mechanism does not work.** Measured against the installed connector rather
than inferred:

| config | absolute path | relative path |
| --- | --- | --- |
| `local` | refused | refused |
| `sandboxed` | ok | **refused** |
| `sandboxed` + `allowedPaths` | -- | refused |
| `sandboxed`, process cwd inside the allowed dir | -- | ok |

This PR's description claimed `local` was rejected because it kills package-local
reads while `sandboxed` keeps them working. That is wrong: `duckdb_config.js:186`
sets `enableExternalAccess: restricted ? false : ...` and `restricted` is true for
both, so `sandboxed` is `local` plus an `allowed_directories` carve-out. The
carve-out does not rescue a relative path, because DuckDB resolves one against the
process working directory for its permission check rather than against
`file_search_path`. So `duckdb.table('data/sales.csv')` is refused, which is how
every real package reads its own files -- the shipped `governed-analytics` and
`storefront` examples included, which is the ~55 Playwright failures.

The cwd row is not a way out: compilation runs in `worker_threads` sharing one
process working directory, so a per-package `chdir` would race.

Worth recording that this PR's own new spec passed 4/4 while the product was
broken, because it used absolute paths.

**And it is the wrong layer anyway.** The capability is only a vulnerability when
one worker serves more than one tenant. A single-tenant operator reading their own
files is using their own machine, and the same capability is what lets a package
read the CSV beside it. Refusing it here breaks legitimate use to enforce a policy
that belongs to whoever builds the shared deployment -- which is the conclusion
#1203 writes down, along with the mechanism that does close it (reject
caller-supplied Malloy before compile; restricted mode is what the query path
already uses) and the network-layer backstop.

The instance-pooling collision this PR hit is real and separate. `sandboxed` forces
`lockConfiguration`, and the connector pools instances on a share key that a stable
per-package path cannot vary, so a reload reconfigures a locked instance.
`duckdb_connection.js:145` already guards that case but gates it on
`effectiveShareable`, which a `:memory:` sandbox is not. That is worth an upstream
issue on its own terms; it is just not load-bearing here any more.

PR #1202, opened 2026-09-19
skills: 38 skills to 17, and navigation that reaches the agent

jswir:
**Draft** until the routing eval runs against this branch versus `main`. That eval is the acceptance gate, not the test suite.

## Why

Every skill's description is loaded at startup; its body loads only if the model picks it. With 38 descriptions in one listing, ten clusters claimed the same request with nothing to adjudicate between them:

- **five** skills quoted "build a dashboard" (`malloy-analysis-report`, `malloy-analyze`, `malloy-notebooks`, `malloy-dashboards`, `malloy-html-data-apps`)
- `malloy-modeling` triggered on "asks about modeling data or specifically mentions Malloy" — a superset of the whole corpus
- `malloy-analysis` competed with the near-homograph `malloy-analyze`

The listing budget was never the problem: 38 descriptions is roughly 9% of it. Ambiguity was.

## What changed

**38 → 17.** Eleven Malloy entries plus six `eval-*` left alone — those are spawned as subagents by `eval-loop` rather than read as reference, so folding them would break that mechanism.

**`malloy-visualization` is new** and owns the question the old corpus could not answer: notebook, dashboard, HTML app, or a chart on a view you already have. Choosing between them *was* the collision; now something owns the choice and routes to the procedure. `malloy-analyze` is retired.

**Twenty skills became reference files** under the entry that needs them.

| entry | absorbs |
|---|---|
| `malloy-analysis` | queries, gotchas-queries, patterns, analysis-pitfalls, phrase-detection |
| `malloy-modeling` | model, define, discover, scope, gotchas-modeling, debug |
| `malloy-visualization` | notebooks, analysis-report, dashboards, html-data-apps (+runtime, +embedding), charts, gotchas-rendering |
| `malloy-materialization` | materialization-tuning |

## Navigation, and why the root index is gone

The `malloy` index skill is deleted. An index that is itself a skill only reaches an agent that chooses to load it — and the measurement says that channel barely fires. On the first full day prod skill telemetry was recording:

| signal | count |
|---|---|
| turns | 226 |
| inlined a skill | **226** |
| loaded a skill | 28 |
| read a `reference/` file | **0** |

So navigation has to live in the text that is always delivered. Every entry now carries a hand-written index of its own reference files with a "when to read this" column.

Paths in those indexes are **relative**, not `.claude/skills/…`. The repo's `hardcodes no host install path` rule caught the first attempt and the rule is right: the host, not the body, should say where a skill lives. ms2data/service#6455 makes the inlining renderer state it, which is what makes a relative pointer resolvable at all. **That PR is a prerequisite** — without it these indexes point at nothing on the app surface.

## Guards

Two new, one removed.

- **`indexes every reference file it ships`** — the direction the existing dangling-pointer check never covered. A file that exists and nothing points at is unreachable and fails silently; that is how two eval skills shipped unreachable from the index for weeks. It is also what makes a hand-written index safe to hand-write.
- **`gives every long reference file a table of contents`** — past roughly a screenful and three headings. A reference file is opened mid-task by an agent that wants one section.
- **removed:** `the index accounts for every skill that ships`, with its premise. Its property is now carried by the per-skill check plus the manifest.

Both new guards mutation-tested: an unindexed file and a stripped TOC each turn exactly their own case red.

## Testing

```
packages/skills            164 pass  0 fail
packages/server mcp/skills  18 pass  0 fail
```

Bundle regenerated: 77 entries (17 skills + 60 reference files).

## Release constraint

Version **0.2.0**, a major for the corpus: **22 published skill names disappear.** Any manifest naming one raises `MissingSkillError` once the pin moves, so this is a coordinated release — publish, then ms2data/agent-skills manifests, then tag, then the service pin — not a merge.


jswir:
Closing: this needs to be eval-driven, not design-driven.

The branch stays on `jswir/skills-consolidation` and nothing is lost. What was wrong was the order — I built a 38→17 fold and then proposed measuring it, which makes the eval a defence of a design rather than the thing that chose it. The collisions are real and the measurement behind the navigation change is real, but "which fold is better" is exactly the question an eval should answer first.

Reopening once there is a harness and a baseline to move.

PR #1205, opened 2026-09-21
feat(skills): score an agent session that ran somewhere else

jswir:
Stacked on #1173. Base is `jswir/eval-harness-ground-truth`; retarget to `main` once that merges.

## Why

The eval loop can only score sessions it ran itself. `run_answerer` spawned the answerer and parsed its stream-json into ledger events in one function, so "get the transcript" and "run the agent" were the same act. A session that ran in a host's own app never entered the loop, which meant the traffic most worth measuring was the traffic we could not see.

This adds the other entry point. Scoring is unchanged: `eval-answer`, `eval-diagnose` and `eval-improve` cannot tell where a session ran.

## How

The unification point is the ledger, reached through a file the harness already reads. Four things were already host-agnostic and are reused rather than duplicated:

- `mcp_payload.entity_hits` already parses the hosted source-centric `GetContextResponse`.
- `resource_json()` already falls back to a bare JSON body "because a hosted MCP may return the JSON body directly".
- `attempt.contaminated` was already `bool or "unknown"`, `"unknown"` meaning no host log exists.
- `--rebuild` already re-derives a whole run from `artifacts/<qid>/answerer.jsonl` with zero model calls.

So: `run_answerer` splits into `capture_answerer` and `derive_attempt`; `log_transcript.build_transcript` turns a host's request-log rows into that same file; `fetch_transcripts.py` pulls the rows per case. Scoring is then the existing `run_baseline.py --rebuild`. No new parser and no new ledger writer.

## What a source can supply decides what a run may claim

| Tier | Source carries | Can measure | Cannot |
|---|---|---|---|
| T1 | the tool calls | retrieval recall, bare-target rate, scoping, construction | anything about the answer |
| T2 | plus the agent's prose | the above, and a judged verdict | contamination |
| T3 | plus a host-side tool log | the above, and contamination decided | nothing |

A spawned run is always T3. **A T1 run has no pass rate**: every attempt scores `verdict: null` with the new reason `no_answer_captured`, decided before the judge is called. Pinned in `run.json` as `source` and `sourceTier`, derived from the attempts, and the tier is the *weakest* any attempt had.

`score_retrieval.score_case` already takes `verdict: str | None`, so T1 needed no new scoring code.

## Verified against a real arm, not just fixtures

Two Publishers (model + isolated truth) over malloy-samples `ecommerce`, semantic retrieval, all 49 goldens verified (45 re-derived, 4 unanswerable skipped).

| Run | Source | Tier | Result |
|---|---|---|---|
| baseline | spawned | T3 | 5 of 6 decided (83%) |
| replay_t1 | logs | T1 | no pass rate, recall 100% |
| replay_t2 | logs | T2 | 5 of 6 decided |

**`flip_table.py` reports 0 flips between the spawned arm and its log-replayed twin.** Same passes, same stable fail, same stable near_match.

## Three defects found by running it, not by reading it

**1. A clean logged run read as a compromised one.** The isolation checks, run over a transcript with no `system/init` event, recorded *"no init event: cannot verify what the answerer held"* as a BREACH. It landed in the contamination histogram, voided the verdict, and printed "isolation breached" about a session where nothing was breached and nothing was checked. Now skipped when the transcript declares no host log, with `contaminated: "unknown"` and no reasons, which is what the ledger contract already said. A test pins that the spawned case still reports it.

**2. The judge was handed the broken query.** The final query is the last call the server *answered*, so a transcript where every call looks successful hands over whatever ran last, which for an agent that made a syntax error and then fixed it is the broken one. Measured on 8 real attempts with no per-call outcomes: 4 of 8 changed their final query, 3 lost their error counts, 1 of 8 tool-call streams matched. Rows now carry `error`; hosts log compile errors and nothing for successes, which is exactly the asymmetry needed. After: 8/8 streams, 0/8 error counts, 7/8 final queries.

**3. A T2 fetch with no prose still claimed `answer_captured: true`**, so the judge would compare an empty answer against the golden and score a real agent as having said nothing. Now downgraded to T1 per session.

The one remaining `final_query` difference is the tier, not a bug: its baseline source was `declared` (the agent printed the query in its answer), and at T1 there is no prose. Documented as the concrete form of "the log cannot see what the agent did after the rows came back".

## Also: `PUBLISHER_MCP_TRACE=retrieval` is a no-op

`eval-loop/SKILL.md` told conductors to enable retrieval tracing and to **refuse a scored run without it**. Nothing reads that variable. The store was written and never merged (`c6cd87e5`; #1101 squash-landed only the `skills/` tree), so at HEAD there is no `mcp_traces` table, no trace tool, and `traceId` is null on every local attempt, while `serve.py` dutifully set it.

Nothing was broken by its absence, because `rankedSummary` is copied onto each `tool_call` at capture precisely so evidence survives without a trace store. So the instruction is corrected rather than the store re-landed.

## Doctrine

`reference/log-scrape.md` lands from the branch it was written on and never merged (cases from traffic). `reference/replaying-a-logged-session.md` is new (scoring a session that already ran). Host specifics stay out of the tree: `--logs-scope` and `--source-*` are arguments, and no vendor table name appears in either script.

## Testing

`bun run test:skills-python` 24/24 green. New: 25 tests in `log_transcript_test.py` (including the round-trip that asserts the *full set* of fields differing between a spawned and a rebuilt transcript, so a future silent loss fails the test), 16 in `fetch_transcripts_test.py`, 14 added to `run_baseline_test.py` (174 total, the pre-existing 160 unedited).

Not run: full `bun run test`, which binds port 4000.

## Out of scope

Publishing the `credible-logs` package. Publisher-side request logging in the same shape, which would make the T1 fetch one query on either side, is a separate change with three policy calls to settle first: persisting `execute_query` responses contradicts `ledger-schema.md:207`, `c6cd87e5` is an explicit prior decision against server-side eval persistence, and the old store evicted by count with no age bound.


jswir:
Superseded by #1207: rebased onto `main` after #1173 merged. GitHub auto-closed this one when its base branch `jswir/eval-harness-ground-truth` was deleted.

PR #1206, opened 2026-09-22
feat(server): a package's index.malloy is its published surface

jswir:
Revives the idea behind #988, which was closed as stale rather than rejected. **This is not that branch rebased.** It is a fresh implementation against current `main`, and it makes the one design decision #988 left open in the opposite direction.

**Base note:** this stacks on #1163, so the diff here is only its own 10 commits. It cannot merge before #1163 does. #1163 is what makes the three-answer table below true.

## What this does

A package's published surface is its `index.malloy`. Put that file at the package root, `import` your models, `export { … }` what you publish. What it exports is what Publisher lists **and** what callers may query. No `publisher.json` key at all.

```
sales/
  publisher.json     { "name": "sales" }
  index.malloy       import "orders.malloy"
                     export { orders }
  orders.malloy      declares `orders` and `orders_staging`
```

`orders_staging` still compiles, and other models can import, join and extend it, but it is not listed and a direct query against it answers 404. `create-malloy-package` scaffolds the file, so a new package is curated from its first boot.

## Where this departs from #988, and why

#988 made the convention curate **listings only** — it hid sources but never refused them, to protect a package that already had a root `index.malloy` from silently losing query access. That is the decision its closing note called "genuinely open".

**That population is empty.** Checked across everything available locally: 14 `malloy-samples` packages (none declare either key, none have an `index.malloy`), the three bundled examples (only `governed-analytics` declares `explores`, explicitly, so the convention never fires), and the one real `index.malloy` in a Malloyyo example repo — which is the *only* model file in that package, so the convention lists it and exposes everything it declares.

So the surface is the boundary, always, which is what the design note under Open questions in `docs/malloyyo-dashboards-design.md` proposed in the first place.

That reversal is mostly deletion. #988 added ~1,900 lines; `exploresFromConvention`, `boundaryDeclared()`, `postLoadWarnings`, `resolvePatchedExploresOrigin`, three warning builders and a 424-line integration spec all existed only because the convention produced a *different kind* of surface. Here it does not, and none of them appear.

**The whole feature, stated once: the convention is a default for one manifest field. Nothing downstream records where a surface came from.** That single sentence is why the diff to `package.ts` is 42 lines rather than 228 — `exploresDeclared`, `exploreSet`, both policy pushes, `getInvalidExplores`, `listModels`, `setPackageMetadata` and `Environment.updatePackage` all read `packageMetadata.explores` and cannot tell a filled-in value from a written one.

## Deprecation, not removal

`explores` and `queryableSources` behave exactly as before. Both are marked `deprecated` in the OpenAPI spec and earn a load-time warning naming the replacement. Nothing is removed.

**`index.malloy` does not replace `queryableSources: "all"`, and its warning says so** rather than advising a switch. `"all"` is the only way to curate listings *without* refusing queries, and a derived surface always enforces the boundary. That author is told to keep both keys.

**Deleting the keys is deliberately out of scope, and not for diff size.** `publisher.json` ignores unknown keys, so deleting the parsing would silently turn a curated, bounded package into an uncurated, unbounded one — `governed-analytics`'s `orders_base` would go from refused to freely queryable with no error and no config change. That is a fail-open access change and deserves its own release, where the right behavior is to *refuse to load* a package still declaring the keys rather than ignore them.

## `examples/governed-analytics` is converted

It was the only package in the repo declaring a curated surface, and it declared a two-**file** one — the shape the convention was least obviously able to express. It can, so `publisher.json` is back to a name and a description. That is the evidence `explores` can eventually be retired rather than merely deprecated.

Two things moved that no unit fixture covered, so a new spec drives the real package and pins both:

- **Every query now enters through an `import`.** Malloy marks imported entries `exported: false` and inlines their struct, which is the shape that once denied *authorized* callers by grafting a gate onto an entry nothing consults (`authorize_import_hop.integration.spec.ts`, whose header notes no other fixture put a gated source behind an import). This example now does, permanently, so the spec asserts the gate still tracks the caller across three different `TENANTS` values.
- **The givens that drive the notebook's Parameters panel are declared in the imported files**, not the entry model. Measured rather than assumed: all three reach `index.malloy`'s surface. The spec pins the full list, because a narrowing there is not a patchable bug — it would mean reverting the conversion.

## Three warnings had to change, and that is one bug not three

A package curated by its `index.malloy` has no `explores` key, so any message whose fix is "add it to `explores`" points at something that does not exist. Three sites said that:

- the **dashboard withholding** warning (adding an `index.malloy` to a package with `dashboards/` withholds every dashboard, because a dashboard file is not something an `index.malloy` can export);
- the **empty-surface** warning (an import-only `index.malloy` exports nothing, so the package looks empty);
- and `"explores": []`, which is the *documented opt-out* and was being told to delete itself.

All three now name a remedy that exists, from one predicate derived from the tree rather than a stored origin. It only ever picks the wording of a remedy, never behavior.

## Verified against a running server, not only in tests

Seven variants of one package served together, plus a gated package, plus a four-hop re-export chain:

| | listed | queryable |
|---|---|---|
| no keys, no index | all 3 files | all 3 |
| `index.malloy` exporting 2 | `index.malloy` | 2, via the surface |
| `explores` omitting the index | `orders.malloy` | 1 |
| `explores: []` + index | all 4 files | all 3 |
| index + `queryableSources: "all"` | `index.malloy` | **all 3** |
| only `reports/index.malloy` | all 4 files | all 3 |

- **Three answers, all three rows:** off the menu **404**, locked **403**, row-gated **200** with the caller's rows (empty is a 200, not an error).
- **The oracle holds.** A source both hidden *and* locked answers **404**, byte-identical to a source that does not exist, whether or not the caller would have passed the lock. Only a *listed* locked source answers 403. Same at `/compile`.
- **`get_context` tracks the surface**, in enumeration and in ranked search, and every `resource_id` it returns round-trips through `execute_query` verbatim — on a convention package that `model_path` is the surface file, which is what keeps agents off a 404.
- **Chains and joins:** a source re-exported through four files stays queryable through the surface; a published source may join an unpublished one and the joined fields read normally. Curation hides a landing point, not a column.
- **Versions:** the same package published four times against a running server (uncurated → convention → widened export → opt-out), reloading each time. Listing, REST and `get_context` tracked every publish, no restart.

Gate: typecheck clean, 4044 server unit pass, 354 integration pass, skills/manifest/bundle/python green. The 35 `config.spec` failures reproduce at the base commit and are unrelated.

## One thing reviewers should push back on if they disagree

`export { … }` governs what an **importing** file may see, which is Malloy's rule rather than Publisher's. So "hidden, not out of reach" means reachable through a file that exports it *or that exports nothing* — not through a file whose `export` omits it, which fails to compile with `Reference to undefined object`. That condition is now in the docs and the new skill. It is the one thing about this convention I would expect to surprise someone mid-migration.


kylenesbit (approved):
Approving. The design holds and the diff is genuinely small for what it does — `resolveExplores` and the boundary itself I have no notes on, and the `export { … }` importer-visibility decision you flagged for pushback reads correct and is documented. Everything below is in the edges the convention now touches. None of it is a reason to hold the merge, but the first two are worth landing before this ships to a user who did not opt into anything.

Verified against `jswir/index-malloy-surface`.

### The scaffolder is the real exposure

It now writes an `index.malloy` into every new package, which breaks two shipped workflows for users who never chose the convention.

**1. `packages/create-malloy-package/templates/model.custom.xlsx.malloy:21` — the xlsx recovery probe is now unrunnable.**
The scaffolded `index.malloy` exports only `{{sourceName}}`, so `sales.malloy` is off the surface. This template tells the user to add `{{sourceName}}_probe` to the model file ("It has to go in the file rather than being sent as a one-off query") and run `{{sourceName}}_probe -> peek`. After this PR that is refused either way: `models/sales.malloy/query` is not a query entry point, and against `models/index.malloy/query` `sales_probe` is not in the export closure. This is the documented fix for the most common spreadsheet failure — header not on row 1 — and the template never mentions `index.malloy`.

**2. `packages/create-malloy-package/src/scaffold.ts:683` — every scaffolded package silently withholds any dashboard added later.**
By this PR's own analysis a dashboard file is not something an `index.malloy` can export, so `dashboards/*.malloy` is never served — only a load warning, whose remedy is to abandon the convention and hand-write an `explores` listing the index plus every dashboard file. `PUT …/models/dashboards/{slug}.malloy` will happily create one that is never routed. Neither the generated `AGENTS.md` nor the `malloy-dashboards` guidance warns a scaffolded-package author before they build one.

### Four warnings give advice that is wrong for a case they match

These are one-line predicate or wording fixes, but three of them point a reader at an action that makes things worse, which is worse than saying nothing.

**3. `packages/server/src/service/package_manifest.ts:466` — `"explores": []` with no `index.malloy` gets inverted advice.**
The `EXPLORES_EMPTY_SUPPRESSES_CONVENTION` branch is gated on `declared.length === 0 && hasIndexModel`. Without an index file — a documented, supported "do not curate" state — the author instead gets `EXPLORES_DEPRECATION`: "put an `index.malloy` at the package root … then delete the key … it curates and enforces exactly as the key does." That last clause is false for an empty array. Following it converts an intentionally uncurated package into a curated-and-bounded one. Key the branch on `declared.length === 0` alone.

**4. `packages/server/src/service/package.ts:2072` — the empty-surface remedy can take a package fully out of service.**
`surfaceIsIndexModel()` is true whenever `models.has("index.malloy") && explores === ["index.malloy"]`, which the doc-comment notes includes a hand-written key. For that package "delete the file, which returns the package to listing every model" is false: the explicit key survives, `exploresDeclared()` stays true, `exploreSet()` = `{"index.malloy"}` matches nothing, `listModels()` returns zero and every model's `isQueryEntryPoint` is false. That is the same "lists and serves nothing" failure the `writePackageManifest` echo guard in this PR was added to prevent — here printed as the recommended fix. Gate it on the manifest key being absent, or word it to cover both.

**5. `packages/server/src/service/package_manifest.ts:498` — the "something is withheld" guard counts notebooks.**
`modelPaths` comes from `filterModelPaths`, which includes `.malloynb`. A package holding only `index.malloy` and, say, `report.malloynb` trips `INDEX_MODEL_IS_THE_SURFACE`, which tells the author sources are being refused — but notebooks are always listed and unaffected by the boundary, so nothing is hidden. Contradicts the comment directly above it ("an index.malloy that is the package's only model withholds nothing") and the matching RELEASE_NOTES claim. Filter the `.some()` to `MODEL_FILE_SUFFIX`.

**6. `packages/server/src/service/package.ts:2122` — two inaccuracies in the broken-surface message.**
It is emitted once per surface file and each copy says that file "is this package's whole discovery surface", so `explores: ["a.malloy","b.malloy"]` with both broken produces two messages each claiming to be the whole surface. And `hidden` (line 2115) is every non-surface `.malloy` including ones that failed to compile themselves, but renders as "the ${hidden.length} that compiled", so the count overstates how many working models went down.

### One doc miss

**7. `docs/givens.md:229`** — the `execute_query` sample still uses `"packageName": "governed-analytics"` with `"modelPath": "orders.malloy"`, which the conversion in this PR takes off that package's surface, so the copy-pasteable call 404s. `docs/authorize.md`, `docs/row-level-access.md`, `docs/discovery-and-access.md` and the example README were all updated to `index.malloy`; this one was not. The introspection link at line 164 points at `orders.malloy` too.

### Checked and correct

The `echoesDerivedSurface` guard in `environment.ts` (its reliance on `formatInvalidExplores` for file existence holds, and `"explores": []` round-trips because `existingManifest.explores` is then defined); resolving `explores` against the on-disk tree *before* the `job.replacement` fixup; `filterModelPaths`/`exploreSet` never matching `.malloynb` against `MODEL_FILE_SUFFIX`; the malformed-`explores` rejection no longer throwing out of `normalizeModelPath`; the RELEASE_NOTES inventory `jq` snippet (`index()` returning `0` is still truthy under `jq -e`); and `brokenSurfaceWarnings()` being cheap enough for the `/status` poll path it was added to.

packages/server/src/service/package_manifest.ts
```diff
@@ -379,3 +382,221 @@ export function queryMetadataParseWarnings(
...
+   const warnings: string[] = [];
+
+   // An array of strings is the only well-formed shape. A non-string element
+   // is malformed too, and is rejected rather than coerced: `String(null)`
+   // would build a surface naming "null", which matches no model, so curation
+   // would switch on over an empty set and the package would list nothing.
+   // Rejecting leaves the package uncurated (or on the convention), which is
+   // the conservative direction. It also stops a non-string element throwing
+   // out of `normalizeModelPath` and failing the whole package load.
+   const wellFormed =
```
Sha-Bang: **Blocking** - `["orders.malloy", 7]` fails `wellFormed`, so `declared` goes `undefined`, and with no root `index.malloy` the function returns no surface at all -- the package loads and serves every source the author curated away. before this change that same manifest threw out of `normalizeModelPath` and failed the load, which the spec comment at `package_manifest.spec.ts:105` records.

the comment here calls uncurated "the conservative direction", but `brokenSurfaceWarnings` in `package.ts` argues the opposite for the same question: *"Deliberately NOT a fallback to uncurated. Falling open on a typo would expose sources the author curated away, which is worse than refusing them."* both can't be right, and a typo in a hand-edited `publisher.json` is exactly when this fires.

could the malformed case keep the valid string entries and warn on the dropped ones, or refuse the load outright -- either way landing on the same side as `brokenSurfaceWarnings`?

packages/server/src/service/package_manifest.ts
```diff
@@ -379,3 +382,221 @@ export function queryMetadataParseWarnings(
...
+ * instead is a 404 on a model that reads as missing rather than withheld.
+ */
+const INDEX_MODEL_IS_THE_SURFACE =
+   `This package has a root "${INDEX_MODEL_NAME}" and no "explores" in ` +
+   `publisher.json, so that file is its published surface: only it is listed, ` +
+   `and sources it does not "export { ... }" are refused by name with a 404. ` +
+   `Other models still compile and are still importable and joinable. If that ` +
+   `is intended, nothing to do. If this package is not meant to be curated, ` +
+   `add "explores": [] to publisher.json to keep every model listed and ` +
+   `queryable, or rename the file.`;
```
Sha-Bang: **Blocking** - this warning ends with "or rename the file" as an opt-out, and `RELEASE_NOTES.md` says "Do **not** rename the file to opt out." the load-time message steers authors into the path the release notes warn against.

it's also the fail-open one, and that half isn't stated anywhere. removing or renaming `index.malloy` takes `hasIndexModel` false, so `resolveExplores` returns `{explores: undefined, warnings: []}` at :487 -- no surface, no warning, and every source that was hidden a moment ago is listed and queryable again. the release notes describe the rename hazard as a 404 plus a dangling import, which is the visible failure; the invisible one is the surface widening. nothing in the specs covers a package losing its index file.

drop "or rename the file" and leave `"explores": []` as the only opt-out? and is the removal case worth its own warning, given every other de-curation path in here is fail-closed and reported?

packages/server/src/service/package.ts
```diff
@@ -2023,20 +2063,82 @@ export class Package {
...
-                  `'export { source_name }' to surface sources on this ` +
-                  `model, or remove it from explores.`,
+                  `Model "${modelPath}" is on this package's discovery ` +
+                  `surface but exposes nothing: its export closure surfaces ` +
+                  `no sources or named queries (typically an import-only ` +
+                  `file, or an export {} that filters everything out). Add ` +
+                  `e.g. 'export { source_name }' to surface sources on this ` +
+                  `model, or ` +
+                  (this.surfaceIsIndexModel()
+                     ? `delete the file, which returns the package to ` +
```
Sha-Bang: **Should-fix** - `surfaceIsIndexModel()` at :351 is true whenever the resolved surface is exactly `["index.malloy"]`, whether the server derived it or the author typed it. so an author who wrote `"explores": ["index.malloy"]` by hand gets told to "delete the file, which returns the package to listing every model" -- and deleting it leaves an explicit `explores` naming a model that no longer exists, so the package lists *nothing*. exact inverse of what the message promises.

the docblock argues origin doesn't matter because "an author who wrote `explores: ["index.malloy"]` edits the same file", which is true of the file and false of the remedy: one of them has a key to remove it from. `explore_visibility.spec.ts:443` pins the convention wording but nothing pins this case.

`resolveExplores` already knows which branch it took -- worth returning that alongside the paths and using it here, rather than re-deriving from the tree? it'd only ever pick wording, so it wouldn't put an origin into behavior.

api-doc.yaml
```diff
@@ -3284,6 +3295,13 @@ components:
             equivalent (everything queryable). Invalid values fall back to
             `"declared"`. Identity-based access is a separate concern — see
             `#(authorize)` and `#(access_filter)`.
+            DEPRECATED, still honored. `"declared"` is already the default, so
```
Sha-Bang, line 3326: **Nit** - the sentence four lines above at 3294 -- "When `explores` is absent there is no curated surface, so both modes are equivalent (everything queryable)" -- is exactly what this block now contradicts. it's in the published spec, so it's the version a client generator reads. delete it or qualify it with "and no root `index.malloy`"?

Sha-Bang (commented):
two blocking, one should-fix, one nit. the theme in the two blocking ones is the same: a package's published surface widens without anyone being told.

contract change worth stating up front, since it isn't derivable from the diff: `PackageMetadata.explores` can now carry a value the server derived rather than one the author wrote, and the response deliberately doesn't distinguish them. no paths added or removed, so nothing moves for a client that just reads the list. but a client that reads `explores` to mean "the author curated this" is now wrong -- ours does, and it reads `publisher.json` off disk rather than through the API, so it sees no surface at all on a convention package. that's ours to fix, not a change request on this PR, but it's worth knowing the design has a blind spot for consumers that read the manifest file rather than your response.

- `packages/server/src/service/package_manifest.ts:442` -- a malformed `explores` now falls open to uncurated where it used to fail the load
- `packages/server/src/service/package_manifest.ts:535` -- the warning recommends the one opt-out the release notes say not to use, and the removal case is silent
- `packages/server/src/service/package.ts:2073` -- "delete the file" is wrong advice when `explores: ["index.malloy"]` was written by hand
- `api-doc.yaml:3294` -- stale sentence contradicting the new text four lines below

packages/server/src/service/package_manifest.ts
jswir: Fixed in fa3a952f — refusing, the second of your two options.

You were right that the two files argued opposite sides of the same question, and `brokenSurfaceWarnings` had the better half of it. A malformed `explores` now throws and fails the package load, like `resolvePackageScope` does for an invalid `scope`.

I went with refuse rather than keeping the valid entries, because salvaging serves a surface the author did not write — guessing at intent on the one key where guessing wrong is an access change. Four options and what each does:

- **ignore** (what it did): widens the surface. `["orders.malloy", 7]` resolved to no surface at all and published everything the key was written to withhold.
- **salvage**: serves a surface nobody wrote.
- **coerce**: `String(null)` builds a surface naming a model that cannot exist, so the package lists nothing while reporting a reason that describes a file rather than the typo.
- **refuse**: neither widens the surface nor invents intent, and an absent package is visible in `loadErrors` where a silently-uncurated one is not.

Verified against a running server rather than just the unit test. At first boot the package is not served (503, absent from the listing) and `loadErrors` carries the message, so the `secret_src` in my fixture never becomes reachable. On reload it answers 503 with the same message and the previously compiled model keeps serving. Both are byte-identical in shape to the existing invalid-`scope` path, including the `Package-load worker pool unavailable:` prefix — which is a pre-existing wart on that path, not something this adds.

A well-formed `[]` is untouched and still the documented opt-out.

packages/server/src/service/package_manifest.ts
jswir: Half fixed in fa3a952f, half deliberately left — flagging rather than half-building it.

**Dropped "or rename the file".** The message now says the opposite, explicitly: `"explores": []` is the opt-out, and *Do NOT rename or delete the file to opt out: that widens the surface silently and breaks every import that names it.* Pinned in the spec with a `not.toMatch(/or rename the file/)` so it cannot drift back.

**The removal case has no warning, and I did not add one.** You are right about the mechanism — removing the file takes `hasIndexModel` false, `resolveExplores` returns `{explores: undefined, warnings: []}`, and the surface widens with nothing said. But `resolveExplores` is a pure function over the manifest and the current file list. It cannot tell "this package never had an index.malloy" from "this package had one until the last save" without something persisting the previous surface across loads, and that state does not exist today.

So it is a real gap and it needs somewhere to keep the previous surface, which felt like more than a review fix should quietly introduce. Happy to build it if you want it in this PR — the natural home looks like the same place `staleCompileErrors` lives, since that is already per-package state the environment keeps across reloads. Say the word and I will, otherwise I would rather it be its own change than a silent addition here.

Worth noting the neighbouring case is better covered than I first thought. I claimed a broken `index.malloy` goes silent; running it showed every author-facing path already reports it — first load fails the package into `loadErrors`, and the watcher, `reload_package` and `?reload=true` all keep the last good model and report `stale: true` with the compile error. I corrected that in bd4cd065.

packages/server/src/service/package.ts
jswir: Fixed in 537c5928, but by wording rather than by returning the origin — flagging that, because you proposed the other shape and I want the option open.

The remedy now reads: *delete the file AND any `"explores"` entry naming it. Deleting the file alone returns the package to listing every model only when no `"explores"` was written; where one was, it would be left naming a model that no longer exists and the package would list nothing.* Pinned by a new test on the hand-written-key package, since as you said only the convention wording was covered.

I also fixed the docblock you quoted. It argued origin does not matter because the author "edits the same file", which is true of the file and false of the remedy — your words, and they were right. It now says the predicate is true for both and that every message it selects has to hold for both.

On returning the branch from `resolveExplores`: I think you are right that it is more correct, and it would make `surfaceIsIndexModel()` honest instead of a re-derivation that happens to be true for two different reasons. I did not do it because it puts an origin into `packageMetadata`, which this design goes out of its way to keep origin-free, and the defect in front of me was the advice rather than the predicate. That is a judgement call on my side, not a disagreement with yours — if you would rather have the honest predicate, it is a small change and I will make it.

Same story one file over in `discoverDashboards`, where the remedy said "declare an `explores`" to an author who already has one; that now reads "declare (or extend)".

api-doc.yaml
jswir: Fixed in f095aed3. Qualified rather than deleted, since the equivalence is still true where there genuinely is no surface — it just is not "when `explores` is absent" any more:

> The two modes are equivalent (everything queryable) only where there is no curated surface at all, which now means no `explores` AND no root `index.malloy` — a package with that file has a surface whether or not the key is present.

Agreed on why it matters: it is the published spec, so it is the version a generator reads, and it was contradicting the block four lines below it.

jswir:
Thanks both — all eleven items are in, and four of Kyle's were defects in fixes I had pushed the day before, which is the useful kind of catch.

Branch was CONFLICTING against the base, which matters more than it reads: GitHub stops dispatching CI on a conflicted PR, so the checks went quiet and looked green when nothing had run (DCO was the only check on the previous head). Merged the base in at 37ca95c7 and CI is running again.

### @Sha-Bang — replies inline on each thread

Both blocking ones are fixed: a malformed `explores` now fails the load (refusing, the second of your two options, reasoning on the thread), and the rename advice is gone. The should-fix and the nit are fixed too. One half I deliberately did **not** build: a warning for the *removal* case. `resolveExplores` cannot tell "never had an index.malloy" from "had one until the last save" without state that persists the previous surface across loads, and that does not exist today. Details and a proposed home on that thread — happy to build it here if you want it in this PR.

### @kylenesbit — all seven

**The scaffolder exposure, both workflows:**

1. **xlsx recovery probe** (`model.custom.xlsx.malloy`) — you were right that it was unrunnable both ways. The probe now goes in `index.malloy` and gets named in `export { ... }`, with the reason stated: raw SQL has to be part of a model, and the surface file is the only model a query can be addressed to.
2. **Dashboards silently withheld** — every scaffolded package's generated `AGENTS.md` now says the surface cannot publish a dashboard, before the author builds one, since the fix is a different curation shape rather than an extra line. `malloy-dashboards` gets it as step 0 of the build sequence, and its troubleshooting bullet covered only the "add index.malloy to a package that has dashboards" direction — the reverse is now the common case, because every scaffolded package has that file.

**The four warnings:**

3. `"explores": []` with no index file now gets its own message instead of the deprecation, whose advice ("it curates and enforces exactly as the key does") is exactly the thing an empty array does not do.
4. Empty-surface remedy — fixed, same thread as Nathan's.
5. The withholding guard counted `.malloynb`. Filtered to `MODEL_FILE_SUFFIX`; it was contradicting the comment directly above it.
6. Broken-surface message — one message per surface now instead of one per file each claiming to be the whole surface, and the collateral count only counts models that actually compiled rather than every hidden `.malloy`.

7. **`docs/givens.md`** — both the sample and the introspection link now name `index.malloy`, with a line that givens declared in an imported file reach the surface unchanged.

### One correction to my own earlier claim

I built and ran this against a live server, and it showed that the broken-`index.malloy` finding I fixed the day before was overstated. I said that case goes silent. It does not, on any path an author uses: first load fails the package into `loadErrors`, and the watcher, `reload_package` and `?reload=true` all go through `Environment.loadPackage`, which keeps the last good model serving and reports `stale: true` **with the compile error**. My unit test only reached the "silent" state by calling `reloadAllModels` directly, which bypasses that handling.

The warning still has a home — the materialization and manifest rebind paths call `reloadAllModels` directly, install a placeholder, and never reach `loadPackage`, so they are the one way to get a package that is serving, is not stale, and exports nothing. Kept the code, scoped the claims to that path, in bd4cd065. Flagging it because I had described it as a broader bug than it is.

### Verified against a running server, not just tests

Three scratch packages plus a second server for the boot case: convention package lists only `index.malloy`; query through the surface 200; off-surface model file 404 `No queryable model`; non-exported source through the surface 404 `No queryable source`; notebook beside `index.malloy` stays quiet; `explores: []` lists everything with the new message; malformed `explores` at first boot leaves the package **unserved** with the message in `loadErrors`, so the hidden source in the fixture never becomes reachable; and the GET→PATCH round trip writes the description without persisting `explores`.

Typecheck and lint clean, 4166 server unit tests pass against the same 35 known `config.spec` failures, integration green. One perf test from the base (`stays linear on the input that made the old one quadratic`) fails under a loaded full run and passes in isolation.

Also resolved the `AGENTS.md` governance bullet in the merge, Kyle — you were right that it was the base's term and not this branch's. Kept the base's `#(authorize)` and this branch's `index.malloy`.

@Sha-Bang on the contract note: understood, and the manifest-reading blind spot is noted — that side is being fixed on your end, so nothing changes here for it.

packages/server/src/service/package_manifest.ts
jswir: Built it after all — 48355e55. You were right that it wanted doing, and it needed less than I thought.

I said `resolveExplores` cannot see this, which is true and is why the warning is not there: it is a pure function over the manifest and the current file list, so it cannot tell a package that never had an `index.malloy` from one that had it until the last save. What I missed is that it does not have to. `Environment._loadOrGetPackageLocked` already holds both the outgoing package and the incoming one on the reload path, so the comparison goes there and the message is handed to the new package to carry on its warnings array. No new persisted state, no worker plumbing.

Only `undefined` counts, and the distinction is the point:

- `undefined` is "no `explores` key and no root `index.malloy`" — what deleting or renaming the file resolves to. Nobody asked for it.
- `[]` is an author writing the documented opt-out. They asked for exactly this and already get a message saying so.

Said once, on the reload that caused it, because it reports a change rather than a state. A later reload of an already-uncurated package has nothing to report, and the spec pins that it goes quiet again — otherwise it becomes permanent noise on every package that has no surface.

One framing correction, and it is [@jswir]'s rather than mine: I had been calling this an access change. It is not. It is curation — what widens is what is listed and what answers by name, and a source gated by `#(authorize)` stays gated. `docs/discovery-and-access.md` says it in this PR: curation hides, `#(authorize)` denies, hiding is not denying. The message and the comments now say curation, not access.

The new integration spec deletes the surface file, reloads, and checks both that the notice appears and that the withheld model really is listed and answering again; it fails with the comparison unwired.

PR #1209, opened 2026-09-22
docs: name the MCP endpoint where non-Claude clients are told to use it

jswir:
"Other clients, and unattended agents" tells Cursor, VS Code, Codex and Claude Desktop users they "take the same endpoint" — without saying what it is.

The URL only appears earlier, inside a `claude mcp add` command in the troubleshooting note for Claude Code. So someone setting up Cursor has to scroll back and extract `http://127.0.0.1:4040/mcp` from a flag on a command meant for a different client.

This states the endpoint inline and names `--mcp_port` / `MCP_PORT`, which were documented only in `docs/ai-agents.md`.

Two lines. The README is otherwise good on MCP — this is the one place it makes the reader work.

## Related

- [#1208](https://github.com/malloydata/publisher/pull/1208) — npm discovery metadata for `@malloy-publisher/server`
- [malloydata/malloy#3096](https://github.com/malloydata/malloy/pull/3096) — surfacing MCP in the language README
- [malloydata.github.io#344](https://github.com/malloydata/malloydata.github.io/pull/344) — MCP docs page rewrite


jswir:
Folded into #1208 — the npm metadata and this README line are the same change (making the MCP server findable and connectable), so they belong in one review.

PR #1210, opened 2026-09-22
An eval set over the storefront example, with verified goldens and a worked run

jswir:
Twelve natural-language questions over the bundled `storefront` package, with goldens derived twice each, plus the write-up of a first run. It is the worked example for `skill:eval-loop`: clone, run one command, and get a score, retrieval recall, a diagnosis per failure and a servable report.

`storefront.malloy` is unchanged. The report proposes model edits; it does not make them.

## What the run found

**11 of 12 (92%).** Four of the five failures I predicted did not happen, and the measurement is the interesting part:

- **net sales** — predicted to report gross, since `total_sales` carries no status filter. The agent found `status` and filtered it.
- **the 2025 signup cohort** — predicted to be unreachable. `customers.signup_date` is a dimension and the agent wrote the two-step join.
- **discount off list** — same prediction, same result. It also avoided the symmetric-aggregate grain trap that caught this set's own author.
- **the duplicate-name trap** — 1,000 customers hold 769 distinct names, so grouping spend by `full_name` merges three Mila Costas into a false top spender of $17,988. The agent grouped by `customer_id`.

Only the **business convention** failed. The company's summer is 25 May to 15 September ($267,423); the agent answered the meteorological June–August window ($213,939, 25% lower). That is the one category no amount of field documentation closes, because the rule exists nowhere in the data.

## A judge fix came out of it

The arm originally printed **100%** with that wrong answer in it. A rubric clause granted `near_match` for stating the window used, and `near_match` is excluded from the pass rate — so a materially wrong number cost the score nothing.

The judge was not at fault: it named the clause that allowed the verdict, which is what rule 7 asks of it. Nothing constrained what a clause may *sanction*. Rule 7 now carries a floor and a test — *would a reader who acted on this figure be wrong?* — and says a rubric clause cannot make a wrong value defensible. Naming your method makes an error diagnosable, not partial credit.

Two of the set's four `near_match` clauses had the defect and now score `no_match`. The other two are sound, and the line is sharp: they grant `near_match` to a figure that is *right* under a reading the question leaves open, missing only its stated basis.

## How the goldens work

Pre-defined and committed, not derived per run — a key derived during the run by the thing being measured is not independent evidence, and a key that moves between runs makes a score delta meaningless. Each value was derived once as SQL over the raw parquet and once as Malloy through a truth package holding the raw tables with no modelling, on axes that differ. `verify_goldens.py` re-derives them before every arm and refuses to spend on a drifted key.

The two-derivation rule earned its place immediately: translating the SQL literally gave `t_products.cost.sum()`, which Malloy correctly evaluates over distinct *products* rather than per order line, so margin rate read 99.6% and discount read **-118%**. One derivation would have shipped both as goldens.

## Reviewing

- `examples/storefront/evals/storefront-tour/README.md` — the exact commands, including five things this run got wrong the first time
- `RESULTS.md` — the write-up
- `runs/run-package-baseline-01/` — the report package: case matrix app and aggregate notebook

**Known gap, stated rather than hidden:** the first run's directory was untracked when an unrelated branch switch removed it, taking the transcripts and the full diagnosis. The report package and console log had already been written, so every figure is sourced — but the arm cannot be re-judged, so the corrected rubric takes effect on the next run and `RESULTS.md` shows both verdicts.


kylenesbit (approved):
Approving. The truth-package split is the right shape and the set is worth having. Everything below is findings rather than objections to the approach — but the first one I would fix before this merges, because the goldens are the product here and this one's evidence contradicts its own rubric.

Verified against `jswir/storefront-eval-tour` @ `34d7af60`. Where a finding is a claim about the data, I queried the committed parquet rather than reasoning about it.

### The golden that argues against itself

**`examples/storefront/evals/storefront-tour/cases.jsonl:10` — the `best-customers` verification note is wrong under both groupings.**

The note reads: *"By spend the top is Delilah Okafor (884, $8,818); by order count it is Amelia Cohen (80 orders). The two rankings disagree, which is the point of the case."*

Queried directly:

- Grouped by `customer_id` — the grain this same file's rubric requires — the top by order count is **Delilah Okafor (884, 43 orders)**, the *same* customer who tops spend ($8,817.56). The rankings agree, so the case's stated premise does not hold.
- Grouped by `full_name`, the top is **Mila Costa (99 orders)**, not Amelia Cohen (80).

There are three distinct `customer_id`s named Amelia Cohen (510, 583, 809), so 80 is a merged total — and she is not top even under the merged grain. The only route to that number is the full-name merge the rubric calls a NO MATCH two lines above, and that `storefront-tour-top-customer-spend` exists to catch. A judge reading this note as evidence will bless exactly the answer the set forbids.

### Two numbers the harness reports that aren't true

**`skills/eval-loop/scripts/run_baseline.py:1306` — the retrieval rung prints a fabricated 100%.**
`cascade()` is a strict funnel: a case whose `coverage` is not `covered`/`ok` is bucketed `unmeasured` and never reaches the recall check, so `not retrieved` is structurally 0 whenever coverage did not run. The new `retrieval_base = total - not covered - no entities named` therefore counts every unmeasured case as a successful retrieval. Reproduced: 12 cases, coverage `unknown`, 3 with real recall 0.5, prints `retrieved? 12 yes, 0 no`. The old line printed `0 yes, 0 no` — confusing, but honest. It fires on two supported paths: `--no-coverage`, and the `measure_coverage` failure path that deliberately returns `None`. The test at `run_baseline_test.py:1230` asserts on a hand-built cascade (`unmeasured: 11` *and* `not retrieved: 1`) that `cascade()` cannot produce, so it does not protect this. The fix belongs in `cascade()` — classify recall for unmeasured-coverage rows — not in the display.

**`skills/eval-loop/templates/eval-run-package/public/app.js:103` — the new `N/M` badge disagrees with the recall metric in the same report.**
`build_run_package.py:394` builds the dots as `required = {e for g in groups(exp) for e in g}` — the flattened union of `required` and every `requiredAnyOf` alternative — while `score_retrieval` computes recall per *group*, where any one member satisfies the group. `storefront-tour-top-category` (1 required + a 3-way `requiredAnyOf`) renders 4 dots, so if the two non-canonical aliases do not come back it shows `2/4` in the red `.short` style on a case whose recall is 2/2 = 100%. Nine of this set's twelve cases carry a `requiredAnyOf`, and the legend added at `index.html:218` promises the fraction is "how many of the entities the right answer needs came back". Count groups, not flattened entities.

### A guaranteed five-minute stall

**`skills/eval-answer/scripts/check_findable.py:383` — the index-wait terminal set names two statuses that do not exist and omits the two that mean "will never be ready".**
The server enum (`api.ts:852`, `api-doc.yaml:3509`) is `indexing | ready | cooldown | too-many-entities`. `"disabled"` and `"unavailable"` are never returned; `cooldown` and `too-many-entities` are not in the break set, so a package over the embedding cap polls the full `--index-wait` (default 300 s, ~150 requests) and can never succeed. Worse, the enum docs state `indexing` "clears once a `get_context` question triggers a sync… it does not clear on its own for a package nothing has queried" — this loop only polls the package resource and never drives a ranking call, unlike `serve.py:warm_retrieval` which issues an MCP call first. On a freshly loaded package that is a guaranteed five-minute stall followed by the warning the wait was written to avoid.

### Tests that don't run, and a table that doesn't render

**`skills/eval-loop/scripts/run_baseline_test.py:1558` — the five new tests never execute in CI.** `CoverageByDefault` and `OffloadedToolResult` are defined *after* the `if __name__ == "__main__": unittest.main()` block at line 1556. `package.json:35` runs these as `python3 "$t"`, so `unittest.main()` calls `sys.exit()` before the classes exist. `python3 run_baseline_test.py` → 161 tests; `python3 -m unittest run_baseline_test` → 166. Move both classes above the guard.

**`skills/eval-diagnose/SKILL.md:219` — two new paragraphs are inserted into the middle of a markdown table.** They land between the `NOT-RETURNED` and `LOW-RANK` rows of the `get_context / model` table, with no blank line before the rows resume. Any renderer — including the skills bundle served as an MCP prompt — absorbs `LOW-RANK`, `AMBIGUOUS`, `GUIDANCE-NOT-RETRIEVED` and `GUIDANCE-DECLINED` into the preceding paragraph as literal pipe text, so four of the six diagnosis codes lose their structure for every agent that reads this skill. Move the prose below the table.

### Smaller

- **`skills/eval-loop/scripts/run_baseline.py:3140`** — coverage-by-default ignores `--only` and fires on `--rejudge`. `measure_coverage` forwards only `--set/--publisher/--environment/--package/--out`, so a `--only one-qid` re-run spends a `claude -p` call on every case in the set; `--rejudge` does not set `a.rebuild` (only `--from` does, line 2789), so it re-measures coverage on a run that answers nothing. Forward `--only` and `--parallel`, and add `a.rejudge` to the skip condition.
- **`examples/storefront-tour-truth/truth.malloy:12`** — the header says "`data/` is a symlink to the model's own `data/`, so the bytes behind a golden are the bytes behind an answer. Copying them would let the two drift." The four files are committed as regular blobs (`100644`), and the package README has a whole "Why `data/` is a copy" section explaining that a symlink does not survive package install. The one comment that would prompt someone to re-verify goldens after a data change is the one that says they cannot drift.
- **`examples/storefront/evals/storefront-tour/as-received/questions.md:3`** — "Ten natural-language questions" in the file the same document declares the source of truth; twelve are listed, and `set.json`, the README and `cases.jsonl` all say twelve. The intro also says three questions ask for a concept `storefront.malloy` does not define, but the set README classes net-sales as reachable.
- **`examples/storefront/eslint.config.mjs:50`** — `evals/.gitignore` excludes both `*/runs/` and `run-package-*/`, but eslint ignores only `**/examples/storefront/evals/*/runs/**`. A run package at `evals/run-package-baseline-01/` — the layout the PR description uses — is gitignored yet still linted, so `lint:examples` fails locally on generated code nobody edits and CI never sees.

### Checked and clean

`packages/skills` is bumped to `0.1.22` and the branch is level with `main`, so the version-bump gate passes (note this takes `0.1.22`, so #1200 will need `0.1.23`). The regenerated `skills_bundle.json` is consistent with the manifest.

PR #1211, opened 2026-09-22
fix(skills): stop teaching query errors, and guard the ones a message cannot fix

jswir:
Three query skills were teaching mistakes that agents make in production, and `malloy-gotchas-queries` was missing guards for errors an agent cannot fix from the message alone. This fixes both. Every claim was checked against a running Publisher before it went in, and several claims already in these skills failed that check.

The error classes come from roughly a thousand classified `execute_query` compile errors that Credible served between 2026-08-01 and 2026-09-21.

## Corrections: things the skills were teaching

- **`malloy-queries` said `month(order_date)` and `year(order_date)` "fail to parse in `where:` because the names are reserved."** They parse and run, including in `where:`. Only the *name* is reserved: `group_by: month is …` fails, and calling `month(ts)` does not. It also contradicted the extraction table in `malloy-gotchas-queries`.
- **`malloy-queries`' own `calculate:` example aliased a field `month`,** which is the reserved-word error the same file warns about.
- **`malloy-queries` said `order_date.month.day` "fails".** Chained truncations compile. Removed the claim, plus the error-table row built on it, whose message (`missing {DAY, HOUR, …}`) the compiler no longer emits.
- **`malloy-analysis-pitfalls` used `seats = null`.** That works (the compiler rewrites it to `IS NULL`), but it is the form the compiler warns about, in an example agents copy. Changed to `is null`.
- **`malloy-gotchas-queries` said commas between clauses are wrong, and that `;` is "not a separator anywhere".** Both are false: commas and `;` between clauses compile. The real trap is a `;` *inside* one clause, which orphans the next field. The two sections are merged into one that says so.
- **The error table spelled `Parser encountered`.** The compiler emits `enountered`, so looking the message up missed. The row now names the common cause: a chart annotation left as the last line inside `{ }`.

## Additions to `malloy-gotchas-queries`

Each covers an error the agent cannot fix from the message alone, which is the design principles' test for prevention content.

| guard | why the message does not fix it |
|---|---|
| a measure is already an aggregate | `Aggregate expression cannot be aggregate` does not name the field |
| window braces bind to the function | `` `partition_by` is not supported for this kind of expression `` does not say where the brace belongs |
| apostrophes and concatenation | an apostrophe reports an error *after* the problem; `++` / `\|\|` report `unexpected '+'` and never mention `concat()` |
| compiles, then refused by the warehouse | the position quoted is in the generated SQL, not the query; and `partition_by:` takes a field name only, so the obvious `::string` fix does not compile either |

They go in `malloy-gotchas-queries` because it is the query-side guard skill in the `analysis` group. The equivalent guards in `malloy-gotchas-modeling` are not in that group, and the group-closure rule forbids referencing them.

**Not added, on purpose:** `= null`, which compiles to `IS NULL` and returns the right rows (a style warning, not an error); reserved words, `where:`/`having:`, `count(distinct)` and unknown functions. Their messages already name the fix, and `malloy-queries` already covers the first two. No new `skill:` references were added, since no focused skill on `main` has one.

## Verification

- **Every WRONG and RIGHT example was run against a local Publisher serving `faa`.** The quoted messages are what it returned.
- **`manifest.spec.ts` 254/254, `skills_bundle.spec.ts` 12/12.** The bundle is regenerated.
- **No version bump.** `main` is already ahead of npm on both packages (`0.1.21` > `0.1.20`, `0.0.17` > `0.0.16`), which is the question `skills-npm.yml` asks.

**A one-off eval.** Twelve questions over `faa` were written so that answering each one naturally leads into one of these errors, with sonnet at low effort. It is a biased sample by construction and the numbers are small, so read the direction, not the digits.

| arm | errored `execute_query` calls |
|---|---|
| `main` skills, as shipped | 17 of 35 (49%) |
| this branch, as shipped | 15 of 31 (48%) |

No difference as shipped, because the answerer opened a skill in only 1–2 of 12 attempts. With the gotchas text pasted into the prompt, so it was certainly read, the new text beat the old on the window-function question (three samples each):

| gotchas text in the prompt | errored calls | used Malloy's `{ partition_by }` |
|---|---|---|
| `main` | 39 of 47 (83%) | 0 of 3 samples |
| this branch | 11 of 27 (41%) | 3 of 3 samples |

With the old text the agent wrote SQL's `over (partition by …)` and never found Malloy's syntax. Most of the other questions came out clean under either version once the text was read.

The eval also caught one regression, which is fixed here. The first draft of the aggregate guard showed only a two-stage pipeline, and the agent copied it for a "top one" question that `order_by` + `limit: 1` answers in one stage. The guard now names the one-stage form, and two re-runs came out clean.

**What that means:** the content is better when it is read, but it will not move production error rates until agents actually open these skills. That is a separate change to the skill descriptions and loading, and it will be its own PR.

## Follow-ups, not in this PR

- **Mirror into `ms2data/agent-skills` by hand.** The skills README requires it until `ms2data/service#6177` lands.
- **Skill descriptions and loading**, per the eval above.
- **The model-authoring skills.** `malloy-document` has 21 `#(doc)` examples written as unquoted prose and none quoted, and unquoted prose is one of the authoring errors in the same corpus.

**Changed in review** (thanks @girishjeswani): the `= null` guard is removed, since it was based on testing against data with no NULLs; the warehouse guard now aliases the cast in `group_by:` because `partition_by: x::string` does not compile; and the window example's denominator is now a group total computed in `aggregate:`, since `all(n)` inside `calculate:` is not one.

One row left alone: *"field is a bar chart, but is not a repeated record"* is a render-time message. It fires when `# bar_chart` sits above a field inside the braces, and the renderer itself suggests moving the tag above `run:`. The existing row is correct.


girishjeswani (approved):
Approving. I compiled all six corrections rather than reading them, and every one holds — including `enountered`, which comes back verbatim as `Parser enountered unexpected statement type 'unimplemented' when it needed 'query statement'`. So that row really never matched. `month()`/`year()` run in `where:`, `.month.day` chains, and commas and `;` between clauses both compile while `;` inside a clause gives exactly the quoted `no viable alternative at input 'charters'`.

One fix worth making before merge, given what this PR is about.

**`partition_by: season::string` does not compile** (`unexpected '::'`). `partition_by:` takes a bare field name only — it rejects any expression, `partition_by: season + 1` included. The working form is to alias the cast in `group_by:` first and partition on the alias. As written, the guard against a warehouse refusal hands the agent a second error it doesn't warn about.

Two smaller ones.

**The `= null` guard's framing.** Malloy's codegen rewrites `= null` to `IS NULL` — `COUNT(CASE WHEN "seats" IS NULL THEN 1 END)` — and against real NULL rows `= null` and `is null` returned identical answers. That's dialect-independent, so "depending on the server this is either rejected or run as a filter that matches nothing" doesn't match what the compiler does. Either name the server or dialect where you saw it misbehave, or soften it to a style rule. The `malloy-analysis-pitfalls` change is right either way.

**The window RIGHT example.** `sum_cumulative(n) { partition_by: region } / all(n)` runs, but `all(n)` is un-partitioned, so `share` is a per-region running total over a whole-result denominator. `all(n) { partition_by: region }` also fails, so that may be forced — worth a word so an agent doesn't copy it expecting a within-region share.

Everything else checks out. The four other guards clear the bar you set for them: `Aggregate expression cannot be aggregate` is four words that identify nothing, and neither `++` nor `||` ever mentions `concat()`. Every other RIGHT example compiles and does what it claims, including the apostrophe case actually matching `Mac's Diner`. Every deletion was of something provably false. Bundle sync 12/12, specs 258/258, DCO signed, descriptions under budget.

On the eval: naming the confound, saying which number to trust, and labelling the pasted-text arm as biased is the right way to report it. It also isn't load-bearing here, since the corrections are checkable by compiling.

jswir:
Thanks for compiling these rather than reading them. All three hold, and each is fixed in d736f436.

**`partition_by: season::string`.** Confirmed: `unexpected '::'`. The guard now casts in `group_by:` (`season_key is season::string`) and partitions on the alias. Checked that it compiles.

**`= null`.** You're right, and the mistake was mine. I tested against `faa`, which has no NULLs in any column, so zero was the correct answer for both forms and I misread it. On a source with real NULLs, `= null` and `is null` both return 90,002 rows, and `!= null` matches `is not null`. A style warning fails the "error the agent can't self-correct" test, so I removed the section rather than softening it. The `malloy-analysis-pitfalls` edit stays.

**The window example.** Worse than un-partitioned: `all(n)` inside `calculate:` isn't a total of any kind, and the "share" ran past 1.9. The RIGHT example now divides by a total computed in `aggregate:` (`region_total is all(count(), region)`). That gives the within-group running share: 0.253 = 8,742 / 34,577 for AA at DFW. One line says why, since `{ partition_by }` doesn't apply to `all()`.

I've corrected the PR description to match.

PR #1213, opened 2026-09-22
skills(malloy-analysis): a description that gets the skill opened

jswir:
Stacked on #1211. One line changed in `skills/malloy-analysis/SKILL.md`, plus the regenerated bundle.

## The problem

Claude Code puts every skill's **name and description** in the prompt, but loads the **body** only when the agent invokes the skill. So the description decides whether the rest of the file is ever read.

On a 79-question eval over a Malloy semantic model, `malloy-analysis` was opened on 8% of answering attempts. Opus opened it on **none of 79**. Everything in the skill was reaching almost none of the work.

The shipped description says what the skill *is*:

> The workflow for answering a data question over a Malloy semantic model - a metric, breakdown, trend or chart - from get_context discovery through execute_query to verified delivery.

An agent that already knows how to answer a data question has no reason to open that.

## The change

> Read before your first get_context call on any business or data question. The required workflow, not background: which source, which rows, how to verify. Skipping it causes confident wrong answers.

Each clause does one job:

- **When** — "before your first get_context call" ties it to a step in the agent's own loop, not a topic.
- **Which questions** — "business or data question". The eval's prompt asks a *business* question; the old wording only matched *data* questions.
- **Status** — "the required workflow, not background".
- **What's inside** — the three things the agent can't guess: which source, which rows, how to verify.
- **Cost of skipping** — "confident wrong answers".

197 characters, inside the 200 that `manifest.spec.ts` allows.

## Measured

15 questions from an eval set over the `agent_analytics` package. Medium effort, same questions, same judge. Only the description differs.

| description | Opus opened skill | Sonnet opened skill |
|---|---|---|
| shipped (182 chars) | 0 / 15 | 3 / 15 |
| "status" only (174) | 6 / 15 | 15 / 15 |
| "contents" only (183) | 5 / 15 | 15 / 15 |
| **this PR (197)** | **11 / 15** (run twice: 11, 11) | **15 / 15** |

Keeping only some of the clauses fixed Sonnet but left Opus under half. Keeping all of them, plus the "business" wording, took Opus to about three in four, the same in both runs.

## What this does not claim

It gets the skill **read**. It does not, on this evidence, make answers **more correct**. Pass counts across these arms moved by a question or two, which is within the run-to-run noise measured on this set. Ship it because a skill should be read, not as a quality fix.

## Two notes for review

- **The value is quoted.** It contains a colon, and the frontmatter test (`frontmatter parses as YAML`) rejects that unquoted. I confirmed with `claude -p --setting-sources project` that the quoted and unquoted forms load as the identical string, so the quoted version is exactly what was measured.
- **The variants were tested on two bases.** The "status" and "contents" variants ran on `main`; the final text ran on this branch, on top of #1211. #1211 changes only skill *bodies*, not any description, so it shouldn't affect whether a skill gets opened.

## Testing

- `packages/skills`: 269 pass, 0 fail
- `packages/server` bundle specs: 18 pass, 0 fail

`docs/agent-skills/authoring-skills.md` already says the description should be "a precise trigger, not a summary." The other skill descriptions may be worth checking against that line; this PR only changes `malloy-analysis`.


girishjeswani (approved):

jswir:
Merged main in (23cfed09). The branch still carried the pre-squash commit of #1211, so main's version now wins for those files and the diff is back to the one description line plus the skills bundle, rebuilt with build_skills_bundle.ts rather than hand-merged. Tests pass: packages/skills (275, including manifest and frontmatter specs), server src/mcp/skills (18), and create-malloy-package skills.spec (6). #1215 is still open, so I parsed the frontmatter with the yaml package directly: the description comes back intact, 197 characters. The version-bump check is left failing on purpose; the bump is deferred to merge time.

Issue #103, opened 2025-05-17
Support mysql connection


aszenz:
Great work on the app, was trying it out, works well, but I'm missing mysql connection support, would be great to add since malloy now supports it quite well

quirkyllama:
MySql supported as of 0.0.75

quirkyllama:
I’ll take a look tomorrow!

quirkyllama:
I integrate the change and confirmed I can connect.
My publisher.connections.json looks like:
 {
        "type": "mysql",
        "name": "mysql",
        "mysqlConnection": {
            "host": XXX
            "port": 3306,
            "database": "mysql",
            "user": "XXX",
            "password": "XXX"
        }
    }

aszenz:
> I integrate the change and confirmed I can connect.
> My publisher.connections.json looks like:
>  {
>         "type": "mysql",
>         "name": "mysql",
>         "mysqlConnection": {
>             "host": XXX
>             "port": 3306,
>             "database": "mysql",
>             "user": "XXX",
>             "password": "XXX"
>         }
>     }
> […](#)

Thanks, seems to be working with new release, i can connect and run queries.

The db connections card is still empty but maybe that's a sep issue.

PR #1214, opened 2026-09-22
docs(skills): stop telling contributors to mirror shared skills into agent-skills

jswir:
Credible now serves every `malloy-*` skill from the published `@malloy-publisher/skills` package (ms2data/service#6226, live in production). This repo is their only copy, so the README instructions to mirror an edit into `ms2data/agent-skills` by hand are now wrong. ms2data/agent-skills#103 deletes the copies on that side.

## Changes

- `skills/README.md`: "Where these come from" now says how a change reaches Credible: publish the package, then bump `skills_version` in service. It says not to open a mirror PR.
- Removed the eval-skill mirroring instructions (record the source SHA, files to delete when mirroring). No Credible manifest lists the eval skills.
- `malloy-dashboards`: Credible's counterpart is now named `credible-dashboards`.
- `AGENTS.md`: "kept in sync with an upstream repo" now names the npm package.

Docs only. `skills/README.md` is excluded from the published tarball, so no skills version bump.


Sha-Bang (approved):
the claims hold — I opened each against the code rather than reading the prose. the routing rule is `agent_skills.py:207` in service (`skill_name == "malloy" or skill_name.startswith("malloy-")`), `skills_version` is the right key in `infrastructure/malloy-versions.json` (`agent_skills_version` is the separate git-tag pin for the Credible half, so naming only `skills_version` here is correct), and "its tests fail on one" is true: I planted a `skills/malloy-sneaky/` directory in ms2data/agent-skills#103 and `test_no_local_malloy_skills` fails on it.

approving. two things, neither blocking.

**follow-up, outside this diff.** `skills/README.md:30` now says nothing reads a `malloy-*` skill from that repo, but `skills/eval-loop/scripts/agent_harness.py:149` still carries an agent-skills mode: `ROLE_MANIFESTS["modeling"] = ("modeling-app", "publisher-local#modeling")`. ms2data/agent-skills#103 deletes `manifests/modeling-app.json`, so from an agent-skills checkout `default_manifest("modeling", ...)` falls through both candidates and returns `"modeling-app"` — a manifest that then exists in neither repo, and `manifest_skills` fails on it. dead path in practice, since the harness only lives here, but it's stale in exactly the direction this PR is cleaning. `platform_target_test.py:220` carries the same assumption in prose ("the `agent-skills` checkout these skills also ship to"). worth a sweep in its own PR rather than widening this one.

**nit.** `skills/README.md:58`, "No Credible manifest lists the eval skills" is true as written — I checked, none of the four do. but "the eval skills" now names two disjoint sets: this repo ships `eval-answer`/`-diagnose`/`-import`/`-improve`/`-judge`/`-loop`/`-report`, and #103 leaves agent-skills with nine `credible-eval-*` skills that share none of those names. one clause saying which set would stop the next reader reconciling them.

not about the content: DCO is red, so this can't merge until the commit carries a sign-off (`git commit --amend -s` and force-push).

jswir:
Thanks — DCO signed off on both commits, and the nit is fixed: the eval paragraph now says the seven here ship nowhere else, and that agent-skills' nine `credible-eval-*` are a different set sharing no name.

On the follow-up: confirmed and out of scope here, agreed. `eval-loop/scripts/agent_harness.py:149` maps `modeling` to `("modeling-app", "publisher-local#modeling")`, and ms2data/agent-skills#103 deletes `manifests/modeling-app.json`, so from an agent-skills checkout both candidates miss and `manifest_skills` fails on a name that then exists in neither repo. `platform_target_test.py:220` carries the same assumption in prose. Its own PR.

PR #1216, opened 2026-09-22
Coverage: show the judge the real model surface, and what the questions mean

jswir:
Stacked on #1210 (`jswir/storefront-eval-tour`), because it needs that set as its oracle. Retarget to `main` once #1210 lands.

Coverage was wrong in the direction that flatters the model. On the storefront tour set it **passed one of the two questions the arm actually got wrong**, and flagged four that passed. A coverage number that misses real gaps is worse than no coverage number, because a failure it passed reads as the agent's fault.

Two causes, measured separately. Neither is "the judge is bad at this".

## It could not see fields the model does not declare

A Malloy source exposes every column of its table, so `products.retail_price` and `customers.signup_date` appear nowhere in the `.malloy` and are fully queryable. The judge is handed source **text**, so it called both absent and returned `COVERAGE` on two questions the model answers.

It now also gets the compiled field list over REST, with the split stated: the text is the authority on what is **documented**, the list on what **exists**, and undocumented is not absent. A `--model` run has no compiled model to ask and now says so, rather than implying the text is the whole surface.

## It did not know what the questions mean

*"How many customers do we have?"* is a coin toss between 974 and 1,000 until you know the business counts the **943** it has delivered to. Not knowing, the judge picked a reading and passed the question — the exact case the set was built to catch.

`set.json` now carries the three conventions this set is asked under, and the prompt states they are definitions rather than readings the judge may substitute its own for.

## Measured

Same model, same 12 cases, before → after:

| | before | after |
|---|---|---|
| real failures caught | 1/2 | **2/2** |
| false gaps | 4 | 5 |
| agrees with the arm | 7/12 | 7/12 |

The error direction that matters is fixed. The headline fell from 58% to 42% because more cases are called gaps, which is why the headline is not the thing to read.

Case by case: `customer-count-basis` ok → CONVENTION (right), `peak-month` AMBIGUOUS → ok (right), `net-sales` CONVENTION → ok (right), and three customer questions ok → CONVENTION (wrong).

## The open question, left open on purpose

Those three regressions are one cause, and it is not the judge's. The convention as written — *a customer is someone delivered to* — is a **global** definition, so a judge holding it reasonably applies it to every customer question. **This set's goldens do not.** `orders-per-customer` is derived over the 974 who ordered, not the 943 delivered to.

So either the convention is global and several goldens are inconsistent with it, or it is scoped to the counting question and conventions need per-case scoping. On the first reading coverage is right and the set is wrong. That is a decision about the set, not something to patch here.

## Not in scope

The capability half of the original diagnosis — letting coverage probe with `get_context`/`execute_query` instead of reasoning from a static dump — is untouched. Worth deciding after this, since the surface fix may have already removed most of what probing would have bought.


girishjeswani:
The coverage work is good and worth saying why first: the accuracy direction is the conservative one on the error that matters. The compiled surface removes false `MISSING` while the prompt still refuses to promote on existence alone (`undocumented is not MISSING`), and the conventions half removes false `MODELLED`, which is the dangerous direction. The `MISSING`/`RULE_UNWRITTEN` split is a reclassification rather than a downgrade — both stay in `FAIL_VERDICTS`, so the headline percentage does not move. The parse-don't-hardcode vocabulary chain works end to end: deleting a row from `eval-diagnose/SKILL.md` fails `diagnose_test`. And reporting the unflattering numbers (false gaps 4 to 5, headline 58% to 42%) alongside the good ones makes the rest of the table easier to trust.

Three things to fix.

## 1. `--verify-figures` cannot work as written

`verify_goldens.py:646` calls `run` without `--output-format stream-json --verbose`. `run_cli` builds its `text` return only from stdout lines that start with `{`, parse as JSON, and carry `type == "assistant"` — its docstring opens "Run one stream-json `claude -p` command". Without those flags `claude -p` emits prose, no line parses, and `text` is `""` on every call. So `body` is never a dict and every figure lands on `verdict="unverifiable", why="the checker returned no readable JSON"`.

Reproduced end to end with a fake `claude` on `PATH`, calling `verify_figures(case, a)` with `run=None` so it takes the real path: the agent returned a valid `contradicted` verdict carrying a query and rows, and the function returned `unverifiable`. Adding the two flags and re-running gave `contradicted, computed: 43`. `check_coverage.py:426-427` already passes them, so the fix is closing that asymmetry:

```python
reply = run(["claude", "-p", prompt, "--model", a.figure_model,
             "--output-format", "stream-json", "--verbose"],
            cwd=None, timeout=300)
```

Two things make it worse than a no-op. `no_events` is `not events`, and with no stream-json `events` is always empty, so every figure burns the full `retries=2` — 3 attempts and 20s+40s of backoff. On the shipped storefront set that is 11 figures to 33 agent calls and about 11 minutes, against a `--help` that promises one call per figure. And `verify()` at `:1174-1187` is an if/else, so passing the flag *replaces* `criteria_figure_findings` rather than adding to it: turning it on today removes a working warning and substitutes a message with the same information content.

**Why the tests miss it:** all five inject `run=lambda...` returning a plain `str`, so `_reply_text`'s tuple branch — the only branch production takes — is never exercised. The docstring ("returns a tuple on the CLI path and a string when stubbed") shows the shapes were known to diverge; what went unnoticed is that the CLI path yields an empty string. One test driving `verify_figures` through the real `run_cli` with a fake `claude` on `PATH` would catch it and stop the class recurring.

**While you are in there:** `_reply_text` (`:660-664`) picks the first non-empty string in the tuple, and `run_cli` returns `(events, text, stderr, attempts, wall)`. When `text` is empty it falls through to **stderr** — `_reply_text(([], '', 'timeout after 300s', 3, 12.0))` returns `'timeout after 300s'`, which then goes to the JSON scanner as if it were the agent's reply. Harmless today since stderr rarely holds JSON, but `_, text, *_ = reply` removes the hazard.

## 2. The rename stops short of four live sites

Commit 2 says the prompt "had been left holding both vocabularies at once, which teaches a judge neither". That is exactly right, and the same condition now exists across the skill files agents are served over MCP:

- `eval-answer/SKILL.md:148` — "`COVERAGE`, `AMBIGUOUS`, `NO-DISAMBIG` and `CONVENTION` mean there exactly what they mean here", now false in three of four names, and describing the file this PR rewrites
- `eval-answer/SKILL.md:201` — "manufactures `COVERAGE` verdicts"
- `eval-report/SKILL.md:306-307` — a paragraph defining `CONVENTION` and `NO-DISAMBIG` as the labels a report should use
- `eval-diagnose/SKILL.md:351` — in this PR's own file

`test_the_prompt_carries_no_retired_code` is the right shape for this; extending that absence check across `skills/**/SKILL.md` would stop it recurring, and is cheap.

Related: `verify_goldens.py:331`'s cause-code regex still lists only retired names, so it will silently miss `RULE_UNWRITTEN` and `MISSING` — worth checking whether that one is load-bearing. And the `NO-DISAMBIG` to `AMBIGUOUS` merge leaves historical ledger diagnoses referencing a code no table defines.

## 3. The wiring is unmutation-covered

Making `judge_case` stop passing `surface` and `conventions` entirely leaves the suite green. So does `main()` never calling `compiled_surface()`, `main()` never reading `set.json` conventions, and dropping `UNDERSPECIFIED` from `FAIL_VERDICTS` (which would silently change the denominator). The pure functions are well guarded; the integration that makes the feature real is not. One test driving `main()` with a temp `set.json` and a stubbed `compiled_entities` would close all four.

`rule_kind` is in the same position: the commit message makes the `unstated`-vs-`guessable` distinction the consequential one, and recording `guessable` instead of `unstated`, or dropping the field, both survive.

## Smaller

**`conventions` is undocumented.** `eval-import/reference/case-format.md:27` enumerates the legal `set.json` fields and was not updated, so an author writing a new set has no way to learn it exists. The per-convention scope sentence is load-bearing — your own open question shows what happens when scoping is wrong — but nothing validates that one is present.

**A degraded run is indistinguishable from a normal one.** `main():750-756` swallows a `compiled_surface` failure to `surface = ""` with a stderr warning. In `--publisher` mode, which is the mode that has a compiled model, a transient REST failure quietly reverts to the old less accurate behaviour and the run JSON records nothing. Worth putting in the artifact.

**Two titles over-claim.** `test_conventions_are_scoped_to_the_questions_that_use_them` asserts two substrings exist in the prompt; scoping is a judge property the test cannot reach (the docstring is honest, the title is not). The `CompiledSurface` docstring says "the judge is shown the fields", where the tests establish that `compiled_surface()` formats a dict.

**Pre-existing, but you are on the line:** `diagnose_test.py:63-66` carries its own defeat in the message — "if it was renamed this test needs the new name". Deriving the code from `skill_codes()` would make it a real guard.

Deletion audit came back clean: every removed assertion in `check_coverage_test.py` has a stronger-or-equal replacement, and the placeholder-set one got strengthened. DCO on all three commits, bundle in sync, no description budgets touched.

jswir:
Thanks, all addressed in `734332a8`..`53c05249`. Two things you didn't flag turned out to be worse than what you did, so those first.

## Two more bugs

**`--verify-figures` never ran at all.** `main()` parsed the flag but never passed it to `verify()`, which builds its own `Namespace`. So `getattr(a, "verify_figures", False)` was always `False`, and your stream-json fix alone would still have done nothing. Now plumbed through, and a test drives `main()` to prove the flag reaches the check.

**Six of this PR's tests never ran in CI.** `CompiledSurface` and `ConventionsInThePrompt` sat below `unittest.main()` in `check_coverage_test.py`. CI runs each file as a script, so the process exited before those classes were defined: 54 tests in CI, 60 under `-m unittest`. Moved the guard, and `test:skills-python` now loads each file with `python3 -m unittest` so a misplaced guard can't hide tests again. Every other file gives the same count either way.

## 1. `--verify-figures`

- Added `--output-format stream-json --verbose`, as you said.
- Also added `--allowedTools Bash(curl:*)`. The prompt tells the checker to query with curl, and a headless `claude -p` call is denied any Bash it wasn't granted, so a fake `claude` can't catch this one.
- Also `--strict-mcp-config`, so account connectors can't answer from other data, and `retry_when=no_text, retries=1`, since this checker is instrumentation.
- The reply is read from `run_cli`'s text slot only (`_, text, stderr, _, _ = run(...)`). `_reply_text` is gone, so stderr can't reach the JSON scanner.
- A verdict outside `confirmed`/`contradicted`/`unverifiable`, or a `confirmed` with no query, is now recorded as `unverifiable`.
- Stubs now return `run_cli`'s tuple, not a bare string. One test runs the real `run_cli` against a fake `claude` on `PATH` and asserts the flags. Taking the stream-json flags out fails it.
- The if/else in `verify()` stays. With the check working, each non-confirmed figure becomes its own finding, which carries more than the old warning did.

## 2. Rename

- Fixed all four sites you listed. Also fixed the prescriptive parts of `coverage-limits.md` and `definition-ledger.md`, and two stale `COVERAGE` mentions in `check_coverage.py`.
- The measured tables in `coverage-limits.md` keep the names they were taken under, with a note mapping old to new.
- `test_no_skill_teaches_a_retired_code` fails on any backticked retired code in any `SKILL.md`.
- The `verify_goldens.py:331` regex is load-bearing: it decides where a rubric's rejecting clause starts, so a rubric written with the new names had its trap figure read as asserted-right. It now lists `RULE_UNWRITTEN` and `MISSING` too, keeping the old names that existing rubrics still use. Tested.
- Historical ledger codes: nothing validates old issue events, only new output. So `ledger-schema.md` now says how to read a retired `primary_code`. I kept the mapping out of eval-diagnose's table on purpose, because a row there would make the retired codes valid again.

## 3. Wiring

`MainWiring` drives `main()` with a temp set, a stubbed REST read and a stubbed judge. I put each of these back into the code one at a time, and each makes a test fail:

- `judge_case` dropping the surface
- `judge_case` dropping the conventions
- `main()` skipping `compiled_surface()`
- `main()` skipping the conventions
- `UNDERSPECIFIED` removed from `FAIL_VERDICTS`
- `rule_kind` defaulting to `guessable`
- `rule_kind` dropped

## Smaller

- **`conventions`** is documented in `eval-import/reference/case-format.md`, with the "Applies to questions that ..." scope sentence. `check_coverage` warns about a convention without one. It can't check that the prose is right, only that a scope is stated.
- **Degraded runs:** the `--out` report records `compiledSurface` (`read`, `failed: <reason>`, `empty: ...`, or `not read: --model ...`) and the conventions the judge was shown.
- **Titles:** renamed to `test_the_prompt_tells_the_judge_conventions_are_scoped`, with a docstring saying judge obedience is out of reach. The `CompiledSurface` docstring now says what it checks and points at `MainWiring` for the rest.
- **`diagnose_test.py:63-66`** reads its codes from `skill_codes()`.

## Run locally on malloy-samples

Built from this branch, with ecommerce and ecommerce-truth on separate servers:

- **`check_coverage --publisher`, all 49 ecommerce cases:** 41% (20/49), `compiledSurface: read`. Agrees with the hand labels on 39/49, against 32/49 in the earlier measurement in `coverage-limits.md` (not a checker score, but it moved the right way).
- **`--verify-figures`** on a scratch copy of the set with two criteria notes. "168,063 customer accounts" came back `confirmed` with `run: users -> { aggregate: n is count() }`. "2,900 returned lines" came back `WRONG, computed 2672`, with its query. That took about 33s. The previous head, on the same data, ignored the flag and printed only the old warning.
- **Conventions:** the scope warning fired on an unscoped one. One mixed result, from single samples: the judge applied a scoped "sold" definition to "distinct products sold" (`RULE_UNWRITTEN`, `arbitrary`), but not to "units sold in 2022", which it also governs. That's the judge property no test here can reach, so I'm noting it rather than claiming it's fixed.

`main` merged in (the bundle was the only conflict, regenerated). 918 Python and 315 skills tests pass locally.

girishjeswani (approved):
Approving. All three findings fixed, and both bugs you found yourself are worse than what I raised — the flag never reaching `verify()` meant my stream-json fix alone would have changed nothing, and six of this PR's own tests never executing in CI is the kind of thing that makes a green suite meaningless.

Verified by running rather than reading:

- **The hidden tests reproduce exactly.** At `044c9dba`, `check_coverage_test.py` gives 54 as a script and 60 as a module; at head both are 68. The `package.json` runner change is the right structural remedy, since it makes the guard's position stop mattering.
- **Flag plumbing and stream-json**, both mutation-proven: removing `verify_figures_flag=args.verify_figures` fails `test_the_flag_reaches_the_figure_check`, and removing the two flags fails the real-`run_cli` test — which takes 20.6s, so it genuinely exercises the retry path rather than short-circuiting. The fake `claude` branching on `if 'stream-json' not in sys.argv` and printing prose reproduces the actual failure mode, which is better than an argv assertion.
- **`_reply_text` is gone** and the positional destructure matches `run_cli`'s 5-tuple. `no_text` is the right predicate for instrumentation and its docstring already draws that distinction.
- **All seven wiring mutations fail**, re-run independently. Zero survivors. `MainWiring` driving the real `main()` with stubbed REST and judge is a genuine harness, not a unit stub.
- **The rename is complete.** Every `SKILL.md` is clean, and each remaining retired code is legitimately retained: the mapping note, the measured tables governed by it, the `ledger-schema.md` historical explanation, the back-compat entries in `_CODE_MARK`. `"COVERAGE & RETRIEVAL"` in the report heading is unrelated vocabulary. Adding a backticked retired code to a `SKILL.md` fails the new guard.

**The `:358` regex was genuinely load-bearing**, and I reproduced the misread rather than taking it on trust. With a rubric reading `The answer is 168,063 customers. Using order_items (2,672,111) is RULE_UNWRITTEN.`, the old list made `accepting_clause()` treat the whole rubric as accepting and emit a spurious "rubric asserts 2,672,111 as right"; with the new codes added the clause cuts correctly and the finding disappears. Keeping the retired names alongside is right, since existing rubrics still use them.

One correction, non-blocking.

**"Every other file gives the same count either way" is not quite true.** I swept all 22 `skills/**/*_test.py` both ways at head and two still differ:

```
verify_goldens_test.py   script=124  module=128
run_baseline_test.py     script=169  module=175
```

Same cause — the guard sits above `HeldGoldensAreNotReDerived` / `ANotQueryableKeyNamesItsRootSource` at `verify_goldens_test.py:1196`, and above `NarrowedRebuildKeepsTheLedger` / `PersistedStubIsTheResult` at `run_baseline_test.py:1673`. Both are pre-existing on `main`, not yours, and all 10 rescued tests pass, so your `package.json` change genuinely covers them and CI is correct now. Worth moving the two guards so the script path agrees too, or splitting them out as a `main` fix — either is defensible.

Two smaller notes.

**The `--allowedTools Bash(curl:*)` test proves the string is present, not that the permission holds.** The hardening is plausible and matches the measured precedent in `check_coverage.py` (72 account connectors reaching a judge, fixed with `--strict-mcp-config`), but a fake `claude` cannot enforce permissions, so the assertion is argv presence. Worth not letting the test's existence imply the behaviour is validated.

**`test_no_skill_teaches_a_retired_code` scans `*/SKILL.md` only, and backticked forms only.** An unbackticked retired code, or one in `reference/*.md`, slips past. Deliberate given the reference files must keep the mapping, but worth knowing the bound.

Your own honest negative — the judge applying the scoped "sold" definition to one question and not another it also governs — is the right thing to have flagged rather than papered over. It does mean the conventions feature ships with known partial obedience, which might want a tracking issue rather than living only in a PR comment.

